PR #112, opened 2026-04-12
Add Infrastructure Dependency Graph and Blast Radius Analysis

### LINKED ISSUE #103: Add Infrastructure Dependency Graph and Blast Radius Analysis
## Summary

Add an **Infrastructure Dependency Graph** engine that automatically tracks resource relationships and enables **blast radius analysis** — answering "what breaks if I delete this resource?"

## Motivation

Cloud providers don't give you a runtime dependency view. Terraform has `graph` but only for planned resources. CloudEmu can build this from actual runtime state, providing unique value for infrastructure testing and understanding.

### Dependency Graph
```go
GetDependencyGraph(ctx) (*DependencyGraph, error)
```
Returns the full graph of all resources and their relationships. Auto-built as resources are created (VPC -> Subnet -> Instance -> Volume, etc.).

### Blast Radius
```go
BlastRadius(ctx, resourceID string) (*ImpactReport, error)
```
Returns all resources that would be affected if the given resource is deleted or goes down. Traverses the dependency graph transitively.

### What-If Analysis
```go
WhatIf(ctx, action string, resourceID string) (*ImpactReport, error)
```
Preview the impact of an action ("delete", "stop", "disconnect") without actually performing it. Returns affected resources, broken connections, and orphaned resources.

### Dependency Chain
```go
DependsOn(ctx, resourceID string) ([]ResourceRef, error)
```
Returns all resources that the given resource depends on (upstream dependencies).

### Dependents
```go
DependedBy(ctx, resourceID string) ([]ResourceRef, error)
```
Returns all resources that depend on the given resource (downstream dependents).

### Export
```go
ExportDOT(ctx) (string, error)    // Graphviz DOT format
ExportMermaid(ctx) (string, error) // Mermaid diagram format
```

## Return Types

```go
type ResourceRef struct {
    ID       string
    Type     string // "vpc", "subnet", "instance", "volume", "security-group", etc.
    Name     string
    Provider string // "aws", "azure", "gcp"
}

type Dependency struct {
    From     ResourceRef
    To       ResourceRef
    Type     string // "contains", "attached-to", "routes-through", "secured-by", "uses"
}

type DependencyGraph struct {
    Resources    []ResourceRef
    Dependencies []Dependency
}

type ImpactReport struct {
    Target           ResourceRef
    Action           string
    DirectlyAffected []ResourceRef   // immediate dependents
    TransitiveImpact []ResourceRef   // full cascade
    OrphanedResources []ResourceRef  // resources that lose their parent
    BrokenConnections []Dependency   // connections that would break
    Summary          string          // human-readable summary
}
```

## Tracked Dependencies

| Parent | Child | Relationship |
|--------|-------|-------------|
| VPC | Subnet | contains |
| VPC | Security Group | contains |
| VPC | Route Table | contains |
| VPC | Internet Gateway | attached-to |
| Subnet | Instance | contains |
| Subnet | NAT Gateway | contains |
| Instance | Volume | attached-to |
| Instance | Security Group | secured-by |
| Instance | Key Pair | uses |
| Instance | Instance Profile | uses |
| Load Balancer | Target Group | routes-to |
| Target Group | Instance | contains |
| Listener | Load Balancer | belongs-to |
| Function | Event Source Mapping | triggered-by |
| Queue | Function (trigger) | triggers |
| DNS Zone | DNS Record | contains |
| Alarm | Notification Channel | notifies |

## Implementation Notes

- Create a new `topology/` package (or `depgraph/`)
- The graph builds incrementally — hook into existing Create/Delete/Attach operations
- Consider an observer pattern: services notify the graph engine when resources change
- Alternatively, build the graph on-demand by scanning all services
- Export formats enable visualization in external tools
- The graph is cross-service: it links compute instances to networking subnets to IAM profiles

## Acceptance Criteria

- [ ] Dependency graph auto-built from resource state
- [ ] `BlastRadius()` returns transitive impact for any resource
- [ ] `WhatIf()` previews delete/stop/disconnect impact
- [ ] `DependsOn()` and `DependedBy()` return direct relationships
- [ ] DOT and Mermaid export for visualization
- [ ] Covers cross-service dependencies (compute + networking + IAM)
- [ ] Works across all 3 providers
- [ ] Tests covering: VPC cascade, instance with volumes/SGs, LB -> TG -> instances chain
- [ ] Linter passes

NitinKumar004:
## Summary

Extends the `topology/` package with dependency graph and blast radius analysis — answering *"what breaks if I delete this resource?"*

Closes #103

## Problem

Cloud infrastructure has complex resource dependencies (VPC -> Subnet -> Instance -> Volume). Deleting a parent resource can cascade failures. There's no way to visualize or analyze these dependencies in mock testing today.

## Solution

A graph engine that scans all resources, builds a dependency graph, and enables impact analysis.

## Architecture

```
                    BuildDependencyGraph()
                           │
          scans all resources from 3 drivers
                           │
     ┌─────────────────────┼────────────────────┐
     │                     │                     │
     ▼                     ▼                     ▼
 Compute              Networking               DNS
 • Instances          • VPCs, Subnets          • Zones
 • (SubnetID,         • Security Groups        • Records
    VPCID,            • Route Tables
    SecurityGroups)   • NAT GW, IGW, Peering
                      • ACLs
                           │
                           ▼
                  ┌─────────────────┐
                  │ DependencyGraph │
                  │  • Resources[]  │
                  │  • Dependencies[]│
                  └────────┬────────┘
                           │
              ┌────────────┼────────────┐
              ▼            ▼            ▼
        BlastRadius   DependsOn    ExportDOT
        DependedBy                 ExportMermaid
```

### BlastRadius Flow

```
BlastRadius("vpc-123")
     │
     ▼
 Build full graph
     │
     ▼
 Find direct dependents of vpc-123:
   → subnet-456 (member-of vpc-123)
   → sg-789 (member-of vpc-123)
   → rt-101 (member-of vpc-123)
     │
     ▼
 Find transitive dependents (BFS):
   → i-abc (member-of subnet-456, secured-by sg-789)
     │
     ▼
 ImpactReport:
   DirectlyAffected: [subnet-456, sg-789, rt-101]
   TransitiveImpact: [i-abc]
   BrokenConnections: [subnet→vpc, sg→vpc, rt→vpc, i→subnet, i→sg]
```

## API

```go
aws := cloudemu.NewAWS()
topo := topology.New(aws.EC2, aws.VPC, aws.Route53)

// Build full graph
graph, _ := topo.BuildDependencyGraph(ctx)

// What breaks if I delete this VPC?
impact, _ := topo.BlastRadius(ctx, "vpc-123")
// impact.DirectlyAffected  → [subnet, sg, route-table]
// impact.TransitiveImpact  → [instance]
// impact.BrokenConnections → [all broken links]

// What does this instance depend on?
deps, _ := topo.DependsOn(ctx, "i-abc")
// → [subnet, vpc, security-group]

// What depends on this subnet?
dependents, _ := topo.DependedBy(ctx, "subnet-456")
// → [instance, nat-gateway]

// Export for visualization
dotOutput := graph.ExportDOT()       // Graphviz
mermaidOutput := graph.ExportMermaid() // Mermaid diagrams
```

## Tracked Relationships

| From | To | Type |
|------|----|------|
| Instance | VPC | member-of |
| Instance | Subnet | member-of |
| Instance | SecurityGroup | secured-by |
| Subnet | VPC | member-of |
| SecurityGroup | VPC | member-of |
| RouteTable | VPC | member-of |
| NATGateway | VPC, Subnet | member-of |
| InternetGateway | VPC | attached-to |
| NetworkACL | VPC | member-of |
| PeeringConnection | VPC, VPC | peers-with |
| DNS Record | DNS Zone | member-of |

## Files

| File | Lines | What |
|------|-------|------|
| `topology/depgraph.go` | 56 | Core types (ResourceRef, Dependency, DependencyGraph, ImpactReport) |
| `topology/graph_builder.go` | 282 | Scans all resources and builds graph |
| `topology/blast_radius.go` | 179 | BlastRadius (BFS), DependsOn, DependedBy |
| `topology/export.go` | 63 | DOT and Mermaid export |
| `topology/topology_test.go` | +192 | 7 new unit tests |
| `cloudemu_test.go` | +82 | 3 integration tests (AWS, Azure, GCP) |

## Tests

- TestBuildDependencyGraph — verifies resource count and dependency relationships
- TestBlastRadius — VPC blast radius includes subnet and instance
- TestDependsOn — instance depends on subnet, VPC, SG
- TestDependedBy — VPC is depended on by subnet, SG, instance
- TestExportDOT — contains expected DOT syntax
- TestExportMermaid — contains expected Mermaid syntax
- TestBlastRadiusNotFound — nonexistent resource returns error
- TestDependencyGraph{AWS,Azure,GCP} — integration across all providers

## Verification

- [x] `go build ./...` — compiles
- [x] 10/10 new tests pass
- [x] Full test suite passes, zero failures
- [x] `golangci-lint` — 0 issues

### DIFF
--- a/topology/blast_radius.go
+++ b/topology/blast_radius.go
@@ -23,11 +23,15 @@ func blastRadius(ctx context.Context, e *Engine, resourceID string) (*ImpactRepo
 	transitive := findTransitiveDependents(graph, resourceID, direct)
 	broken := findBrokenConnections(graph, resourceID)
 
+	allAffected := collectAffectedIDs(resourceID, direct, transitive)
+	orphaned := findOrphanedResources(graph, allAffected)
+
 	return &ImpactReport{
 		Target:            target,
 		Action:            "delete",
 		DirectlyAffected:  direct,
 		TransitiveImpact:  transitive,
+		OrphanedResources: orphaned,
 		BrokenConnections: broken,
 		Summary: fmt.Sprintf(
 			"deleting %s (%s) affects %d direct and %d transitive resources",
@@ -176,4 +180,135 @@ func findBrokenConnections(graph *DependencyGraph, resourceID string) []Dependen
 	}
 
 	return broken
+}
+
+// collectAffectedIDs gathers the target + direct + transitive IDs into a set.
+func collectAffectedIDs(rootID string, direct, transitive []ResourceRef) map[string]bool {
+	affected := map[string]bool{rootID: true}
+
+	for _, r := range direct {
+		affected[r.ID] = true
+	}
+
+	for _, r := range transitive {
+		affected[r.ID] = true
+	}
+
+	return affected
+}
+
+// findOrphanedResources returns resources outside the affected set whose
+// every parent dependency points into the affected set.
+func findOrphanedResources(graph *DependencyGraph, affected map[string]bool) []ResourceRef {
+	var orphaned []ResourceRef
+
+	for _, r := range graph.Resources {
+		if affected[r.ID] {
+			continue
+		}
+
+		if isOrphaned(graph, r.ID, affected) {
+			orphaned = append(orphaned, r)
+		}
+	}
+
+	return orphaned
+}
+
+// isOrphaned checks whether all of a resource's parent dependencies point
+// into the affected set (meaning it would lose every parent).
+func isOrphaned(graph *DependencyGraph, resourceID string, affected map[string]bool) bool {
+	hasParent := false
+
+	for i := range graph.Dependencies {
+		dep := &graph.Dependencies[i]
+		if dep.From.ID != resourceID {
+			continue
+		}
+
+		if dep.Type != "member-of" && dep.Type != "attached-to" && dep.Type != "belongs-to" {
+			continue
+		}
+
+		hasParent = true
+
+		if !affected[dep.To.ID] {
+			return false
+		}
+	}
+
+	return hasParent
+}
+
+// whatIf previews the impact of an action on a resource.
+func whatIf(ctx context.Context, e *Engine, action, resourceID string) (*ImpactReport, error) {
+	switch action {
+	case "delete":
+		return blastRadius(ctx, e, resourceID)
+	case "stop":
+		return whatIfStop(ctx, e, resourceID)
+	case "disconnect":
+		return whatIfDisconnect(ctx, e, resourceID)
+	default:
+		return nil, cerrors.Newf(cerrors.InvalidArgument, "unsupported action: %s", action)
+	}
+}
+
+// whatIfStop computes the impact of stopping a resource (e.g. an instance).
+// Stopping doesn't remove dependencies but breaks connections that require a running state.
+func whatIfStop(ctx context.Context, e *Engine, resourceID string) (*ImpactReport, error) {
+	graph, err := e.BuildDependencyGraph(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	target, found := findResource(graph, resourceID)
+	if !found {
+		return nil, cerrors.Newf(cerrors.NotFound, "resource %s not found in graph", resourceID)
+	}
+
+	broken := findBrokenConnections(graph, resourceID)
+	direct := findDirectDependents(graph, resourceID)
+
+	return &ImpactReport{
+		Target:            target,
+		Action:            "stop",
+		DirectlyAffected:  direct,
+		BrokenConnections: broken,
+		Summary: fmt.Sprintf(
+			"stopping %s (%s) breaks %d connections",
+			target.ID, target.Type, len(broken),
+		),
+	}, nil
+}
+
+// whatIfDisconnect computes the impact of disconnecting a resource
+// (e.g. removing a peering connection or detaching an internet gateway).
+func whatIfDisconnect(ctx context.Context, e *Engine, resourceID string) (*ImpactReport, error) {
+	graph, err := e.BuildDependencyGraph(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	target, found := findResource(graph, resourceID)
+	if !found {
+		return nil, cerrors.Newf(cerrors.NotFound, "resource %s not found in graph", resourceID)
+	}
+
+	broken := findBrokenConnections(graph, resourceID)
+	direct := findDirectDependents(graph, resourceID)
+	affected := collectAffectedIDs(resourceID, direct, nil)
+	orphaned := findOrphanedResources(graph, affected)
+
+	return &ImpactReport{
+		Target:            target,
+		Action:            "disconnect",
+		DirectlyAffected:  direct,
+		OrphanedResources: orphaned,
+		BrokenConnections: broken,
+		Summary: fmt.Sprintf(
+			"disconnecting %s (%s) breaks %d connections, orphans %d resources",
+			target.ID, target.Type, len(broken), len(orphaned),
+		),
+	}, nil
 }
--- a/topology/blast_radius.go
+++ b/topology/blast_radius.go
@@ -1,0 +1,179 @@
+package topology
+
+import (
+	"context"
+	"fmt"
+
+	cerrors "github.com/stackshy/cloudemu/errors"
+)
+
+// blastRadius computes the impact of removing or modifying the given resource.
+func blastRadius(ctx context.Context, e *Engine, resourceID string) (*ImpactReport, error) {
+	graph, err := e.BuildDependencyGraph(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	target, found := findResource(graph, resourceID)
+	if !found {
+		return nil, cerrors.Newf(cerrors.NotFound, "resource %s not found in graph", resourceID)
+	}
+
+	direct := findDirectDependents(graph, resourceID)
+	transitive := findTransitiveDependents(graph, resourceID, direct)
+	broken := findBrokenConnections(graph, resourceID)
+
+	return &ImpactReport{
+		Target:            target,
+		Action:            "delete",
+		DirectlyAffected:  direct,
+		TransitiveImpact:  transitive,
+		BrokenConnections: broken,
+		Summary: fmt.Sprintf(
+			"deleting %s (%s) affects %d direct and %d transitive resources",
+			target.ID, target.Type, len(direct), len(transitive),
+		),
+	}, nil
+}
+
+// dependsOn returns all resources that the given resource directly depends on.
+func dependsOn(ctx context.Context, e *Engine, resourceID string) ([]ResourceRef, error) {
+	graph, err := e.BuildDependencyGraph(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if _, found := findResource(graph, resourceID); !found {
+		return nil, cerrors.Newf(cerrors.NotFound, "resource %s not found in graph", resourceID)
+	}
+
+	refs := collectDepsFrom(graph, resourceID)
+
+	return refs, nil
+}
+
+// collectDepsFrom returns the To side of every dependency whose From matches the ID.
+func collectDepsFrom(graph *DependencyGraph, resourceID string) []ResourceRef {
+	var refs []ResourceRef
+
+	for i := range graph.Dependencies {
+		dep := &graph.Dependencies[i]
+		if dep.From.ID == resourceID {
+			refs = append(refs, dep.To)
+		}
+	}
+
+	return refs
+}
+
+// dependedBy returns all resources that directly depend on the given resource.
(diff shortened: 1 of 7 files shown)


topology/blast_radius.go
```diff
@@ -0,0 +1,314 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	target, found := findResource(graph, resourceID)
+	if !found {
+		return nil, cerrors.Newf(cerrors.NotFound, "resource %s not found in graph", resourceID)
+	}
+
+	direct := findDirectDependents(graph, resourceID)
```
NitinKumar004, line 22: **`whatIfStop` reports only direct dependents — transitive impact is dropped.** It calls `findDirectDependents` and never `findTransitiveDependents`, so `TransitiveImpact` is always empty. Stop an instance that's a target-group member routed to by a listener: the target group shows as affected, but the listener/LB that transitively depend on it are never surfaced, so a caller concludes the LB is unaffected.

NitinKumar004 (commented):
## Review: Infrastructure Dependency Graph + Blast Radius

Rebased onto current `development` (one trivial test-comment conflict resolved) — now mergeable/clean, builds, and the full `go test ./...` tree is green. Nice, genuinely useful feature (issue #103) and the API is backward-compatible (`topology.New` stays variadic, no existing caller breaks). The BFS correctly guards a `visited` set so peering cycles can't recurse infinitely, and nil optional drivers are skipped. Findings below are marked inline.

### Correctness
The most important is `whatIfStop` dropping transitive impact — combined with DNS node-ID collisions, orphan-detection edge-type logic, and dangling edges, the blast-radius output can be wrong in ways the tests don't catch.

### Tests are too shallow to protect this
`TestOrphanedResources` never inspects `report.OrphanedResources`; `TestWhatIfStop`/`TestWhatIfDisconnect` assert only that `BrokenConnections`/`Summary` are non-empty; `TestBlastRadius` merges direct+transitive and only checks `Contains`, so it can't tell a direct hit from a transitive one. Every correctness finding above passes green. Please pin expected blast-radius/orphan contents.

### Design / reuse
- Inventory walking duplicates `resourcediscovery/` (inline) — the two will drift; and the per-service adders are near-identical copy-paste that a `resourcediscovery`-backed node set + an edge-derivation table would collapse.
- Dependency-type vocabulary (`member-of`/`secured-by`/`peers-with`/…) is ad-hoc string literals spread across `graph_builder.go` with no documented enumeration — worth defining as typed constants.
- The feature (and the pre-existing `CanConnect`/`Resolve`) isn't documented in `docs/services.md`.

Solid foundation — I'd fix the correctness findings + strengthen the tests before merge. Happy to take the transitive-impact and DOT-escaping fixes directly if useful.

topology/graph_builder.go
```diff
@@ -0,0 +1,589 @@
...
+	zoneID string,
+	zoneRef ResourceRef,
+) error {
+	records, err := e.dns.ListRecords(ctx, zoneID)
+	if err != nil {
+		return err
+	}
+
+	for _, rec := range records {
+		recRef := ResourceRef{ID: rec.Name, Type: "dns-record", Name: rec.Type}
```
NitinKumar004, line 298: **DNS record nodes key on bare `rec.Name`, causing node collisions.** A record's identity is `(zoneID, name, type)`. An `A` and `AAAA` record both named `api.example.com` collapse to one node, and the plain-string ID can collide with any other resource whose ID is a bare name (roles/alarms/instance-profiles). `BlastRadius("api.example.com")` then conflates unrelated resources. Key the node on name+type (+zone), like `resolve.go` does.

topology/blast_radius.go
```diff
@@ -0,0 +1,314 @@
...
+func isOrphaned(graph *DependencyGraph, resourceID string, affected map[string]bool) bool {
+	hasParent := false
+
+	for i := range graph.Dependencies {
+		dep := &graph.Dependencies[i]
+		if dep.From.ID != resourceID {
+			continue
+		}
+
+		if dep.Type != "member-of" && dep.Type != "attached-to" && dep.Type != "belongs-to" {
```
NitinKumar004, line 229: **`isOrphaned` only inspects `member-of`/`attached-to`/`belongs-to` edges.** A resource whose real dependencies are `secured-by`/`peers-with`/`routes-to` is never orphaned even when all of them are destroyed; conversely a multi-parent resource is judged orphaned as soon as its `member-of` parent is affected even if a `secured-by` parent survives. The orphan set both under- and over-reports depending on edge-type mix.

topology/graph_builder.go
```diff
@@ -0,0 +1,589 @@
...
+	tgRef ResourceRef,
+) {
+	targets, err := e.loadbalancer.DescribeTargetHealth(ctx, tgARN)
+	if err != nil {
+		return
+	}
+
+	for _, th := range targets {
+		g.Dependencies = append(g.Dependencies, Dependency{
+			From: ResourceRef{ID: th.Target.ID, Type: "instance"},
```
NitinKumar004, line 402: **Dangling edges to nodes that were never added.** Target-health (and ESM at ~503) edges reference an `instance`/`queue` node by ID that only exists if independently enumerated. An IP/cross-account/filtered target `i-123` yields an edge whose `From` node has no `ResourceRef`; `ExportDOT`/`ExportMermaid` emit an edge from an undeclared node and `BlastRadius(tg)` reports `i-123` even though `findResource("i-123")` is not-found.

topology/export.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+// ExportDOT generates a Graphviz DOT representation of the dependency graph.
+func (g *DependencyGraph) ExportDOT() string {
+	var b strings.Builder
+
+	b.WriteString("digraph CloudEmu {\n")
+
+	for _, r := range g.Resources {
+		label := fmt.Sprintf("%s\\n%s", r.Type, r.ID)
```
NitinKumar004, line 15: **DOT label double-escapes the newline.** `fmt.Sprintf("%s\\n%s", ...)` builds a literal backslash-n, then `%q` escapes the backslash again, so Graphviz shows the literal text `vpc\n10.0.0.0/16` on one line instead of two. Build the label without the pre-escaped `\n` (let `%q` produce the escape, or write the label unquoted with a real `\n`).

topology/depgraph.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// BuildDependencyGraph scans all compute, networking, and DNS resources
+// and returns a graph of their relationships.
+func (e *Engine) BuildDependencyGraph(ctx context.Context) (*DependencyGraph, error) {
+	return buildGraph(ctx, e)
+}
+
+// GetDependencyGraph is an alias for BuildDependencyGraph.
+func (e *Engine) GetDependencyGraph(ctx context.Context) (*DependencyGraph, error) {
```
NitinKumar004, line 44: **`GetDependencyGraph` is a pure alias for `BuildDependencyGraph`.** Both call `buildGraph(ctx, e)`. Two public methods doing the same thing is inconsistent with the rest of `topology.Engine` (`CanConnect`/`TraceRoute`/`Resolve` are single verbs) and only the alias's own test uses it. Drop one.

topology/blast_radius.go
```diff
@@ -0,0 +1,314 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	cerrors "github.com/stackshy/cloudemu/errors"
+)
+
+// blastRadius computes the impact of removing or modifying the given resource.
+func blastRadius(ctx context.Context, e *Engine, resourceID string) (*ImpactReport, error) {
```
NitinKumar004, line 11: **The full graph is rebuilt from every driver on every query.** `blastRadius`/`dependsOn`/`dependedBy`/`whatIfStop`/`whatIfDisconnect` each call `BuildDependencyGraph` (re-issuing Describe calls to all drivers) then do linear scans. Iterating impacts over many resources is quadratic-plus driver traffic. Consider exposing build-once/query-many — e.g. `Impact(id)` methods on `*DependencyGraph` (which already owns `ExportDOT`/`ExportMermaid`).

topology/graph_builder.go
```diff
@@ -0,0 +1,589 @@
...
+	if provider == "" {
+		return
+	}
+
+	for i := range g.Resources {
+		g.Resources[i].Provider = provider
+	}
+}
+
+func addVPCs(ctx context.Context, e *Engine, g *DependencyGraph) error {
```
NitinKumar004, line 65: **Resource-inventory walking duplicates `resourcediscovery/`.** `addVPCs`/`addSubnets`/`addInstances`/`addFunctions`/DNS walking re-read the same drivers `resourcediscovery` already walks (and it already handles e.g. the bucket-tag NotFound race the graph builder doesn't). Two inventory walkers over the same drivers will drift. Consider consuming `resourcediscovery.Engine.ListAll` for the node set and having the graph builder own only edge derivation.

Issue #208, opened 2026-06-11
feat: Azure Databricks data-plane SDK-compat (clusters, instance pools, jobs, permissions)


NitinKumar004:
Add the Databricks data-plane REST API (workspace URL, databricks-sdk-go) on top of the ARM workspace management built in #164.

Operations:
- Instance Pools: create, get, list, edit, delete
- Clusters: create, get, list, edit, start, restart, delete/terminate, permanent-delete
- Jobs: create, get, list, update/reset, delete, run-now
- Permissions: get, set, update (clusters, instance-pools, jobs)
- SQL Warehouses: create, get, list, edit, start, stop, delete

Issue #164, opened 2026-05-01
feat: Azure Databricks SDK-compat handler


NitinKumar004:
**Goal**: Add an SDK-compat handler so the real `armdatabricks` client works against cloudemu.

**Operations**:
- Workspaces: CreateOrUpdate, Get, List, ListByResourceGroup, Update, Delete
- VirtualNetworkPeerings: CreateOrUpdate, Get, ListByWorkspace, Delete
- PrivateEndpointConnections: Get, List, Update, Delete

PR #215, opened 2026-06-13
feat: Azure Databricks workspace data plane (compute, jobs, governance, identity)

NitinKumar004:
Builds out the full Azure Databricks **workspace data plane** on top of the ARM workspace management (#164) and the data-plane core (#208), so the real `github.com/databricks/databricks-sdk-go` `WorkspaceClient` works end-to-end against the in-memory backend — a user just points the SDK `Host` at a local URL.

## Design

Two complementary layers, both reachable through one `azureserver.New(...)` server:

1. **Core data plane** — clusters, instance pools, jobs, runs, cluster policies, libraries, permissions. These follow CloudEmu's standard layering: a `databricks/driver.DataPlane` interface → a portable `databricks.DataPlane` wrapper (recording, metrics, rate limiting, error injection, latency) → an in-memory implementation on the shared provider `Mock` (so state is owned by `cloud.Databricks`) → the `server/azure/databricks` ARM/data-plane handler.

2. **Self-contained family handlers** — secrets, tokens, git credentials, repos, DBFS, workspace (notebooks/dirs), SQL warehouses, pipelines, serving endpoints, SCIM, Unity Catalog, UC storage. Each lives in its own package under `server/azure/databricks/<family>/`, owns its in-memory state, and claims a **disjoint `/api` path prefix**. This keeps every family independently buildable/testable and avoids one monolithic god-interface.

**Routing.** Each handler's `Matches()` claims a disjoint prefix (keyed on the resource path segment), so handlers coexist on a single `server.Server` and registration order is unconstrained. They're all registered behind the `DatabricksDataPlane` toggle via a `registerDatabricksDataPlane` helper.

**Wire fidelity.** Each handler speaks the exact Databricks REST shapes the SDK emits — command-style (`/secrets/put`, `/token/create`), REST-by-id (`/repos/{id}`), nested (`/jobs/runs/*`, `/policies/clusters/*`), Unity Catalog (`/api/2.1/unity-catalog/...`), and SCIM v2 envelopes. LRO endpoints (cluster/warehouse/pipeline create, etc.) return terminal states so the SDK's pollers resolve. Errors use the Databricks `{error_code, message}` envelope with correct HTTP statuses.

## Coverage

| Family | Operations |
|--------|-----------|
| Clusters | create, get, list, edit, delete (terminate), permanent-delete, start, restart, resize, pin, unpin, list-node-types, spark-versions, list-zones |
| Instance pools | create, get, list, edit, delete |
| Cluster policies | create, get, list, edit, delete |
| Libraries | install, uninstall, cluster-status, all-cluster-statuses |
| Jobs | create, get, list, update, reset, delete, run-now, submit, cancel-all |
| Runs | get, list, cancel, get-output, delete, repair |
| Permissions | get, set, update |
| Secrets | scopes (create/delete/list), secrets (put/get/delete/list), ACLs (put/get/delete/list) |
| Tokens | create, list, delete |
| Git credentials | create, get, list, update, delete |
| Repos | create, get, list, update, delete |
| DBFS | mkdirs, put, create/add-block/close (block upload), read, get-status, list, delete |
| Workspace | import, export, list, mkdirs, delete, get-status |
| SQL warehouses | create, get, list, edit, start, stop, delete |
| Pipelines | create, get, list, update, delete, start-update, stop |
| Serving endpoints | create, get, list, update-config, delete |
| SCIM | users / groups / service principals — create, get, list, update, delete |
| Unity Catalog | catalogs, schemas, tables — create/get/list/update/delete |
| UC storage | metastores, external locations, storage credentials, volumes — create/get/list/update/delete |

## Testing

- **Per-family roundtrip tests** — each family spins an `httptest` server and drives its full lifecycle through the real `databricks-sdk-go` `WorkspaceClient` (53 SDK tests).
- **One-server integration test** (`TestSDKFullWorkspaceDataPlane`) — registers the entire data plane on a single Azure server and exercises a real create/get/list/lifecycle/delete for every family through one client, proving the handlers coexist and route correctly.
- `go build ./...`, `go test -race ./...`, and `golangci-lint run` all pass.

## Review fixes (in this PR)
- **dbfs `sliceRange`**: guarded an int64 overflow on `read?length=<huge>` that could panic the read handler.
- **repos update**: branch and tag are now mutually exclusive checkout refs (tag no longer clobbers branch).

### DIFF
--- a/server/azure/databricks/dbfs/dbfs.go
+++ b/server/azure/databricks/dbfs/dbfs.go
@@ -1,0 +1,687 @@
+// Package dbfs implements the Databricks DBFS data-plane REST API
+// (the /api/2.0/dbfs/... surface) as a server.Handler, backed by an in-memory
+// filesystem. Point the real github.com/databricks/databricks-sdk-go
+// WorkspaceClient at a server registered with this handler and w.Dbfs.Put,
+// Read, List, Mkdirs, Delete, GetStatus, and the block-upload helpers
+// (WriteFile via create/add-block/close) work end-to-end.
+//
+// Covered endpoints:
+//
+//	POST /api/2.0/dbfs/mkdirs        {path}
+//	POST /api/2.0/dbfs/put           {path, contents(base64), overwrite}
+//	POST /api/2.0/dbfs/create        {path, overwrite} -> {handle}
+//	POST /api/2.0/dbfs/add-block     {handle, data(base64)}
+//	POST /api/2.0/dbfs/close         {handle}
+//	POST /api/2.0/dbfs/delete        {path, recursive}
+//	GET  /api/2.0/dbfs/get-status    ?path=        -> {path, is_dir, file_size}
+//	GET  /api/2.0/dbfs/list          ?path=        -> {files:[...]}
+//	GET  /api/2.0/dbfs/read          ?path=&offset=&length= -> {bytes_read, data}
+package dbfs
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"net/http"
+	"path"
+	"sort"
+	"strconv"
+	"strings"
+	"sync"
+)
+
+const maxBodyBytes = 5 << 20
+
+// Resource segment that this handler claims (3rd path segment after split).
+const resDbfs = "dbfs"
+
+// minSegs is the [api, ver, dbfs, action] segment count required to route.
+const minSegs = 4
+
+// Action path segments.
+const (
+	actMkdirs    = "mkdirs"
+	actPut       = "put"
+	actCreate    = "create"
+	actAddBlock  = "add-block"
+	actClose     = "close"
+	actDelete    = "delete"
+	actGetStatus = "get-status"
+	actList      = "list"
+	actRead      = "read"
+)
+
+// Databricks error codes.
+const (
+	codeNotFound     = "RESOURCE_DOES_NOT_EXIST"
+	codeExists       = "RESOURCE_ALREADY_EXISTS"
+	codeInvalidParam = "INVALID_PARAMETER_VALUE"
+	codeNotFoundPath = "ENDPOINT_NOT_FOUND"
+	codeMalformed    = "MALFORMED_REQUEST"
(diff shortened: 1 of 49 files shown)


server/azure/databricks/dbfs/dbfs.go
```diff
@@ -0,0 +1,687 @@
...
+
+// sliceRange returns data[offset:offset+length], clamped to bounds. A length of
+// 0 means read to end of file.
+func sliceRange(data []byte, offset, length int64) []byte {
+	if offset >= int64(len(data)) {
+		return nil
+	}
+
+	end := int64(len(data))
+	if length > 0 && offset+length < end {
```
NitinKumar004: **Bug (panic / DoS): integer overflow in `sliceRange`.** With a crafted read like `GET /api/2.0/dbfs/read?path=f&offset=1&length=9223372036854775807`, `offset+length` overflows int64 to a negative value, so `offset+length < end` is true and `end` becomes negative — `data[offset:end]` then panics with "slice bounds out of range" and crashes the server goroutine.

Fix: only clamp when the sum doesn't overflow (`want := offset+length; want >= offset && want < end`). (Fixed in a follow-up commit on this PR.)

server/azure/databricks/repos/repos.go
```diff
@@ -0,0 +1,337 @@
...
+
+		return
+	}
+
+	if req.Branch != "" {
+		rp.branch = req.Branch
+	}
+
+	if req.Tag != "" {
+		rp.branch = req.Tag
```
NitinKumar004: **Bug: `tag` overwrites `branch`.** A repo is checked out to a branch *or* a tag (mutually exclusive). Here a PATCH with `{"tag":"v1"}` writes the tag value into the `branch` field, so `GetRepo` then reports `branch:"v1"` — and a request carrying both fields silently lets tag clobber branch. Fix: track `tag` separately and clear the other ref when one is set. (Fixed in a follow-up commit on this PR.)

PR #222, opened 2026-06-14
Add AWS SageMaker In-Memory Emulation And SDK-Compat Server

thzgajendra:
## Summary
Adds **Amazon SageMaker AI** to the emulator — the first machine-learning platform service — across both surfaces (typed Go API and SDK-compat HTTP server), following the existing three-layer pattern (`<svc>/driver` → `providers/aws/sagemaker` → `server/aws/sagemaker`).

**Full-parity coverage (121 ops), all exposed over the SDK-compat HTTP server:**
- **Jobs (all kinds):** Training, Processing, Transform, HyperParameterTuning, AutoML V2, Labeling, Compilation — Create/Describe/List/Stop
- **Inference:** Model, EndpointConfig, Endpoint (+ UpdateEndpoint, weights/capacities), InferenceComponent + runtime `InvokeEndpoint`/`InvokeEndpointAsync`
- **Model registry:** versioned ModelPackage / ModelPackageGroup
- **Studio:** Domain, UserProfile, Space, App
- **Notebooks:** NotebookInstance (+ Start/Stop), lifecycle configs, code repositories
- **HyperPod clusters** (+ node listing/describe)
- **Feature Store:** control plane + online-store runtime (PutRecord/GetRecord/DeleteRecord)
- **Pipelines** (+ executions), Experiments, Trials, and tagging

**Wire protocols:** awsJson1_1 control plane (`X-Amz-Target: SageMaker.*`), restJson1 runtime (`/endpoints/{name}/invocations`), and featurestore-runtime REST (`/FeatureGroup/{name}`). Registered before the catch-all S3 handler.

**Behavioral integration:** auto-metrics to CloudWatch via `SetMonitoring` — `JobsCreated`, `ResourcesCreated` (endpoints/notebooks/clusters/…), `Invocations` and `ModelLatency`.

## Test plan
- [x] `go build ./...`, `go vet`, `golangci-lint run` — clean (0 issues)
- [x] `go test ./...` — 139 packages pass, 0 failures
- [x] Unit tests for every resource family
- [x] Real-SDK round-trip tests for **every family** against `aws-sdk-go-v2/service/sagemaker`, `sagemakerruntime`, and `sagemakerfeaturestoreruntime`
- [x] Manually verified against a locally-running server with the `aws` CLI (training jobs, model→endpoint→invoke, clusters+nodes, notebooks, feature-store put/get)

## Notes
- `dupl` is excluded for `server/aws/sagemaker/` in `.golangci.yml`: the ~100 per-operation decode/encode shims are intentionally uniform boilerplate, not extractable logic.
- Resource Discovery / cost / chaos integration for SageMaker remains a follow-up (engine-level work; the existing Bedrock service also lacks these).
- Next per roadmap: GCP Vertex AI and Azure AI Foundry/ML.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -157,6 +157,12 @@ linters:
           - goconst
           - mnd
         path: _test\.go
+      # The SageMaker SDK-compat handlers are ~100 near-identical decode/encode
+      # shims (one per operation); dupl flags the shared shape, which is
+      # intentional boilerplate, not extractable logic.
+      - linters:
+          - dupl
+        path: server/aws/sagemaker/
       - linters:
           - revive
         text: "exported (.+) should have comment" # TODO fix errors reported by this and remove this line
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -157,12 +157,6 @@ linters:
           - goconst
           - mnd
         path: _test\.go
-      # The SageMaker SDK-compat handlers are ~100 near-identical decode/encode
-      # shims (one per operation); dupl flags the shared shape, which is
-      # intentional boilerplate, not extractable logic.
-      - linters:
-          - dupl
-        path: server/aws/sagemaker/
       - linters:
           - revive
         text: "exported (.+) should have comment" # TODO fix errors reported by this and remove this line
--- a/providers/aws/sagemaker/cluster.go
+++ b/providers/aws/sagemaker/cluster.go
@@ -8,9 +8,18 @@ import (
 	"github.com/stackshy/cloudemu/sagemaker/driver"
 )
 
+// maxClusterInstancesPerGroup bounds the per-group instance count so an
+// oversized (copy-paste or hostile) create can't make a later ListClusterNodes
+// allocate unbounded node structs. The real HyperPod ceiling is in this range.
+const maxClusterInstancesPerGroup = 6758
+
 func (m *Mock) CreateCluster(_ context.Context, cfg driver.ClusterSpec) (*driver.Cluster, error) {
 	if cfg.ClusterName == "" {
 		return nil, errors.New(errors.InvalidArgument, "clusterName is required")
+	}
+
+	if err := validateInstanceGroups(cfg.InstanceGroups); err != nil {
+		return nil, err
 	}
 
 	if m.clusters.Has(cfg.ClusterName) {
@@ -30,9 +39,25 @@ func (m *Mock) CreateCluster(_ context.Context, cfg driver.ClusterSpec) (*driver
 	m.setTags(arn, cfg.Tags)
 	m.emitResourceCreated("Cluster")
 
-	out := *c
+	return cloneCluster(c), nil
+}
 
-	return &out, nil
+// validateInstanceGroups rejects negative or over-cap instance counts before
+// they are stored.
+func validateInstanceGroups(groups []driver.ClusterInstanceGroupSpec) error {
+	for _, g := range groups {
+		if g.InstanceCount < 0 {
+			return errors.Newf(errors.InvalidArgument, "instance group %q has negative InstanceCount", g.GroupName)
+		}
+
+		if g.InstanceCount > maxClusterInstancesPerGroup {
+			return errors.Newf(errors.InvalidArgument,
+				"instance group %q InstanceCount %d exceeds the maximum of %d",
+				g.GroupName, g.InstanceCount, maxClusterInstancesPerGroup)
+		}
+	}
+
+	return nil
 }
 
 func toInstanceGroups(in []driver.ClusterInstanceGroupSpec) []driver.ClusterInstanceGroup {
@@ -44,15 +69,26 @@ func toInstanceGroups(in []driver.ClusterInstanceGroupSpec) []driver.ClusterInst
 	return out
 }
 
+// cloneCluster deep-copies the instance-group and tag slices.
+func cloneCluster(in *driver.Cluster) *driver.Cluster {
+	out := *in
+	if in.InstanceGroups != nil {
+		out.InstanceGroups = make([]driver.ClusterInstanceGroup, len(in.InstanceGroups))
+		copy(out.InstanceGroups, in.InstanceGroups)
+	}
+
+	out.Tags = copyTags(in.Tags)
+
+	return &out
+}
+
 func (m *Mock) DescribeCluster(_ context.Context, name string) (*driver.Cluster, error) {
 	c, ok := m.clusters.Get(name)
 	if !ok {
 		return nil, errors.Newf(errors.NotFound, "cluster %q not found", name)
 	}
 
-	out := *c
-
-	return &out, nil
+	return cloneCluster(c), nil
 }
 
 func (m *Mock) ListClusters(_ context.Context) ([]driver.Cluster, error) {
@@ -60,7 +96,7 @@ func (m *Mock) ListClusters(_ context.Context) ([]driver.Cluster, error) {
 	out := make([]driver.Cluster, 0, len(all))
 
 	for _, v := range all {
-		out = append(out, *v)
+		out = append(out, *cloneCluster(v))
 	}
 
 	return out, nil
@@ -72,12 +108,16 @@ func (m *Mock) UpdateCluster(_ context.Context, name string, groups []driver.Clu
 		return nil, errors.Newf(errors.NotFound, "cluster %q not found", name)
 	}
 
-	c.InstanceGroups = toInstanceGroups(groups)
-	c.Status = driver.ClusterInService
+	if err := validateInstanceGroups(groups); err != nil {
+		return nil, err
+	}
 
-	out := *c
+	updated := *c
+	updated.InstanceGroups = toInstanceGroups(groups)
+	updated.Status = driver.ClusterInService
+	m.clusters.Set(name, &updated)
 
-	return &out, nil
+	return cloneCluster(&updated), nil
 }
 
 func (m *Mock) DeleteCluster(_ context.Context, name string) error {
--- a/providers/aws/sagemaker/cluster.go
+++ b/providers/aws/sagemaker/cluster.go
@@ -1,0 +1,132 @@
+package sagemaker
+
+import (
+	"context"
+	"strconv"
+
+	"github.com/stackshy/cloudemu/errors"
+	"github.com/stackshy/cloudemu/sagemaker/driver"
+)
+
+func (m *Mock) CreateCluster(_ context.Context, cfg driver.ClusterSpec) (*driver.Cluster, error) {
+	if cfg.ClusterName == "" {
+		return nil, errors.New(errors.InvalidArgument, "clusterName is required")
+	}
+
+	if m.clusters.Has(cfg.ClusterName) {
+		return nil, errors.Newf(errors.AlreadyExists, "cluster %q already exists", cfg.ClusterName)
+	}
+
+	arn := m.arn("cluster/" + cfg.ClusterName)
+	c := &driver.Cluster{
+		ClusterName:    cfg.ClusterName,
+		ClusterARN:     arn,
+		Status:         driver.ClusterInService, // synchronous Creating -> InService
+		InstanceGroups: toInstanceGroups(cfg.InstanceGroups),
+		CreationTime:   m.now(),
+		Tags:           copyTags(cfg.Tags),
+	}
+	m.clusters.Set(cfg.ClusterName, c)
+	m.setTags(arn, cfg.Tags)
+
+	out := *c
+
+	return &out, nil
+}
+
+func toInstanceGroups(in []driver.ClusterInstanceGroupSpec) []driver.ClusterInstanceGroup {
+	out := make([]driver.ClusterInstanceGroup, 0, len(in))
+	for _, g := range in {
+		out = append(out, driver.ClusterInstanceGroup(g))
+	}
+
+	return out
+}
+
+func (m *Mock) DescribeCluster(_ context.Context, name string) (*driver.Cluster, error) {
+	c, ok := m.clusters.Get(name)
+	if !ok {
+		return nil, errors.Newf(errors.NotFound, "cluster %q not found", name)
+	}
+
+	out := *c
+
+	return &out, nil
+}
+
+func (m *Mock) ListClusters(_ context.Context) ([]driver.Cluster, error) {
+	all := m.clusters.All()
+	out := make([]driver.Cluster, 0, len(all))
+
+	for _, v := range all {
+		out = append(out, *v)
+	}
+
+	return out, nil
+}
+
+func (m *Mock) UpdateCluster(_ context.Context, name string, groups []driver.ClusterInstanceGroupSpec) (*driver.Cluster, error) {
+	c, ok := m.clusters.Get(name)
+	if !ok {
+		return nil, errors.Newf(errors.NotFound, "cluster %q not found", name)
+	}
+
+	c.InstanceGroups = toInstanceGroups(groups)
+	c.Status = driver.ClusterInService
+
+	out := *c
+
+	return &out, nil
+}
+
+func (m *Mock) DeleteCluster(_ context.Context, name string) error {
+	if !m.clusters.Has(name) {
+		return errors.Newf(errors.NotFound, "cluster %q not found", name)
+	}
+
+	m.clusters.Delete(name)
+
+	return nil
+}
+
+// ListClusterNodes synthesizes one node per instance, deterministically, from
+// the cluster's instance-group configuration.
+func (m *Mock) ListClusterNodes(_ context.Context, clusterName string) ([]driver.ClusterNode, error) {
+	c, ok := m.clusters.Get(clusterName)
+	if !ok {
+		return nil, errors.Newf(errors.NotFound, "cluster %q not found", clusterName)
+	}
+
+	out := make([]driver.ClusterNode, 0)
+
+	for _, g := range c.InstanceGroups {
+		for i := 0; i < g.InstanceCount; i++ {
+			out = append(out, driver.ClusterNode{
+				NodeID:       g.GroupName + "-" + strconv.Itoa(i),
+				GroupName:    g.GroupName,
+				InstanceType: g.InstanceType,
+				Status:       "Running",
+				LaunchTime:   c.CreationTime,
+			})
+		}
+	}
+
+	return out, nil
+}
+
+func (m *Mock) DescribeClusterNode(ctx context.Context, clusterName, nodeID string) (*driver.ClusterNode, error) {
+	nodes, err := m.ListClusterNodes(ctx, clusterName)
+	if err != nil {
+		return nil, err
+	}
+
+	for i := range nodes {
+		if nodes[i].NodeID == nodeID {
+			out := nodes[i]
+
+			return &out, nil
+		}
+	}
+
+	return nil, errors.Newf(errors.NotFound, "cluster node %q not found", nodeID)
+}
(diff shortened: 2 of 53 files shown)


providers/aws/sagemaker/inference.go
```diff
@@ -0,0 +1,326 @@
...
+		return nil, errors.Newf(errors.NotFound, "endpoint %q not found", name)
+	}
+
+	ec, ok := m.endpointConfigs.Get(configName)
+	if !ok {
+		return nil, errors.Newf(errors.InvalidArgument, "endpoint config %q not found", configName)
+	}
+
+	ep.ConfigName = configName
+	ep.Variants = ec.ProductionVariants
```
NitinKumar004: **Shared-slice aliasing corrupts other resources.** `ep.Variants = ec.ProductionVariants` (and `CreateModel` storing `cfg.Containers`, `CreateEndpointConfig` storing `cfg.ProductionVariants`) makes the stored endpoint share the **same backing array** as the endpoint-config. `UpdateEndpointWeightsAndCapacities` then mutates `ep.Variants[i]` in place (L232–235), so a weight/capacity change on one endpoint also alters the source `EndpointConfig` and every other endpoint created from it. `out := *ep` (L213) is only a shallow copy, so the returned value still aliases the stored slice. Deep-copy the variant/container slices both when storing and when returning.

providers/aws/sagemaker/jobs.go
```diff
@@ -0,0 +1,529 @@
...
+// --- shared helpers ---
+
+// stopJob applies mutate to a stored job, or returns NotFound.
+func stopJob[T any](store *memstore.Store[*T], name, kind string, mutate func(*T)) error {
+	job, ok := store.Get(name)
+	if !ok {
+		return errors.Newf(errors.NotFound, "%s %q not found", kind, name)
+	}
+
+	mutate(job)
```
NitinKumar004: **In-place mutation of the stored pointer → data race.** `memstore.Store.Get` returns the **stored** `*T` (its `RLock` is released on return), so `mutate(job)` writes shared state with no lock held. A concurrent `Describe*`/`List*` (which does `out := *job`) racing this write is a real data race under `go test -race`. The same pattern repeats in `Stop*`/`Update*` across `cluster.go`, `pipeline.go`, `registry.go`, `notebook.go`. Follow the bedrock copy-then-`Set` pattern: `j := *job; mutate(&j); store.Set(name, &j)` so readers always see either the old or the new pointer atomically.

providers/aws/sagemaker/registry.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	m.packages.Set(arn, p)
+	m.setTags(arn, cfg.Tags)
+
+	out := *p
+
+	return &out, nil
+}
+
+func (m *Mock) nextPackageVersion(group string) int {
```
NitinKumar004, line 111: **Non-atomic version assignment → ARN collision.** `nextPackageVersion` scans `packages.All()` for the max version, then `CreateModelPackage` builds the ARN from `max+1` and `Set`s it — but the read and the write are not under a single lock. Two concurrent `CreateModelPackage` calls for the same group both compute `N+1`, produce the identical ARN, and the second `Set` silently overwrites the first (a package is lost). Track a per-group version counter mutated atomically, or guard the read-compute-write in one mutex.

providers/aws/sagemaker/notebook.go
```diff
@@ -0,0 +1,214 @@
...
+
+func (m *Mock) StartNotebookInstance(_ context.Context, name string) error {
+	return m.setNotebookStatus(name, driver.NotebookInService)
+}
+
+func (m *Mock) StopNotebookInstance(_ context.Context, name string) error {
+	return m.setNotebookStatus(name, driver.NotebookStopped)
+}
+
+func (m *Mock) setNotebookStatus(name, status string) error {
```
NitinKumar004: **No lifecycle guard on Start/Stop.** `setNotebookStatus` writes `nb.Status` unconditionally, so `StartNotebookInstance` on an already-`InService`/`Pending` notebook (or `Stop` on a `Stopped` one) silently succeeds. Real SageMaker rejects these, so a test asserting the illegal-transition error passes against AWS but not the emulator. (The same unlocked in-place mutation as the stop/update paths applies here too.)

providers/aws/sagemaker/cluster.go
```diff
@@ -0,0 +1,133 @@
...
+func (m *Mock) ListClusterNodes(_ context.Context, clusterName string) ([]driver.ClusterNode, error) {
+	c, ok := m.clusters.Get(clusterName)
+	if !ok {
+		return nil, errors.Newf(errors.NotFound, "cluster %q not found", clusterName)
+	}
+
+	out := make([]driver.ClusterNode, 0)
+
+	for _, g := range c.InstanceGroups {
+		for i := 0; i < g.InstanceCount; i++ {
```
NitinKumar004, line 144: **Unbounded allocation from unvalidated `InstanceCount`.** `ListClusterNodes` does `for i := 0; i < g.InstanceCount; i++ { append(...) }` with no upper bound, and `InstanceCount` comes straight from the create request unvalidated. A create with a huge `InstanceCount` (copy-paste or hostile input) makes a later `ListClusterNodes`/`DescribeClusterNode` allocate billions of structs and OOM the server. Validate `InstanceCount` at create time.

server/aws/sagemaker/inference.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	resp := map[string]any{
+		"ModelName":        model.ModelName,
+		"ModelArn":         model.ModelARN,
+		"ExecutionRoleArn": model.RoleARN,
+		"CreationTime":     epoch(model.CreationTime),
+	}
+	if len(model.Containers) > 0 {
+		resp["PrimaryContainer"] = fromContainer(model.Containers[0])
```
NitinKumar004: **`describeModel` returns both `PrimaryContainer` and `Containers`.** A model created via `PrimaryContainer` should describe back with `PrimaryContainer` set and `Containers` empty; a model created via `Containers` (inference pipeline) is the reverse. Here both are populated from the same single stored slice, so a client that treats a non-empty `Containers` as "this is a pipeline model" misclassifies a plain single-container model. Track which field the model was created with and echo only that one.

providers/aws/sagemaker/featurestore.go
```diff
@@ -0,0 +1,139 @@
...
+
+	for _, fv := range record {
+		if fv.Name == fg.RecordIdentifierName {
+			recordID = fv.Value
+
+			break
+		}
+	}
+
+	if recordID == "" {
```
NitinKumar004: **Empty-string record identifier is rejected as "missing".** `PutRecord` sets `recordID` from the matching feature value, then `if recordID == ""` returns `InvalidArgument` — conflating "identifier feature absent" with "identifier present but the empty string". A record whose identifier value is legitimately `""` can never be written (and `GetRecord` then always 404s). Check for **presence** of the identifier feature in the record, not a non-empty value.

providers/aws/sagemaker/sagemaker.go
```diff
@@ -0,0 +1,141 @@
+// Package sagemaker provides an in-memory mock implementation of Amazon
+// SageMaker AI: training/processing/transform/tuning/AutoML/labeling/
+// compilation jobs, the model-and-endpoint inference stack, the model
+// registry, Studio, notebook instances, HyperPod clusters, Feature Store and
+// pipelines. Asynchronous jobs complete synchronously (straight to a terminal
```
NitinKumar004, line 5: **Missing portable API layer (3-layer architecture).** Every other service ships a `<service>/<service>.go` portable wrapper (e.g. `bedrock/bedrock.go`) that wraps the driver with the cross-cutting concerns — recorder, metrics, rate limiting, error injection, latency. SageMaker has only `sagemaker/driver/` + this provider + the server handler; the middle portable layer is absent (there is no `sagemaker/sagemaker.go`). Consequences: SageMaker gets none of metrics/recording/rate-limit/error-injection/latency that all 15 other services get, and the rates added to `cost/cost.go` have no portable call-path to record against. Per CLAUDE.md's 3-layer design this layer should exist. (The "mirrors the Bedrock mock" doc is misleading — Bedrock also has the portable wrapper.)

server/aws/sagemaker/errors.go
```diff
@@ -0,0 +1,26 @@
...
+// (written by wire.WriteJSONError) to pick the typed error.
+func writeDriverError(w http.ResponseWriter, err error) {
+	switch {
+	case cerrors.IsNotFound(err):
+		wire.WriteJSONError(w, http.StatusBadRequest, "ResourceNotFound", err.Error())
+	case cerrors.IsAlreadyExists(err):
+		wire.WriteJSONError(w, http.StatusBadRequest, "ResourceInUse", err.Error())
+	case cerrors.IsInvalidArgument(err):
+		wire.WriteJSONError(w, http.StatusBadRequest, "ValidationException", err.Error())
+	case cerrors.IsFailedPrecondition(err):
```
NitinKumar004, line 21: **Verify `ValidationException` deserializes as a typed SDK error.** `IsInvalidArgument`/`IsFailedPrecondition` map to `"ValidationException"`. The other mappings (`ResourceNotFound`, `ResourceInUse`) match shapes the `service/sagemaker` SDK models, but I couldn't confirm SageMaker models a `ValidationException` for these operations. If it doesn't, the awsJson1_1 deserializer won't match it against any modeled error and the client gets a generic `*smithy.GenericAPIError` instead of a typed exception — so `errors.As(err, &types.…{})` branches won't fire. Please confirm against the SDK's modeled error shapes and add a roundtrip assertion on the returned error type.

.golangci.yml
```diff
@@ -157,6 +157,12 @@ linters:
           - goconst
           - mnd
         path: _test\.go
+      # The SageMaker SDK-compat handlers are ~100 near-identical decode/encode
+      # shims (one per operation); dupl flags the shared shape, which is
+      # intentional boilerplate, not extractable logic.
+      - linters:
+          - dupl
+        path: server/aws/sagemaker/
```
NitinKumar004: **`dupl` disabled for the whole `server/aws/sagemaker/` dir.** No other SDK-compat handler dir (bedrock, eks, rds — similar decode/encode boilerplate) carves out `dupl`. This silences duplication detection across the whole package permanently, not just the per-op shims — so genuinely extractable duplication (e.g. the ~9 near-identical `List*` handlers, the 3-way job-router split in `jobs_more.go`) won't be flagged going forward. Prefer extracting the shared shapes (a generic `writeSummaries`/projector helper) and keeping `dupl` enabled, or scope the exclusion to the specific shim files that truly can't be deduped.

NitinKumar004:
## Review summary

Thanks for this — it's a big, well-structured addition and the real-SDK round-trip tests are great to see. I left 10 inline comments; here's the overall picture.

**The one thing to address before merge** is a systemic pattern in the provider: it mutates the pointer returned by `memstore.Store.Get` in place. `Get` returns the **stored** `*T` and releases its lock on return, so every `Stop*`/`Update*`/`set*Status` path (jobs, cluster, pipeline, registry, notebook, endpoints) writes shared state with no lock held — a real data race against concurrent `Describe*`/`List*` under `go test -race`. Relatedly, stored slice fields (`Variants`, `Containers`, `ProductionVariants`) are aliased rather than copied, so a weight/capacity update on one endpoint corrupts the source `EndpointConfig` and sibling endpoints, and `out := *x` shallow copies still share those backing arrays. The fix is the existing bedrock pattern: copy the struct, mutate the copy, `Set` it back, and deep-copy slices on store/return.

**Correctness edges** worth fixing: non-atomic model-package version → ARN collision; notebook Start/Stop accepting illegal transitions; `ListClusterNodes` allocating an unvalidated `InstanceCount`; `describeModel` returning both `PrimaryContainer` and `Containers`; FeatureStore `PutRecord` rejecting a legitimately empty identifier value.

**Architecture / consistency**: there's no portable API layer (`sagemaker/sagemaker.go`) like every other service has — so SageMaker gets none of the metrics/recording/rate-limit/error-injection/latency wrapping, and the `cost/cost.go` rates have no call-path to record against. Worth adding to match the 3-layer design. Also: please double-check `ValidationException` is a modeled SDK error (otherwise those errors won't deserialize as typed), and reconsider disabling `dupl` for the whole `server/aws/sagemaker/` dir — extracting the shared list/route shapes would let it stay on.

None of this detracts from the breadth here; the core resources round-trip well. Happy to re-review once the store-mutation pattern is addressed.

server/aws/sagemaker/errors.go
thzgajendra: Confirmed against the SDK's modeled error shapes and verified by round-trip test.

`service/sagemaker/types` models exactly: `ConflictException`, `ResourceInUse`, `ResourceLimitExceeded`, `ResourceNotFound` — and **not** `ValidationException`. Key point: typed-error deserialization is **per-operation** — the awsJson1_1 deserializer only matches an `X-Amzn-ErrorType` to a typed error that the *specific* operation models; everything else surfaces as `*smithy.GenericAPIError`.

So:
- `NotFound → ResourceNotFound` is modeled and now asserted (`errors.As(&smtypes.ResourceNotFound{})` fires) in `TestSDKDescribeMissingTrainingJob`.
- `AlreadyExists → ResourceInUse` is modeled, but only on the ops that declare it (e.g. not `CreateModel`, which is why an assertion there returns generic).
- `InvalidArgument/FailedPrecondition → ValidationException` is intentionally **generic** — SageMaker does not model `ValidationException` per-op, so real SageMaker clients also receive a generic error for it. Mapping it to a typed exception would be *less* faithful, not more.

No code change beyond adding the typed-error round-trip assertion; the mappings are correct as-is. Left a comment in the test documenting the per-operation matching behavior.

thzgajendra:
Thanks for the thorough review — addressed all ten in commit c9e1a68. Summary:

| # | Comment | Resolution |
|---|---------|-----------|
| 1 | Shared-slice aliasing | Deep-copy ProductionVariants/Containers on store **and** return (`copyVariants`/`copyContainers` + `clone*` helpers); updates operate on copies. |
| 2 | In-place mutation → data race | Copy-then-Set everywhere (`stopJob` generic helper + UpdateEndpoint/WeightsAndCapacities, UpdateCluster, UpdatePipeline, StopPipelineExecution, UpdateModelPackage, notebook transitions). |
| 3 | Non-atomic package version → ARN collision | `nextPackageVersion`+`Set` now guarded by a `pkgMu` mutex. |
| 4 | No lifecycle guard on notebook Start/Stop | `transitionNotebook` rejects illegal transitions with `FailedPrecondition`. |
| 5 | Unbounded `InstanceCount` | Validated at CreateCluster/UpdateCluster against a cap (negative rejected too). |
| 6 | DescribeModel returns both container fields | Track `Pipeline` (created via Containers vs PrimaryContainer); echo only the one used. |
| 7 | Empty-string record identifier rejected | PutRecord now checks **presence** of the identifier feature, not a non-empty value. |
| 8 | Missing portable Layer-1 | Added `sagemaker/sagemaker.go` wrapping `driver.Service` with recorder/metrics/rate-limit/error-injection/latency (mirrors `bedrock/bedrock.go`); fixed the misleading "mirrors the Bedrock mock" doc. |
| 9 | Verify `ValidationException` typed error | Replied inline — per-op matching; ResourceNotFound modeled+asserted, ValidationException is generic-by-design (matches real AWS). |
| 10 | Blanket `dupl` exclusion | Removed it; extracted a generic `writeSummaries` helper for the ~25 `List*` handlers and kept `dupl` enabled, annotating only the residual per-op create/describe/router shims. |

`go build ./... && go test ./...` green (153 packages, 0 failures); `golangci-lint` clean.

providers/aws/sagemaker/sagemaker.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// copyContainers returns a deep copy of a container-definition slice.
+func copyContainers(in []driver.ContainerDefinition) []driver.ContainerDefinition {
+	if in == nil {
+		return nil
+	}
+
+	out := make([]driver.ContainerDefinition, len(in))
+	copy(out, in)
```
NitinKumar004: **Residual aliasing — `Environment` map not deep-copied.** Nice work moving everything to clone-then-`Set`; this one spot still aliases. `copyContainers` does `copy(out, in)`, which copies each `ContainerDefinition` struct but leaves its `Environment map[string]string` pointing at the **same backing map**. So mutating `model.Containers[i].Environment[k]` on a value returned by `DescribeModel`/`ListModels` (or on the caller's original `cfg.Containers`) still corrupts the stored model — the same class as the endpoint-variant bug, just one level deeper. `copyStrMap` already exists; apply it per element:

```go
out := make([]driver.ContainerDefinition, len(in))
for i, c := range in {
    c.Environment = copyStrMap(c.Environment)
    out[i] = c
}
return out
```

NitinKumar004:
## Re-review (after "Address SageMaker Review Feedback")

Excellent turnaround — this is a thorough response. Build, full tests, and `golangci-lint` are all clean, and the `dupl` carve-out is gone (re-enabled dir-wide, suppressed only with targeted per-function `//nolint:dupl`). Status of the earlier findings:

**Resolved (8):**
- Endpoint/config/model slice aliasing → clone-then-`Set` with `cloneEndpoint`/`cloneEndpointConfig`/`cloneModel`; `UpdateEndpointWeightsAndCapacities` now copies the variant slice before mutating. ✅
- In-place store-pointer mutation across jobs/cluster/pipeline/registry/notebook → all copy-then-`Set` now (the systemic race is gone). ✅
- `nextPackageVersion` → serialized under `pkgMu`, no more version/ARN collision. ✅
- Notebook lifecycle → `transitionNotebook` enforces allowed source states. ✅
- `ListClusterNodes` unbounded alloc → `validateInstanceGroups` bounds `InstanceCount` at create and update. ✅
- `describeModel` → switches on a new `Pipeline` flag, emits `PrimaryContainer` xor `Containers`. ✅
- FeatureStore `PutRecord` → uses a `found` bool, so an empty-string identifier value is accepted. ✅
- Portable layer → `sagemaker/sagemaker.go` + wrappers added, full `driver.Service`, bedrock-style `do()` pipeline. ✅ (And to be clear: it not being auto-wired into the provider factory is **correct** — `bedrock.NewBedrock` isn't either; the factory exposes the driver-backed mock for every service. No change needed.)

**One residual** (left an inline comment): `copyContainers` deep-copies the container slice but not each container's `Environment` map — that map is still aliased, so the original aliasing class isn't fully closed for model environments. One-line fix with the existing `copyStrMap`.

**Minor / optional:**
- `ValidationException`: you documented (in a test comment) that it deserializes as a generic error — good that it's confirmed. A tiny roundtrip assertion on an actual validation error would lock it in.
- `StartNotebookInstance` allows starting from `Failed`; real SageMaker only allows `Stopped`. Edge-case, not blocking.

Really solid revision — once the `Environment` map copy is in, this looks good to me.

providers/aws/sagemaker/sagemaker.go
thzgajendra: Good catch — fixed in 649ac88. `copyContainers` now deep-copies each container's `Environment` via the existing `copyStrMap`, so neither a `DescribeModel`/`ListModels` result nor the caller's original `cfg.Containers` aliases the stored map.

Also addressed the two optional notes: added a typed-error round-trip assertion + comment for the per-operation `ValidationException`/`ResourceNotFound` matching, and restricted `StartNotebookInstance` to the `Stopped` source state only (dropped `Failed`), matching real SageMaker (1818089).

NitinKumar004 (approved):
LGTM ✅

Every item from both review rounds is resolved and verified on the latest head (`28b1943`):
- The systemic store-pointer mutation race and slice/`Environment`-map aliasing are gone (clone-then-`Set` throughout).
- `nextPackageVersion` serialized; notebook lifecycle guards (Start only from `Stopped`); `InstanceCount` bounded; `describeModel` PrimaryContainer-xor-Containers; FeatureStore empty-identifier handling.
- Portable layer added (`sagemaker/sagemaker.go` + wrappers), matching the bedrock convention.
- `dupl` carve-out removed; `ValidationException`-as-generic pinned by a test.

`go build`, full SageMaker tests, and `golangci-lint` are all clean. Nice work on the breadth and the turnaround.

PR #231, opened 2026-06-14
Add GCP Vertex AI Emulation (Driver, Provider, SDK-Compat REST Server)

thzgajendra:
## Summary

Adds full Google Cloud Vertex AI (`aiplatform.googleapis.com`) emulation across all three layers, with behavioral integration matching every other service.

- **Driver + in-memory provider** — the complete Vertex AI surface (~128 operations) spanning datasets, model registry (versions/evaluations), endpoints + online prediction, custom/batch/HPO/tuning jobs, training & pipeline jobs, the Gemini `generateContent` runtime + cached contents, Feature Store (legacy + Registry + online stores/views), Vector Search, ML metadata, tensorboards, schedules, and notebook runtimes.
- **Portable Layer-1 wrapper** (`vertexai/vertexai.go`) — wraps the driver with recorder, metrics, rate limiting, error injection and latency simulation through a single `do()` pipeline, exactly like `bedrock`/`sagemaker`.
- **SDK-compat REST server** — now serves **all 20 collections** (previously only 5 core ones). Real `google.golang.org/api/aiplatform/v1` clients (or any REST caller) configured with a custom endpoint hit it the same way they hit `{location}-aiplatform.googleapis.com`. Control-plane mutations return done `google.longrunning.Operation`s; job-family creates are synchronous.
- **Cross-cutting parity** — `chaos.WrapVertexAI` injects failures on submission/runtime calls; default Vertex cost rates added.

## Provider correctness (SageMaker review lessons applied proactively)

- Deep-copy slices/maps on store and return (`clone.go`); no shared backing arrays or maps.
- Copy-then-Set on every mutator — stored pointers are never mutated in place (fixes the `UndeployModel` traffic-split aliasing class of bug).
- Lifecycle guards: schedule Pause/Resume and notebook-runtime Start/Stop reject illegal transitions (`FailedPrecondition`).
- `FindNeighbors` bounds its allocation against unvalidated/hostile counts.
- `ServeHTTP` dispatches via a route table to stay within cyclomatic-complexity limits.

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` — all green
- [x] `golangci-lint run` clean on every touched package (provider, Layer-1, server, chaos, cost)
- [x] SDK-compat roundtrip tests start a real `httptest` server and exercise every new family over HTTP (models/versions/evaluations, jobs, pipelines, tuning, cached contents, Feature Store + online read/write, Feature Registry, online stores/views + fetch, Vector Search deploy + findNeighbors, metadata/tensorboard/schedule/notebook lifecycle)
- [x] Layer-1 wrapper tests cover recording, error injection, latency, result forwarding
- [x] Chaos wrapper tests cover injected-failure and pass-through paths

### DIFF
--- a/providers/gcp/vertexai/datasets.go
+++ b/providers/gcp/vertexai/datasets.go
@@ -28,16 +28,14 @@ func (m *Mock) CreateDataset(_ context.Context, cfg driver.DatasetConfig) (*driv
 		Name:              name,
 		DisplayName:       cfg.DisplayName,
 		MetadataSchemaURI: cfg.MetadataSchemaURI,
-		Labels:            cfg.Labels,
+		Labels:            copyLabels(cfg.Labels),
 		CreateTime:        now,
 		UpdateTime:        now,
 	}
 	m.datasets.Set(name, ds)
 	m.emitMetric("dataset/count", 1, map[string]string{"location": orLocation(cfg.Location)})
 
-	out := *ds
-
-	return m.doneOp(cfg.Location, name), &out, nil
+	return m.doneOp(cfg.Location, name), cloneDataset(ds), nil
 }
 
 func (m *Mock) GetDataset(_ context.Context, name string) (*driver.Dataset, error) {
@@ -46,9 +44,7 @@ func (m *Mock) GetDataset(_ context.Context, name string) (*driver.Dataset, erro
 		return nil, errors.Newf(errors.NotFound, "dataset %q not found", name)
 	}
 
-	out := *ds
-
-	return &out, nil
+	return cloneDataset(ds), nil
 }
 
 func (m *Mock) ListDatasets(_ context.Context, location string) ([]driver.Dataset, error) {
@@ -56,7 +52,7 @@ func (m *Mock) ListDatasets(_ context.Context, location string) ([]driver.Datase
 
 	for _, ds := range m.datasets.All() {
 		if location == "" || locationOf(ds.Name) == location {
-			out = append(out, *ds)
+			out = append(out, *cloneDataset(ds))
 		}
 	}
 
@@ -69,14 +65,16 @@ func (m *Mock) PatchDataset(_ context.Context, name, displayName string) (*drive
 		return nil, errors.Newf(errors.NotFound, "dataset %q not found", name)
 	}
 
+	// Copy-then-Set: never mutate the stored pointer in place.
+	updated := *ds
 	if displayName != "" {
-		ds.DisplayName = displayName
+		updated.DisplayName = displayName
 	}
 
-	ds.UpdateTime = m.now()
-	out := *ds
+	updated.UpdateTime = m.now()
+	m.datasets.Set(name, &updated)
 
-	return &out, nil
+	return cloneDataset(&updated), nil
 }
 
 func (m *Mock) DeleteDataset(_ context.Context, name string) (*driver.Operation, error) {
--- a/providers/gcp/vertexai/datasets.go
+++ b/providers/gcp/vertexai/datasets.go
@@ -1,0 +1,106 @@
+package vertexai
+
+import (
+	"context"
+	"strings"
+
+	"github.com/stackshy/cloudemu/errors"
+	"github.com/stackshy/cloudemu/vertexai/driver"
+)
+
+// locationOf extracts the location segment from a resource name, or the
+// default when absent.
+func locationOf(name string) string {
+	parts := strings.Split(name, "/")
+	for i, p := range parts {
+		if p == "locations" && i+1 < len(parts) {
+			return parts[i+1]
+		}
+	}
+
+	return defaultLocation
+}
+
+func (m *Mock) CreateDataset(_ context.Context, cfg driver.DatasetConfig) (*driver.Operation, *driver.Dataset, error) {
+	now := m.now()
+	name := m.resName(cfg.Location, "datasets", m.newID())
+	ds := &driver.Dataset{
+		Name:              name,
+		DisplayName:       cfg.DisplayName,
+		MetadataSchemaURI: cfg.MetadataSchemaURI,
+		Labels:            cfg.Labels,
+		CreateTime:        now,
+		UpdateTime:        now,
+	}
+	m.datasets.Set(name, ds)
+	m.emitMetric("dataset/count", 1, map[string]string{"location": orLocation(cfg.Location)})
+
+	out := *ds
+
+	return m.doneOp(cfg.Location, name), &out, nil
+}
+
+func (m *Mock) GetDataset(_ context.Context, name string) (*driver.Dataset, error) {
+	ds, ok := m.datasets.Get(name)
+	if !ok {
+		return nil, errors.Newf(errors.NotFound, "dataset %q not found", name)
+	}
+
+	out := *ds
+
+	return &out, nil
+}
+
+func (m *Mock) ListDatasets(_ context.Context, location string) ([]driver.Dataset, error) {
+	out := make([]driver.Dataset, 0)
+
+	for _, ds := range m.datasets.All() {
+		if location == "" || locationOf(ds.Name) == location {
+			out = append(out, *ds)
+		}
+	}
+
+	return out, nil
+}
+
+func (m *Mock) PatchDataset(_ context.Context, name, displayName string) (*driver.Dataset, error) {
+	ds, ok := m.datasets.Get(name)
+	if !ok {
+		return nil, errors.Newf(errors.NotFound, "dataset %q not found", name)
+	}
+
+	if displayName != "" {
+		ds.DisplayName = displayName
+	}
+
+	ds.UpdateTime = m.now()
+	out := *ds
+
+	return &out, nil
+}
+
+func (m *Mock) DeleteDataset(_ context.Context, name string) (*driver.Operation, error) {
+	if !m.datasets.Has(name) {
+		return nil, errors.Newf(errors.NotFound, "dataset %q not found", name)
+	}
+
+	m.datasets.Delete(name)
+
+	return m.doneOp(locationOf(name), name), nil
+}
+
+func (m *Mock) ImportData(_ context.Context, name, _ string) (*driver.Operation, error) {
+	if !m.datasets.Has(name) {
+		return nil, errors.Newf(errors.NotFound, "dataset %q not found", name)
+	}
+
+	return m.doneOp(locationOf(name), name), nil
+}
+
+func (m *Mock) ExportData(_ context.Context, name, _ string) (*driver.Operation, error) {
+	if !m.datasets.Has(name) {
+		return nil, errors.Newf(errors.NotFound, "dataset %q not found", name)
+	}
+
+	return m.doneOp(locationOf(name), name), nil
+}
--- a/providers/gcp/vertexai/endpoints.go
+++ b/providers/gcp/vertexai/endpoints.go
@@ -1,0 +1,143 @@
+package vertexai
+
+import (
+	"context"
+
+	"github.com/stackshy/cloudemu/errors"
+	"github.com/stackshy/cloudemu/vertexai/driver"
+)
+
+func (m *Mock) CreateEndpoint(_ context.Context, cfg driver.EndpointConfig) (*driver.Operation, *driver.Endpoint, error) {
+	now := m.now()
+	name := m.resName(cfg.Location, "endpoints", m.newID())
+	ep := &driver.Endpoint{
+		Name:         name,
+		DisplayName:  cfg.DisplayName,
+		Description:  cfg.Description,
+		TrafficSplit: map[string]int{},
+		Labels:       cfg.Labels,
+		CreateTime:   now,
+		UpdateTime:   now,
+	}
+	m.endpoints.Set(name, ep)
+	m.emitMetric("endpoint/count", 1, map[string]string{"location": orLocation(cfg.Location)})
+
+	out := *ep
+
+	return m.doneOp(cfg.Location, name), &out, nil
+}
+
+func (m *Mock) GetEndpoint(_ context.Context, name string) (*driver.Endpoint, error) {
+	ep, ok := m.endpoints.Get(name)
+	if !ok {
+		return nil, errors.Newf(errors.NotFound, "endpoint %q not found", name)
+	}
+
+	out := *ep
+
+	return &out, nil
+}
+
+func (m *Mock) ListEndpoints(_ context.Context, location string) ([]driver.Endpoint, error) {
+	out := make([]driver.Endpoint, 0)
+
+	for _, ep := range m.endpoints.All() {
+		if location == "" || locationOf(ep.Name) == location {
+			out = append(out, *ep)
+		}
+	}
+
+	return out, nil
+}
+
+func (m *Mock) DeleteEndpoint(_ context.Context, name string) (*driver.Operation, error) {
+	if !m.endpoints.Has(name) {
+		return nil, errors.Newf(errors.NotFound, "endpoint %q not found", name)
+	}
+
+	m.endpoints.Delete(name)
+
+	return m.doneOp(locationOf(name), name), nil
+}
+
+//nolint:gocritic // dm matches the driver signature; copied on entry.
+func (m *Mock) DeployModel(_ context.Context, endpoint string, dm driver.DeployedModel) (*driver.Operation, *driver.Endpoint, error) {
+	ep, ok := m.endpoints.Get(endpoint)
+	if !ok {
+		return nil, nil, errors.Newf(errors.NotFound, "endpoint %q not found", endpoint)
+	}
+
+	if dm.ID == "" {
+		dm.ID = m.newID()
+	}
+
+	updated := *ep
+	updated.DeployedModels = append(append([]driver.DeployedModel{}, ep.DeployedModels...), dm)
+	updated.TrafficSplit = map[string]int{dm.ID: 100}
+	updated.UpdateTime = m.now()
+	m.endpoints.Set(endpoint, &updated)
+
+	out := updated
+
+	return m.doneOp(locationOf(endpoint), endpoint), &out, nil
+}
+
+func (m *Mock) UndeployModel(_ context.Context, endpoint, deployedModelID string) (*driver.Operation, *driver.Endpoint, error) {
+	ep, ok := m.endpoints.Get(endpoint)
+	if !ok {
+		return nil, nil, errors.Newf(errors.NotFound, "endpoint %q not found", endpoint)
+	}
+
+	kept := make([]driver.DeployedModel, 0, len(ep.DeployedModels))
+
+	for _, d := range ep.DeployedModels {
+		if d.ID != deployedModelID {
+			kept = append(kept, d)
+		}
+	}
+
+	updated := *ep
+	updated.DeployedModels = kept
+	delete(updated.TrafficSplit, deployedModelID)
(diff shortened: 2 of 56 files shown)


providers/gcp/vertexai/models.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	return out, nil
+}
+
+func (m *Mock) PatchModel(_ context.Context, name, displayName, description string) (*driver.Model, error) {
+	model, ok := m.models.Get(name)
+	if !ok {
+		return nil, errors.Newf(errors.NotFound, "model %q not found", name)
+	}
```
NitinKumar004, line 60: **In-place mutation of the stored pointer (regression of the #222 finding).** `m.models.Get` returns the **live** stored `*Model`, and `PatchModel` writes `model.DisplayName` / `model.Description` / `model.UpdateTime` directly on it (L67–74) before copying to `out`. That's a data race against a concurrent `GetModel`/`ListModels` (under `-race`) and a torn read even single-threaded. `PatchDataset` (datasets.go:73) has the identical bug. Other mutators here already use copy-then-`Set` — do the same: `updated := *model; updated.DisplayName = …; m.models.Set(name, &updated)`.

providers/gcp/vertexai/endpoints.go
```diff
@@ -0,0 +1,137 @@
...
+		dm.ID = m.newID()
+	}
+
+	// Deep-copy working set so the stored endpoint is never mutated in place.
+	updated := cloneEndpoint(ep)
+	updated.DeployedModels = append(updated.DeployedModels, dm)
+	updated.TrafficSplit = map[string]int{dm.ID: 100}
+	updated.UpdateTime = m.now()
+	m.endpoints.Set(endpoint, updated)
+
```
NitinKumar004, line 79: **`DeployModel` overwrites the whole `TrafficSplit`.** `updated.TrafficSplit = map[string]int{dm.ID: 100}` discards traffic for every previously-deployed model. Deploy A then B → `DeployedModels` correctly holds [A, B] but `TrafficSplit` becomes `{B:100}`, so A is deployed yet has no split entry. Real Vertex multi-model endpoints keep both. Merge into the existing split (and rebalance) instead of replacing it.

providers/gcp/vertexai/endpoints.go
```diff
@@ -0,0 +1,137 @@
...
+			kept = append(kept, d)
+		}
+	}
+
+	updated.DeployedModels = kept
+	delete(updated.TrafficSplit, deployedModelID)
+	updated.UpdateTime = m.now()
+	m.endpoints.Set(endpoint, updated)
+
+	return m.doneOp(locationOf(endpoint), endpoint), cloneEndpoint(updated), nil
```
NitinKumar004, line 128: **`UndeployModel` mutates a shared map in place.** `updated := *ep` is a shallow copy, so `delete(updated.TrafficSplit, deployedModelID)` deletes from the **same map** the stored endpoint (and any `*Endpoint` previously returned by `GetEndpoint`, which also shallow-copies) point at. Concurrent access races; sequentially it silently mutates earlier snapshots. Clone the map before deleting: `m := maps.Clone(ep.TrafficSplit); delete(m, id); updated.TrafficSplit = m`.

providers/gcp/vertexai/datasets.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func (m *Mock) CreateDataset(_ context.Context, cfg driver.DatasetConfig) (*driver.Operation, *driver.Dataset, error) {
+	now := m.now()
+	name := m.resName(cfg.Location, "datasets", m.newID())
+	ds := &driver.Dataset{
+		Name:              name,
+		DisplayName:       cfg.DisplayName,
+		MetadataSchemaURI: cfg.MetadataSchemaURI,
+		Labels:            copyLabels(cfg.Labels),
```
NitinKumar004, line 31: **Caller's `Labels` map is stored by reference.** `Labels: cfg.Labels` aliases the caller's map into the store (same at endpoints.go:18 and models.go:23). After create, if the caller mutates or reuses `cfg.Labels`, the stored resource's labels change underneath it, and a concurrent `List*` racing that write is a data race. Clone on store: `Labels: maps.Clone(cfg.Labels)`.

providers/gcp/vertexai/featurestore.go
```diff
@@ -0,0 +1,337 @@
...
+	out := *f
+
+	return &out, nil
+}
+
+func (m *Mock) ListFeatures(_ context.Context, parent string) ([]driver.Feature, error) {
+	out := make([]driver.Feature, 0)
+
+	for k, f := range m.features.All() {
+		if len(k) > len(parent) && k[:len(parent)] == parent {
```
NitinKumar004: **Child listing uses a bare prefix match without a `/` delimiter.** `k[:len(parent)] == parent` (also L183, L287) treats the parent name as a raw string prefix, so siblings whose IDs share a prefix leak across listings — e.g. `ListFeatures(.../featureGroups/fg1)` also returns features of `.../featureGroups/fg10`. Match `parent + "/"` (like `ListOperations` does) so only true children match. Reachable with caller-supplied IDs.

providers/gcp/vertexai/genai.go
```diff
@@ -0,0 +1,200 @@
...
+		return nil, errors.New(errors.InvalidArgument, "model is required")
+	}
+
+	name := m.resName(cfg.Location, "cachedContents", m.newID())
+	cc := &driver.CachedContent{
+		Name:        name,
+		Model:       cfg.Model,
+		DisplayName: cfg.DisplayName,
+		CreateTime:  m.now(),
+		ExpireTime:  m.now(),
```
NitinKumar004: **`CreateCachedContent` ignores `TTLSeconds` → created already-expired.** `ExpireTime: m.now()` equals `CreateTime`, and `cfg.TTLSeconds` (and `Contents`/`SystemInstruction`) are dropped. A client creating a cache with `TTLSeconds: 3600` reads back `expireTime == now`, so any `expireTime > now` validity check treats it as immediately expired. Set `ExpireTime = now + TTL` and persist the contents.

providers/gcp/vertexai/vertexai.go
```diff
@@ -0,0 +1,154 @@
+// Package vertexai provides an in-memory mock implementation of Google Cloud
```
NitinKumar004, line 1: **Missing portable API layer — the exact gap #222 was sent back for.** There's no `vertexai/vertexai.go` implementing `driver.VertexAI` with the recorder / metrics / rate-limit / error-injection / latency wrappers (the `do()` pipeline). `server/gcp/vertexai/handler.New` wraps the raw provider `*Mock` directly, so Vertex gets none of the cross-cutting concerns every other service (and the `sagemaker/sagemaker.go` you just added in #222) provides. Please add the portable wrapper package to match the 3-layer design.

server/gcp/vertexai/responses.go
```diff
@@ -0,0 +1,47 @@
...
+func writeCErr(w http.ResponseWriter, err error) {
+	gcprest.WriteCErr(w, err)
+}
+
+func decode(w http.ResponseWriter, r *http.Request, v any) bool {
+	return gcprest.DecodeJSON(w, r, v)
+}
+
+// writeOp writes a long-running Operation as the GCP JSON envelope.
+func writeOp(w http.ResponseWriter, op *driver.Operation) {
```
NitinKumar004, line 27: **Done-operation `response` only carries `{name}` and no `@type`.** For `done:true` LROs the aiplatform SDK unmarshals `op.Response` into the typed result (e.g. `Model` / `DeployModelResponse`), so a caller reading `deployedModel.id` or `model.versionId` off the operation result gets empty values. Populate `response` with the actual resource object (and the `@type`).

Related: the LRO **poll** path `GET /v1/projects/{p}/locations/{l}/operations/{id}` isn't in `vertexCollections`, and the GKE handler (registered before Vertex in server/gcp/gcp.go) already claims `operations` — so any Vertex op that isn't done-on-arrival is unpollable (routes to GKE → 404). It works today only because creates return done inline; worth a shared operations router (or at least a note) before any non-done op is introduced.

providers/gcp/vertexai/models.go
```diff
@@ -0,0 +1,142 @@
...
+
+	return []driver.Model{*cloneModel(model)}, nil
+}
+
+func (m *Mock) DeleteModelVersion(_ context.Context, name string) (*driver.Operation, error) {
+	return m.doneOp(locationOf(name), name), nil
+}
```
NitinKumar004, line 109: **`DeleteModelVersion` is a no-op that reports success.** Unlike `DeleteModel` (L80, which checks `Has` and deletes), this returns a done Operation without verifying existence or removing anything. Deleting a non-existent version returns success (real Vertex returns NotFound), and a real version is never actually removed — `ListModelVersions` would still show it. Add the existence check + actual delete.

NitinKumar004:
## Review — GCP Vertex AI

Big, clean addition and the breadth is impressive; it builds, tests pass, and lint is clean. I left 9 inline comments. The themes:

**Correctness (please address before merge):**
- **In-place mutation of stored pointers** in `PatchModel` / `PatchDataset` — the same data-race/aliasing pattern #222 was sent back for. Use copy-then-`Set`.
- **Endpoint traffic** — `DeployModel` overwrites the whole `TrafficSplit` (a 2nd deploy drops the 1st model's share); `UndeployModel` deletes from a shared map in place.
- **`Labels` maps** stored by reference across datasets/endpoints/models — clone on store.
- **Feature Store child listings** use a bare string-prefix match (no `/` delimiter) → sibling-prefix leakage (`fg1` vs `fg10`).
- **`CreateCachedContent`** ignores `TTLSeconds`, so caches are born expired.
- **`DeleteModelVersion`** is a success-returning no-op.

**Wire fidelity:**
- Done-operation `response` carries only `{name}` (no typed result / `@type`), so SDK callers reading a resource off the LRO result get empty fields.
- `:streamGenerateContent` returns a single JSON object; the SDK expects an array/SSE stream of chunks.
- LRO **poll** path isn't served and would route to GKE (registered first) — works today only because creates return done inline; fragile if any non-done op is added.

**Architecture:**
- **No portable API layer** (`vertexai/vertexai.go`) — the identical gap from #222, so Vertex has no recorder/metrics/rate-limit/error-injection/latency. This is the one I'd most want fixed for consistency.

**Scope / maintainability (non-blocking, worth noting):**
- ~90 driver methods across Feature Store / Vector Search / pipelines / tuning / metadata have **no REST route and no tests** (~25% provider coverage). The PR scopes the handlers as incremental, which is fine — but that's a lot of unreachable surface; consider trimming the driver to what's served, or land handlers+tests alongside.
- Vector Search ignores the uploaded vectors and `FindNeighbors` returns synthetic neighbors — fine as a stub, but the unused `FeatureVector` field is misleading.
- List/Get/Delete and Cancel are copy-pasted across ~10 families (a generic helper already exists for cancel); and `List*` has no pageSize/pageToken despite the real APIs being paginated.

Really nice work overall — the structure mirrors the GCP pattern well. Once the mutation/aliasing fixes and the portable layer land, this is in good shape; the unserved-families scoping is the main judgment call for you.

providers/gcp/vertexai/models.go
thzgajendra: Fixed. \`PatchModel\` and \`PatchDataset\` now copy-then-\`Set\` (\`updated := *model; updated.DisplayName = …; m.models.Set(name, &updated)\`) and return clones, so concurrent \`Get\`/\`List\` snapshots are never mutated. Added \`TestPatchModelDoesNotMutateSnapshot\`. (commit 019a1d5)

providers/gcp/vertexai/endpoints.go
thzgajendra: Fixed. \`DeployModel\`/\`UndeployModel\` now rebalance traffic evenly across every deployed model via \`evenTrafficSplit\` (totals 100, never drops a deployed model from the split). Added \`TestDeployModelKeepsTrafficForAllModels\`. (commit 019a1d5)

providers/gcp/vertexai/endpoints.go
thzgajendra: Fixed. \`UndeployModel\` works on a deep \`cloneEndpoint\` copy (which clones \`TrafficSplit\`/\`DeployedModels\`) before mutating, so the stored map and any prior \`GetEndpoint\` snapshot are untouched. (commits in the hardening pass + 019a1d5)

providers/gcp/vertexai/datasets.go
thzgajendra: Fixed in the hardening pass — datasets/endpoints/models now store \`copyLabels(cfg.Labels)\` (a clone) and return clones. Added \`TestCreateLabelsAreCloned\` proving a post-create mutation of the caller map does not affect the stored resource.

providers/gcp/vertexai/featurestore.go
thzgajendra: Fixed. All three child listings now match \`strings.HasPrefix(k, parent+"/")\` so \`fg1\` no longer leaks \`fg10\`'s children. Added \`TestChildListingPrefixDelimiter\`. (commit 019a1d5)

providers/gcp/vertexai/genai.go
thzgajendra: Fixed. \`CreateCachedContent\` now sets \`ExpireTime = createTime + TTLSeconds\` (default 3600s) and persists \`Contents\`/\`SystemInstruction\` (added to the driver type, deep-copied). Added \`TestCachedContentTTLAndContents\`. (commit 019a1d5)

providers/gcp/vertexai/vertexai.go
thzgajendra: Done. Added the portable \`vertexai/vertexai.go\` Layer-1 wrapper implementing \`driver.VertexAI\` with the recorder/metrics/ratelimit/inject/latency \`do()\` pipeline (+ wrappers_*.go and tests), mirroring \`sagemaker/sagemaker.go\`.

server/gcp/vertexai/responses.go
thzgajendra: Fixed. Done LROs now carry the typed result in \`response\` with its \`@type\` (e.g. \`google.cloud.aiplatform.v1.Model\` with \`versionId\`, \`DeployModelResponse.deployedModel\`) via a new \`writeResourceOp\` helper across all create/deploy handlers; roundtrip test asserts it. For the poll-path collision: documented in handler.go that the operations collection is intentionally left to the GKE handler since every Vertex op is done-on-arrival with its result inlined, so polling is never needed (with a note to add a shared operations router before any non-done op). (commit 4e164ce)

providers/gcp/vertexai/models.go
thzgajendra: Fixed. \`DeleteModelVersion\` now resolves the base name, checks \`Has\`, and deletes — returning NotFound for a missing version. Added \`TestDeleteModelVersionActuallyDeletes\`. (commit 019a1d5)

server/gcp/vertexai/endpoints.go
```diff
@@ -0,0 +1,254 @@
...
+	out, cerr := h.svc.RawPredict(r.Context(), endpoint, body)
+	if cerr != nil {
+		writeCErr(w, cerr)
+
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	_, _ = w.Write(out)
```
NitinKumar004: **New lint failure — `gosec` G705 (XSS via taint) here.** `golangci-lint` now reports 1 issue on `w.Write(out)` because the bytes originate from the request body (rawPredict echo). It's a false positive — the response Content-Type is `application/json`, not HTML — but the repo requires 0 lint issues, so CI fails as-is. Add a directive with a reason:

```go
//nolint:gosec // G705: echoes predict input as application/json, not HTML — no XSS surface
_, _ = w.Write(out)
```

server/gcp/vertexai/genai.go
```diff
@@ -0,0 +1,143 @@
...
+
+// runGenAI dispatches generateContent / countTokens for either a publisher
+// model path or an endpoint resource name.
+func (h *Handler) runGenAI(w http.ResponseWriter, r *http.Request, model, action string) {
+	var req wireGenerateRequest
+	if !decode(w, r, &req) {
+		return
+	}
+
+	switch action {
```
NitinKumar004, line 121: **`:streamGenerateContent` still returns a single object.** It shares the `generateContent` case and emits one `GenerateContentResponse` via `writeJSON`. The real endpoint streams a JSON **array** of response chunks (or SSE), so an SDK iterating the stream fails to decode a lone object / yields zero chunks. Return an array (even a single-element one) for the stream action, or split the candidate text into a couple of chunks.

NitinKumar004:
## Re-review — nearly there 👍

Great turnaround. Verified against the current code, all 9 inline findings from the first pass are resolved:

- copy-then-`Set` in `PatchModel`/`PatchDataset`; `DeployModel`/`UndeployModel` now rebalance via `evenTrafficSplit` and clone the map; `Labels` cloned on store; Feature Store child filters use a `/` delimiter; `CreateCachedContent` honors TTL; LRO responses carry typed `@type` results; `DeleteModelVersion` checks existence.
- The **portable layer** (`vertexai/vertexai.go`) is in, and REST handlers + chaos/cost wiring now cover all families — that also closes the earlier "unreachable surface" concern.

Two small things left (inline):
1. **CI-blocking:** a new `gosec` G705 on `endpoints.go:253` (rawPredict echo) — false positive, just needs a `//nolint:gosec` with a reason so `golangci-lint` is clean again.
2. **`:streamGenerateContent`** still returns a single object instead of a streamed array — SDK stream decoders expect an array/SSE.

Once those two land it's good to go from my side.

server/gcp/vertexai/endpoints.go
thzgajendra: Fixed at the source rather than suppressing: \`rawPredict\` now re-encodes the echoed payload through \`json\` (decode → \`writeJSON\`) instead of writing the request-derived bytes verbatim, so the request→response taint flow is broken and gosec G705 no longer fires — no \`//nolint\` needed (which also keeps nolintlint happy across golangci-lint versions). Non-JSON payloads are wrapped in \`{"raw": …}\` so the response stays valid JSON. (commit da9d368)

server/gcp/vertexai/genai.go
thzgajendra: :streamGenerateContent now returns a JSON array of response chunks (single-element) instead of a lone object, so SDK stream decoders iterate it correctly. Split into its own case from generateContent; added \`TestStreamGenerateContentReturnsArray\` asserting the body decodes as a JSON array. (commit da9d368)

NitinKumar004 (approved):
LGTM ✅

All findings from both review rounds are resolved and verified on the latest head (`da9d368`):
- Provider correctness: copy-then-`Set` in Patch*, `evenTrafficSplit` rebalancing + map cloning in Deploy/UndeployModel, `Labels` cloned on store, Feature Store child filters delimited, `CreateCachedContent` honors TTL, `DeleteModelVersion` checks existence.
- Wire: LRO responses carry typed `@type` results; `:streamGenerateContent` returns a JSON array; rawPredict re-encodes the payload (breaks the G705 taint flow at the source — clean fix).
- Architecture: portable `vertexai/vertexai.go` layer added; REST handlers now cover all families; chaos + cost wired.

`go build`, full Vertex tests, and `golangci-lint` are all clean. Nice work on the breadth and the fast iterations.

PR #252, opened 2026-07-11
feat(dns): Secrets-style SDK-compat servers for Route 53, Azure DNS, and Cloud DNS

NitinKumar004:
## Summary

Wires the existing `dns` driver through the SDK-compat HTTP layer across all three providers — the **DNS slice of #143**. Real cloud SDK clients drive the in-memory DNS driver over HTTP with only an endpoint change. Follows the structure and quality bar of the just-merged Secrets slice (#238).

## AWS Route 53 (`server/aws/route53`)
REST/XML. **Operations:** CreateHostedZone, GetHostedZone, ListHostedZones, DeleteHostedZone, ChangeResourceRecordSets (CREATE/UPSERT/DELETE), ListResourceRecordSets.
- `Matches` = path prefix `/2013-04-01/hostedzone` — disjoint from the JSON-RPC (`X-Amz-Target`), query-protocol (`Action=`), and other REST-root handlers. Registered **before S3** (the permissive REST fallback).
- Returns bare driver zone IDs (the aws-sdk binds the ID straight into the request path); `ChangeResourceRecordSets` returns an immediate `INSYNC` ChangeInfo.

## Azure DNS (`server/azure/dns`)
ARM `Microsoft.Network/dnsZones`. **Operations:** zones CreateOrUpdate/Get/Delete/List/ListByResourceGroup; record sets CreateOrUpdate/Get/Delete/ListByDnsZone, converting the driver's flat `Values []string` to/from Azure's typed per-type arrays for A/AAAA/CNAME/TXT/NS/PTR.
- `Matches` = ARM provider `Microsoft.Network` **and** resource type `dnsZones` — disjoint from the existing `network` handler (which claims `virtualNetworks`/`networkSecurityGroups`). Registered **before BlobStorage**.
- Zones addressed by name in the URL → resolved to driver IDs via a `resolveZoneID` scan.

## GCP Cloud DNS (`server/gcp/clouddns`)
`dns/v1` REST/JSON. **Operations:** managedZones create/get/list/delete, changes.create (additions/deletions applied atomically, deletions first), resourceRecordSets.list.
- `Matches` = `/dns/v1/projects/{p}/managedZones[...]` — a distinct URL space from the `/v1/projects/` family (Firestore/IAM/SecretManager). Registered **before Firestore/GCS**.
- `ManagedZone.Id` is a uint64 in the SDK, so the `zone-<uuid>` driver ID is folded to a stable numeric string via FNV-64a (the one place a driver ID isn't echoed verbatim).

[...]

Issue #143, opened 2026-04-29
Wire remaining driver surfaces through SDK-compat HTTP layer


NitinKumar004:
Goal: bring SDK-compat HTTP coverage in line with the underlying portable drivers.

## Within wired services

**Storage (S3 / Blob / GCS):** multipart uploads, lifecycle policies, versioning, bucket policies, CORS, encryption, object tagging, bucket tagging, presigned URLs.

**Database (DynamoDB / Cosmos / Firestore):** TTL, streams/change feed, GSIs (Create/Delete/Describe/List), BatchWrite + BatchGet for Cosmos and Firestore, TransactWriteItems for Cosmos and Firestore.

**Compute (Azure VMs / GCP Instances) — AWS EC2 already has these:** auto-scaling groups, scaling policies, spot / preemptible, launch templates.

**Networking (Azure / GCP) — AWS EC2 already has these:** route tables, network ACLs, internet gateways, NAT gateways, VPC peering, flow logs, VPC endpoints, elastic IPs.

**Monitoring:** Azure Monitor + GCP Cloud Monitoring metric data flow (PutMetricData / GetMetricData / ListMetrics — currently alarm/policy CRUD only); notification channels and alarm history across all 3 providers.

## Entire services with no SDK-compat at all

Drivers exist; no HTTP server for any provider yet.

1. Serverless (Lambda / Functions / Cloud Functions)
2. IAM
3. DNS (Route 53 / Azure DNS / Cloud DNS)
4. Load balancer (ELB / Azure LB / GCP LB)
5. Message queue (SQS / Service Bus / Pub-Sub)
6. Cache (ElastiCache / Azure Cache / Memorystore)
7. Secrets (Secrets Manager / Key Vault / Secret Manager)
8. Logging (CloudWatch Logs / Log Analytics / Cloud Logging)
9. Notification (SNS / Notification Hubs / FCM)
10. Container registry (ECR / ACR / Artifact Registry)
11. Event bus (EventBridge / Event Grid / Eventarc)

NitinKumar004:
Status update — the "entire services with no SDK-compat" list has shrunk considerably:

**Done since this issue was filed:**
1. ~Serverless~ — `server/aws/lambda`, `server/azure/functions`, `server/gcp/cloudfunctions`
2. ~IAM~ — all 3 providers (`server/{aws,azure,gcp}/iam`)
[...]

PR #254, opened 2026-07-11
feat(loadbalancer): SDK-compat servers for ELBv2, Azure LB, and GCP LB

NitinKumar004:
## Summary
Wires the existing `loadbalancer` driver through the SDK-compat HTTP layer across all three providers — the Load Balancer slice of #143. Real cloud SDK clients drive the in-memory driver with only an endpoint change. Follows the merged Secrets (#238) and DNS (#252) slices.

## Providers
- **AWS ELBv2** (`server/aws/elbv2`) — AWS query protocol. Load balancers, target groups, listeners, rules, RegisterTargets/DeregisterTargets/DescribeTargetHealth. Matches the ELBv2 action set, registered before EC2 (disjoint actions).
- **Azure LB** (`server/azure/loadbalancer`) — ARM `Microsoft.Network/loadBalancers`. Backend pools → target groups, load-balancing rules → listeners (pool scoping via internal `cloudemu:*` tags, stripped on output). Matches only `loadBalancers` (disjoint from the network/dns handlers).
- **GCP LB** (`server/gcp/loadbalancer`) — compute REST `backendServices` + global `forwardingRules`; a forwarding rule referencing a backend service also creates a listener. Matches only those resource types, disjoint from the existing compute/networks handlers (mutating ops return `compute#operation` envelopes the compute handler already polls).

## Tests
Real-SDK roundtrip per provider (`elasticloadbalancingv2`, `armnetwork.LoadBalancersClient`, `compute/apiv1`) against `httptest.NewServer` (TLS for Azure). `go build ./...` ✓, `go vet` clean, full `go test ./server/...` green fresh.

## Known gaps (documented, follow-up)
- `ModifyListener`, Get/PutLBAttributes, SetTargetHealth are on the driver but not exposed on the wire (outside core lifecycle); registered targets report the driver's default `initial` health.
- Azure: scheme inferred (defaults internal/Standard SKU), frontend IP configs synthesized.
- GCP: global forwarding rules only; healthChecks/targetPools/urlMaps not wired.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -35,6 +35,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/ec2 v1.297.1
 	github.com/aws/aws-sdk-go-v2/service/ecr v1.58.4
 	github.com/aws/aws-sdk-go-v2/service/eks v1.83.0
+	github.com/aws/aws-sdk-go-v2/service/elasticloadbalancingv2 v1.56.0
 	github.com/aws/aws-sdk-go-v2/service/iam v1.53.10
 	github.com/aws/aws-sdk-go-v2/service/lambda v1.90.1
 	github.com/aws/aws-sdk-go-v2/service/neptune v1.44.5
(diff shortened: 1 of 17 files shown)


NitinKumar004:
## Code review — findings

Multi-angle review (correctness across all three handlers + dispatch/shadowing). Dispatch is clean: ELBv2 registers before EC2 with a disjoint action set; Azure `loadBalancers` and GCP `backendServices`/`forwardingRules` are disjoint from their neighbors.

**Fixed in this PR (`9d049df`):**
- **Describe by a non-existent `Name` returned the whole fleet** — `resolveLBArns`/`resolveTGArns` fell through to an empty ARN list, which the driver treats as "all". Now returns `LoadBalancerNotFound` / `TargetGroupNotFound`.
- **`RuleNotFound` misclassified as `ListenerNotFound`** — the rule's not-found message embeds the parent listener ARN, so the substring check matched "listener" first. Now checks "rule" before "listener".
- **GCP forwarding rule with a missing backend service** silently created a dangling rule, and a failed `CreateListener` was swallowed. Both now surface as errors.

**Deferred to #259 (driver-model, affect the Portable API too):**
- Azure backend-pool→target-group ARN can collide across load balancers with certain names.
- Azure ARM PUT is create-only for nested pools/rules (not a declarative full replace); a rule re-pointed on an existing port is dropped.
- List output ordering is nondeterministic (memstore map iteration).

**Noted (not a bug today):** every `AlreadyExists` maps to `DuplicateLoadBalancerName`, but the bundled driver never returns `AlreadyExists`, so it's latent against a conformant driver.

PR #255, opened 2026-07-11
feat(eventbus): SDK-compat servers for EventBridge, Event Grid, and Eventarc

NitinKumar004:
SDK-compat servers for the **Event** service across AWS/Azure/GCP — the Event slice of #143, following the merged Secrets/DNS pattern. Real-SDK roundtrip tests per provider; `go build ./...`, `go vet`, and the full `go test ./server/...` tree pass. See the implementing commit for per-provider ops, driver→wire mapping, and documented gaps.
Notes: Azure Event Grid publish is a data-plane endpoint (not in the ARM SDK) so PutEvents isn't reachable via the Azure slice; GCP Eventarc's rich trigger model is approximated onto the driver's bus/rule/target — documented in the package docs.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v6 v6.6.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/databricks/armdatabricks v1.1.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/dns/armdns v1.2.0
+	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/eventgrid/armeventgrid/v2 v2.3.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/mysql/armmysqlflexibleservers v1.2.0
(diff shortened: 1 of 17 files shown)


NitinKumar004:
## Code review — findings

Multi-angle review. Dispatch clean (EventBridge `X-Amz-Target: AWSEvents.`; EventGrid `Microsoft.EventGrid`; Eventarc `/…/triggers` disjoint from Firestore/etc.). Error-code mapping and real-SDK typed-error assertions verified good across all three.

**Fixed in this PR (`33357ee`):**
- **GCP `createTrigger` left an orphaned rule** if `PutTargets` failed (a failed Create then blocked a re-Create with the same id). Now rolls back the rule on failure.

**Deferred to #259 (driver-model):**
- AWS `PutEvents` can't report per-entry results on partial failure — the driver's `PublishResult` exposes only `FailCount`, not which entries failed, so the response can't align `len(input)` entries with per-entry error codes like real EventBridge.
- Azure `ListByResourceGroup`/`ListBySubscription` ignore scope (driver `ListEventBuses` has no scope arg) → returns all topics.
- Azure `CreateOrUpdate` on an existing topic drops updated tags (no driver `Update`).
- List ordering nondeterministic.

**Documented approximation:** GCP Eventarc trigger fields beyond the driver's bus/rule/target model (serviceAccount, labels, non-CloudRun destinations) don't round-trip.

PR #256, opened 2026-07-11
feat(logging): SDK-compat servers for CloudWatch Logs, Log Analytics, and Cloud Logging

NitinKumar004:
SDK-compat servers for the **Logging** service across AWS/Azure/GCP — the Logging slice of #143, following the merged Secrets/DNS pattern. Real-SDK roundtrip tests per provider; `go build ./...`, `go vet`, and the full `go test ./server/...` tree pass. See the implementing commit for per-provider ops, driver→wire mapping, and documented gaps.
Notes: also makes the shared `server/wire/azurearm` resourceGroups path keyword case-insensitive. Azure data-plane log query (api.loganalytics.io) and GCP log buckets/sinks are out of scope (documented).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,6 +18,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/eventgrid/armeventgrid/v2 v2.3.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/mysql/armmysqlflexibleservers v1.2.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork v1.1.0
+	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/operationalinsights/armoperationalinsights v1.0.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/postgresql/armpostgresqlflexibleservers v1.1.0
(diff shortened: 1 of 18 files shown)


NitinKumar004:
## Code review — findings

Multi-angle review incl. the shared `server/wire/azurearm` change (case-insensitive `resourceGroups`) — verified safe: it only relaxes the structural keyword match, not resource-group *names*, and every existing handler parses identically. Dispatch clean. AWS CloudWatch Logs handler is solid (typed errors, epoch-millis round-trip).

**Fixed in this PR (`93e9132`):**
- **GCP `logName` filter matched as an unanchored substring** — a filter mentioning e.g. `logName_suffix`, or `logName` after another clause, mis-parsed the target log. Now matched as a whole token.

**Deferred to #259 (driver-model):**
- Azure workspace `id` / `customerId` are built from a hardcoded `rg-default` / account instead of the request path — should use `azurearm.BuildResourceID` like the DNS handler.
- Azure `CreateOrUpdate` on an existing workspace drops updated retention/tags (no driver `Update`).
- GCP `entries:list` with a `logName` filter but absent `resourceNames` yields `logName: projects//logs/<id>` (empty project).
- List/entry ordering nondeterministic; malformed entry timestamps coerced to ingest-time `now`.

PR #257, opened 2026-07-11
feat(cache): SDK-compat servers for ElastiCache, Azure Cache, and Memorystore

NitinKumar004:
SDK-compat servers for the **Cache** service across AWS/Azure/GCP — the Cache slice of #143, following the merged Secrets/DNS pattern. Real-SDK roundtrip tests per provider; `go build ./...`, `go vet`, and the full `go test ./server/...` tree pass. See the implementing commit for per-provider ops, driver→wire mapping, and documented gaps.
Notes: control-plane (cluster/instance lifecycle) only — the Redis data plane (Set/Get/Incr/...) has no cloud-SDK surface and is intentionally out of scope (documented).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork v1.1.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/operationalinsights/armoperationalinsights v1.0.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/postgresql/armpostgresqlflexibleservers v1.1.0
+	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/redis/armredis/v3 v3.3.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resourcegraph/armresourcegraph v0.9.0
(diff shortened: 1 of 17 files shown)


NitinKumar004:
## Code review — findings

Multi-angle review. ElastiCache-vs-EC2 dispatch clean (disjoint actions, `TestSDKElastiCacheDoesNotShadowEC2` is a real assertion); Azure `Microsoft.Cache` and GCP Memorystore `instances` are disjoint from their neighbors; error mapping + typed-error assertions verified good. Scope: control plane only (Redis data plane intentionally out of scope — no cloud-SDK surface).

**Deferred to #259 (driver-model / latent):**
- **GCP: GKE shadows Memorystore's operation-poll path** — both claim `/v1/projects/{p}/locations/{l}/operations/{op}` and GKE registers first. Latent today because Memorystore LROs return `done:true` inline so the SDK never polls, but the "disjoint" docstring is inaccurate; `GET …/operations` (collection) also returns a fabricated done-op. Narrow Memorystore to `instances`-only or make operations service-aware.
- Azure `CreateOrUpdate` on an existing cache drops updated tags/SKU (no driver `Update`).
- List ordering nondeterministic (memstore map iteration).

No self-contained correctness bug required an in-PR fix; the items above are shared driver-model gaps.

PR #258, opened 2026-07-11
feat(notification): SDK-compat servers for SNS, Notification Hubs, and FCM

NitinKumar004:
SDK-compat servers for the **Notification** service — the Notification slice of #143, following the merged Secrets/DNS pattern. Real-SDK roundtrip tests per provider; `go build ./...`, `go vet`, and the full `go test ./server/...` tree pass.

Notes: SNS maps ARN↔name and is publish-as-store (no fan-out delivery). Azure models namespace→topic and hub→`{ns}/{hub}` topic. GCP **FCM is intentionally `messages:send`→Publish only** — FCM's REST surface has no topic/subscription CRUD, so none is fabricated (documented).


NitinKumar004:
## Code review — findings

Multi-angle review. Dispatch clean (SNS disjoint actions before EC2; `Microsoft.NotificationHubs` distinct provider; FCM `messages:send` suffix disjoint from Firestore/PubSub/IAM). SNS error mapping verified against the real SDK deserializers (`NotFound`→`NotFoundException`). Azure `{ns}/{hub}` composite key is collision-safe (ARM path can't contain `/` in a segment).

**Fixed in this PR (`a1e9ed9`):**
- **FCM `validateOnly` was ignored** — a dry-run send auto-created the topic, published, and emitted metrics (the opposite of the flag's contract). Now returns a fabricated message name with no side effects when `validateOnly` is set.

**Deferred to #259 (driver-model / fidelity):**
- FCM token/condition-addressed sends funnel into one global `_fcm_token_target` topic (docstrings claim per-project/per-request).
- Azure `CreateOrUpdate` on an existing namespace/hub drops updated tags/TTL (no driver `Update`).
- SNS `GetTopicAttributes` omits always-present attributes (`SubscriptionsPending`, `Owner`, …).
- List ordering nondeterministic.

PR #265, opened 2026-07-12
feat(azure): Storage Queues + Table Storage SDK-compat

### LINKED ISSUE #243: feat: Azure Storage Queues + Table Storage SDK-compat handlers
Blob is covered but Azure's other two Azurite headline services — Queue Storage and Table Storage — are not. Direct parity win for Azure Go teams.

**Build:** `server/azure/queue` and `server/azure/table` handlers speaking the Azure Storage REST protocols for `azqueue` and `aztables`. Queue can reuse the messagequeue driver; Table needs a light entity/partition-key store (mirror the shared-sub-backend pattern used by the Kubernetes data plane).

**Acceptance:** real `azqueue` enqueue/dequeue and `aztables` upsert/query-entity roundtrips green; `go test ./server/...` green.

NitinKumar004:
Closes #243. `server/azure/queue` (REST/XML over the messagequeue driver) and `server/azure/table` (OData/JSON over a new light `tablestorage` driver). Real `azqueue`/`aztables` roundtrip tests + a no-shadowing test (Blob+Queue+Table together). One documented non-disjoint case: `GET /?comp=list` is identical for list-queues vs list-containers (Azure separates by hostname); Queue handler owns it when both are wired. Full `go test ./server/...` green.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.21.1
 	github.com/Azure/azure-sdk-for-go/sdk/containers/azcontainerregistry v0.2.3
 	github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos v1.4.2
+	github.com/Azure/azure-sdk-for-go/sdk/data/aztables v1.4.1
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/appservice/armappservice/v3 v3.0.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/authorization/armauthorization/v3 v3.0.0-beta.3
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/cognitiveservices/armcognitiveservices v1.8.0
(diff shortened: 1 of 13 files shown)


NitinKumar004:
## Code review — findings

Multi-angle review. OData key parsing (quotes/commas/special chars), the `pk\x00rk` composite key, queue name→URL resolution, and message round-trip all verified clean. Insert/get/delete/conflict/404 driver semantics correct.

**Fixed (`85ee5f6`):**
- **UpsertEntity now works:** a PUT/MERGE with no `If-Match` against a missing row inserts (insert-or-replace/merge) instead of 404 — matching the common `aztables` UpsertEntity call.
- Corrected the misleading Queue `Matches` docstring: `/{queue}/messages` and `/?comp=list` are **not** strictly disjoint from Blob.

**Documented / inherent (not cleanly fixable):**
- Queue vs Blob collide on `/{name}/messages` (a blob named `messages`) and `GET /?comp=list` — Azure separates these only by the `{account}.queue` vs `.blob` hostname, invisible behind cloudemu's single endpoint. Documented in the handler; needs host-based routing to resolve.
- `$filter` degrades to match-all on an unparseable expression (data hazard for unsupported operators); enqueue advertises the MessageID as the PopReceipt; `parseKeyPredicate` accepts a single-key predicate (wrong status code).

PR #270, opened 2026-07-19
E2E campaign: storage + database domains — 9 bugs found and fixed, ~110 new e2e tests

NitinKumar004:
## Summary

First two iterations of the end-to-end test campaign (see `E2E_CAMPAIGN.md` for the full matrix, per-cell methodology, and issue log). Every domain cell is tested the way real users drive it: portable Go API journeys plus the REAL SDKs (`aws-sdk-go-v2`, azblob/azcosmos, Google clients) against the httptest servers, asserting SDK-decoded values and typed errors.

## Bugs found & fixed

**Storage (3):**
1. S3 `PutObject` response omitted the `ETag` header (real S3 always returns it).
2. Azure Put Blob ignored `x-ms-blob-content-type` — SDK uploads stored as `application/octet-stream`.
3. GCP multipart upload trimmed trailing `-`/CRLF bytes from payloads (data corruption) — now byte-exact RFC 2046 parsing.

**Database (6):**
4. **Systemic, all three portable drivers:** Scan/Query paginated over a randomized map snapshot with offset tokens — multi-page reads duplicated and dropped items. Fixed with a stable item-key sort before pagination.
5. DynamoDB wire had no `ExclusiveStartKey`/`LastEvaluatedKey` — SDK users could never page past page 1. Implemented.
6. DynamoDB `PutItem` ignored `ConditionExpression` — `attribute_not_exists` create-if-absent silently overwrote. Now enforced with `ConditionalCheckFailedException` (unsupported forms → `ValidationException`).
7. Cosmos continuation-page queries (`Content-Type: application/query+json`) were misrouted as document creates → 400; query paging (`x-ms-max-item-count`/`x-ms-continuation`) implemented.
8. Firestore `batchGet` framed one JSON array per entry — clients decoded only the first document.
9. Firestore `:commit` dropped `currentDocument` preconditions — conditional create/update never failed. Now 409/404 per real semantics.

## Tests added
~110 e2e tests across 12 cells (2 domains × 3 providers × 2 surfaces): full lifecycles, typed errors, pagination continuation, multipart, TTL-under-fake-clock, streams/change feeds, GSIs, conditional writes, unicode keys, batch ops.

## Known deviations needing a maintainer call
Logged in `E2E_CAMPAIGN.md` under "Human decisions needed" — e.g. S3 delete-missing-key returns 404 vs real S3's idempotent 204; non-empty bucket delete returns 500 (triggers SDK retry backoff) vs real 409; `ScanForward` accepted but never applied; Azure hard-coded blob ETag.

## Test plan
- Full suite: `go test ./... -count=1` — **0 failures** (190 packages).
- `go vet` clean on all touched packages.
- Every fix is locked by a test that fails on pre-fix code (verified for storage by the iteration reviewer; database fixes were each reproduced before fixing).


NitinKumar004:
## Code review — 8-angle pass (line-by-line, removed-behavior, cross-file, reuse, simplification, efficiency, altitude, conventions)

**Overview:** this PR adds ~110 SDK-driven and portable-API e2e suites for the storage and database domains and fixes 9 emulator bugs. The review examined the product-code hunks (3 drivers, 6 wire handlers) plus the new test files. Findings below; everything under "Fixed" landed in 684e67c and CI is green.

### Fixed in this PR (confirmed findings)

1. **`server/aws/dynamodb/handler.go` — `checkCondition` ignored `ExpressionAttributeNames`.** The idiomatic `ConditionExpression: "attribute_not_exists(#pk)"` + `{"#pk":"id"}` resolved the field to the literal `#pk`, so the guard always passed and conditional PutItem silently overwrote existing items. Now resolves aliases via `resolveAttrName`, parses the expression once, and rejects compound/malformed forms with `ValidationException` instead of mis-evaluating them.
2. **All three NoSQL drivers swallowed `pagination.Paginate` errors** (`page, _ :=`). A malformed/expired continuation token — wire-reachable via Cosmos `x-ms-continuation` — returned an empty 200 page, which clients read as "end of results" (silent data loss). Now returns `InvalidArgument` → 400.
3. **`paginateWire` silently restarted from item 0 on an unknown/stale `ExclusiveStartKey`** (e.g. anchor item deleted between pages), re-serving already-consumed items. Now rejected with `ValidationException`.
4. **`ScanIndexForward` was never applied** — pre-existing, but the new stable sort made descending queries deterministically wrong. Now honored at the wire layer (reverse before paging).
5. **`sortByItemKey` was copy-pasted into three drivers with a per-comparison `fmt.Sprintf` comparator** (O(M log M) Sprintf calls). Replaced with one shared `driver.SortStableByKey` using decorate-sort (one key computation per item).
6. **Key-separator drift:** `paginateWire.keyOf` used `pk|sk` while the drivers use `pk:sk` — a silent-mismatch trap for future maintainers. Aligned to `:`.
7. **Cosmos query default page size was the whole collection** (`1<<30`) when `x-ms-max-item-count` was absent. Now defaults to 100 with a continuation token, matching real Cosmos and the driver default.
8. **GCS multipart content-type precedence inverted:** the media part's `Content-Type` (often `application/octet-stream` from SDKs) beat `metadata.contentType`; real GCS is the reverse. Fixed.
9. **Five lines exceeded CONTRIBUTING.md's 140-char limit** in `s3_lifecycle_test.go`. Wrapped.

### Known limitations / follow-ups (not fixed here, by design)

- **DynamoDB wire pagination materializes the full result set per page** (`fetchAll` + handler-side slicing) — O(N) per page. The right fix is key-based continuation in the driver; also affects the `ConsumedReadCapacityUnits` metric and `ScannedCount`, which now reflect the full matched set.
- **Numeric sort keys order lexicographically** (`"10" < "9"`) in the stable sort — diverges from real DynamoDB numeric ordering.
- **GSI queries**: `LastEvaluatedKey` is shaped from base-table keys and ordering is by base key, not the index key.
- **Pre-existing gaps adjacent to (not introduced by) this PR:** Firestore `runQuery` ignores `where`/`limit` (100-doc cap), `updateDocument` ignores `updateMask` (full replace), Cosmos GET document list caps at 100 with no continuation.
- **Architectural:** the "stable input ordering" invariant that offset tokens require should eventually be enforced centrally (in `internal/pagination` or `memstore`) rather than per-driver — any future driver that feeds `.All()` into `Paginate` without sorting reintroduces the duplicate/dropped-pages bug.

Verification: `go vet` clean, full suite `go test ./... -count=1` 0 failures, all CI checks green after the fix commit.

Issue #259, opened 2026-07-11
Driver-model fidelity gaps across the #143 SDK-compat slices (list scope, upsert updates, ordering, Azure IDs)


NitinKumar004:
Follow-up from the code review of the #143 SDK-compat slices (LB #254, Event Bus #255, Logging #256, Cache #257, Notification #258). The self-contained correctness bugs were fixed in those PRs; the items below are **recurring limitations rooted in the shared service-driver abstractions**, so they affect the Portable API too and are best fixed once at the driver layer rather than per-handler. (Companion to the DNS follow-up #253.)

**Cross-cutting driver-model gaps (seen in most/all slices)**
- **Scope-less `List*`.** `logging.ListLogGroups`, `eventbus.ListEventBuses`, `cache.ListCaches`, `notification.ListTopics` take no scope arg, so the Azure `ListByResourceGroup` / `ListBySubscription` (and GCP per-project list) handlers return every resource regardless of resource group / subscription / project, and Azure re-labels them under the requested scope. Needs scope-aware driver lookups.
- **No `Update` → tag/field drop on upsert.** Azure `CreateOrUpdate` on an existing EventGrid topic / Log Analytics workspace / Redis cache / Notification Hub namespace silently discards the request's new tags / retention / SKU and echoes the stored resource with 200. Same class as DNS #253. Needs an `Update*` (or upsert) driver method.
- **Nondeterministic list ordering.** All `List*` iterate `memstore.All()` (a Go map), so multi-item responses come back in random order (real APIs default to a defined order, e.g. reverse-time for log entries). Repo-wide pattern (DNS/route53 too) — worth a deterministic sort at the memstore/driver layer.
- **Azure resource IDs baked from a fixed `rg-default`.** `providers/azure/loganalytics` (and peers) build `ResourceID` from a hardcoded resource group / account, so the ARM `id` (and `customerId`) returned don't match the request's subscription/RG. Handlers should rebuild the ID from the request path via `azurearm.BuildResourceID` (as the DNS handler does).

**Per-slice items deferred here (not quick self-contained fixes)**
- **Event Bus (AWS):** `PutEvents` can't report per-entry results — the driver's `PublishResult` exposes only `FailCount`, not which entries failed, so a partial-failure batch can't return `len(input)` aligned entries with per-entry error codes as real EventBridge does. Needs per-entry results on the driver.
[...]

PR #274, opened 2026-07-19
Deep audit: real-cloud error semantics, fake-clock fidelity, and boundary isolation

NitinKumar004:
## Summary

A deep end-to-end audit pass: first the five wire deviations flagged (and deferred) in the PR #270 review, then a codebase-wide hunt for systemic bug **classes** rather than per-domain journeys. Every fix is tested; two new regression tests lock the systemic classes. Full suite `go test ./... -count=1` = 0 failures, `go vet` clean.

## Fixed — known wire deviations (commit 1)

- **S3 `DeleteObject` is idempotent** for a missing key (204, real-S3 semantics) — defensive-delete SDK code now works. A missing **bucket** is `NoSuchBucket` everywhere instead of `NoSuchKey` (discriminated via the driver's structured error `Message`, not string-matching the formatted error).
- **Non-empty bucket deletion is a client error, not a 500**: S3 → `409 BucketNotEmpty`, GCS → `409 conflict` (the old `InternalError` triggered SDK retry backoff). **Azure matches real Azure instead**: `DeleteContainer` deletes blobs recursively and succeeds.
- **Azure Blob returns real ETags and clock-driven `Last-Modified`** — the shared hard-coded `"0x8DAB0"` and wall-clock `time.Now()` are gone from Put Blob and the container endpoints (per-container ETags derive from creation time).
- **S3 `PutObject` reads the ETag back from the driver** instead of recomputing sha256 inline — one source of truth for the algorithm. (Trade-off: puts now also tick the Head request metric.)

## Fixed — systemic classes (commit 2)

- **Wall-clock leaks that break the fake clock**: all 15 `time.Now()` sites across the lambda / Azure Functions / Cloud Functions mocks (functions, versions, aliases, layers) now use `opts.Clock`. Deterministic-time tests previously got real timestamps.
- **Reference-shared maps**: storage drivers clone object metadata on write **and** on every `ObjectInfo` read; database drivers clone item maps at every public boundary (`PutItem`/batch/transact in, `GetItem`/`BatchGetItems` out, and Query/Scan page outputs centrally in `driver.PageOrdered`). Caller mutations can no longer corrupt stored state — and stream/change-feed images now capture isolated snapshots.
- **Swallowed-error sweep** of `providers/`, `server/`, `services/`: no remaining real cases (survivors are hash writes, ok-pattern lookups, and state-returning calls).

## Regression tests
`TestMetadataIsolation` (mutate passed + returned maps → store unaffected) and `TestTimestampsUseConfiguredClock` (fake clock pinned to 2025-01-01 → no wall-clock leak). Lifecycle tests that pinned the old deviations (delete-missing-key 404, `NoSuchKey`-for-bucket, 500-on-non-empty, Azure non-empty-delete failure) were flipped to assert real-cloud semantics.

## Review notes
Self-reviewed the diff for the classic traps: recursive-delete loop termination, `maps.Clone(nil)` behavior, clone placement relative to stream-event capture, and error-type discrimination through `*cerrors.Error` (wrapped-error safe via `errors.As`).

### DIFF
--- a/providers/aws/dynamodb/dynamodb.go
+++ b/providers/aws/dynamodb/dynamodb.go
@@ -3,6 +3,7 @@ package dynamodb
 import (
 	"context"
 	"fmt"
+	"maps"
 	"strconv"
 	"strings"
 	"sync"
@@ -151,6 +152,7 @@ func (m *Mock) PutItem(_ context.Context, table string, item map[string]any) err
 
 	key := itemKey(td.config, item)
 	oldItem, hadOld := td.items.Get(key)
+	item = maps.Clone(item)
 	td.items.Set(key, item)
 	m.recordStreamEvent(td, oldItem, item, hadOld)
 	m.mu.Unlock()
@@ -187,7 +189,7 @@ func (m *Mock) GetItem(_ context.Context, table string, key map[string]any) (map
 	m.emitMetric("ConsumedReadCapacityUnits", 1, dims)
 	m.emitMetric("SuccessfulRequestCount", 1, dims)
 
-	return item, nil
+	return maps.Clone(item), nil
 }
 
 // UpdateItem applies partial updates to an existing item.
@@ -407,6 +409,7 @@ func (m *Mock) BatchPutItems(_ context.Context, table string, items []map[string
 	for _, item := range items {
 		key := itemKey(td.config, item)
 		oldItem, hadOld := td.items.Get(key)
+		item = maps.Clone(item)
 		td.items.Set(key, item)
 
 		m.recordStreamEvent(td, oldItem, item, hadOld)
@@ -430,7 +433,7 @@ func (m *Mock) BatchGetItems(_ context.Context, table string, keys []map[string]
 
 	for _, key := range keys {
 		if item, ok := td.items.Get(itemKey(td.config, key)); ok {
-			results = append(results, item)
+			results = append(results, maps.Clone(item))
 		}
 	}
 
@@ -643,6 +646,7 @@ func (m *Mock) applyTransactPuts(td *tableData, puts []map[string]any) {
 	for _, item := range puts {
 		key := itemKey(td.config, item)
 		oldItem, hadOld := td.items.Get(key)
+		item = maps.Clone(item)
 		td.items.Set(key, item)
 		m.recordStreamEvent(td, oldItem, item, hadOld)
 	}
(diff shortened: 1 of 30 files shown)


NitinKumar004:
## /review — 8-angle pass (6 finder agents, cross-verified)

**Overview:** this PR fixes the five deferred wire deviations and two systemic classes (fake-clock leaks, reference-shared maps). The review found the fixes correct in direction but incomplete at the edges; everything below marked **fixed** landed in 4aa57a0, CI green.

### Findings fixed
1. **`UpdateItem` still returned the stored map by reference** (all 3 NoSQL drivers) — the one boundary the isolation sweep missed; mutating its result corrupted the store *and* stream images. → clones now.
2. **`bucketMissing` missed CopyObject's `source/destination bucket …` wording** — copy against a missing bucket still said `NoSuchKey`. → discriminates by excluding the `object …` form, then matching any `…bucket` wording.
3. **Put read-back TOCTOU:** a successful PUT could answer 404 if a concurrent delete (or a chaos fault scoped to `HeadObject`) raced the ETag read-back. → falls back to computing the sha256 ETag from the stored body; a successful PUT never fails.
4. **`copyBlob` still emitted the removed hard-coded `0x8DAB0` ETag + wall-clock time** — copy-then-If-Match would 412 immediately. → real ETag/Last-Modified via read-back.
5. **Container ETags collided** (crc32 of second-resolution `CreatedAt` — identical for *all* containers under a pinned fake clock), and `getContainerProperties` served a bogus 200 ETag for missing containers. → name+createdAt hash; 404 `ContainerNotFound`.
6. **`emptyContainer` race modes:** a blob deleted mid-purge surfaced as `BlobNotFound` for a *container* delete, and a racing writer could spin the loop forever. → NotFound-tolerant, pass-bounded.
7. **Same aliasing class one directory over:** lambda/Functions/Cloud Functions stored and returned `Environment`/`Tags` by reference — in files this PR touched for the clock fix. → cloned at create/update/get.
8. **`ListObjects` cloned metadata for every match before pagination** — a paged scan of a 100k-object bucket allocated 100k clones per page. → clones only the returned page.
9. **Convention (CONTRIBUTING "all 3 providers … same behaviors / unit tests per provider"):** the clock and isolation regression tests existed only for AWS. → mirrored to Azure + GCP siblings.
10. `cloneItems` doc comment overpromised — clones are shallow; nested values remain shared. → documented as a deliberate limitation.

### Noted, deliberately not fixed here (right home is #259 / follow-ups)
- Deeper altitude for several of these: `HeadBucket` and `PutObject`-returns-info on the storage driver, a driver-level recursive `DeleteBucketForce`, a resource-kind field on `cerrors` (replacing message-prefix discrimination), and a single memstore chokepoint for copy-on-write — all fit the #259 driver-model work.
- DynamoDB TTL lazy-delete on `GetItem` emits no stream REMOVE event (pre-existing, adjacent).
- Chaos faults injecting `NotFound` on `DeleteObject` are now absorbed by the real-S3 idempotency mapping — mirrors real S3, but chaos experiments targeting that op should use a different code.

**Verification:** full suite `go test ./... -count=1` 0 failures, `go vet` clean, no added line exceeds 140 chars, all CI checks green after the fix commit.

Issue #253, opened 2026-07-11
DNS driver-model fidelity gaps (zone scoping, UpdateZone, TXT chunks, batch atomicity)


NitinKumar004:
Follow-up from the DNS SDK-compat review (#252). These are fidelity gaps rooted in the shared `dns/driver.DNS` model, so they affect the Portable API too, not just the HTTP handlers — worth fixing at the driver level rather than per-handler.

**Zone identity & scoping**
- `dns/driver` has no zone **name uniqueness** — two zones with the same name resolve nondeterministically (map iteration order) in every handler's `resolveZoneID`. Real Route 53/Azure/Cloud DNS reject duplicates (or scope by id).
- `ZoneInfo` carries no **resource-group / project** attribution, so Azure `ListByResourceGroup` and GCP `managedZones.list` return every zone across RGs/projects (and Azure re-labels each under the requested RG). Needs a scope field on the driver.
- No `GetZoneByName` / indexed lookup — every zone/record op does an O(n) `ListZones` scan (duplicated across the Azure and GCP handlers).

**Zone mutation**
- No `UpdateZone` method → Azure `CreateOrUpdate` on an existing zone silently drops tag/zoneType updates (returns the stale zone with 200).

**Record fidelity**
- Flat `Values []string` can't represent Azure **TXT multi-chunk** entries (`[][]string`); chunks are joined on the wire, so `["a","b"]` round-trips to `["ab"]` (DKIM/SPF).
- `DeleteRecord(zoneID, name, type)` ignores rrdatas/TTL, so Cloud DNS deletions don't enforce exact-match of the existing record set the way the real API does.
- No `UpsertRecord` primitive → the check-then-act upsert is reimplemented (and slightly racy) in the Route 53 and Azure handlers.

**Batch atomicity**
- No transaction primitive → GCP `changes.create` can't be truly atomic. #252 added an up-front validation pass so a bad batch fails before mutating, but a concurrent writer between validate and apply can still race.

Acceptance: extend `dns/driver.DNS` (name/id + scope-aware lookups, `UpdateZone`, `UpsertRecord`, structured TXT), update all three providers + the handlers, and add roundtrip tests covering TXT chunks, per-scope listing, tag updates, and duplicate-name rejection.

NitinKumar004:
Closing — all listed DNS driver-model fidelity gaps are fixed, each as a focused PR with real-SDK regression tests:

- **Record-listing determinism** → #280
[...]

PR #276, opened 2026-07-19
fix: scope-aware listing, real CreateOrUpdate, request-derived ARM IDs (#259 — PR 2)

NitinKumar004:
Part 2 of the #259 series (after #275 landed deterministic list ordering). Covers items 1, 2, and 4 across the four #143 domains: **logging, eventbus, cache, notification**.

## What changed

**New `services/scope` package** — `Scope{Subscription, ResourceGroup, Project}` with wildcard `Matches` semantics: a zero scope on a resource means unscoped (visible everywhere), an empty filter field means "don't filter on this".

**Item 1 — scope-aware listing.** Each domain driver's `List*` now takes a `scope.Scope` filter, and each provider stores the creating request's scope on the resource. The Azure handlers pass the subscription/resource-group parsed from the URL, GCP handlers pass the request project, AWS handlers pass a zero scope (account-global, as before). Result: `ListByResourceGroup` no longer returns other groups' resources, and a GCP list no longer returns other projects' resources.

**Item 2 — CreateOrUpdate applies updates.** New `Update*` driver method per domain with ARM PUT semantics (mutable fields: tags, retention, node type, display name). The Azure CreateOrUpdate handlers now do Get → Update-with-request-fields instead of echoing the stale stored resource.

**Item 4 — request-derived ARM IDs.** Azure resource IDs are built from the request's subscription/resource group instead of hardcoded `rg-default` (fixed in loganalytics and notificationhubs providers; eventgrid topic IDs now derive from the creating scope). Note: the `rg-default` in `providers/azure/keyvault` is dead on the wire — the keyvault handler builds secret IDs from the request path — so it was left alone.

Portable wrappers (`services/<domain>`) keep their public signatures unchanged, passing a zero scope internally — portable-API users see everything, exactly as before.

## Testing

12 new real-SDK wire tests (3 per domain, in `server/azure/<svc>/scope_update_sdk_test.go`), using the same `httptest` + real Azure SDK bootstraps as the existing roundtrip suites:

- `TestSDKScopedListing` — resources created in `rg-team-a` don't appear in `rg-team-b`'s pager
- `TestSDKUpsertAppliesUpdates` — CreateOrUpdate on an existing resource applies the request's tags; a follow-up Get confirms
- `TestSDKResourceIDMatchesRequestScope` — the returned ARM id carries the request's subscription and resource group

Plus `services/scope` unit tests. Full `go test ./...` and `go vet ./...` clean; eventgrid packages verified stable with `-count=3`.

## Remaining for PR 3

#253 DNS deep pass (zone name uniqueness, structured TXT records, exact-match delete, upsert, batch atomicity).

Closes nothing yet — #259 stays open until PR 3 lands.

### DIFF
--- a/cloudemu_test.go
+++ b/cloudemu_test.go
@@ -1861,7 +1861,7 @@ func testCacheWithDriver(t *testing.T, ctx context.Context, d cachedriver.Cache)
 	}
 
 	// List caches
-	caches, err := d.ListCaches(ctx)
+	caches, err := d.ListCaches(ctx, scope.Scope{})
 	if err != nil {
 		t.Fatalf("ListCaches: %v", err)
 	}
@@ -2403,7 +2403,7 @@ func testNotificationWithDriver(t *testing.T, ctx context.Context, d notifdriver
 	}
 
 	// List topics
-	topics, err := d.ListTopics(ctx)
+	topics, err := d.ListTopics(ctx, scope.Scope{})
 	if err != nil {
 		t.Fatalf("ListTopics: %v", err)
 	}
@@ -2581,7 +2581,7 @@ func TestCacheTagsPersistence(t *testing.T) {
 			}
 
 			// Verify tags appear in ListCaches
-			caches, err := p.d.ListCaches(ctx)
+			caches, err := p.d.ListCaches(ctx, scope.Scope{})
 			if err != nil {
 				t.Fatalf("ListCaches: %v", err)
 			}
@@ -3017,7 +3017,7 @@ func TestEventBusOperations(t *testing.T) {
 			}
 
 			// ListEventBuses
-			buses, err := p.d.ListEventBuses(ctx)
+			buses, err := p.d.ListEventBuses(ctx, scope.Scope{})
 			if err != nil {
 				t.Fatalf("ListEventBuses: %v", err)
 			}
--- a/cloudemu_test.go
+++ b/cloudemu_test.go
@@ -2,6 +2,7 @@ package cloudemu
 
 import (
 	"context"
+	"github.com/stackshy/cloudemu/v2/services/scope"
 	"sort"
 	"sync/atomic"
 	"testing"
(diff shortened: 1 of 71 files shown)


NitinKumar004:
## Review (8 angles: line-scan, removed-behavior, cross-file, reuse, simplification, efficiency, altitude, conventions)

Reviewed the full 71-file diff by hand. **No blocking defects.** The four domain implementations are correct and consistent with each other and with the logging template:

- `Update*` methods clone `Tags` on write (`maps.Clone`), copy the stored struct before returning, and apply `Scope` only when non-zero — no shared-map aliasing back into the store.
- List filtering happens **inside** the driver (`SortedValues()` → `Scope.Matches` → append), so it runs before any handler-side pagination — page sizes and continuation stay correct.
- ARM IDs are built from the creating scope with a fallback to the request path (`toRedisJSON`, `toTopicJSON`), so `ListBySubscription` responses with an empty request RG still carry a valid resource-group segment.
- Azure `CreateOrUpdate`-on-existing now does Get → `Update`-with-request-fields (eventgrid answers 201 because armeventgrid's SDK rejects 200 for `Topics.CreateOrUpdate`; the others answer 200) — verified by `TestSDKUpsertAppliesUpdates` in each domain.

### Two notes (both pre-existing design choices, not regressions — no change made)

1. **Name-keyed stores.** Resources are keyed by name only (`store.Get(cfg.Name)`), repo-wide and predating this series. Consequence: two resource groups can't hold same-named resources in one emulator instance, and a `CreateOrUpdate` for an existing name in a *different* RG takes the Update path and reassigns the resource's scope rather than creating a distinct one. Fixing this means scope-aware store keys across every provider — out of scope for #259 PR 2; noted as a follow-up.

2. **ARM PUT tag semantics.** `Update*` applies tags only when the request sends a non-nil `Tags` (merge-on-absence); a real ARM PUT replaces tags wholesale. An explicit empty map still clears them, so the only divergence is "tags field omitted entirely" keeping the old set. Deliberate and consistent across all four domains.

`go build ./...`, `go vet ./...`, `gofmt -l`, and full `go test ./...` all clean; eventgrid verified stable at `-count=3`.

Issue #250, opened 2026-07-11
feat: go:embed fixture-seeding loader for drivers


NitinKumar004:
Let teams check declarative JSON/YAML fixtures into their repo and seed cloudemu deterministically. Low effort; pairs with the admin/seed API.

**Build:** a `seed` package that loads `go:embed`-ed fixture files (buckets, tables, secrets, instances, …) into any wired driver set.

**Acceptance:** a test embeds a fixtures dir, seeds the AWS drivers, and reads the resources back via real SDK clients.

NitinKumar004:
Done in #288 — a `seed` package loads JSON fixtures (buckets, tables, secrets, instances) into a driver Target, provider-agnostic via the shared driver interfaces, with `LoadFS` for go:embed. Acceptance met: a test embeds a fixtures dir, seeds the AWS drivers, and reads the resources back via real S3 + DynamoDB SDK clients. Also wired into the standalone server's `/_cloudemu/seed` endpoint.

Issue #248, opened 2026-07-11
feat: Testcontainers-Go module for cloudemu


NitinKumar004:
A `testcontainers-go` module is the idiomatic way Go teams discover and adopt emulators — high visibility for engagement. Depends on the Docker image + #224.

**Build:** a `modules/cloudemu` Testcontainers module that starts the image and exposes typed endpoint accessors per provider (AWS/Azure/GCP base endpoints).

**Acceptance:** an example test spins up the container and runs a real `aws-sdk-go-v2` S3 roundtrip through it.

Issue #247, opened 2026-07-11
feat: publish an official Docker image (Dockerfile + GHCR publish workflow)


NitinKumar004:
There is no Dockerfile today. A published image is the biggest reach-expander — it unblocks non-Go users and the #224 standalone-server mode. Depends on / pairs with #224.

**Build:** a multi-stage `Dockerfile` building `cmd/cloudemu serve` (from #224), a GitHub Actions workflow publishing to GHCR on tag, and a `docker-compose.yml` example.

**Acceptance:** `docker run cloudemu` starts the server and a real SDK client (endpoint override) drives it successfully.

Issue #224, opened 2026-06-14
Feature: standalone server mode ('minikube for cloud') so out-of-process apps can point real SDK clients at cloudemu


NitinKumar004:
### Use case
I'd love to use cloudemu the way `minikube` is used for Kubernetes: run it as a **long-lived local process** that my app's **separate, real service processes** point their cloud SDK clients at — so live flows (e.g. resource discovery) actually run against cloudemu and results show up in the app's own UI, end-to-end, with no real cloud.

Today this isn't quite reachable out of the box, and the docs intentionally scope cloudemu to in-process test usage (`httptest.NewServer`, "keep in `_test.go`", "never import from production"). That's perfect for tests — this is an **additive** ask for a local dev-cloud mode.

### What's missing for the minikube-style flow
1. **A standalone runner** — there's no `cmd/` / no `ListenAndServe` anywhere, so each user must hand-roll `http.ListenAndServe(":port", azureserver.New(Drivers{...}))`. A first-class `cloudemu serve` (per-provider, stable configurable ports) would be the `minikube start` equivalent.
2. **TLS for a real listener** — the Azure SDK requires HTTPS; in tests `httptest.NewTLSServer` auto-certs. A standalone server needs a self-signed cert + a documented trust path (CA file, or an insecure-skip-verify flag) for out-of-process clients.
3. **A documented, stable auth/cloud-config story for external clients** — i.e. how to point `azcore`'s `cloud.Configuration` (ResourceManager endpoint + **authority/token endpoint**) at the running server. There are already `login.microsoftonline`/`GetToken` references internally; exposing the token/authority endpoint as a stable, documented standalone endpoint (so `azidentity` against a fake cred just works over the network) would close the loop. Same idea for AWS/GCP auth.
4. **State seeding + optional persistence** — state is in-process memory, so a freshly-started server is empty. A way to seed resources at startup (config file / admin API) and optionally persist across restarts — like minikube's persistent cluster — so discovery flows find something.
5. **An endpoints/“cloud-config” bundle** — the kubeconfig equivalent: one artifact listing the emulated endpoints so an app can be pointed at the whole emulated cloud at once.

### Why it's valuable
This turns cloudemu from an excellent in-process test double into a full **local dev cloud** for multi-service apps — you could bring up your real services against it and exercise live, UI-visible workflows without real accounts or Docker. The existing test-only mode stays exactly as is.

Totally understand if this is out of scope vs. the library's focus — wanted to capture the use case. Thanks!

[...]

PR #277, opened 2026-07-25
feat: standalone server mode — cloudemu serve (#224)

NitinKumar004:
Implements the packaging half of #224 — the "minikube for cloud" standalone mode. The HTTP core was already reachable (each `server.New()` returns an `http.Handler`); this makes it a first-class, out-of-process server a real user can run and point any SDK at.

**Purely additive** — the in-process test-double API (`cloudemu.NewAWS()`, `awsserver.New(Drivers{…})`) is untouched.

## What's new

**`cmd/cloudemu serve`** — one binary boots all three providers on stable ports:

| Provider | Port | Proto |
|----------|------|-------|
| AWS | 4566 | HTTP (LocalStack's port) |
| Azure | 4568 | HTTPS (ARM SDK requires TLS) |
| GCP | 4569 | HTTP |
| Kubernetes | 4570 | HTTP (shared EKS/AKS/GKE data-plane) |

- **Self-signed TLS** generated in memory for Azure (SANs cover localhost + loopback + `--tls-host`); or supply your own with `--tls-cert`/`--tls-key`.
- **Endpoints bundle** printed on startup and written as JSON with `--endpoints-file`, so an app can be wired at the whole emulated cloud at once. Startup banner prints per-SDK snippets.
- **Config flags** map to existing `config.Option`s: `--account-id`, `--region`, `--project-id`, `--latency`. `--providers=aws,gcp` starts a subset; `--host 0.0.0.0` exposes on the network (default local-only).
- Graceful shutdown on SIGINT/SIGTERM; optional `--log-requests`; fail-fast port binding before the banner.

**Auto-wiring helpers** — `DriversFrom(provider)` + `NewFromProvider(provider)` in each `server/*` package turn a fully-constructed provider into a running server without hand-mapping 27/36/22 driver fields. Providers now expose their identity (`AccountID`/`Region`, `SubscriptionID`/`Region`, `ProjectID`/`Region`).

## Testing

`cmd/cloudemu/serve_test.go` builds the binary and runs it **as a separate process**, then drives:
- a real `aws-sdk-go-v2` S3 client (CreateBucket + ListBuckets) over the HTTP socket, and
- a real `azure-sdk-for-go` ARM client (namespace CreateOrUpdate + Get) over the **self-signed HTTPS** socket.

Plus per-provider `NewFromProvider` roundtrip tests. Full `go test ./...`, `go vet ./...`, `gofmt` all clean.

## Deferred (follow-ups)

- **State seeding / persistence** — needs a cross-service state-import loader (#250). The server starts empty each run; noted in the docs.
- **Docker image (#247)** and **Testcontainers module (#248)** build directly on this binary and are the natural next steps.

Docs: `docs/standalone-server.md` (quickstart, ports, per-SDK wiring, trust-the-cert) + README section.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,6 @@ CloudEmu_Research_Paper.docx
 _paper_imgs/
 generate_paper.py
 .claude/
+
+# standalone server binary
+/cloudemu
--- a/README.md
+++ b/README.md
@@ -70,6 +70,20 @@ client.PutObject(ctx, &s3.PutObjectInput{ /* … */ }) // hits the in-memory bac
 
 Equivalent setups for Azure (`azureserver.New`) and GCP (`gcpserver.New`) are in [docs/sdk-server.md](docs/sdk-server.md).
 
+### Standalone server
+
+Prefer a long-lived process you point out-of-process apps at? Run the emulator
+as a server and point any SDK — any language — at the printed endpoints:
+
+```sh
+go run ./cmd/cloudemu serve
+#   AWS         http://127.0.0.1:4566
+#   Azure       https://127.0.0.1:4568   (self-signed TLS)
+#   GCP         http://127.0.0.1:4569
+```
+
+Full flags, port scheme, and per-SDK wiring: [docs/standalone-server.md](docs/standalone-server.md).
+
 The snippet above is a quick taste. To adopt cloudemu in a real app, don't write a demo — wire it into your existing client and tests so your real code runs against it. See [docs/integration.md](docs/integration.md).
 
 ## Or use the Go API directly
--- a/cmd/cloudemu/endpoints.go
+++ b/cmd/cloudemu/endpoints.go
@@ -1,0 +1,57 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+)
+
+// endpointSet is the machine-readable bundle of URLs a client points at. Empty
+// fields (a provider that wasn't started) are omitted from the JSON.
+type endpointSet struct {
+	AWS        string `json:"aws,omitempty"`
+	Azure      string `json:"azure,omitempty"`
+	GCP        string `json:"gcp,omitempty"`
+	Kubernetes string `json:"kubernetes,omitempty"`
+}
+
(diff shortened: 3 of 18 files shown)


NitinKumar004:

Reviewed the full diff by hand, with emphasis on the two highest-risk areas for this kind of change: the mechanical `DriversFrom` mappings and the server lifecycle.

### Verified

- **`DriversFrom` completeness** — every field of each `Drivers` struct is populated: AWS 27/27, Azure 36/36, GCP 22/22 (mechanical field-diff, zero missing). A dropped field would mean a service silently 404s in standalone mode; none are dropped.
- **Same-type mappings** (the one class Go's type system can't catch) are correct: AWS `RDS`→`p.RDS` and `Redshift`→`p.Redshift` are distinct; the Azure compute family (`VirtualMachines`/`Disks`/`Snapshots`/`Images`/`SSHPublicKeys`) and the paired control/data-plane fields (`Databricks`, `AzureAI`, `Search`) intentionally share one mock, matching the in-process wiring.
- **Shared K8s** — `d.K8sAPI` is consumed by each `server.New` (registered when non-nil), so injecting one `kubernetes.NewAPIServer()` across all three providers is correct.
- **Lifecycle** — all listeners bind *before* serving (port clash fails fast, before the banner promises endpoints that never came up); earlier listeners are closed on a mid-loop bind failure; buffered `errCh` so serve goroutines never block; `Shutdown` runs on every server on SIGINT/SIGTERM; `http.ErrServerClosed` is filtered.
- **Security posture** — default bind is `127.0.0.1` (network exposure is opt-in via `--host`); `InsecureSkipVerify` appears only in the test (annotated `#nosec G402`), never in the binary.

### Cleanups applied during review

- Removed an unused `providers` param from `printBanner`.
- Moved the test-only `waitReady` helper out of the non-test `endpoints.go` into `serve_test.go`.
- Set `MinVersion: tls.VersionTLS12` on both TLS config paths.

**No blocking defects.** `go build ./...`, `go vet ./...`, `gofmt`, and full `go test ./...` clean; the out-of-process integration test drives real AWS (HTTP) and Azure (self-signed HTTPS) SDK clients against the running binary and passes.

PR #279, opened 2026-07-25
fix: back Azure Service Bus namespaces with state — tags + scoped listing (#278)

### LINKED ISSUE #278: Azure Service Bus namespaces are stubbed — no tags, empty ListByResourceGroup
The Service Bus **namespace** control plane in `server/azure/servicebus/handler.go` is a stub with no backing state, so real SDK clients get wrong results:

- **`Get` drops tags** — `createNamespace` never persists or echoes `req.Tags` (there's no `Tags` field on the namespace resource), so a namespace created with tags comes back with none.
- **`ListByResourceGroup` is always empty** — `listNamespaces` returns `[]` with the comment *"We don't track namespaces in the driver."*
- `getNamespace` returns a synthetic `Succeeded` namespace regardless of what was created.

The *queue* control plane (driver-backed) works; only the virtual *namespace* resource is stubbed.

### Repro
Point a real `armservicebus` client at the emulator, `BeginCreateOrUpdate` a namespace with `Tags: {env: prod}`, then `Get` (tags empty) and `NewListByResourceGroupPager` (namespace absent).

### Fix
Give namespaces driver-backed state — persist name, tags, and creation scope (subscription/resource group), and filter `ListByResourceGroup` by it. Same pattern as the #259 series (scope-aware listing + tags on create/update).

NitinKumar004:
Fixes #278.

Surfaced by an out-of-process real-SDK run against `cloudemu serve`: Service Bus namespaces were a stub, so a real `armservicebus` client got wrong results.

## Problem

`server/azure/servicebus/handler.go` didn't track namespaces:
- `createNamespace` never persisted or echoed `req.Tags` (no `Tags` field existed),
- `getNamespace` returned a synthetic `Succeeded` namespace regardless of what was created,
- `listNamespaces` always returned `[]` ("We don't track namespaces in the driver").

(The *queue* control plane, which is driver-backed, already worked — only the virtual *namespace* container was stubbed.)

## Fix

Give the handler an in-memory namespace store keyed by name (namespace names are globally unique in Azure):
- **CreateOrUpdate** stores location, tags, SKU, and the request's subscription/resource-group scope, and re-applies them on a repeat PUT (ARM PUT semantics).
- **Get** returns the stored record — and a proper `404 ResourceNotFound` when the namespace doesn't exist, instead of a fake `Succeeded`.
- **List** filters by the request path's subscription/resource-group through `services/scope` (the same scope semantics as the #259 series), in deterministic sorted order.

## Testing

New real-SDK regression `TestSDKNamespaceTagsAndScopedListing` (armservicebus over TLS): tags survive create→Get, `ListByResourceGroup` returns only that group's namespaces, and a deleted namespace 404s on Get. Existing namespace/queue/data-plane tests still pass.

Verified end-to-end: the same real-user run that found this now passes 19/19 across S3, DynamoDB, SQS, GCS, and Service Bus against the running `cloudemu serve` binary. Full `go test ./...`, `go vet ./...`, `gofmt` clean.

### DIFF
--- a/server/azure/servicebus/handler.go
+++ b/server/azure/servicebus/handler.go
@@ -27,12 +27,15 @@ import (
 	"context"
 	"encoding/json"
 	"io"
+	"maps"
 	"net/http"
 	"strings"
 
 	cerrors "github.com/stackshy/cloudemu/v2/errors"
+	"github.com/stackshy/cloudemu/v2/internal/memstore"
 	"github.com/stackshy/cloudemu/v2/server/wire/azurearm"
 	mqdriver "github.com/stackshy/cloudemu/v2/services/messagequeue/driver"
+	"github.com/stackshy/cloudemu/v2/services/scope"
 )
 
 const (
@@ -46,12 +49,13 @@ const (
 
 // Handler serves ARM Service Bus + raw-HTTP data-plane requests.
 type Handler struct {
-	mq mqdriver.MessageQueue
+	mq         mqdriver.MessageQueue
+	namespaces *memstore.Store[namespaceState]
 }
 
 // New returns a Service Bus handler backed by mq.
 func New(mq mqdriver.MessageQueue) *Handler {
-	return &Handler{mq: mq}
+	return &Handler{mq: mq, namespaces: memstore.New[namespaceState]()}
 }
 
 // Matches accepts ARM Microsoft.ServiceBus/namespaces[...] paths plus
@@ -84,7 +88,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	// /providers/Microsoft.ServiceBus/namespaces (subscription-level list)
 	if rp.ResourceName == "" {
-		h.listNamespaces(w, r)
+		h.listNamespaces(w, rp)
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


NitinKumar004:

Reviewed the change; it's small and self-contained. No blocking defects.

Verified:
- **Thread-safety** — namespace state lives in `memstore.Store` (RWMutex-guarded), same as every other store in the tree.
- **ARM PUT semantics** — a repeat PUT overwrites the stored record, so tags/SKU/scope are re-applied (not merged), matching real CreateOrUpdate.
- **404 on missing** — `getNamespace` now returns `ResourceNotFound` for an unprovisioned namespace instead of a fake `Succeeded`; no existing test does a GET-without-PUT, and queue routes never touch namespace state, so nothing regresses.
- **Scope filtering** reuses `services/scope` — subscription-level lists match all groups in the subscription, RG-level lists match only that group; deterministic order via `SortedValues()`.

Cleanups applied during review:
- `listNamespaces` now takes `rp` from the router instead of re-parsing the path (drops an ignored error return).
- Tags are `maps.Clone`d on store, matching the boundary-isolation pattern from the #259 series.

`go build`, `go vet`, `gofmt`, full `go test ./...` clean; the real-user run against `cloudemu serve` passes 19/19.

NitinKumar004:
Applied review fixes:
- **Get/Delete now scope-checked** — a request naming the wrong subscription/resource group returns `404` on Get and does not delete on Delete (was: name-only lookup, wrong-scope requests wrongly succeeded). Shared `namespaceInScope` helper, reused by List for consistency.
- **SKU deep-copied** on store (`cloneSKU`), matching the `maps.Clone` isolation already applied to tags.
- Regression extended: wrong-RG Get → 404, wrong-RG Delete leaves the namespace intact.

Full `go test ./...`, vet, gofmt clean.

PR #280, opened 2026-07-25
fix(dns): deterministic record listing across the #143 slices (#259)

NitinKumar004:
The last unswept ordering gap from #259. PRs #275/#272 made list ordering deterministic across the storage/database/logging/eventbus/cache/notification slices, and DNS **zone** listing was fixed — but DNS **record** listing was missed.

## Problem

`ListRecords` in all three DNS providers iterated `memstore` in map order (`Filter()` / `All()`), so a real client got records back in a different order on every call.

## Fix

Iterate `SortedValues()` and filter to the requested zone, so records come back in a stable order keyed by `zoneID:name:type[:setID]` — matching the deterministic ordering real DNS APIs return, and consistent with how zones (and every other #143 slice) already list.

Touched: `providers/aws/route53`, `providers/azure/azuredns`, `providers/gcp/clouddns`.

## Testing

Per-provider `TestListRecordsOrderingDeterministic` (create a zone + several records, assert `ListRecords` returns identical order across repeated calls). Full `go test ./...`, `go vet`, `gofmt` clean.

## Remaining #253 items (separate follow-ups)

This closes the #259 ordering item for DNS. The broader #253 fidelity work — zone-name uniqueness + scope attribution, `UpdateZone`, structured TXT chunks (`[][]string`), and change-batch atomicity — is larger and independent; I'll take it as its own PR(s).

### DIFF
--- a/providers/aws/route53/ordering_test.go
+++ b/providers/aws/route53/ordering_test.go
@@ -40,3 +40,48 @@ func TestListOrderingDeterministic(t *testing.T) {
 		}
 	}
 }
+
+// TestListRecordsOrderingDeterministic locks the #259 ordering guarantee for
+// record listing: ListRecords must return the same, defined order on every
+// call (map iteration randomness must never reach the wire).
+func TestListRecordsOrderingDeterministic(t *testing.T) {
+	m := newTestMock()
+	ctx := context.Background()
+
+	zone, err := m.CreateZone(ctx, driver.ZoneConfig{Name: "example.com"})
+	if err != nil {
+		t.Fatalf("create zone: %v", err)
+	}
+
+	for _, name := range []string{"zeta", "alpha", "mid", "beta", "omega"} {
+		if _, err := m.CreateRecord(ctx, driver.RecordConfig{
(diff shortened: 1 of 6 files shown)


NitinKumar004:

Reviewed the full diff and audited sibling read paths. The `ListRecords` change is correct — records filter on the `ZoneID` field (not a key-prefix), so cross-zone contamination is impossible, and `SortedValues()` keys by `zoneID:name:type[:setID]`.

**One sibling gap found and fixed (pushed):** `GetRecord`'s weighted-record fallback still iterated `m.records.All()` and returned the *first* map-order match — so a name+type with several weighted records (distinct set IDs, no plain record) resolved to a *different* record on different calls, the exact defect class this PR set out to remove. Fixed to iterate `SortedValues()` and return the lowest set ID deterministically, across all three providers, with a `TestGetRecordWeightedDeterministic` regression.

Audited the remaining `.All()` uses — `DeleteZone` (cascade-delete all zone records) and `DeleteRecord` (set-ID sweep) delete *every* match, so iteration order is irrelevant there. No other order-sensitive read path remains.

Full `go test ./...`, `go vet`, `gofmt` clean.

PR #281, opened 2026-07-25
fix(dns): scope-aware zone listing + CreateOrUpdate applies tags (#259 DNS slice)

NitinKumar004:
Completes the #259 fidelity work for the DNS slice — the same template PR #276 applied to logging/eventbus/cache/notification. With #280 (record-listing determinism) already merged, this closes out #259 for DNS.

## What changed

**Driver** (`services/dns/driver`): `ZoneConfig`/`ZoneInfo` gain a `Scope` field; `ListZones` takes a `scope.Scope` filter; new `UpdateZone` applies mutable fields (tags) with ARM CreateOrUpdate-on-existing semantics.

**Providers** (route53 / azuredns / clouddns): store the creating scope on each zone, filter `ListZones` by it in deterministic `SortedValues()` order, implement `UpdateZone`. azuredns derives its stored ARM id's subscription/resource-group from the scope.

**Handlers**:
- Azure `createOrUpdateZone` now applies the request's tags via `UpdateZone` on an existing zone instead of echoing the stale one; lists pass the request's `{subscription, resourceGroup}` scope.
- GCP lists filter by the request's project.
- AWS stays account-global (zero scope) — Route53 hosted zones aren't RG/project-scoped.

The portable wrapper keeps its no-arg `ListZones` (passing zero scope internally, like the other domains); the chaos wrapper threads the filter; topology resolves across all zones.

## Testing

New real-SDK tests: Azure scoped listing (rg-team-a vs rg-team-b), upsert-applies-tags, request-derived zone id; GCP per-project zone isolation; AWS provider-level `UpdateZone` (applies tags / NotFound). Full `go test ./...`, `go vet`, `gofmt` clean.

## Closes
Finishes the #259 series (ordering in #275/#280, scope+upsert+IDs in #276 and here). The broader #253 enhancements — zone-name uniqueness, structured TXT chunks (`[][]string`), and change-batch atomicity — remain as separate follow-ups.

### DIFF
--- a/features/chaos/wrappers_baseline_test.go
+++ b/features/chaos/wrappers_baseline_test.go
@@ -104,7 +104,7 @@ func TestWrapDNSBaseline(t *testing.T) {
 
 	z, _ := d.CreateZone(ctx, dnsdriver.ZoneConfig{Name: "b.test."})
 	_, _ = d.GetZone(ctx, z.ID)
-	_, _ = d.ListZones(ctx)
+	_, _ = d.ListZones(ctx, scope.Scope{})
 	rec := dnsdriver.RecordConfig{ZoneID: z.ID, Name: "x.b.test.", Type: "A", TTL: 300, Values: []string{"1.2.3.4"}}
 	_, _ = d.CreateRecord(ctx, rec)
 	_, _ = d.GetRecord(ctx, z.ID, "x.b.test.", "A")
--- a/features/chaos/wrappers_dns.go
+++ b/features/chaos/wrappers_dns.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	dnsdriver "github.com/stackshy/cloudemu/v2/services/dns/driver"
+	"github.com/stackshy/cloudemu/v2/services/scope"
 )
 
 // chaosDNS wraps a DNS driver. Hot-path: zone and record CRUD.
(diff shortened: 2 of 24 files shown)


NitinKumar004:

Reviewed the full 24-file diff. The scope/list/UpdateZone template matches PR #276, and completeness is verified (all three providers implement the new `ListZones` signature + `UpdateZone`; full build/vet/suite green).

**One real bug found and fixed (pushed):** `UpdateZone` matched zones by **name only**, but Azure/GCP legitimately allow the *same zone name in different resource groups/projects*. A `CreateOrUpdate` for `shared.com` in rg-B, when `shared.com` already existed in rg-A, took the update path and **hijacked the rg-A zone** (reassigned its scope) instead of creating a distinct rg-B zone. Fixed:
- `UpdateZone` now matches on name **and** scope (azuredns + clouddns).
- The Azure `createOrUpdateZone` handler now does try-update-then-create (falls through to Create on NotFound) instead of a name-only existence probe.
- Regression `TestSDKSameNameZonesStayIndependent`: two `shared.com` zones in different RGs keep their own tags after both are created.

**Noted, not fixed here (pre-existing, broader):** `resolveZoneID` resolves a zone by name across all scopes, so a GET/record op for a same-named zone can resolve to an arbitrary one of them. That's a read-path fidelity gap that predates this PR and touches every record handler path; better as its own change. The regression above asserts via the (correctly scope-filtered) list path to avoid depending on it.

Full `go test ./...`, `go vet`, `gofmt` clean.

PR #282, opened 2026-07-25
fix(dns): scope Azure resolveZoneID to the request's resource group (#253)

NitinKumar004:
Follow-up to #281 (the DNS scope/upsert work), closing the read-path counterpart of the same bug.

## Problem

Azure `resolveZoneID` scanned **all** zones by name across every scope:

```go
zones, _ := h.dns.ListZones(ctx, scope.Scope{})  // every subscription/RG
```

The same DNS zone name legitimately exists in more than one resource group, so a `Get` or any record-set operation for such a name could resolve to a zone in a **different** group — returning the wrong zone's id, tags, and records.

## Fix

Scope the scan to the request's subscription + resource group (parsed from the ARM path). `resolveZoneID` now takes the `*azurearm.ResourcePath` and filters `ListZones` by `{subscription, resourceGroup}`; the six call sites pass `rp`. GCP's `resolveZoneID` already scoped by project — this brings Azure in line.

## Testing

New real-SDK regression `TestSDKGetResolvesWithinRequestScope`: create `dup.com` in two resource groups with different tags, then `Get` each group's `dup.com` and confirm each returns its own tag (not the other group's). Existing DNS zone/record/error SDK tests still pass. Full `go test ./...`, `go vet`, `gofmt` clean.

## #253 status
This clears the `resolveZoneID` item noted on #281. Remaining #253 follow-ups: zone-name uniqueness, structured TXT chunks (`[][]string`), and change-batch atomicity.

### DIFF
--- a/server/azure/dns/operations.go
+++ b/server/azure/dns/operations.go
@@ -51,7 +51,7 @@ func (h *Handler) createOrUpdateZone(w http.ResponseWriter, r *http.Request, rp 
 }
 
 func (h *Handler) getZone(w http.ResponseWriter, r *http.Request, rp *azurearm.ResourcePath) {
-	id, err := h.resolveZoneID(r.Context(), rp.ResourceName)
+	id, err := h.resolveZoneID(r.Context(), rp)
 	if err != nil {
 		azurearm.WriteCErr(w, err)
 		return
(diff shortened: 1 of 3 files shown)


NitinKumar004:

Small, focused change; no findings.

Verified:
- All 6 `resolveZoneID` call sites now pass `rp` (grep confirms no name-only caller remains); build/vet clean.
- Record-set paths carry the zone in `rp.ResourceName` and the record in `rp.SubResource*`, so scoping on `{subscription, resourceGroup}` is correct for both zone and record ops.
- **No cross-API regression:** a zone created via the portable API has a zero scope, and `Scope.Matches` returns true for a zero source scope against any filter — so scoped wire lookups still find portable-created zones.
- ARM zone/record paths always include the resource group, so scoping never hides a legitimately-addressed zone.

Regression `TestSDKGetResolvesWithinRequestScope` fails without the fix (Get resolves to an arbitrary same-named zone) and passes with it. Full `go test ./...`, `go vet`, `gofmt` clean.

PR #283, opened 2026-07-25
fix(dns): chunk Azure TXT values into ≤255-byte character-strings (#253)

NitinKumar004:
Next #253 follow-up — TXT record chunking fidelity.

## Problem

A DNS TXT record is a sequence of character-strings, each capped at 255 bytes; Azure models this as `TxtRecords: [{value: [chunk1, chunk2, …]}]`. But `toRecordSetProperties` emitted each logical value as a **single** array element:

```go
props.TxtRecords = append(props.TxtRecords, txtRecordJSON{Value: []string{v}})
```

So a TXT value longer than 255 bytes (a DKIM key, a long SPF record) went out as one oversized chunk — invalid on the Azure wire.

## Fix

Split TXT values into ≤255-byte chunks on output (`chunkTXT`). The input side already rejoins the chunks into the logical value (`strings.Join(t.Value, "")`), so a long value that comes in as valid ≤255 chunks round-trips back to the same chunks.

**Scope:** Azure-only. Route53 and GCP Cloud DNS carry TXT values in presentation format (the SDK sends the quoted string as-is) and store/return them verbatim, so they need no chunking here.

## Design note

I kept the driver's `Values []string` flat (one logical value per TXT record) rather than widening it to `[][]string`, and chunk at the wire layer where the 255-byte rule actually lives — the same place every other record type's wire shape is built. This fixes the real fidelity gap (valid chunked TXT on the wire) without a TXT-specific field rippling through every record type and provider. Happy to go to `[][]string` if you want byte-exact preservation of arbitrary user-chosen chunk boundaries.

## Testing

`TestSDKTXTRecordChunking`: a 355-byte TXT value round-trips as ≤255-byte chunks whose concatenation equals the original. Full `go test ./...`, `go vet`, `gofmt` clean.

## #253 status
Remaining follow-ups: zone-name uniqueness on create, and change-batch atomicity.

### DIFF
--- a/server/azure/dns/scope_update_sdk_test.go
+++ b/server/azure/dns/scope_update_sdk_test.go
@@ -127,6 +127,53 @@ func TestSDKSameNameZonesStayIndependent(t *testing.T) {
 	}
 }
 
+// TestSDKTXTRecordChunking asserts that a TXT value longer than 255 bytes is
+// returned as valid ≤255-byte character-strings whose concatenation preserves
+// the original value — Azure rejects a single oversized chunk.
+func TestSDKTXTRecordChunking(t *testing.T) {
+	zones, records := newDNSClients(t)
+	ctx := context.Background()
+
+	if _, err := zones.CreateOrUpdate(ctx, testRG, "txt.com", armdns.Zone{Location: to.Ptr("global")}, nil); err != nil {
+		t.Fatalf("Zones.CreateOrUpdate: %v", err)
+	}
+
+	chunkA := strings.Repeat("a", 255)
+	chunkB := strings.Repeat("b", 100)
+	want := chunkA + chunkB // 355-byte logical value
+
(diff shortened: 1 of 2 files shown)


NitinKumar004:

No findings. Verified:
- **Boundary math** — exercised `chunkTXT` at lengths 0/1/255/256/510/511: every chunk ≤255 bytes, exact reassembly, no off-by-one (255 stays one chunk; 256 → [255,1]; 510 → [255,255]; 511 → [255,255,1]).
- **Octet-correct** — TXT character-strings are octet sequences per the DNS spec, so byte-slicing (`s[:255]`) is the right unit, not runes.
- **Scope** — only the TXT case of `toRecordSetProperties` changed; input join and all other record types untouched; existing TXT/record tests still pass.

Regression `TestSDKTXTRecordChunking` fails pre-fix (single 355-byte chunk) and passes after. Full `go test ./...`, `go vet`, `gofmt` clean.

PR #284, opened 2026-07-25
fix(dns): make Route53 ChangeResourceRecordSets atomic (#253)

NitinKumar004:
Next #253 follow-up — change-batch atomicity.

## Problem

`changeResourceRecordSets` applied a batch's changes in a loop and returned on the first failure:

```go
for i := range req.ChangeBatch.Changes {
    if err := h.applyChange(...); err != nil { writeChangeErr(w, err); return }
}
```

So a batch that failed on, say, the third change left the first two **committed** — a half-applied batch. Real Route 53 validates the entire batch and applies nothing if any change is invalid (`InvalidChangeBatch`).

## Fix

Validate every change against the zone's current record sets *before* mutating anything, folding each change's effect into a simulated state so intra-batch ordering is respected (a DELETE followed by a CREATE of the same record set is allowed). Only if the whole batch is valid are the changes applied. This mirrors the pre-validation the **GCP Cloud DNS** `changes` handler already does — so both wire surfaces are now atomic. (Azure DNS has no batch API; record sets are individual PUTs, each already atomic.)

Validation rejects: an unknown action, a CREATE of an already-present set, and a DELETE of a missing set — the deterministic InvalidChangeBatch cases. A mid-apply transient/injected error remains out of scope (same caveat noted in the GCP handler).

## Testing

`TestSDKRoute53ChangeBatchAtomic`: a batch of `[CREATE keep, DELETE ghost(missing)]` is rejected, and `keep` is confirmed **absent** afterward — proving nothing half-applied. Fails pre-fix, passes after. Full `go test ./...`, `go vet`, `gofmt` clean.

## #253 status
Remaining: zone-name uniqueness on create (the last item).

### DIFF
--- a/server/aws/route53/operations.go
+++ b/server/aws/route53/operations.go
@@ -3,6 +3,7 @@ package route53
 import (
 	"encoding/xml"
 	"net/http"
+	"strings"
 	"time"
 
 	cerrors "github.com/stackshy/cloudemu/v2/errors"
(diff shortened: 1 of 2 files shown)


NitinKumar004:

Audited whether validation covers **every** deterministic apply-time failure — otherwise a change could still half-apply after validation passed.

Walked the driver's failure modes:
- **zone not found** → `validateChangeBatch` calls `ListRecords` first, which errors on a missing zone → whole batch rejected before any apply. ✓
- **CREATE conflict / DELETE missing** → caught by the simulated `present` map (keyed on name+type+setID, matching the driver's `recordKey`). ✓
- **empty name/type** (`CreateRecord` returns InvalidArgument) → this was **not** pre-caught by presence-only validation, so a malformed CREATE could half-apply. **Fixed:** added a name/type non-empty check to the pre-flight pass.
- **UPSERT** → `upsertRecord` resolves to update-or-create against real state; both succeed once presence is validated. ✓

With the name/type guard added, every deterministic `CreateRecord`/`UpdateRecord`/`DeleteRecord` error is now caught before mutation, so the batch is genuinely all-or-nothing. The only residual is a mid-apply transient/injected (chaos) error — out of scope for batch validity, same caveat as the GCP handler.

Intra-batch ordering verified: DELETE→CREATE of the same set is allowed (delete clears `present`, create then succeeds); CREATE→DELETE likewise.

`TestSDKRoute53ChangeBatchAtomic` passes; full `go test ./...`, `go vet`, `gofmt` clean.

PR #285, opened 2026-07-25
fix(dns): enforce GCP managed-zone name uniqueness within a project (#253)

NitinKumar004:
Final #253 follow-up — zone-name uniqueness.

## Problem

GCP Cloud DNS `clouddns` keyed zones by a random id, so the same managed-zone name could be created any number of times in one project. Real Cloud DNS requires the managed-zone name to be **unique within a project**.

## Fix

`CreateZone` now rejects a duplicate name in the same project with `AlreadyExists` (mapped to HTTP 409 by the GCP wire layer). The check is gated on a non-empty project scope, so it targets the wire path exactly and leaves the portable API (which creates with a zero scope) unaffected.

**Scope:** GCP-only. AWS Route 53 legitimately allows duplicate zone names (each gets a distinct id — correct as-is), and Azure DNS is already unique by construction (its ARM id is `f(subscription, resource group, name)`, so a same-name+scope create maps to the same key, and the handler routes an existing zone to `UpdateZone`).

## Testing

`TestSDKCloudDNSDuplicateZoneName`: creating `dup-zone` twice in the same project — the second create fails. Full `go test ./...`, `go vet`, `gofmt` clean.

## #253 complete
With this, all #253 items are addressed: record-listing determinism (#280), scope-aware zone listing + `UpdateZone` (#281), `resolveZoneID` read-path scoping (#282), TXT ≤255-byte chunking (#283), change-batch atomicity (#284), and zone-name uniqueness (this PR).

### DIFF
--- a/providers/gcp/clouddns/dns.go
+++ b/providers/gcp/clouddns/dns.go
@@ -52,6 +52,18 @@ func (m *Mock) CreateZone(_ context.Context, cfg driver.ZoneConfig) (*driver.Zon
 		return nil, cerrors.New(cerrors.InvalidArgument, "zone name is required")
 	}
 
+	// Managed-zone names are unique within a project. Reject a duplicate in the
+	// same project (the wire always carries one); the portable API, which
+	// creates with a zero scope, is unaffected.
+	if cfg.Scope.Project != "" {
+		for _, z := range m.zones.SortedValues() {
+			if z.Name == cfg.Name && z.Scope.Project == cfg.Scope.Project {
+				return nil, cerrors.Newf(cerrors.AlreadyExists,
+					"managed zone %q already exists in project %q", cfg.Name, cfg.Scope.Project)
+			}
+		}
+	}
+
 	id := idgen.GenerateID("zone-")
 
 	tags := make(map[string]string, len(cfg.Tags))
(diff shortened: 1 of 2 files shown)


NitinKumar004:

No blocking findings.

Verified:
- **Right identity** — uniqueness is keyed on the managed-zone `Name` (not `dnsName`), matching Cloud DNS, so two zones may share a DNS domain but not a resource name.
- **Wire path always checked** — the handler's `Matches` requires a non-empty project (`rt.project != ""`), so every create carries a project and hits the gate; the portable API (zero scope) is intentionally skipped, and existing zero-scope provider tests are unaffected.
- **Correct semantics** — GCP Create is not an upsert, so rejecting a duplicate (409) is the right behavior; AWS (duplicate names allowed) and Azure (unique by id construction) are correctly left alone.

Minor, non-blocking: the check-then-set isn't atomic (two concurrent creates of the same name could race), the same concurrency caveat already documented on the GCP `changes` handler. A name index/lock would be overkill for the in-memory emulator; real usage is single-writer per test.

`TestSDKCloudDNSDuplicateZoneName` passes; full `go test ./...`, `go vet`, `gofmt` clean.

Issue #244, opened 2026-07-11
feat: admin/state HTTP API (reset, seed, snapshot, restore) on the SDK-compat server


NitinKumar004:
The single biggest DX unlock for the SDK-compat surface: out-of-process / parallel-test users currently have no way to reset or seed emulator state between tests. Pairs with #224 (standalone server) and #107 (snapshots).

**Build:** a small control-plane handler mounted at `/_cloudemu/{reset,seed,snapshot,restore}` on `server.Server`. `reset` clears all wired drivers; `seed` bulk-loads fixtures; `snapshot`/`restore` reuse the state model from #107. The `recorder` package already provides call capture; this adds state lifecycle.

**Acceptance:** HTTP roundtrip test seeds resources, reads them back via a real SDK client, calls reset, and confirms the backend is empty.

NitinKumar004:
Done. The `/_cloudemu/*` control plane is implemented in `server/admin/admin.go` — `reset`, `seed`, `snapshot`, `restore` — with state serialization in `persist/`, wired into the standalone server (`cmd/cloudemu`). Closing.

Issue #107, opened 2026-04-12
Add Time Travel with State Snapshots and Fork


NitinKumar004:
## Summary

Add the ability to snapshot, restore, replay, and fork cloud state — enabling "what-if" testing and state management.

## Core Operations

```go
// Snapshot entire cloud state
snapshot := provider.Snapshot()

// Restore to a previous state
provider.Restore(snapshot)

// Fork state for parallel what-if testing
forked := provider.Fork()
// mutations on forked don't affect original

// Record and replay
recording := provider.StartRecording()
// ... perform operations ...
provider.StopRecording()
recording.Replay(provider, replay.WithSpeed(10)) // 10x speed
```

## Use Cases

- **Snapshot before destructive tests**: take snapshot, run test that deletes everything, restore
- **What-if testing**: fork state, try different approaches on each fork, compare results
- **Regression replay**: record a production-like session, replay against new code
- **Deterministic debugging**: replay exact sequence of operations that caused a bug

## Implementation Notes

- Snapshot serializes all memstore state across all services to a portable format (JSON or gob)
- Restore deserializes and replaces all memstore contents
- Fork creates a deep copy of all state
- Recording captures operation name, inputs, timestamps (already partially done via recorder)
- Replay re-executes recorded operations against a fresh or existing provider

## Acceptance Criteria

- [ ] Snapshot captures complete state across all services
- [ ] Restore replaces state and operations work against restored state
- [ ] Fork creates independent copy (mutations don't cross)
- [ ] Recording captures all operations with inputs
- [ ] Replay re-executes operations in order
- [ ] Tests for snapshot/restore roundtrip, fork independence, replay accuracy
- [ ] Linter passes

NitinKumar004:
Partial.

**Done:** snapshot + restore. Whole-emulator state serialization lives in `persist/`, exposed over the control plane at `/_cloudemu/{snapshot,restore}` (`server/admin`). Two of the three acceptance items — snapshot and restore — are covered.

**Left:** **Fork** (independent deep-copy for what-if testing) and **record/replay** (the recorder captures calls but can't replay them yet).

NitinKumar004:
Snapshot registry + rewind + fork (isolated by construction, over the full-surface persist ExportAll/RestoreAll, exposed on the `/_cloudemu/snapshot/{name}` admin surface) shipped in #926.

[...]

PR #287, opened 2026-07-25
feat: /_cloudemu/reset control plane for standalone-server test isolation (#244)

NitinKumar004:
Implements the **reset** half of #244 — the biggest DX gap left after the standalone server (#224): a long-lived `cloudemu serve` process shared by an out-of-process or parallel test suite had no way to get a clean slate between tests.

## What's new

A control plane at `/_cloudemu` on the standalone server (on by default; `--admin=false` to disable):

- **`POST /_cloudemu/reset`** — rebuild every provider (and the shared Kubernetes data-plane) to empty state and swap it in atomically. In-flight requests finish against the old state; new requests see the fresh one. A POST to any provider's port resets the whole emulator.
- **`GET /_cloudemu/health`** — liveness.
- **`/_cloudemu/seed`** — reserved, returns `501` (bulk seeding needs the cross-service fixture loader in #250; until then, create resources with your SDK after reset).

## Why this shape

New `server/admin` package: a hot-swappable `Backend` + a `Control` that routes the `/_cloudemu` prefix and passes everything else through. `reset` reuses the existing `NewAWS(opts)` / `DriversFrom` construction path to build a fresh provider+server and swap it in — so **no mock, wire-handler, driver, or core `server.Server` code changes**. The state lifecycle lives only in the standalone binary's wiring, exactly where out-of-process reset belongs (in-process users just construct a fresh provider).

## Testing

- **Acceptance (out-of-process, real SDK):** `serve_test.go` boots the binary, creates an S3 bucket with a real `aws-sdk-go-v2` client, `POST`s `/_cloudemu/reset`, and confirms `ListBuckets` is empty — #244's stated acceptance.
- `server/admin` unit tests: backend hot-swap, reset routing, and method/route table (reset POST-only, health 200, seed 501, unknown 404).
- Full `go test ./...`, `go vet`, `gofmt` clean.

## Follow-ups
`seed` (with #250) and `snapshot`/`restore` (with #107) remain; the endpoints are namespaced so they slot in without breaking callers.

### DIFF
--- a/cmd/cloudemu/endpoints.go
+++ b/cmd/cloudemu/endpoints.go
@@ -26,7 +26,7 @@ func (e endpointSet) writeFile(path string) error {
 
 // printBanner writes the startup summary: the live endpoints and copy-paste
 // snippets for pointing each SDK at them.
-func printBanner(w io.Writer, e endpointSet) {
+func printBanner(w io.Writer, e endpointSet, adminOn bool) {
 	fmt.Fprintln(w, "cloudemu — standalone server")
 	fmt.Fprintln(w, "────────────────────────────")
 	if e.AWS != "" {
@@ -51,6 +51,10 @@ func printBanner(w io.Writer, e endpointSet) {
 	}
 	if e.Azure != "" {
 		fmt.Fprintf(w, "  Azure cloud.Configuration ResourceManager endpoint = %q (trust the cert or skip verify)\n", e.Azure)
+	}
+	if adminOn {
+		fmt.Fprintln(w)
+		fmt.Fprintln(w, "Reset state between tests: POST /_cloudemu/reset")
 	}
 	fmt.Fprintln(w)
 	fmt.Fprintln(w, "Press Ctrl-C to stop.")
--- a/cmd/cloudemu/serve.go
+++ b/cmd/cloudemu/serve.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"flag"
 	"fmt"
@@ -10,11 +11,13 @@ import (
 	"os"
 	"os/signal"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
 	"github.com/stackshy/cloudemu/v2"
 	"github.com/stackshy/cloudemu/v2/config"
+	"github.com/stackshy/cloudemu/v2/server/admin"
 	awsserver "github.com/stackshy/cloudemu/v2/server/aws"
 	azureserver "github.com/stackshy/cloudemu/v2/server/azure"
 	gcpserver "github.com/stackshy/cloudemu/v2/server/gcp"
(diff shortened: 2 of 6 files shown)


NitinKumar004:

No blocking defects. Verified the risk areas:

- **Concurrent swap** — `Backend` guards its handler with an `RWMutex`: `ServeHTTP` takes an `RLock` only to read the handler ref, then releases it *before* calling `ServeHTTP`, so a `reset` swap never blocks in-flight requests and an in-flight request completes against the handler it already holds. Locked in by `TestBackendConcurrentSwap` (8 readers serving while 200 swaps run) under `-race` — clean.
- **No orphaned goroutines** — audited the provider mocks: none spawn long-lived goroutines (no `go func()`/ticker/`AfterFunc`), so `reset` discarding the old provider just lets it be GC'd; nothing leaks.
- **Ordering** — `rebuild()` runs synchronously before any listener serves, so a `Backend` is never hit while its handler is still nil.
- **Scope isolation** — `reset` rebuilds a fresh Kubernetes data-plane and re-injects it into every provider, so post-reset the providers and the shared cluster are consistent (no dangling old-cluster ref).
- **Core untouched** — the wire handlers, drivers, mocks, and `server.Server` are unchanged; the state lifecycle lives only in `server/admin` + the binary wiring.

One deliberate design note for reviewers: `reset` is **global** (a POST to any provider's port rebuilds the whole emulator) — that's the intended "clean slate between tests" semantic, and it's gated behind `--admin` (default on; disable for exposed binds).

`go test ./...`, `go test -race ./server/admin`, `go vet`, `gofmt` all clean; out-of-process acceptance test (real S3 SDK → reset → empty) passes.

NitinKumar004:
## Review round 2 — findings fixed

Ran a 3-angle review (correctness/concurrency, removed-behavior, security/tests). Removed-behavior came back clean (the serve.go refactor preserved fail-fast binding, graceful shutdown, TLS, shared-k8s injection, endpoints/banner, `--admin` bypass, request logging). Four real findings, all now fixed:

1. **`rebuild()` wasn't serialized** (concurrency) — two concurrent `POST /_cloudemu/reset` calls could interleave and leave providers wired to different Kubernetes instances. **Fixed:** `rebuild` now holds a mutex and **builds every fresh handler before swapping any** — so concurrent resets serialize into consistent sets and a construction panic can't leave a half-swapped state.
2. **Unauthenticated reset on a non-loopback bind** — with `--host 0.0.0.0` anyone on the network could wipe state. **Fixed:** startup warning when `--admin` is on and the bind host isn't loopback, pointing at `--admin=false`.
3. **Under-scoped tests.** **Fixed:** the acceptance test now also creates a **GCP** bucket and confirms a reset issued on the **AWS** port clears it (proves reset is global), plus a new `admin-reset-concurrent` subtest fires 6 parallel resets and confirms the server stays usable (exercises the rebuild mutex).
4. **Inaccurate comment** claiming `--admin=false` yields a 501 for control paths (AWS S3's catch-all actually serves them). **Fixed:** comment corrected.

Non-issue: the global-reset semantic (any port resets the whole emulator) is intended and documented.

`go test ./...`, `go test -race ./server/admin`, `go vet`, `gofmt` all clean; the out-of-process acceptance + concurrent + cross-provider tests pass.

PR #288, opened 2026-07-25
feat: declarative fixture seeding — seed package + /_cloudemu/seed (#250, #244)

NitinKumar004:
Implements **#250** (go:embed fixture loader) and makes the **#244** `/_cloudemu/seed` endpoint real — closing the seed half of the admin control plane.

## `seed` package (#250)

A provider-agnostic fixture loader. Fixtures name resource *kinds* — `buckets` (+objects), `tables` (+items), `secrets`, `instances` — and `Apply` writes them through the **driver interfaces every provider implements**, so the same fixture file seeds AWS, Azure, or GCP depending on the `Target` you pass:

```go
//go:embed testdata/fixtures.json
var fixtures embed.FS

f, _ := seed.LoadFS(fixtures, "testdata/fixtures.json")
seed.Apply(ctx, f, seed.Target{Storage: aws.S3, Database: aws.DynamoDB, Secrets: aws.SecretsManager, Compute: aws.EC2})
```

A fixture naming a kind whose driver is nil is a clear error, not a silent skip.

## `/_cloudemu/seed` endpoint (#244)

`POST /_cloudemu/seed` applies a fixture to the provider on that port (was 501). It shares the reset mutex, so seed and reset can't run against each other's half-built state; the Kubernetes port (no seedable drivers) still returns 501.

## Design

Seeding goes through the driver interfaces rather than a new per-service serialization layer — consistent with the repo's principle that adding a service never touches shared machinery. Covers the four resource kinds the #250 ticket names; the schema is additive, so more kinds slot in without breaking fixtures.

## Testing

- **#250 acceptance:** embed a fixtures dir, seed the AWS drivers, read `app-data/config.yaml` and the `users` item back through **real S3 + DynamoDB SDK** clients (secrets/instances verified via their drivers).
- `server/admin` unit tests for the seed route (200 + applied count, 400 on seeder error, 501 when disabled, POST-only).
- Out-of-process serve test: `POST /_cloudemu/seed` a fixture, read the seeded object back via the S3 SDK.
- Full `go test ./...`, `go vet`, `gofmt`, `-race` on `server/admin` clean.

## Remaining
`snapshot`/`restore` (with the #107 state model) is the only unshipped part of the control plane; the endpoints stay namespaced for it.

### DIFF
--- a/cmd/cloudemu/serve.go
+++ b/cmd/cloudemu/serve.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/stackshy/cloudemu/v2"
 	"github.com/stackshy/cloudemu/v2/config"
+	"github.com/stackshy/cloudemu/v2/seed"
 	"github.com/stackshy/cloudemu/v2/server/admin"
 	awsserver "github.com/stackshy/cloudemu/v2/server/aws"
 	azureserver "github.com/stackshy/cloudemu/v2/server/azure"
@@ -120,7 +121,10 @@ func runServe(args []string) error {
 		k8sBackend = admin.NewBackend(nil)
 	}
 
-	var rebuildMu sync.Mutex
+	var (
+		rebuildMu sync.Mutex
+		targets   map[string]seed.Target // provider → current drivers, for seeding
+	)
 	rebuild := func() {
 		// Serialise resets so two concurrent /_cloudemu/reset calls can't
 		// interleave and leave providers wired to different Kubernetes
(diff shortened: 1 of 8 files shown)


NitinKumar004:

No blocking defects. Verified the risk areas:

- **seed vs reset concurrency** — `seedFor` reads `targets[provider]` and runs `seed.Apply` while holding the same `rebuildMu` that `rebuild()` (reset) holds, so a seed and a reset are mutually exclusive and neither sees the other's half-built state. `targets` is written and read only under that lock → no race (`-race` clean on `seed` and `server/admin`).
- **Same instances on the wire** — the `Target` is built from the *same* `cloud` provider used for the served `Drivers`, so seeding mutates the exact mocks the server serves; the out-of-process test confirms a seeded object is readable via the S3 SDK.
- **Store-level safety** — `Apply` calls driver methods that take the mocks' own locks; it never re-enters `rebuild`, so no deadlock with in-flight requests.
- **Honest failure** — a fixture naming a kind with a nil `Target` driver errors instead of silently skipping; a mid-fixture failure is fail-fast (documented; reset + retry). Body capped at 32 MiB.
- **Endpoint routing** — seed is POST-only (405 otherwise), 501 when no seeder is wired (Kubernetes port), 400 on a bad fixture or seeder error.

Design note: seeding goes through the driver interfaces (not a new per-service serialization layer), consistent with the repo's "adding a service never touches shared machinery" principle. Covers the four kinds the #250 ticket names; the schema is additive.

`go test ./...`, `go vet`, `gofmt`, `-race` (seed + admin) all clean.

NitinKumar004:
## Review round 2 — 3-angle review, findings fixed

Ran seed-correctness, endpoint/concurrency, and removed-behavior/tests/security angles. Third angle confirmed no correctness regressions (all `NewControl` callers updated, reset/health/passthrough unchanged, #250 acceptance genuinely met). Four real findings, all fixed:

1. **Silent data loss** — a fixture with an empty/missing `partitionKey` (or empty object key / names) was accepted, then items collapsed to a single store key, dropping all but the last while reporting success. **Fixed:** `Fixtures.Validate` runs before any write and rejects missing required fields — so bad fixtures fail with **nothing created** (also removes the half-seed/retry footgun for the validation case).
2. **Lock-hold blocking resets** — `seed.Apply` ran under the shared `rebuildMu`, so a large fixture (or `--latency`) blocked every provider's reset for its full duration. **Fixed:** the Target is read under the lock but `Apply` runs outside it.
3. **Oversize body truncated** — a >32 MiB fixture was silently truncated into a misleading JSON parse error. **Fixed:** reads one byte past the cap and returns **413**.
4. **Misleading count** — reported top-level entries, not resources. **Fixed:** `ResourceCount` counts objects, items, and instance `count`.

Added error-path tests (malformed JSON, nil driver, validation-before-write, ResourceCount) — the seed package had only a happy-path test before.

By-design, documented: `Apply` isn't transactional across driver writes (can't roll back wire creates), so seeding a *non-empty* backend can partially apply — the guidance is reset-then-seed onto a fresh backend.

`go test ./...`, `go vet`, `gofmt`, `-race` (seed + admin) all clean.

PR #296, opened 2026-07-27
feat(aws): close provisioning-lifecycle gaps and AWS fidelity fixes

aryanmehrotra:
## What this is

A set of AWS-surface additions and fidelity fixes, found by pointing a real infrastructure provisioner at cloudemu and fixing whatever it hit until full provision-and-teardown lifecycles ran green.

Every AWS-facing change is covered by a round-trip test driving the real `aws-sdk-go-v2` client against the emulator, rather than hand-built request forms — several bugs here were in the gap between what the SDK sends and what the server read.

## Motivation

I wanted to exercise a provisioner end-to-end without a cloud account. cloudemu got remarkably far already; this closes the gaps that stopped a full lifecycle, plus the fidelity issues that surfaced along the way.

## New surface

**EC2 / VPC**
- `DescribeAvailabilityZones`
- Elastic IP handlers (`AllocateAddress`, `ReleaseAddress`, `DescribeAddresses`) — the driver already implemented these; only the wire layer was missing
- Route-table associations (`AssociateRouteTable`, `DisassociateRouteTable`) and the `associationSet` projection on `DescribeRouteTables`
- `DeleteRoute`, `DeleteRouteTable`, `DeleteInternetGateway` — also wire-only gaps
- Elastic network interfaces, with NAT gateways holding one for their lifetime
- Main route tables, created with each VPC
- `ModifyVpcAttribute`

**RDS / ElastiCache**
- DB subnet groups and cache subnet groups
- Replication groups

**IAM / SSM / ELBv2**
- AWS-managed policies resolve without a preceding `CreatePolicy`
- AWS-published `/aws/service/.../ami-id` parameters resolve without a preceding `PutParameter`
- SSM Run Command (`SendCommand`, `GetCommandInvocation`)
- `ModifyLoadBalancerAttributes`, `DescribeLoadBalancerAttributes`, `DescribeTags`

**Kubernetes data plane**
- `serve` advertises the real data-plane endpoint, so `DescribeCluster` no longer returns the `EKS-DATAPLANE-NOT-IMPLEMENTED` sentinel
- EKS returns a parseable self-signed CA (the previous placeholder broke client-go at `NewForConfig`)
- API discovery, and protobuf rejected with 415 rather than surfacing as a JSON decode error
- `PodDisruptionBudget` (policy/v1) and an `/openapi/v2` stub

## Fidelity fixes

These are behaviour changes, each one a case where the emulator agreed with a caller that AWS would have rejected — or disagreed where AWS would have agreed.

- **`DescribeLoadBalancers` on a missing ARN** returned an empty list; real ELBv2 returns `LoadBalancerNotFound`. A caller polling for a delete to settle spun until timeout over a load balancer that was already gone.
- **Duplicate-resource error codes** were generic. The message read correctly but the SDK surfaces the *code* as a typed error, so `errors.As(&types.DBSubnetGroupAlreadyExistsFault{})` never matched and idempotent re-runs hard-failed. ElastiCache had no per-resource mapping at all.
- **`DescribeNetworkInterfaces`** silently returned nothing for an unrecognised filter; real EC2 answers `InvalidParameterValue`. Silence is the worst option here — a caller draining a VPC reads it as "nothing left" and proceeds to a delete that fails with `DependencyViolation`.
- **`DescribeTags`** resolved every ARN as a load balancer, so a target group ARN came back empty.
- **Route-table resolution in `features/topology`** took whichever table was listed first for the VPC. That was only right while a VPC had exactly one; it now follows the cloud — explicit subnet association wins, main table is the fallback.
- **`SendCommand`** reported `Success`; real SSM reports `Pending` and the caller learns the outcome from `GetCommandInvocation`.

## Design notes

**Optional capabilities.** Subnet groups, replication groups and Run Command are AWS-only concepts. Rather than widening the shared driver interfaces, they are declared in the service driver packages and discovered by type assertion. Drivers that do not implement them answer `InvalidAction`, which is the truthful response for a cloud with no such resource — and no Azure or GCP driver needed to change.

**Real catalogues, not permissive stand-ins.** AWS-managed policies and published AMI parameters are matched against the sets AWS actually publishes. Accepting anything well-formed would turn a false failure into a false success: a typo like `AmazonEKSClusterPolicyy` would attach cleanly here and fail only in production. A missing catalogue entry is a loud one-line fix instead.

## Known limitation

**SSM Run Command does not execute anything.** An emulated instance has no guest OS, so invocations report success with empty output. Targets *are* validated — an unknown instance is `InvalidInstanceId` — but the script is not. This is stated in the `RunCommand` interface doc so it cannot be mistaken for real execution.

## Two existing tests changed

Both asserted emulator behaviour rather than AWS behaviour:

1. `providers/aws/elb`: expected `DescribeLoadBalancers` on an unknown ARN to return an empty list. Real ELBv2 errors.
2. `features/topology`: `TestTraceRoute` created a route table, never associated it with the subnet, then asserted traffic routed through it. It passed only because the emulator picked the sole table available; in the real cloud that subnet uses the main table and the destination is unreachable. The test now associates the table, and a new case pins the unassociated path.

## Testing

- `go build ./...`, `go vet ./...` clean
- `go test -race ./...` clean
- Round-trip tests against the real `aws-sdk-go-v2` clients for every new AWS action
- Exercised live: full provision *and* teardown of EKS clusters, MySQL, PostgreSQL, Redis and VM-with-load-balancer stacks, verified by querying the emulator directly rather than trusting caller status — every resource back to zero afterwards

Happy to split this into smaller PRs by area if that is easier to review.


providers/aws/vpc/eni.go
```diff
@@ -0,0 +1,114 @@
...
+	VPCID        string
+	SubnetID     string
+	Status       string
+	AttachmentID string
+	Description  string
+	Tags         map[string]string
+}
+
+// DescribeNetworkInterfaces returns ENIs matching the given IDs, or all if empty.
+func (m *Mock) DescribeNetworkInterfaces(_ context.Context, ids []string) ([]driver.NetworkInterface, error) {
```
thzgajendra, line 33: `DescribeNetworkInterfaces` returns `[], nil` for an explicit-but-unknown ID — real EC2 returns `InvalidNetworkInterfaceID.NotFound`. This is the same silent-empty pattern the PR fixes for `DescribeLoadBalancers`; a caller polling an ENI delete reads empty as "gone" and stops early. Related: the ENI apparatus here (attach/release, NAT holds one) is never consulted by `DeleteVPC`/`DeleteSubnet`, so a VPC with a live ENI still deletes cleanly instead of returning `DependencyViolation` — the teardown-ordering bug this tracking was meant to catch goes uncaught.

thzgajendra (commented):
## Review — provisioning-lifecycle gaps & AWS fidelity fixes

Reviewed end-to-end against cloudemu's layering (driver → provider → server handler → dispatch/registration), with a focus on wire-protocol correctness, error fidelity, the Kubernetes data plane, tests, and cross-provider consistency. **Verified locally:** `go build ./...` ✓, `go vet ./...` ✓, AWS SDK round-trip tests pass ✓ — but `gofmt` fails on 3 files.

Genuinely strong PR — the SDK round-trip test discipline (`errors.As` against typed faults), the optional-capability pattern for subnet/replication groups + Run Command, the topology route-table-by-association fix, and cross-provider parity are all well done. A few things to address:

### 🔴 High
1. **`gofmt` fails → lint CI blocker.** `providers/aws/rds/rds.go`, `services/kubernetes/state.go`, `services/kubernetes/wire.go`. `gofmt -w` fixes it.
2. **The ENI teardown-drain isn't enforced at delete.** `providers/aws/vpc/vpc.go` `DeleteVPC`/`DeleteSubnet` delete unconditionally — they never consult `m.enis`, so a VPC holding a live NAT-gateway ENI deletes cleanly. Real EC2 returns `DependencyViolation`. The `attachManagedENI`/`releaseManagedENIs` apparatus and the test comments ("refuses a VPC delete issued too early") describe a contract the code doesn't implement (same in azure/gcp). This is the emulator agreeing with a caller AWS would reject — the class this PR set out to fix. Suggest: in `DeleteVPC`/`DeleteSubnet`, scan `m.enis` and return `FailedPrecondition` → `DependencyViolation` if any remain, with a round-trip test for the refusal.
3. **Silent-empty `Describe` on a missing resource** (the exact anti-pattern the PR fixes for LBs): `providers/aws/elb/elb.go:180` `DescribeTargetGroups` returns `[], nil` for a missing ARN — should be `TargetGroupNotFound`, mirroring the `DescribeLoadBalancers` fix right above it. New `DescribeNetworkInterfaces` (`providers/aws/vpc/eni.go`) has the same silent-empty on an unknown explicit ID (should be `InvalidNetworkInterfaceID.NotFound`).
4. **EKS endpoint is HTTP but a real CA is advertised.** The k8s data plane listens on plain HTTP, yet `DescribeCluster` returns a valid self-signed CA (`cmd/cloudemu/serve.go`, `providers/aws/eks/eks.go`). A caller building a `rest.Config` from `Endpoint`+`CertificateAuthority` directly (not via the kubeconfig path, which sets `insecure-skip-tls-verify`) fails the TLS handshake — a plausible CA for a nonexistent TLS server is worse than an empty one. Either serve k8s over TLS with the generated CA, or emit an empty CA + document `InsecureSkipTLSVerify`.
5. **PDB discovery advertises verbs it doesn't implement.** `services/kubernetes/discovery.go:185` lists `watch` and `patch` for PodDisruptionBudgets, but `services/kubernetes/pdb.go` implements neither — `?watch=true` returns a list (breaks client-go Reflectors/informers) and `PATCH` → 405 (breaks `kubectl patch pdb` / `helm upgrade`). Advertise read-only verbs, or implement watch/patch.

### 🟡 Medium
6. **Error-code fidelity mismatches** (comment/text says one code, the SDK receives another): RDS/ElastiCache subnet-group "unsupported" claims `InvalidAction` but the handler maps it to `InvalidParameterValue`; SSM `SendCommand` to a missing instance claims `InvalidInstanceId` but surfaces as `ParameterNotFound`. The promised code lives only in the message string, not what `errors.As` matches.
7. **`NumCacheClusters` parse error swallowed** → silent coercion to 1 instead of a `ValidationError` (`server/aws/elasticache/replicationgroup.go`).
8. **Design:** ENI + `ModifyVPCAttribute` are added to the *core* `Networking` interface, inconsistent with the optional-capability pattern this PR uses (correctly) for subnet groups / replication groups / Run Command. The cost: azure/gcp carry **byte-identical real ENI implementations** — triplicated logic to keep in sync. Consider an optional `NetworkInterfaces`/`VPCAttributes` capability interface (or one shared helper). Also `ModifyVPCAttribute(id, *bool, *bool)` → a `VPCAttributeUpdate` struct for additive evolution.
9. **`DeleteCluster` returns the sentinel endpoint** — it deregisters the UID before reading the endpoint, so the delete response shows `EKS-DATAPLANE-NOT-IMPLEMENTED` instead of the endpoint the cluster had (`providers/aws/eks/eks.go`).

### ⚪ Low / nits
- **Tests:** azure/gcp ENI + VPC-attribute, `DescribeAvailabilityZones`/`regionFromRequest`, EIP `address.go`, and k8s discovery/PDB have **no tests**; the EC2 handler tests string-scrape XML rather than driving the real SDK client (weaker than the elasticache/rds/elbv2/ssm round-trips); no idempotent re-create-after-delete for the datastore groups.
- **docs/services.md:** SSM Run Command + public-AMI params are undocumented; ELBv2 action count not bumped.
- **Stale comments:** `server/aws/ec2/route_table.go` "later phases" (peering is already implemented); `server/aws/elbv2/handler.go` `//nolint:gocyclo // 15 cases` (now 18); doubled `LoadBalancerNotFound:` prefix in the elb message; the EKS CA comment quotes a client-go error string it doesn't actually produce.
- **Discovery:** the new resources (ENI, Elastic IP, RDS/ElastiCache subnet groups, replication groups) aren't added to `resourcediscovery/walkers.go`, so they won't appear in Resource Explorer 2 / Resource Graph / Cloud Asset. Fine for `Describe`-driven teardown; a discovery-driven cleanup would miss them — worth a follow-up.
- Add `var _ Interface = (*Mock)(nil)` assertions for the new optional interfaces so signature drift is a compile error rather than a silent `InvalidAction`.

Nothing structural is wrong — wiring is complete across all layers and the build is clean. Fix #1–#3 (and ideally the k8s #4/#5) and this closes the lifecycle it set out to. Nice work.

services/kubernetes/discovery.go
```diff
@@ -0,0 +1,193 @@
...
+		{"pods", "pod", "Pod", true, rwVerbs(), []string{"po"}},
+		{"secrets", "secret", "Secret", true, rwVerbs(), nil},
+		{"serviceaccounts", "serviceaccount", "ServiceAccount", true, rwVerbs(), []string{"sa"}},
+		{"services", "service", "Service", true, rwVerbs(), []string{"svc"}},
+	}
+}
+
+func policyResources() []apiResource {
+	return []apiResource{
+		{"poddisruptionbudgets", "poddisruptionbudget", "PodDisruptionBudget", true, rwVerbs(), []string{"pdb"}},
```
thzgajendra: PDB discovery advertises `rwVerbs()` (includes `watch` and `patch`), but `pdb.go` implements neither: `?watch=true` returns a plain list (breaks client-go Reflectors/informers — they expect a watch stream), and `PATCH` returns 405 (breaks `kubectl patch pdb` / `helm upgrade`). Either advertise a read-only verb set here, or implement the watch broadcaster + a `patchPDB` handler (as `deployment.go` does).

server/aws/rds/subnetgroup.go
```diff
@@ -0,0 +1,158 @@
...
+
+type deleteDBSubnetGroupResponse struct {
+	XMLName  xml.Name         `xml:"DeleteDBSubnetGroupResponse"`
+	Xmlns    string           `xml:"xmlns,attr"`
+	Metadata responseMetadata `xml:"ResponseMetadata"`
+}
+
+// subnetGroups reports whether the configured driver models subnet groups.
+// They are an AWS-only resource, so a driver for another cloud legitimately
+// does not, and the honest answer there is InvalidAction rather than a
```
thzgajendra, line 55: The comment says the honest answer is `InvalidAction`, but the error is built with `InvalidArgument` and the RDS handler maps that to the wire code `InvalidParameterValue` — so `InvalidAction` only appears in the message text, not the code the SDK surfaces via `errors.As`. Same shape in the ElastiCache subnet-group path, and SSM `SendCommand` (comment says `InvalidInstanceId`, code delivers `ParameterNotFound`). Either map to the code the comment promises, or update the comment to the delivered code.

NitinKumar004 (commented):
## Review — provisioning-lifecycle gaps + AWS fidelity

Strong, SDK-verified PR. Build is green and the error-code fidelity work checks out against the real `aws-sdk-go-v2` types. Flagging the issues that automated gates won't catch — grouped by priority. (Skipping lint/style; CI covers those.)

### 🟠 Correctness / concurrency / cross-provider consistency

**1. ENI drain enforced only on AWS (mirror-rule violation).**
AWS `DeleteVPC`/`DeleteSubnet` (`providers/aws/vpc/vpc.go:158,228`) refuse with `DependencyViolation` while a managed ENI is attached. Azure (`providers/azure/vnet/vnet.go:144,190`) and GCP (`providers/gcp/gcpvpc/vpc.go:114,160`) create managed ENIs on their NAT gateways but never check them on teardown — there's no `attachedENIIn` equivalent. So the drain semantics this PR builds and tests on AWS are silently un-exercised on the other two providers, and the ENI record leaks after the network is gone. No Azure/GCP ENI drain tests were added.

**2. `DescribeNetworkInterfaces` NotFound-vs-empty diverges across providers.**
AWS returns `InvalidNetworkInterfaceID.NotFound` for an explicitly-named unknown ID (`providers/aws/vpc/eni.go:36` — the point of 59a8a89). Azure/GCP `eni.go:29` still return an empty slice. A drain loop that reads empty as "already gone" terminates early on Azure/GCP — the exact bug the AWS fix targets.

**3. `LBAttributes.Extra` shared-map aliasing → concurrent-map-write panic.**
`GetLBAttributes` (`providers/aws/elb/elb.go:363`) returns `&attrs`, a shallow copy whose new `Extra` map aliases the map stored in `m.attrs`. `modifyLoadBalancerAttributes` (`server/aws/elbv2/attributes.go`) then writes `attrs.Extra[key] = value` via `applyLBAttribute` **outside `attrsMu`**, and `PutLBAttributes` re-stores the same map. Two overlapping `ModifyLoadBalancerAttributes` on the same LB (or a Modify racing a Describe that reads `Extra`) → "concurrent map writes" panic, killing the server goroutine under `cloudemu serve`. The first-ever modify is safe (fresh map); it bites on the 2nd+. Fix: deep-copy `Extra` in Get/Put, or hold the lock across read-modify-write.

**4. K8s discovery advertises verbs the PDB handler doesn't implement.**
`services/kubernetes/discovery.go:120` advertises `patch`+`watch` for `poddisruptionbudgets`, but `services/kubernetes/pdb.go` implements only GET/PUT/DELETE (no PATCH, no `?watch=` stream). `kubectl apply` (uses PATCH) → 405; an informer watch → decode hang. This contradicts the file's own "advertise only what we serve" invariant — PDB is the lone outlier vs core/deployments.

**5. Service Networking param-less DELETE wipes ALL connections.**
`server/gcp/servicenetworking/handler.go` defaults `network` to `"-"` when the query param is absent, and `remove("-")` resets the whole map. A DELETE with no query string destroys every connection across all networks in the process, not just the caller's. Consider requiring an explicit network for delete.

**6. ENI struct-field mutation race.**
`DetachNetworkInterface` writing `eni.AttachmentID`/`Status` races with `toENIInfo` reads and `attachedENIIn`, because `memstore.All()` copies the map but values are `*eniData` pointers (confirmed with `go test -race`). This mirrors a pre-existing pattern (`eip.go`) rather than a new class — lower priority than #3, but real under concurrency.

### 🟡 Fidelity / robustness (lower priority)

- **`resolveVPCID` can still return blank** (`providers/aws/{rds,elasticache}/subnetgroup.go`): a non-existent member subnet → empty `DescribeSubnets` → blank `VpcId` → re-introduces the teardown-leak this PR targets. Real AWS rejects with `InvalidSubnet`.
- **`ReleaseAddress`/`DescribeAddresses` by `PublicIp` is non-functional** (`server/aws/ec2/address.go:74,95`): EIPs are keyed by allocation ID, so a `PublicIp` value always misses → NotFound. The comment claims both are supported.
- **New GCP handlers bypass the 1 MiB body cap**: `networks/addresses.go`, `networks/routers.go`, `servicenetworking`, `azure/resourcegroups` use bare `json.NewDecoder(r.Body)` while sibling `networks/handler.go` uses the bounded `gcprest.DecodeJSON`.
- **`errUnsupported*` comment says "InvalidAction" but emits code `InvalidParameterValue`** (`server/aws/{rds,elasticache}/subnetgroup.go`) — misleading, though unreachable in practice.
- **SSM eks-optimized AMI entry won't resolve**: real name is `.../recommended/image_id`, not `/ami-id`; also `GetParametersByPath`/`DescribeParameters` aren't wired to the synth path.

### 🔵 Structural note
The core `Networking` driver interface gained 4 required methods (`ModifyVPCAttribute`, `DescribeNetworkInterfaces`, `DetachNetworkInterface`, `DeleteNetworkInterface`) — a compile break for any external implementer. RDS/Cache/SSM/LB kept the optional type-asserted-capability pattern; widening Networking directly is defensible (all clouds have NICs/DNS attrs) but inconsistent with that principle.

### ✅ Verified correct (no action)
Error-code strings exact-match the SDK and the `errors.As` round-trips genuinely prove it; factory wiring is nil-safe and correctly ordered; the topology route-table change is well-guarded (explicit association → main table → fallback) and the `TestTraceRoute` change reflects real behavior; the `DescribeLoadBalancers` NotFound change has no surviving internal dependents; IAM/SSM catalogue restrictions are strictly additive vs `development`.

providers/aws/eks/tls.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// ServingTLSConfig returns a TLS configuration for the Kubernetes data plane,
+// carrying a leaf signed by the CA clusters advertise.
+//
+// Serving the data plane with this makes the advertised CA true: a caller can
+// validate the endpoint against it, which is what a rest.Config built from a
+// cluster description does. Serving plain HTTP instead leaves the CA describing
+// a TLS server that does not exist.
+func ServingTLSConfig(hosts []string) (*tls.Config, error) {
```
thzgajendra, line 89: `ServingTLSConfig` is correct in isolation (leaf signed by the same CA `stubCertificate()` returns), but it has **zero call sites** — only comments reference it. The k8s data plane is still served over plain HTTP: `cmd/cloudemu/serve.go:150` does `SetBaseURL("http://"+…)`, `:267` advertises `eps.Kubernetes = "http://…"`, and the listener uses `srv.Serve` (not `ServeTLS`). So the advertised real CA still doesn't certify the endpoint — a `rest.Config` from `Endpoint`+`CertificateAuthority` will fail the TLS handshake, same as before. Either wire this into the k8s listener (`ServeTLS` + `https://` base URL), or keep HTTP with an empty CA + `insecure-skip-tls-verify` and drop the "CA certifies the endpoint" claim.

providers/aws/eks/tls.go
```diff
@@ -0,0 +1,121 @@
...
+// The CA private key is therefore retained rather than discarded, so
+// ServingTLSConfig can mint a leaf for whoever serves the data plane.
+var (
+	caOnce sync.Once
+	caPEM  string
+	caCert *x509.Certificate
+	caKey  *rsa.PrivateKey
+)
+
+func initCA() {
```
thzgajendra: `initCA` returns on any key/cert-gen error leaving `caPEM`/`caCert`/`caKey` empty, and it runs under `sync.Once` so it never retries. Since `ServingTLSConfig` is guarded but `stubCertificate()` isn't, a generation failure means every cluster advertises an empty `CertificateAuthority` with no error surfaced — the same "placeholder that breaks client-go" class this change set out to fix. Consider `sync.OnceValues` returning the error and propagating it.

thzgajendra (commented):
## Re-review (head `75d5e6b`, +13 commits since the last pass)

Thanks for the thorough turnaround — re-reviewed with the same 8-way pass. **Verified locally: `go build ./...` ✓, `go vet ./...` ✓, `gofmt` clean ✓ (was 3 files), affected test suites pass.** Most prior findings are genuinely fixed; one is fixed-but-into-dead-code, and there are a few new items.

### ✅ Prior findings — fixed (verified in source, not by commit message)
- **gofmt** (3 files) → clean.
- **ENI teardown enforcement** → `DeleteVPC`/`DeleteSubnet` now return `DependencyViolation`/`FailedPrecondition` while an ENI is attached, correct scope; `DeleteNetworkInterface` refuses an attached ENI. (Azure/GCP correctly don't enforce — ENIs are now an *optional capability* they don't model.)
- **Silent-empty Describe** → `DescribeTargetGroups` and `DescribeNetworkInterfaces` now return the proper NotFound.
- **Error-code fidelity** → subnet-group "unsupported" now delivers `InvalidAction`; SSM SendCommand-missing-instance delivers `InvalidInstanceId` (verified the handler mapping, not just the comment).
- **NumCacheClusters** → now validated (`InvalidArgument` on bad input), no silent default.
- **Optional-capability refactor** → excellent. ENI/`VPCAttribute` are now type-asserted optional interfaces with graceful `InvalidAction` fallbacks; the triplicated `eni.go`/`vpcattribute.go` in azure/gcp are **deleted** (−392/+198); `ModifyVPCAttribute` takes a `VPCAttributeUpdate` struct; ENI-status consts declared once; `var _` compile-assertions added. Fully resolves the prior design finding.
- **DeleteCluster** → reads the endpoint before deregistering; no more sentinel in the delete response.
- **PDB verbs** → discovery now advertises only implemented verbs (no `watch`/`patch`), pinned by `TestPDBAdvertisesOnlyImplementedVerbs`. 
- **Stale comments** (route_table "later phases", elbv2 "15 cases") → gone.

### 🔴 Still broken — EKS CA fix is wired into dead code
The `feat(eks): make the advertised CA certify the data plane` machinery is correct in isolation (`providers/aws/eks/tls.go`: `initCA` mints a real CA; `ServingTLSConfig` mints a leaf signed by it) — **but `ServingTLSConfig` has zero call sites** (only referenced in comments). The k8s listener is still started with `srv.Serve` (plain HTTP), `SetBaseURL("http://…")` (`cmd/cloudemu/serve.go:150`), and `eps.Kubernetes = "http://…"` (`:267`). So the contract "the advertised CA certifies the endpoint" is still false at runtime: a `rest.Config` built from `Endpoint`+`CertificateAuthority` still hits a TLS handshake against a non-TLS server. Two agents flagged this independently; I confirmed the grep. **Fix:** either serve the k8s data plane over TLS using `ServingTLSConfig([]string{c.host})` (set `https://`), or keep HTTP but emit an empty CA + `insecure-skip-tls-verify` (as `kubeconfig.go` already does) and drop the "CA certifies the endpoint" claim.

### 🟠 New / partial
- **Discovery type strings fabricated (partial fix).** Walkers now surface ENI + Elastic IP (good), but `resourcegraph/handler.go` `portableToAzureType` and `cloudasset/filter.go` `portableToGCPAssetType` have **no case** for the new types, so they emit `networking/networkinterface` / `networking/elasticip` instead of `microsoft.network/networkInterfaces` / `compute.googleapis.com/Address`. `arn.go` already knows the correct names, so the `type`/`assetType` field disagrees with the ARN. A client filtering RG/Cloud Asset by real type won't find them.
- **Discovery error-handling (HIGH-ish).** `walkNetworkInterfaces` (`services/resourcediscovery/walkers.go`) swallows any driver error and returns 0 ENIs silently, while the new EIP block in `walkNetworking` does the opposite — aborts the **entire** walk on a `DescribeAddresses` error (an EIP-driver error would blank out storage/compute discovery). Make ENI propagate its error like the other walkers, and don't let an EIP error abort the whole inventory.
- **Silent-swallow of malformed bodies (MEDIUM).** Azure resource-group `PUT` and GCP service-networking upsert coerce a malformed JSON body to `{}` and report `Succeeded` — real ARM/GCP return 400. Distinguish empty body (allow) from present-but-malformed (400).
- **`initCA` silent failure (MEDIUM).** On a key/cert-gen error `initCA` leaves the CA empty with no signal (under `sync.Once`, never retried), so every cluster advertises an empty CA — the same "unparseable placeholder" class the round set out to kill. Surface the error.
- **ENI coverage vs comments (LOW/fidelity).** Only NAT gateways attach a managed ENI, but comments in `providers/aws/vpc/eni.go` and the `NetworkInterface` type doc claim load balancers and managed databases hold ENIs too. So an LB/RDS left in a VPC won't block a delete the way the comments promise. Wire ELB/RDS ENI attach, or narrow the comments to "NAT gateways only".

### 🟡 Comment rot (new)
- `providers/aws/eks/eks.go:115-134` has **two orphaned doc comments**: one for `stubCertificate` (moved to `tls.go`) now stranded above `newUpdateID`, and one for `generateSelfSignedCA` — a function that **exists nowhere** (the real one is `initCA`). Delete/move them.
- `eks.go:1-6` package doc still says the data plane is "out of scope … placeholder values" — stale now that it's wired.
- `services/kubernetes/discovery.go:21-24` claims the advertised set is "derived from" `ServeHTTP`'s dispatch, but it's a hand-maintained list and has already drifted (`endpoints` is served but not advertised). Reword to the real invariant.

### ⚪ Tests
Good additions this round (ENI-hold, route-table lifecycle, AZ, EIP, subnet groups, ELBv2 attrs, SSM, IAM, `TestPDBAdvertisesOnlyImplementedVerbs`). Still untested and shipping dark:
- **EKS CA-signs-dataplane** (highest value — no test decodes the advertised CA and verifies it signs the `ServingTLSConfig` leaf / covers the SAN; a pure unit test, no server needed).
- **GCP Cloud Routers, GCP addresses, GCP/Azure Service Networking, Azure resource groups/subscriptions** — new handlers with zero tests.
- Discovery ENI/EIP surfacing; the `DeleteVPC`/`DeleteSubnet` DependencyViolation branch (not directly unit-tested); re-create-after-delete for datastore groups; the unsupported-optional-capability `InvalidAction` fallback path.
- Note: the new EC2 tests string-scrape XML rather than driving the real SDK client — weaker than the RDS/ElastiCache/ELBv2/SSM typed-error round-trips (a wrong-but-non-200 error code would still pass).

### Verdict
Strong iteration — the optional-capability refactor and the error-fidelity/gofmt/PDB/DeleteCluster fixes are all solid and several are pinned by good tests. The one thing that genuinely still doesn't work is the **EKS CA/TLS contract** (fixed into dead code); I'd treat that + the discovery `walkNetworkInterfaces` error-swallow as the blockers, with the type-string mapping and malformed-body handling close behind.

NitinKumar004 (commented):
## Re-review (HEAD `43bdcc6`)

Thanks for the fast turnaround — re-reviewed the 6 new commits. Verified fixed, still-open, and a couple of new items. (Leaving pure lint/formatting to CI as before.)

### ✅ Confirmed fixed
- **Networking interface widening → optional capabilities** (`9ce4efd`): ENI / VPC-attribute ops are now type-asserted optional capabilities; deletions are complete (no dangling refs), discovery is panic-free (comma-ok), and the prior **mirror-rule divergence is genuinely resolved** — Azure/GCP NAT gateways no longer create phantom ENIs that nothing could drain. Nice.
- **PDB discovery verbs pinned** (`b0e2ec4`): advertised set is now `{get,list,create,update,delete}`, matching what the handler serves (no `watch`/`patch`), and the new test pins it strongly. My prior finding is resolved.
- **SSM `InvalidInstanceId`** (`runcommand.go`): unknown/unmanaged instance now maps to `InvalidInstanceId` instead of a `ParameterNotFound` leak; precise and tested.
- **`errUnsupported` code** now emits `InvalidAction` directly via `writeUnsupported` (was `InvalidParameterValue`), and `NumCacheClusters` is now rejected when unparseable rather than silently defaulted.

### 🟠 New this round
**1. EKS `ServingTLSConfig` is unwired dead code — the advertised CA is still cosmetic.** *(providers/aws/eks/tls.go:79)*
The commit *"make the advertised certificate authority certify the data plane"* adds the cert machinery but nothing calls `ServingTLSConfig` (verified: zero callers). The data-plane server at `cmd/cloudemu/serve.go:267` is a plain `http.Server` (no `TLSConfig`), advertised as `http://`. So a client that builds a `rest.Config` from `Endpoint` + `CertificateAuthority` and dials `https` still fails the handshake, and if it dials `http` the CA is pointless — the exact prior complaint persists, and the comment at `eks.go:308` ("certifies this endpoint…") describes behavior that doesn't exist. Either wire `ServingTLSConfig` into the data-plane listener, or drop the exported function + the claim. (Cert code itself is correct if wired: `sync.Once`-cached, `IsCA`/KeyUsage/SANs all right, no panic path.)

**2. New `resourcediscovery` EIP/interface walkers are untested.** *(services/resourcediscovery/walkers.go)*
`walkNetworking` (EIPs) and `walkNetworkInterfaces` (optional-capability) are correct and wired, but there's no test referencing `DescribeAddresses`/`DescribeNetworkInterfaces` — new code paths uncovered against the 90% target. Also LOW: `walkNetworkInterfaces` does `return nil` on *any* `DescribeNetworkInterfaces` error, hiding genuine errors, unlike the NotFound-only skips in `walkStorage`/`walkDatabase`.

### 🔴 Still open from the prior review (not addressed by these commits)
- **`LBAttributes.Extra` shared-map → concurrent-map-write panic** — `providers/aws/elb/elb.go:363` still returns `&attrs` aliasing the stored `Extra` map, and `modifyLoadBalancerAttributes` writes into it outside `attrsMu`. Two overlapping `ModifyLoadBalancerAttributes` on the same LB (or Modify racing Describe) → runtime panic under `serve`.
- **GCP Service Networking param-less DELETE wipes ALL connections** — `server/gcp/servicenetworking/handler.go` still defaults `network` to `"-"` → `remove("-")` clears the whole map.
- **New GCP handlers bypass the 1 MiB body cap** — `networks/addresses.go`, `networks/routers.go`, `azure/resourcegroups` still use bare `json.NewDecoder(r.Body)` vs the bounded `gcprest.DecodeJSON`.
- **`ReleaseAddress`/`DescribeAddresses` by `PublicIp` non-functional** — EIPs are keyed by allocation id, so a `PublicIp` value always misses; the comment claims both are supported.
- **`resolveVPCID` can still be blank** on a non-existent member subnet (real AWS → `InvalidSubnet`), re-opening the teardown-leak this work targets.

### 🔵 Low / optional
- The portable `*networking.Networking` wrapper defines the optional methods unconditionally, so type-asserting the *wrapper* to `NetworkInterfaces`/`VPCAttributes` always succeeds then errors at call time — contradicts "wrapper sees the same capability set as the driver." Harmless today (real consumers assert the concrete driver); worth a comment.
- `driver.go` says an unmodeled driver "answers InvalidAction," but the portable layer returns `InvalidArgument`; only the HTTP layer emits literal `InvalidAction`.
- Follow-up: PDB was pinned, but other k8s resources in `discovery.go` still advertise `watch`+`patch` — worth checking they're all actually served.

NitinKumar004 (commented):
## Re-review @ `372fb10` — TLS fix confirmed ✅

The K8s data plane is now served over TLS with a leaf signed by the advertised CA, and `initCA` propagates failures. That closes the `ServingTLSConfig` dead-code / cosmetic-CA item.

Two **MEDIUM** items still open:

**1. `LBAttributes.Extra` → concurrent-map-write panic.** `providers/aws/elb/elb.go:354` `GetLBAttributes` returns `&attrs`, whose `Extra` map aliases the map stored in `m.attrs`. `modifyLoadBalancerAttributes` then writes `attrs.Extra[key]=value` outside `attrsMu`, and `PutLBAttributes` re-stores the same map. Two overlapping `ModifyLoadBalancerAttributes` on the same LB (or a Modify racing a Describe that reads `Extra`) → "concurrent map writes" panic under `serve`. Fix: deep-copy `Extra` in Get/Put, or hold the lock across read-modify-write.

**2. Service Networking param-less DELETE wipes ALL connections.** `server/gcp/servicenetworking/handler.go:104-109` — `network()` returns `"-"` when the query param is absent, and `remove` treats `"-"` as `h.connections = map[...]{}`. A DELETE with no `?network=` clears every network's connections. Consider requiring an explicit network for delete.

thzgajendra (commented):
Final re-review at `86a89a6`. **All five findings from my earlier rounds are resolved** (verified against source): unknown-ENI-id → `InvalidNetworkInterfaceID.NotFound`; RDS subnet-group → `InvalidAction`/`DBSubnetGroupNotFoundFault` at the right status; `ServingTLSConfig` now wired into a real TLS listener; `initCA` no longer bricks under `sync.Once` (fails loudly via `OnceValues`+error); PDB discovery trimmed to only the verbs it implements. The Service-Networking data-loss delete and the ENI race from the second review are both correctly fixed, and I found no other destructive empty-filter defaults.

## MEDIUM — I'd fix before merge (same class the second-review commit claims to have closed)
The ENI race fix added `eniMu` because `memstore.Store[*T]` hands out the shared pointer. That exact pattern is still **unguarded in two other new VPC handlers**, so the "fixed a race" claim is incomplete — both fail `-race` under concurrent requests:

- **`providers/aws/vpc/vpcattribute.go:24,28` — `ModifyVPCAttribute`.** `m.vpcs` is `memstore.Store[*vpcData]`, so `Get` returns the shared pointer; mutating `v.EnableDNSSupport` / `v.EnableDNSHostnames` with no lock races a concurrent `DescribeVpcs` (and two `ModifyVPCAttribute` lose-update).
- **`providers/aws/vpc/routetable.go:123,142,152` — `CreateRoute`/`DeleteRoute`/`toRouteTableInfo`.** `m.routeTables` is `memstore.Store[*routeTableData]`; `CreateRoute` appends to `rt.Routes` and `DeleteRoute` re-slices it unlocked, while `toRouteTableInfo` reads/copies it unlocked → data race + torn/realloc'd slice on concurrent create+describe, and lost route on two concurrent creates.

Same fix as ENIs — a mutex around the read-modify-write, or store these by value.

## Non-blocking
- **LB attributes** (`server/aws/elbv2` + `providers/aws/elb/elb.go`): the crash is fixed, but `GetLBAttributes`(RLock)→`PutLBAttributes`(Lock) drops the lock between read and write, so two overlapping `ModifyLoadBalancerAttributes` now silently lose one update. Worth an atomic update.
- **Delete-in-use not blocked** for RDS/ElastiCache subnet groups + replication groups (real AWS: `InvalidDBSubnetGroupStateFault` / `CacheSubnetGroupInUse`).
- **IAM `GetPolicy`** calls `ensureAWSManagedPolicy` (check-then-act) *before* taking the lock → two concurrent first-references double-create; SSM's `PutParameter` `SetIfAbsent` is the pattern to copy.
- **EIP `AssociateAddress`/`DisassociateAddress`** have driver methods but no query-API wire handler (the description lists them as wired).
- **No regression tests for the three second-review fixes** — a `-race` test for the LB crash + the route/vpc-attr races, and a delete-without-target test for Service Networking, would stop silent regressions. The EKS advertised-CA→leaf chain also isn't verified (`serve_test.go` uses `InsecureSkipVerify`).

Everything else — AWS error-code fidelity (EC2/ELBv2/ElastiCache/SSM/IAM), EC2/VPC lifecycle (EIP state machine, main-route-table invariants, ENI drain/DependencyViolation, topology-by-association), GCP/Azure operation/ARM envelopes + scoping, and vendor-neutral fixtures — checks out. Close; the incomplete race sweep is the one thing undercutting the "fixed a race" claim.

NitinKumar004 (approved):
## Approving — re-review @ `86a89a6`

Both MEDIUM items from the last pass are now genuinely fixed and verified:
- **`LBAttributes.Extra`** — `Get`/`PutLBAttributes` copy the map on both ends (`copyStringMap`), so the concurrent-map-write panic is gone.
- **Service Networking DELETE** — now requires an explicit `network` param, no more blanket wipe.

Also confirmed fixed: the ENI field-mutation race (guarded), `ReleaseAddress` by `PublicIp`, the EKS `/image_id` AMI leaf, the `walkNetworkInterfaces` error-surfacing (with new tests), and the GCP body-cap on the routers/addresses handlers.

The substantive review is clear — approving. Two trivial **lint-gate** items to clean up before merge (they fail `golangci-lint`, so CI will block otherwise):

1. `providers/aws/vpc/eni.go:21` — `eniMu` is a package-level global (`gochecknoglobals`). Prefer moving it to a field on `Mock` (a package global is shared across emulator instances anyway); a `//nolint` also works.
2. `features/topology/connectivity.go:203` — `findRouteTableForVPC` is now unused (`unused`); delete the dead wrapper.

Minor, non-blocking: two body-decode sites still bare (`servicenetworking/handler.go:92`, `azure/resourcegroups/handler.go:91`) — the sibling handlers use the bounded `gcprest.DecodeJSON`.

Nice work on the turnaround.

aryanmehrotra:
Pushed `e591afc`. Both rounds addressed.

## MEDIUM — the race sweep was incomplete (@thzgajendra)

Correct, and the framing was right: guarding ENIs alone didn't close the class. `memstore.Store[*T]` hands back the stored pointer, so the same read-modify-write existed at every other site using that pattern.

Swept: `ModifyVPCAttribute`, `CreateRoute`/`DeleteRoute`/`toRouteTableInfo`, and their describe paths. One more not in the list — **`AssociateAddress`/`DisassociateAddress` on elastic IPs** — had the identical bug and is now guarded too.

The mutex moved from a package global onto `Mock`, which closes this and @NitinKumar004's `gochecknoglobals` item in one change.

**The tests were confirmed to fail without the fix.** With the guards stripped, `-race` reports 10 race/lost-update signals across `providers/aws/vpc/race_test.go`; with them, clean. `TestCreateRoute_ConcurrentCreatesAllLand` asserts the final route count so the lost-update half is pinned independently of the race detector.

## Lint items (@NitinKumar004)

1. `eniMu` — now a field on `Mock`, per your reasoning about instance sharing.
2. `findRouteTableForVPC` — deleted.

One correction: these wouldn't have blocked CI. `ci.yml` runs lint as `golangci-lint run ... || echo '::warning::golangci-lint reported findings (advisory)'`, so the step can't fail the build. Fixed regardless — the `Mock` field is better than a global on its own merits — but worth knowing when triaging the other 244 repo-wide findings.

Also flagging one of mine that the same linter reports: `providers/aws/eks/tls.go:29` `loadCA`. Left deliberately — it's a `sync.OnceValues` singleton and one CA per process is the intended semantic; moving it onto a struct would make the CA vary per instance. Happy to `//nolint` it if you'd rather it be silent.

## Non-blocking items

- **LB attributes** — added `UpdateLBAttributes`, a read-modify-write under one lock, as an optional capability (type assertion) so other providers need no implementation; the Get/Put path stays as fallback. Test asserts all 16 concurrent updates survive and that the returned `Extra` can't reach stored state.
- **Delete-in-use** — RDS and ElastiCache subnet groups now refuse while occupied (`InvalidDBSubnetGroupStateFault` / `CacheSubnetGroupInUse`). Tests cover the refusal, the succeed-once-empty path, and an unrelated group still deleting, so the guard isn't over-broad.
- **IAM `GetPolicy`** — `SetIfAbsent`, matching the SSM `PutParameter` pattern you pointed at.
- **EIP `AssociateAddress`/`DisassociateAddress`** — wire handlers added. You were right that the description overstated this; the file's own header comment claimed they were wired too. Both accept `PublicIp` as well as the ID, as AWS does.
- **Bare decodes** — `servicenetworking` and `azure/resourcegroups` now cap the body with their package's `MaxBodyBytes`. The decode stays tolerant of an absent body, which both rely on.
- **Regression tests** — Service Networking delete-without-target (all three spellings: no query string, empty param, `-`), asserting no other connection is destroyed; and the EKS chain now completes a **real handshake trusting only the advertised CA**, no `InsecureSkipVerify`, plus a vacuity guard so an empty root pool can't make it pass.

Full suite green under `-race`.

thzgajendra (commented):
Re-reviewed at `e591afcff` — **every item from my final review is resolved.** Verified against source, not just the commit message.

**MEDIUM (the incomplete race sweep) — FIXED, and done systematically.** The read-modify-write on a shared `memstore.Store[*T]` pointer is now guarded everywhere it occurred, not just ENIs: `ModifyVPCAttribute`, `CreateRoute`/`DeleteRoute`, and EIP `AssociateAddress`/`DisassociateAddress` all take `m.mu.Lock()`, and the describe paths take `RLock()`. Nice touch: the mutex moved from a package global onto `Mock`, so it's no longer shared across emulator instances.

**Non-blocking items — all addressed too:**
- LB attributes: new `UpdateLBAttributes` does the whole RMW under `attrsMu` (optional capability, Get/Put fallback kept) — the lost-update is closed, and the returned `Extra` map is copied so the caller can't alias stored state.
- Delete-in-use: subnet groups now refuse while something is placed in them (`InvalidDBSubnetGroupStateFault` / `CacheSubnetGroupInUse`).
- IAM managed-policy materialization uses `SetIfAbsent`, so two concurrent first-references can't hand out mismatched ARNs.
- EIP `AssociateAddress`/`DisassociateAddress` now have the wire handlers the file header claimed.
- Regression tests are real: `attributes_race_test.go` runs 16×2 concurrent `UpdateLBAttributes` and asserts all land (genuine lost-update test, not sequential) + a map-independence test; `subnetgroup_inuse_test.go` covers refuse-while-in-use / succeeds-once-empty / unrelated-doesn't-block / unknown-is-not-found; EKS TLS test added. The commit notes the race tests were confirmed to fail (10 race/lost-update reports) with the guards removed.

From my side everything's cleared — feel free to dismiss my earlier change-requesting notes. Leaving this as a comment per usual; the merge call is yours. Thorough turnaround.

thzgajendra (approved):

Issue #295, opened 2026-07-27
Resource-discovery parity: surface the resources cloudemu already emulates (only 7–8 types today)


thzgajendra:
## Summary

cloudemu has a two-layer coverage picture, and the two are badly out of sync:

- **Emulation (API) layer — broad.** ~23 service categories are implemented across AWS/Azure/GCP with real SDK-compat wire handlers (only 3 known per-cell gaps).
- **Discovery (inventory) layer — very narrow.** `resourcediscovery` walks only **5 drivers** and emits only **8 resource types** total. Every SDK-compat inventory API (Resource Explorer 2, Resource Graph, Cloud Asset) is a 1:1 re-label of those same 8.

So even though RDS/EKS/ELB/ElastiCache/SNS/SQS/Bedrock/SageMaker etc. are **emulated**, **none of them appear when a client enumerates resources** via RE2 / Resource Graph / Cloud Asset. The discovery layer should surface the resources cloudemu already models (and, over time, the broader provider catalog).

Evidence: `resourcediscovery/walkers.go` (6 walkers, 8 type constants L30-39), `resourcediscovery/engine.go` L102-130; `server/aws/resourceexplorer2/handler.go` + `filter.go`; `server/azure/resourcegraph/kql.go` (`mapAzureType`) + `handler.go`; `server/gcp/cloudasset/filter.go` (`mapGCPAssetType`).

## Resource types surfaced in discovery TODAY (exhaustive)

| Engine type | AWS RE2 | Azure Resource Graph | GCP Cloud Asset |
|---|---|---|---|
| Instance | `compute:instance` | `microsoft.compute/virtualmachines` | `compute.googleapis.com/Instance` |
| VPC | `networking:vpc` | `microsoft.network/virtualnetworks` | `compute.googleapis.com/Network` |
| Subnet | `networking:subnet` | `microsoft.network/subnets` | `compute.googleapis.com/Subnetwork` |
| SecurityGroup | `networking:securitygroup` | `microsoft.network/networksecuritygroups` | `compute.googleapis.com/Firewall` |
| Bucket | `storage:bucket` | `microsoft.storage/storageaccounts` | `storage.googleapis.com/Bucket` |
| Table | `database:table` | `microsoft.documentdb/databaseaccounts` | `firestore.googleapis.com/Database` |
| Function | `serverless:function` | `microsoft.web/sites` | `cloudfunctions.googleapis.com/Function` |
[...]

PR #297, opened 2026-07-28
feat(discovery): surface Kubernetes clusters and node groups (EKS/GKE/AKS)

thzgajendra:
## What

Surfaces managed Kubernetes clusters and their node groups in resource discovery, so they appear in Resource Explorer 2 / Azure Resource Graph / GCP Cloud Asset alongside the other inventory types. Addresses the Kubernetes item of the discovery-parity backlog in #295.

- **EKS** — cluster + nodegroup
- **GKE** — cluster + node pool
- **AKS** — managed cluster + agent pool

| Portable | AWS RE2 | Azure Resource Graph | GCP Cloud Asset |
|---|---|---|---|
| `kubernetes/Cluster` | `kubernetes:cluster` | `microsoft.containerservice/managedclusters` | `container.googleapis.com/Cluster` |
| `kubernetes/NodeGroup` | `kubernetes:nodegroup` | `.../managedclusters/agentpools` | `container.googleapis.com/NodePool` |

## How

- **`services/resourcediscovery/`** — a `KubernetesClusters` capability with a provider-neutral `DiscoveredCluster` projection, a single `walkKubernetes` walker, `Cluster`/`NodeGroup` type constants, and per-provider ARN/ID builders. The discovery engine stays free of provider imports; each provider wires a thin adapter over its cluster mock (mirrors how Databricks is walked).
- **`providers/{aws,gcp,azure}`** — `eksDiscovery` / `gkeDiscovery` / `aksDiscovery` adapters projecting each cloud's cluster mock (clusters + node-group names, region, tags) onto `DiscoveredCluster`, wired into the discovery `Drivers`.
- **Inventory maps** — Azure Resource Graph (`kql.go` `mapAzureType` + `handler.go` `portableToAzureType`) and GCP Cloud Asset (`filter.go`) gain the ↔ mappings; AWS RE2 handles it via the generic `service:type`.

## Test plan

- `go build ./...`, `go vet ./...`, `gofmt -l` — clean.
- **Walker unit tests** (`kubernetes_walk_test.go`): per-provider surfacing, ARN shape, region fallback, error propagation.
- **End-to-end per-provider tests** (`providers/{aws,gcp,azure}/kubernetes_discovery_test.go`): create real EKS/GKE/AKS clusters + node groups through the mocks and assert they surface via the wired `ResourceDiscovery`, with region/tags.
- **Map-table cases** added to the Azure/GCP filter tests.
- All touched suites pass.

## Notes

- Scope is clusters + node groups only (not Fargate profiles / addons), per the #295 backlog item.
- ARNs are best-effort canonical identifiers (consistent with the existing discovery ARN builders).

Refs #295

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -1190,9 +1190,24 @@ Shared in-memory K8s API server registered by every cluster from any provider. U
 
 ## 19. Resource Discovery
 
-**Engine:** `services/resourcediscovery/` — a cross-service inventory engine that walks the Compute, Networking, Storage, Database, and Serverless drivers of any provider and returns a normalized `Resource` view (provider, service, type, ID, ARN/URN, region, tags, created-at). Auto-wired by every provider factory and exposed as `Provider.ResourceDiscovery`.
+**Engine:** `services/resourcediscovery/` — a cross-service inventory engine that walks the Compute, Networking, Storage, Database, Serverless, Databricks, and Kubernetes drivers of any provider and returns a normalized `Resource` view (provider, service, type, ID, ARN/URN, region, tags, created-at). Auto-wired by every provider factory and exposed as `Provider.ResourceDiscovery`.
 
 **SDK-compat handlers:** AWS Resource Explorer Two + Resource Groups Tagging API, Azure Resource Graph, and GCP Cloud Asset Inventory. All three sit on top of the same engine, so a tag written through any one path is visible through the others.
+
+**Surfaced resource types** (portable `service/Type` → per-provider inventory string):
+
+| Portable | AWS RE2 | Azure Resource Graph | GCP Cloud Asset |
+|---|---|---|---|
+| `compute/Instance` | `compute:instance` | `microsoft.compute/virtualmachines` | `compute.googleapis.com/Instance` |
+| `networking/VPC` · `Subnet` · `SecurityGroup` · `NetworkInterface` · `ElasticIP` | `networking:*` | `microsoft.network/*` | `compute.googleapis.com/*` |
+| `storage/Bucket` | `storage:bucket` | `microsoft.storage/storageaccounts` | `storage.googleapis.com/Bucket` |
+| `database/Table` | `database:table` | `microsoft.documentdb/databaseaccounts` | `firestore.googleapis.com/Database` |
+| `serverless/Function` | `serverless:function` | `microsoft.web/sites` | `cloudfunctions.googleapis.com/Function` |
+| `databricks/Workspace` | — | `microsoft.databricks/workspaces` | — |
+| `kubernetes/Cluster` | `kubernetes:cluster` | `microsoft.containerservice/managedclusters` | `container.googleapis.com/Cluster` |
+| `kubernetes/NodeGroup` | `kubernetes:nodegroup` | `microsoft.containerservice/managedclusters/agentpools` | `container.googleapis.com/NodePool` |
+
+Kubernetes clusters (EKS/GKE/AKS) and their node groups (nodegroups / node pools / agent pools) are surfaced via a `KubernetesClusters` discovery adapter each provider wires in over its cluster mock.
 
 ### Engine (`services/resourcediscovery/`)
 
--- a/providers/aws/aws.go
+++ b/providers/aws/aws.go
@@ -3,8 +3,10 @@ package aws
 
 import (
 	"context"
+	"strings"
 
 	"github.com/stackshy/cloudemu/v2/config"
+	cerrors "github.com/stackshy/cloudemu/v2/errors"
 	"github.com/stackshy/cloudemu/v2/providers/aws/awsiam"
 	"github.com/stackshy/cloudemu/v2/providers/aws/bedrock"
 	"github.com/stackshy/cloudemu/v2/providers/aws/cloudwatch"
@@ -13,6 +15,7 @@ import (
 	"github.com/stackshy/cloudemu/v2/providers/aws/ec2"
 	"github.com/stackshy/cloudemu/v2/providers/aws/ecr"
 	"github.com/stackshy/cloudemu/v2/providers/aws/eks"
+	eksdriver "github.com/stackshy/cloudemu/v2/providers/aws/eks/driver"
 	"github.com/stackshy/cloudemu/v2/providers/aws/elasticache"
 	"github.com/stackshy/cloudemu/v2/providers/aws/elb"
 	"github.com/stackshy/cloudemu/v2/providers/aws/eventbridge"
@@ -30,11 +33,19 @@ import (
 	"github.com/stackshy/cloudemu/v2/services/resourcediscovery"
 )
 
+// eksClusters is the slice of the EKS mock that discovery reads — an interface
+// so the vanished-cluster (NotFound) skip can be tested with a fake.
+type eksClusters interface {
+	ListClusters(ctx context.Context) ([]string, error)
+	DescribeCluster(ctx context.Context, name string) (*eksdriver.Cluster, error)
+	ListNodegroups(ctx context.Context, clusterName string) ([]string, error)
+}
+
 // eksDiscovery adapts the EKS mock to the resourcediscovery KubernetesClusters
 // capability, so EKS clusters and their node groups surface in Resource
 // Explorer. Kept in the provider package (not services/) to avoid inverting
 // the layering — the discovery engine stays free of provider imports.
-type eksDiscovery struct{ m *eks.Mock }
+type eksDiscovery struct{ m eksClusters }
 
 func (a eksDiscovery) DiscoverClusters(ctx context.Context) ([]resourcediscovery.DiscoveredCluster, error) {
 	names, err := a.m.ListClusters(ctx)
@@ -45,15 +56,25 @@ func (a eksDiscovery) DiscoverClusters(ctx context.Context) ([]resourcediscovery
 	out := make([]resourcediscovery.DiscoveredCluster, 0, len(names))
 
 	for _, name := range names {
-		// Fail loud: a Describe/list error for a cluster ListClusters just
-		// returned would otherwise silently drop it (or its node groups) from
-		// the inventory — the "silent-empty hides inventory" pattern.
+		// Discovery is a shared, polled surface, so a DeleteCluster can race
+		// between ListClusters and these per-cluster reads. A vanished cluster
+		// (NotFound) is correct to omit — skip it rather than fail the whole
+		// walk, which engine.List would propagate into every provider's
+		// inventory. Any other error is real and propagates.
 		c, err := a.m.DescribeCluster(ctx, name)
+		if cerrors.IsNotFound(err) {
+			continue
+		}
+
 		if err != nil {
 			return nil, err
 		}
 
 		ngs, err := a.m.ListNodegroups(ctx, name)
+		if cerrors.IsNotFound(err) {
+			continue
+		}
+
 		if err != nil {
 			return nil, err
 		}
@@ -61,6 +82,9 @@ func (a eksDiscovery) DiscoverClusters(ctx context.Context) ([]resourcediscovery
 		dc := resourcediscovery.DiscoveredCluster{Name: name, NodeGroups: ngs}
 		if c != nil {
 			dc.ARN = c.ARN // use the EKS mock's own ARN verbatim
+			// Keep Region in step with the verbatim ARN so the node-group ARN
+			// (built from Region) and Resource.Region can't diverge from it.
+			dc.Region = eksRegionFromARN(c.ARN)
 			dc.Tags = c.Tags
 		}
 
@@ -68,6 +92,20 @@ func (a eksDiscovery) DiscoverClusters(ctx context.Context) ([]resourcediscovery
 	}
 
 	return out, nil
+}
+
+// eksRegionFromARN pulls the region out of an EKS ARN
+// (arn:aws:eks:<region>:<account>:cluster/<name>). Returns "" if unparseable,
+// which leaves the walker to fall back to the engine's default region.
+func eksRegionFromARN(arn string) string {
+	const regionField = 3
+
+	parts := strings.Split(arn, ":")
+	if len(parts) <= regionField {
+		return ""
+	}
+
+	return parts[regionField]
 }
 
 // Provider holds all AWS mock services.
--- a/providers/aws/aws.go
+++ b/providers/aws/aws.go
@@ -45,14 +45,23 @@ func (a eksDiscovery) DiscoverClusters(ctx context.Context) ([]resourcediscovery
 	out := make([]resourcediscovery.DiscoveredCluster, 0, len(names))
 
 	for _, name := range names {
-		dc := resourcediscovery.DiscoveredCluster{Name: name}
-
-		if c, cerr := a.m.DescribeCluster(ctx, name); cerr == nil && c != nil {
-			dc.Tags = c.Tags
+		// Fail loud: a Describe/list error for a cluster ListClusters just
+		// returned would otherwise silently drop it (or its node groups) from
+		// the inventory — the "silent-empty hides inventory" pattern.
+		c, err := a.m.DescribeCluster(ctx, name)
+		if err != nil {
+			return nil, err
 		}
 
-		if ngs, ngErr := a.m.ListNodegroups(ctx, name); ngErr == nil {
-			dc.NodeGroups = ngs
+		ngs, err := a.m.ListNodegroups(ctx, name)
+		if err != nil {
+			return nil, err
+		}
+
+		dc := resourcediscovery.DiscoveredCluster{Name: name, NodeGroups: ngs}
+		if c != nil {
+			dc.ARN = c.ARN // use the EKS mock's own ARN verbatim
+			dc.Tags = c.Tags
 		}
 
 		out = append(out, dc)
--- a/providers/aws/aws.go
+++ b/providers/aws/aws.go
@@ -2,6 +2,8 @@
 package aws
 
 import (
+	"context"
+
 	"github.com/stackshy/cloudemu/v2/config"
 	"github.com/stackshy/cloudemu/v2/providers/aws/awsiam"
 	"github.com/stackshy/cloudemu/v2/providers/aws/bedrock"
@@ -27,6 +29,37 @@ import (
 	"github.com/stackshy/cloudemu/v2/providers/aws/vpc"
 	"github.com/stackshy/cloudemu/v2/services/resourcediscovery"
 )
+
+// eksDiscovery adapts the EKS mock to the resourcediscovery KubernetesClusters
+// capability, so EKS clusters and their node groups surface in Resource
+// Explorer. Kept in the provider package (not services/) to avoid inverting
+// the layering — the discovery engine stays free of provider imports.
+type eksDiscovery struct{ m *eks.Mock }
+
+func (a eksDiscovery) DiscoverClusters(ctx context.Context) ([]resourcediscovery.DiscoveredCluster, error) {
+	names, err := a.m.ListClusters(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	out := make([]resourcediscovery.DiscoveredCluster, 0, len(names))
+
+	for _, name := range names {
+		dc := resourcediscovery.DiscoveredCluster{Name: name}
+
+		if c, cerr := a.m.DescribeCluster(ctx, name); cerr == nil && c != nil {
+			dc.Tags = c.Tags
+		}
+
+		if ngs, ngErr := a.m.ListNodegroups(ctx, name); ngErr == nil {
+			dc.NodeGroups = ngs
+		}
+
+		out = append(out, dc)
+	}
+
+	return out, nil
+}
 
 // Provider holds all AWS mock services.
 type Provider struct {
@@ -112,6 +145,7 @@ func New(opts ...config.Option) *Provider {
 			Storage:    p.S3,
 			Database:   p.DynamoDB,
 			Serverless: p.Lambda,
+			Kubernetes: eksDiscovery{p.EKS},
 		},
 	)
 
--- a/providers/aws/eks_discovery_skip_test.go
+++ b/providers/aws/eks_discovery_skip_test.go
@@ -1,0 +1,93 @@
+package aws
+
+import (
+	"context"
+	"testing"
+
+	cerrors "github.com/stackshy/cloudemu/v2/errors"
+	eksdriver "github.com/stackshy/cloudemu/v2/providers/aws/eks/driver"
+)
+
+// fakeEKS lets a test drive DescribeCluster/ListNodegroups outcomes per cluster
+// name without racing a real DeleteCluster.
+type fakeEKS struct {
+	names      []string
+	describe   func(name string) (*eksdriver.Cluster, error)
+	nodegroups func(name string) ([]string, error)
+}
+
+func (f fakeEKS) ListClusters(context.Context) ([]string, error) { return f.names, nil }
+
+func (f fakeEKS) DescribeCluster(_ context.Context, name string) (*eksdriver.Cluster, error) {
+	return f.describe(name)
+}
+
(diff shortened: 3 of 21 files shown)


NitinKumar004 (commented):
## Review — surface Kubernetes clusters & node groups (EKS/GKE/AKS)

Clean, well-layered, well-tested feature — adapters live in the provider packages so the engine stays free of provider imports (mirrors Databricks), and coverage spans all three providers (walker unit + e2e + map tables). Build green. A few things worth a look:

### 🟠 MEDIUM — GCP ARN embeds the engine region, not the cluster's
`walkKubernetes` (`walkers.go:311`) computes `region := c.Region` (fallback `e.region`) for the `Resource.Region` field, but the ARN builders don't receive it — `kubernetesClusterARN(c.Name)` / `kubernetesNodeGroupARN(c.Name, ng)` embed `e.region` (`arn.go:123,137`).

A GKE cluster in `us-central1` on an engine defaulted to a different region surfaces with `Region: us-central1` but `ARN: …/locations/<engine-region>/clusters/prod` — the identifier contradicts the region field, for both the cluster and its node pools. AWS stays consistent (the adapter doesn't set `Region`, so both sides use `e.region`); Azure's `AzureID` embeds no region, so only GCP is affected.

Not caught by tests — the ARN assertion is a substring (`"clusters/prod"`) and the region check is on `Resource.Region` only. Suggest threading `region` into both ARN builders and adding an exact-ARN assertion.

### 🟡 LOW
- **AWS `eksDiscovery` swallows `DescribeCluster`/`ListNodegroups` errors** (`aws.go:66,70`). A `ListNodegroups` failure silently drops that cluster's node groups from the inventory — the same "silent-empty hides inventory" pattern just addressed for `walkNetworkInterfaces` in #296. The cluster still surfaces, so it's softer, but a failed listing is invisible.
- **Azure ARN uses `azureDefaultResourceGroup`, not the cluster's real RG** (`arn.go:127`). The AKS cluster knows its RG (`c.ResourceGroup`), but `DiscoveredCluster` carries no RG field, so the canonical ID points at the default group. Consistent with the other Azure discovery ARNs, so arguably by-design — flagging since the real value is available and dropped.
- **AWS adapter doesn't surface region** and ignores the EKS mock's real `Cluster.ARN`, rebuilding a best-effort one — harmless today, but inconsistent with the AKS/GKE adapters that set `Location`.
- **`NodeGroup` resources carry no tags**, and `ID` is the bare name (collides across clusters; the ARN disambiguates). Fine for scope.
- **Minor test gap:** GCP has a round-trip map test; Azure's `portableToAzureType` for kubernetes is only exercised in the reverse (`mapAzureType`) direction.

### ✅ Nits aside, nicely done
Symmetric tested type maps, cluster tags copied defensively via `copyTags`, node groups inherit the cluster region, docs updated. Only the GCP ARN/region mismatch is worth fixing before merge.

thzgajendra:
Thanks for the thorough review @NitinKumar004 — all addressed in 3f998b4:

- **🟠 GCP ARN embeds engine region, not the cluster's** — fixed. `kubernetesClusterARN`/`kubernetesNodeGroupARN` now take the resolved region (and resource group), so GCP self-links embed the cluster's own region for both the cluster and its node pools. Added an exact-ARN assertion (`…/locations/us-west-2/clusters/prod`) so the region field and the ARN can't diverge again.
- **AWS `eksDiscovery` swallows `DescribeCluster`/`ListNodegroups` errors** — fixed. The adapter now fails loud (propagates the error) instead of silently dropping a cluster's node groups — matching the `walkNetworkInterfaces` fix from #296.
- **Azure ARN uses `azureDefaultResourceGroup`, not the real RG** — fixed. `DiscoveredCluster` gained a `ResourceGroup` field; the AKS adapter sets it from `c.ResourceGroup` and the ARM ID now uses the cluster's real group (falling back to the default only when empty). E2e test asserts `resourceGroups/rg-1/`.
- **AWS adapter rebuilds the ARN instead of using the mock's real one** — fixed. `DiscoveredCluster` gained an `ARN` field; the EKS adapter passes the mock's own `Cluster.ARN`, which the walker uses verbatim.
- **Azure `portableToAzureType` only exercised in reverse** — fixed. Added a forward-direction test covering both kubernetes types.

Left as-is per your "fine for scope" call: node groups carry no tags, and the node-group `ID` is the bare name (the ARN disambiguates). Happy to add those if you'd prefer.

Verified: `go build`/`go vet`/`gofmt` clean, all touched suites pass.

NitinKumar004 (commented):
## Follow-up (deeper pass) — supersedes the region point above

Tracing the adapters into the `eks`/`aks`/`gke` mocks and the engine's `List` reframes the earlier "GCP region mismatch" as something sharper: a **canonical-identifier collision**.

### 🟠 MEDIUM — the ARN builders drop the dimension that distinguishes clusters

The K8s ARN builders ignore the cluster's real scope and use engine-level defaults:
- **GCP** `kubernetesClusterARN`/`kubernetesNodeGroupARN` embed `e.region` (`arn.go:123,137`), not the cluster's `c.Region`/`Location` — even though the walker surfaces `c.Region` correctly in `Resource.Region`.
- **Azure** embeds `azureDefaultResourceGroup` (`arn.go:127`), not the cluster's real `ResourceGroup` — which is available on `aks.ManagedCluster` (`aks.go:40`) and is exactly the key that distinguishes clusters (`clusterKey(rg, name) = rg + "/" + name`).

`idgen.AzureID` embeds the resource group; `idgen.GCPID` embeds the `locations/<e.region>` segment. Consequences:

1. **ARN contradicts the surfaced `Region`** (GCP): a GKE cluster in `us-central1` on an engine defaulted elsewhere gets `Region: us-central1` but `ARN: …/locations/<engine-region>/clusters/prod`.
2. **Identity collision.** The mocks support same-name clusters in different scopes (AKS keys on `rg+name`; GKE `ListClusters` spans locations). Two distinct clusters both named `prod` produce identical ARNs — GCP `…/locations/<engine-region>/clusters/prod`, Azure `…/resourceGroups/<default>/…/managedClusters/prod` — and `engine.List` does no dedup (it appends every match), so the inventory shows two entries with the same "canonical" ID. Node-group ARNs inherit the defect.

AWS is unaffected (EKS names are unique per account and everything stays in `e.region`).

**Suggested fix:** thread the cluster's location into the GCP builders (`c.Region`), and add a `ResourceGroup` field to `DiscoveredCluster` populated by the AKS adapter (`c.ResourceGroup`) and passed to both Azure builders. Add an exact-ARN test with two same-name clusters in different scopes — the current tests can't catch this (substring ARN check + single-scope fixtures).

### 🟡 LOW (new from this pass)
- Tags maps and `AgentPoolNames`/`NodePoolNames` slices returned by the mocks alias internal state (shallow struct copies). The walker `copyTags` protects the emitted `Resource`, but the read happens after the mock's lock is released — a theoretical race with a concurrent `UpdateCluster`. Pre-existing aliasing pattern, not introduced here; noting since the discovery walk widens who reads it.
- Confirmed the adapters call the right all-scope methods: `aks.ListClusters` spans all resource groups, `gke.ListClusters("")` spans all locations. Good.

The earlier LOW items (AWS adapter error-swallowing, no node-group tags, Azure round-trip test gap) still stand.

thzgajendra:
Thanks for the deeper pass @NitinKumar004 — the collision framing is a better way to see it. Status:

- **🟠 Identity collision (same-name clusters in different scopes → identical ARN)** — root cause fixed in 3f998b4 (region threaded into the GCP builders, `ResourceGroup` added to `DiscoveredCluster` and threaded into the Azure builders). eba1a20 adds the exact-ARN tests you asked for: `TestResourceDiscoveryGKESameNameDistinctARNs` (two `prod` clusters in `us-central1` + `europe-west1`) and `TestResourceDiscoveryAKSSameNameDistinctARNs` (two `prod` clusters in `rg-a` + `rg-b`) — each asserts **two distinct ARNs**, each embedding its own region / resource group. These fail against the pre-fix builders.
- **🟡 Aliasing of tags/`*Names` slices** — the emitted inventory is alias-safe: the walker copies tags via `copyTags` and node-group IDs are strings copied by value, so no alias escapes the discovery boundary. The residual is the mocks' `List*` returning shallow copies under-then-outside their lock — pre-existing and shared by every reader, so I've left it to a mock-level fix rather than widening this PR.
- Earlier LOWs: AWS adapter now fails loud on `DescribeCluster`/`ListNodegroups` errors (3f998b4); `portableToAzureType` forward-direction test added (3f998b4); node-group tags / bare `ID` left per your "fine for scope" call.

`go build`/`vet`/`gofmt` clean, all touched suites green.

NitinKumar004 (commented):
## Re-review @ `eba1a20` — collision fix confirmed; one new robustness item

Reviewed the fix (`3f998b4`) and collision tests (`eba1a20`) against the mocks and the engine's `List`.

### ✅ Identifier-collision fix — complete and well-pinned
- GCP `kubernetesClusterARN`/`NodeGroupARN` now embed the cluster's own `region`; Azure `DiscoveredCluster` gained `ResourceGroup`, threaded via `azureResourceGroupOrDefault`; AWS uses the EKS mock's verbatim `c.ARN`.
- Tests are exactly right — `TestResourceDiscoveryGKESameNameDistinctARNs` (2 regions) and `...AKSSameNameDistinctARNs` (2 resource groups) prove two same-name clusters yield **2 distinct ARNs**, each embedding its own scope, and the GKE test asserts `ARN` contains `locations/<r.Region>/…` (region↔ARN consistency). Azure forward-direction `portableToAzureType` test gap closed too.

### 🟠 MEDIUM (new — introduced by the fix) — AWS discovery aborts the *whole* inventory on a concurrent cluster delete
The AWS adapter went from best-effort to fail-loud: it now calls `DescribeCluster` + `ListNodegroups` per name from `ListClusters` and does `return nil, err` on any error (`providers/aws/aws.go`). Since `engine.List` propagates any walker error (`engine.go`: `if err != nil { return nil, err }`), the effect is:

- `ListClusters` returns names, releases its `RLock`. A concurrent `DeleteCluster` (discovery is a shared surface — RE2 / Resource Graph / Cloud Asset all poll the same engine) removes cluster `X`; the following `DescribeCluster(X)` returns `NotFound` → the **entire `ResourceDiscovery.List` fails**, dropping every provider's resources, not just Kubernetes.

The old code silently surfaced the cluster with empty detail (cosmetic gap); this over-corrects — a benign teardown-during-scan now becomes a total-inventory error. A genuinely-vanished cluster should be *skipped* (correct to omit), while real errors propagate. GCP/Azure are immune (single `ListClusters`, no per-item `Describe`).

Suggested: in `eksDiscovery`, treat `cerrors.IsNotFound(err)` from `DescribeCluster`/`ListNodegroups` as "skip this cluster, continue"; propagate only other errors.

### 🟡 LOW
- AWS cluster ARN (verbatim, built with the EKS mock's `m.opts.Region`) vs the node-group ARN / `Resource.Region` (which fall back to the engine's `e.region`, since the adapter doesn't set `dc.Region`) can embed different regions if those configs ever diverge. Equal in practice; setting `dc.Region` explicitly would make it robust.
- Node groups still carry no tags; mock-returned tag maps / node-group slices still alias stored state (the cluster `Resource` is protected by `copyTags`; node-group slice is read-only) — pre-existing class, unchanged.

Collision fix is exactly right; the only thing worth addressing from this pass is the fail-loud TOCTOU.

thzgajendra:
Thanks @NitinKumar004 — good catch on the blast radius. Fixed in 7bf85c2.

**🟠 MEDIUM — concurrent delete aborting the whole inventory:** the fail-loud `eksDiscovery` I added last round over-corrected. A `DeleteCluster` racing between `ListClusters` and the per-cluster `DescribeCluster`/`ListNodegroups` reads makes those reads return `NotFound`, and since `engine.List` propagates any walker error, that benign case would drop *every* provider's resources, not just the vanished cluster. Now `eksDiscovery` treats `cerrors.IsNotFound` as "omit this cluster, continue" and propagates only genuine (non-NotFound) errors — same skip-on-NotFound precedent as `walkStorage`/`walkDatabase`. (GKE/AKS are immune: single `ListClusters`, no per-item Describe.)

**🟡 LOW — region/ARN divergence:** now set `DiscoveredCluster.Region` from the cluster's verbatim ARN, so the node-group ARN (built from `Region`) and `Resource.Region` stay in step with the cluster ARN instead of silently falling back to the engine default if configs ever differ.

Also backed the adapter with a small `eksClusters` interface so the NotFound-skip and error-propagation paths are covered by a fake (`TestEKSDiscoverySkipsVanishedCluster`, `TestEKSDiscoveryPropagatesRealErrors`). `build`/`vet`/`gofmt` clean, full discovery + resourcegraph suites green.

NitinKumar004 (commented):
## Re-review @ `7bf85c2` — both round-3 items fixed; converged

The new commit resolves both prior findings, correctly and with tests.

### ✅ Verified
- **TOCTOU:** `DescribeCluster`/`ListNodegroups` now `continue` on `cerrors.IsNotFound` (skip a vanished cluster) and propagate every other error. Confirmed `IsNotFound(nil) == false`, so live clusters aren't skipped — and the fake-backed tests pin both the skip (Describe *and* ListNodegroups paths) and real-error propagation.
- **Region consistency:** `dc.Region = eksRegionFromARN(c.ARN)` — the node-group ARN and `Resource.Region` now derive from the same verbatim cluster ARN. `eksRegionFromARN` is partition-safe (region is field 3 for `aws`/`aws-us-gov` alike) and falls back to the engine region on an empty/malformed ARN.
- The `eksClusters` interface + `fakeEKS` are the right way to cover a race that can't be triggered deterministically.

### Converged
No remaining substantive findings. Across the passes, the ARN collision, region/resource-group threading, error-swallowing, the fail-loud TOCTOU, and region consistency are all fixed and test-pinned; layering / determinism / error-mapping / routing were clean throughout. Node-groups carry no tags and mock tag-map aliasing remain the only by-design LOW leftovers.

### One CI-gate nudge (non-substantive)
Adding the two `kubernetes/*` cases tips both Azure type-map switches over the complexity gate: `portableToAzureType` (`handler.go:255`) and `mapAzureType` (`kql.go:292`) now report `gocyclo` 11 (>10) locally — trivially fixed by making each a `map` lookup instead of a `switch`. Worth a quick confirm against CI's linter version.

Nice work — approvable on substance.

thzgajendra:
Thanks @NitinKumar004 — CI-gate nudge handled in 27a64ca.

Both Azure type-map switches are now package-level map lookups instead of `switch` statements:
- `portableToAzureType` → `portableToAzureTypeMap` (service/type key → ARM type; unmapped falls back to `strings.ToLower(key)`, unchanged).
- `mapAzureType` → `azureToPortableType` (Azure type → `portableResourceType{service, typ}`; the zero value gives the same `("", "")` match-none for unmapped types).

Verified with `gocyclo -over 10` — nothing flagged; top complexity in the package is now 8 (`applyWhere`/`applyLimit`). Adding future type pairs no longer raises complexity. `build`/`vet`/`gofmt` clean, resourcegraph suite green.

NitinKumar004 (commented):
## 360° review @ `27a64ca` — is it done "the cloudemu way"?

Functionally correct (all earlier correctness findings are fixed and test-pinned). This pass is about convention/architecture fit.

### 1. Architecture fit — the discovery-wiring idiom
cloudemu's established pattern (Databricks + the 5 portable services) is: the engine holds a **shared driver interface** and the walker calls it directly — `walkDatabricks` → `e.drivers.Databricks.ListWorkspaces(ctx)`, projecting the driver's own type. That works because Databricks has a shared `services/databricks/driver`.

#297 uses a **different** idiom: a `KubernetesClusters` capability interface + a neutral `DiscoveredCluster` projection + a **per-provider adapter** (`eksDiscovery`/`aksDiscovery`/`gkeDiscovery`) wired in each provider factory.

**Verdict — a justified, documented deviation, but a second pattern rather than the idiomatic one:**
- ✅ Acceptable because there is no shared managed-cluster driver — EKS/GKE/AKS control-plane mocks live in provider packages with heterogeneous APIs, and the engine must stay free of provider imports. Adapter-in-provider honors the layering rule (no inversion).
- ⚠️ Not the *most* cloudemu-idiomatic: the strictly-consistent form would be a shared `services/kubernetes` managed-cluster driver (normalized `Cluster`) that all three mocks implement, so the engine holds it like `Databricks` and `walkKubernetes` mirrors `walkDatabricks`. As written, projection/business logic (list + per-cluster Describe + node-group listing + NotFound-skip + ARN region-parse) lives in the **composition root** (`providers/{aws,azure,gcp}/*.go`), which historically only *wires* services. Correct layer, heavier factory, divergent idiom. A judgment call, not a defect.

### 2. Not proper per cloudemu — should fix
The `gocyclo` fix (`27a64ca`) traded that gate violation for a **`gochecknoglobals`** one:
- `server/azure/resourcegraph/handler.go:256` `portableToAzureTypeMap`
- `server/azure/resourcegraph/kql.go:294` `azureToPortableType`

Both are package-level map globals with **no `//nolint`**. `.golangci.yml` enables `gochecknoglobals` and CLAUDE.md requires `//nolint:gochecknoglobals // reason` for globals — the repo's own `elbActions` (`elbv2/handler.go`) annotates a static lookup table exactly that way. So the map refactor is the right idea but needs `//nolint:gochecknoglobals // static lookup table` (or build the maps function-local). As-is it fails the 0-lint gate.

### 3. What is proper (all good)
- ARNs via `idgen` (`AWSARN`/`AzureID`/`GCPID`), region/RG threaded so identifiers are scope-unique; `cerrors` with correct NotFound-skip-vs-propagate.
- Tests across all 3 providers; provider tests use plain `t.Fatalf/Errorf` (no testify, per convention); collision/skip/region cases pinned; symmetric tested type maps both directions.
- `copyTags` on emitted resources (avoids the aliasing class); named const in `eksRegionFromARN`, partition-safe parse; clean layering (engine free of provider imports, compile-time interface satisfaction).

### Bottom line
Correct and safe; ~85% "the cloudemu way." The one genuinely not-proper item is the un-annotated global maps (add the `//nolint`, or make them function-local). The adapter architecture is an acceptable, well-documented deviation; the strictly-idiomatic form is a shared managed-cluster driver so it mirrors `walkDatabricks` — worth considering, not required.

thzgajendra:
Thanks @NitinKumar004 — fixed in 1b331df.

**Item 2 (the genuinely not-proper one):** the two static type-map globals now carry the repo convention `//nolint:gochecknoglobals // static lookup table`, matching `elbActions` in `elbv2/handler.go`. Verified with `golangci-lint run --enable-only gochecknoglobals ./server/azure/resourcegraph/...` → `0 issues`; `build`/`gofmt` clean. Kept them package-level (vs function-local) so they're built once, consistent with `elbActions`.

**Item 1 (architecture idiom):** noted and agreed — the capability-interface + per-provider-adapter idiom is a deliberate deviation because there's no shared managed-cluster driver (the three control-plane mocks have heterogeneous APIs and the engine must stay free of provider imports). The strictly-idiomatic form — a shared `services/kubernetes` driver so `walkKubernetes` mirrors `walkDatabricks` — is worth a follow-up but is a larger refactor across all three providers; leaving it out of this PR's scope.

NitinKumar004 (commented):
## 360° re-review (all-way) @ `1b331df`

### ✅ Prior "not-proper" item — fixed
`1b331df` annotates both type-map globals with `//nolint:gochecknoglobals // static lookup table` — the CLAUDE.md-sanctioned form, matching the `elbActions` precedent. `server/azure/resourcegraph` is now lint-clean; docs are accurate.

### 🟡 The one remaining shortfall — coverage stops one layer short of its own precedent
The pattern this PR mirrors, **Databricks**, has a handler-level end-to-end test — `TestSDKResourceGraph_DatabricksIndexing` — proving a workspace is retrievable through the *actual* Resource Graph query handler via the real SDK.

Kubernetes coverage stops below that:
- **unit** — `TestPortableToAzureType` / `TestMapAzureType` / `TestMapGCPAssetType` (type translation in isolation), and
- **engine-level** — provider tests call `p.ResourceDiscovery.List(...)` directly.

There is no `TestSDKResourceGraph_KubernetesIndexing` (nor GCP Cloud Asset / AWS RE2 equivalent) that drives a real query — e.g. `resources | where type == "microsoft.containerservice/managedclusters"` — through the HTTP handler and asserts a cluster row returns. So the handler↔engine seam for the new kubernetes types is covered piecewise but not end-to-end, even though the service it copied sets that precedent.

Suggested: add a `TestSDKResourceGraph_KubernetesIndexing` (and a Cloud Asset filter equivalent) mirroring `TestSDKResourceGraph_DatabricksIndexing`.

### 🔵 Architecture note (unchanged, acceptable)
Still a second discovery idiom (capability adapter + `DiscoveredCluster` projection) vs. the engine-holds-shared-driver pattern (Databricks). Justified by the absence of a shared managed-cluster driver and documented — acceptable deviation. Strictly-idiomatic target remains a shared `services/kubernetes` cluster driver so `walkKubernetes` mirrors `walkDatabricks`.

### ✅ Everything else — proper
`idgen` ARNs (scope-unique), `cerrors` with correct NotFound-skip/propagate, all-3-provider tests, no-testify provider convention, `copyTags`, named consts, partition-safe ARN parse, clean layering, symmetric type maps both directions, `Query{Type:...}` filter works.

### Bottom line
Lint-clean and correct. The only convention shortfall left is the missing handler-level SDK indexing test for kubernetes that its Databricks precedent has — worth adding; the adapter architecture is an accepted, documented deviation.

thzgajendra:
Thanks @NitinKumar004 — closed the coverage gap in e8c7a84.

Added a handler-level SDK round-trip test on each of the three discovery surfaces, mirroring `TestSDKResourceGraph_DatabricksIndexing` (drives the real client → in-memory handler → engine, no engine shortcut):

- **Azure Resource Graph** — `TestSDKResourceGraph_KubernetesIndexing`: seeds an AKS cluster + agent pool, asserts both appear in a bare `Resources` query, that `where type == 'microsoft.containerservice/managedclusters'` narrows to just the cluster (with an ARM-shaped id embedding `resourceGroups/rg-1/`), and that an unmodeled type matches none (not all).
- **GCP Cloud Asset** — `TestSDKCloudAsset_KubernetesIndexing`: seeds a GKE cluster + node pool, asserts both surface in `searchAllResources`, and that `assetType:container.googleapis.com/Cluster` (and `assets.list` with the same assetType) narrows to just the cluster.
- **AWS Resource Explorer 2** — `TestSDKResourceExplorer2_KubernetesIndexing`: seeds an EKS cluster + node group, asserts both surface in an unscoped `Search`, that `service:kubernetes` narrows to the two EKS resources, and that the cluster ARN is EKS-shaped (`:eks:` … `:cluster/prod`).

All pass; `build`/`vet`/`gofmt` clean, all three server suites green.

NitinKumar004 (approved):
## Approving @ `e8c7a84`

The handler-level indexing gap from the last pass is closed — `e8c7a84` adds end-to-end SDK tests across all three discovery surfaces, each driving the real client against an `httptest` server and asserting both surfacing and filter-narrowing (with a negative case):
- `TestSDKResourceGraph_KubernetesIndexing` (Azure Resource Graph)
- `TestSDKCloudAsset_KubernetesIndexing` (GCP Cloud Asset)
- `TestSDKResourceExplorer2_KubernetesIndexing` (AWS Resource Explorer 2)

All three pass; the discovery/provider suites pass.

Everything raised across the review rounds is resolved:
- ARN identifier collisions (GCP region / Azure resource group threaded; distinct-ARN tests)
- fail-loud TOCTOU → skip vanished clusters (`IsNotFound`), propagate real errors
- region↔ARN consistency (`eksRegionFromARN`)
- `gochecknoglobals` on the type-map globals annotated per convention
- handler-level SDK indexing tests added (this commit)

`server/azure/resourcegraph` and the kubernetes code are lint-clean. (The two remaining `services/resourcediscovery` lint hits — `walkers.go` gci, `arn.go` goconst — come from `80b4e20`/#296, already on `development`, not this PR.)

Architecture note stands as an accepted, documented choice: the discovery adapter + `DiscoveredCluster` projection is a reasonable deviation from the shared-driver (`walkDatabricks`) idiom given there's no shared managed-cluster driver — not a blocker.

LGTM.

PR #304, opened 2026-07-30
Add GCP AlloyDB full-parity support

thzgajendra:
## Objective
Add **GCP AlloyDB** as a first-class managed database server in cloudemu, at full parity (service + connectivity + metrics + cost + all child resources), reusing the `relationaldb` driver.

> **⚠️ Stacked on #303.** This branch is cut from `feat/sql-full-parity` (PR #303), which carries the full relationaldb driver AlloyDB depends on (Users/Databases/etc.) and isn't on `development` yet. **Review the single AlloyDB commit `a37e2b2`** (`git show a37e2b2`) — the rest of the diff is #303. Once #303 merges, this diff collapses to AlloyDB only.

## What we found
DocumentDB/Neptune are already emulated via the RDS query-protocol handler; the genuinely-missing managed DB servers are MemoryDB/Timestream/Keyspaces (AWS), Spanner/Bigtable/**AlloyDB** (GCP), Managed Cassandra (Azure). AlloyDB is the first because it maps cleanly onto the relational driver.

## How it's built
- **Provider** `providers/gcp/alloydb`: implements `RelationalDB` (AlloyDB cluster→`Cluster`, instance PRIMARY/READ_POOL/SECONDARY→`Instance`, backup→`ClusterSnapshot`, restore) + `Users` + `Databases`, plus a new optional **`rdsdriver.AlloyDB`** capability (mirrors the `ManagedInstances` precedent) for AlloyDB-specifics: rich cluster/instance create, cross-region **secondary + promote**, instance **failover/restart**, continuous/automated-backup + maintenance config, and `*Info` accessors. Cloud Monitoring metrics on instance create.
- **Server** `server/gcp/alloydb`: `alloydb.googleapis.com/v1` REST handler (clusters[/instances|/users], backups, LRO operations, custom methods `:promote`/`:createsecondary`/`:restore`/`:failover`/`:restart`) using the SDK types for wire fidelity.
- **Wiring**: `providers/gcp/gcp.go` (monitoring + combined CloudSQL+AlloyDB relational **discovery** adapter, `TypeAlloyDBCluster`); dedicated **cost** rates; docs updated.

## Built to the #303 review lessons (from the start)
Copy-on-read (clone Tags/slices on every return path; aliasing test), copy-on-write + single-lock RMW, cascade delete, lifecycle **state guards** (promote/secondary preconditions, invalid instance-type + `/`-name rejection), metrics-on-create, dedicated cost keys, provider unit + SDK round-trip + **wire-level error-mapping (404/409/400)** tests.

## Known limitation (documented)
AlloyDB's REST paths (`/v1/projects/{p}/locations/{l}/clusters…`) are **identical to GKE's**, and the SDK requests carry no service-distinguishing header — so the two can't be multiplexed on one server. `Drivers.AlloyDB` is therefore opt-in (left nil in `DriversFrom`; injected in place of GKE), exactly like the existing `K8sAPI` pattern. Standalone AlloyDB usage is fully functional and tested.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `go test ./...` green
- [x] `golangci-lint` clean on changed packages
- [x] Provider unit tests (lifecycle, aliasing, cascade, validation, secondary/promote, metrics) — ~83%
- [x] Server SDK round-trip + wire-error tests via the real `google.golang.org/api/alloydb/v1` client

## Risk & rollback
Additive: a new provider + server package + one optional driver capability + opt-in wiring. No existing behavior changes (AlloyDB nil by default). Rollback = revert `a37e2b2`.


NitinKumar004 (commented):
## Deep end-to-end lifecycle review — GCP AlloyDB (commit `a37e2b2`)

Reviewed the single AlloyDB commit in an isolated worktree (gates, static/coverage, provider lifecycle, server/SDK/LRO, and cross-cutting wiring, with adversarial verification). **The whole lifecycle is implemented and round-trips through the real `alloydb/v1` SDK — but it is not gap-free.**

**Gates: green** — `build ✓ vet ✓ test ✓ -race ✓ gofmt ✓ go mod tidy ✓`; lint clean on the AlloyDB files. **Discovery has no CloudSQL regression** (the combined adapter delegates to the unchanged CloudSQL path then appends AlloyDB); cost keys are consistent with the Aurora/managed-SQL precedent; docs op-counts match the 8-method capability; copy-on-read, single-lock RMW, promote consistency, LRO fidelity, and body decoding are all verified solid.

### Lifecycle scorecard
| Stage | Status |
|---|---|
| Cluster create/get/list/patch/delete | works — but delete doesn't detach secondaries |
| Instance create/get/list/patch/delete | works — but no one-primary rule |
| Secondary + promote | promote guards correct; dangling link on *delete* |
| Failover / restart | no PRIMARY guard (no-op; invalid calls accepted) |
| Backup / restore | complete (backups correctly survive cluster delete) |
| Users / Databases | complete except UpdateUser route missing |
| LRO / custom methods / wire errors | round-trip via real SDK (404/409 tested; 400 not) |

### Medium (should fix before merge)
- **GKE-vs-AlloyDB routing footgun** — the "AlloyDB in place of GKE" contract is comment-only; the most-likely wiring silently breaks GKE. *(inline)*
- **Dangling SECONDARY link on PRIMARY delete** — the exact #303 replica class the PR aimed to avoid; promote is fine, delete isn't. *(inline)*
- **One-primary-per-cluster not enforced** — two PRIMARY instances can be created in one cluster. *(inline)*
- **UpdateUser is dead** — the driver method is wired but no PATCH route, so `Users.Patch` → 405. *(inline)*

### Low
- `FailoverInstance`/`RestartInstance` have no PRIMARY/type guard — failover on a READ_POOL silently succeeds. *(inline)*
- Unknown `:verb` mis-dispatches, and `promoteCluster` ignores `r.Method` → `GET .../clusters/{c}:promote` **promotes** (a state change on a read).
- `DescribeClusters`/`DescribeInstances` list-all use `.All()` (unordered) while snapshots/users/dbs sort — inconsistent determinism within the same PR (`clusters.go:76`, `instances.go:122`).
- **8 banner-comment blocks** (`// ---- X ----`) across the AlloyDB files — contradicts the project's no-banner-comments convention.
- Coverage below the 90% pillar: `server/gcp/alloydb` 70.2%, provider 82.8%; the 400 error path isn't wire-tested; the aliasing test covers only `DescribeClusters`.
- `patchCluster`/`patchInstance` ignore the capability-assertion `ok` (nil-panic if a plain `RelationalDB` were ever registered — unreachable today); `toWireUser.Name` is the bare id, not the full resource path; `operations` GET returns a bare `done` with no `response`.
- go.mod: the `google.golang.org/api` v0.277→v0.291 bump drags a broad transitive update (otel, grpc, `golang.org/x/*`) — tidy-clean and building, but wider blast radius than AlloyDB.

**Verdict: comment.** Additive and gate-green, with the full lifecycle working — but the GKE routing footgun (top priority) and the dangling-secondary / two-primaries / dead-UpdateUser Mediums are worth closing before merge.

server/gcp/gcp.go
```diff
@@ -159,6 +166,14 @@ func New(d Drivers) *server.Server {
 		srv.Register(cloudsql.New(d.CloudSQL))
 	}
 
+	// AlloyDB matches /v1/projects/{p}/locations/{l}/{clusters|backups|
+	// operations}/... — the cluster/operations paths are identical to GKE's, so
+	// the two are mutually exclusive on one server. Registered before GKE so an
+	// AlloyDB-configured server (GKE nil) works; DriversFrom leaves AlloyDB nil.
+	if d.AlloyDB != nil {
```
NitinKumar004, line 182: **[Medium] GKE-vs-AlloyDB routing footgun.** AlloyDB and GKE are registered as independent `if != nil` blocks (AlloyDB **first**), and dispatch is first-match-wins. But `DriversFrom` returns `GKE: p.GKE` (non-nil) and `AlloyDB: nil` (`fromprovider.go:22,37`), and AlloyDB's `Matches` claims the same `/locations/.../clusters` + `/operations` paths as GKE. → *The natural enable — `d := DriversFrom(p); d.AlloyDB = p.AlloyDB` — leaves **both** non-nil, so AlloyDB shadows GKE's cluster/operations paths and the GKE control plane silently breaks.* The "in place of GKE" contract lives only in comments; there's no helper/guard/test. Suggest a `DriversFromWithAlloyDB` that nils GKE, or a guard in `New` rejecting both non-nil.

providers/gcp/alloydb/clusters.go
```diff
@@ -0,0 +1,326 @@
...
+	delete(m.clusterExtra, id)
+	m.deleteClusterChildren(id)
+
+	return nil
+}
+
+// deleteClusterChildren removes instances, users and databases under a cluster.
+// The caller holds the write lock. The trailing '/' prefix guard prevents
+// "foo" from sweeping "foo2/…".
+func (m *Mock) deleteClusterChildren(cluster string) {
```
NitinKumar004, line 172: **[Medium] Dangling SECONDARY link on PRIMARY delete.** `deleteClusterChildren` removes prefix-keyed instances/users/databases but never scans `clusterExtra` for other clusters whose `PrimaryCluster == id`. → *`CreateSecondaryCluster(PrimaryCluster="P")` then `DeleteCluster("P")` leaves the secondary with `clusterExtra[S].PrimaryCluster == "P"` pointing at a deleted cluster; `AlloyDBClusterInfo(S)` then reports a ghost primary.* This is the same replica-linkage class fixed in #303 — real AlloyDB blocks deleting a primary with live secondaries or detaches them. Detach (or reject) on delete, and add a test for primary-delete-with-live-secondary.

providers/gcp/alloydb/alloydb_native.go
```diff
@@ -0,0 +1,303 @@
...
+
+	out := cloneCluster(cluster)
+
+	return &out, nil
+}
+
+// CreateAlloyDBInstance creates an instance of the given type under a cluster.
+//
+//nolint:gocritic // cfg matches the AlloyDB capability signature.
+func (m *Mock) CreateAlloyDBInstance(
```
NitinKumar004, line 219: **[Medium] One-primary-per-cluster not enforced.** `CreateAlloyDBInstance` (and the base `CreateInstance`, which stamps `InstanceType: PRIMARY` unconditionally) checks duplicate ID and valid type but not whether the cluster already has a PRIMARY. → *Creating two PRIMARY instances in one cluster both succeed; real AlloyDB permits exactly one primary.* Also worth guarding: a SECONDARY instance type should only live in a SECONDARY cluster, and READ_POOL `NodeCount` is stored unvalidated (0 accepted).

server/gcp/alloydb/users.go
```diff
@@ -0,0 +1,92 @@
...
+
+import (
+	"net/http"
+
+	alloydb "google.golang.org/api/alloydb/v1"
+
+	rdsdriver "github.com/stackshy/cloudemu/v2/services/relationaldb/driver"
+)
+
+func (h *Handler) serveUsers(w http.ResponseWriter, r *http.Request, p *alloyPath) {
```
NitinKumar004, line 11: **[Medium] UpdateUser is unreachable (dead capability).** `driver.Users.UpdateUser` is implemented and functional, and the real SDK exposes `Users.Patch`, but the item-level router here only handles GET/DELETE. → *`svc.Projects.Locations.Clusters.Users.Patch(...)` → 405 METHOD_NOT_ALLOWED.* Add an `http.MethodPatch` case that calls `UpdateUser`.

providers/gcp/alloydb/alloydb_native.go
```diff
@@ -0,0 +1,303 @@
...
+	m.emitInstanceMetrics(cfg.ClusterID, cfg.ID, cpuMetricRunning, connRunning)
+
+	out := cloneInstance(inst)
+
+	return &out, nil
+}
+
+// FailoverInstance triggers a failover of a REGIONAL primary instance. The
+// instance stays available; the action just verifies it exists.
+func (m *Mock) FailoverInstance(_ context.Context, clusterID, instanceID string) (*rdsdriver.Instance, error) {
```
NitinKumar004, line 287: **[Low] `FailoverInstance`/`RestartInstance` have no PRIMARY/type guard.** Both share `instanceAction`, which only verifies the instance exists and returns a clone — no instance-type check, no state transition. → *`FailoverInstance` on a READ_POOL instance silently succeeds, whereas real AlloyDB `instances.failover` is valid only on the PRIMARY.* The no-op is defensible, but the missing precondition means an invalid call is accepted.

server/gcp/gcp.go
thzgajendra: Fixed in c1cfafd. `New` now panics fast if both `AlloyDB` and `GKE` are non-nil (they share `/locations/.../clusters` paths), and added `DriversFromWithAlloyDB(p)` which enables AlloyDB and nils GKE — so the natural enable path is safe and the ambiguous one fails loudly instead of silently shadowing GKE.

providers/gcp/alloydb/clusters.go
thzgajendra: Fixed in c1cfafd. `DeleteCluster` now calls `detachSecondaries(id)`, which scans `clusterExtra` and clears `PrimaryCluster` on any SECONDARY pointing at the deleted primary — so `AlloyDBClusterInfo(secondary)` no longer reports a ghost primary. Covered by `TestPrimaryDeleteDetachesSecondaries`.

providers/gcp/alloydb/alloydb_native.go
thzgajendra: Fixed in c1cfafd. Both native `CreateAlloyDBInstance` and base `CreateInstance` now reject a second PRIMARY (`clusterHasPrimary` → FailedPrecondition). Also added: a SECONDARY instance requires a SECONDARY cluster, and READ_POOL requires `nodeCount > 0`. Covered by `TestOnePrimaryPerCluster`.

server/gcp/alloydb/users.go
thzgajendra: Fixed in c1cfafd. Added an `http.MethodPatch` case in the user item router → `UpdateUser`; `Users.Patch(...)` now returns 200. Covered by `TestSDKAlloyDBUserPatchAndGuards`.

providers/gcp/alloydb/alloydb_native.go
thzgajendra: Fixed in c1cfafd. `FailoverInstance` now requires the target be a PRIMARY (FailedPrecondition otherwise); `RestartInstance` stays valid on any type. Covered by `TestFailoverRequiresPrimary`.

thzgajendra:
Thanks for the deep lifecycle pass — all findings addressed in `c1cfafd` (per-comment replies inline). Beyond the 4 Mediums + the failover Low, I also closed the other Lows you flagged:

- **Unknown `:verb` mis-dispatch / GET-promotes-on-read** — custom methods (`:promote`/`:failover`/`:restart`) are now POST-only; a non-POST is 405 and an unknown verb 404, so `GET .../clusters/{c}:promote` no longer promotes.
- **List determinism** — `DescribeClusters`/`DescribeInstances` now use `SortedValues()`, matching the snapshot/user/db paths.
- **Banner comments** — removed all 8 `// ---- X ----` blocks (project no-banner convention).
- **`patchCluster`/`patchInstance` ignored the capability `ok`** — now return 501 UNIMPLEMENTED instead of risking a nil-panic.
- **`toWireUser.Name`** — now the full resource path (`projects/…/clusters/…/users/…`), not the bare id.
- **Coverage** — added tests for one-primary, secondary-in-secondary-cluster, READ_POOL nodeCount, failover-requires-primary, primary-delete-detaches-secondary, `Users.Patch`, GET-`:promote` rejection, the **400** wire path, and instance output-aliasing. Provider ~84%, server ~70%.

Deliberately left as-is (noted, not bugs):
- **`operations` GET returns bare `done:true`** with no `response` — the mutating calls already return a terminal operation carrying the resource, so the poll is only a fallback; fine for SDK round-trips.
- **go.mod transitive bump** — `alloydb/v1` requires `google.golang.org/api` ≥ 0.291, which pulls the wider update; unavoidable for the new SDK, tidy-clean and green.

Gates: `build ✓ vet ✓ test ✓ (full) golangci-lint ✓ (0 issues on changed pkgs)`.

NitinKumar004 (approved):

Verified `c1cfafd` against the earlier review; every finding is resolved, correctly and with tests. Gates green (`build ✓ vet ✓ test ✓ -race ✓ gofmt ✓ go mod tidy ✓`; lint clean on the AlloyDB files).

**Medium — all fixed:**
- **GKE-vs-AlloyDB routing footgun** → `DriversFromWithAlloyDB` helper that nils GKE, plus a fail-fast `panic` in the server wiring if both are enabled — the ambiguous state can no longer silently shadow GKE.
- **Dangling SECONDARY link on PRIMARY delete** → `detachSecondaries` clears the primary link on every secondary before delete (the #303 replica class, now closed on the delete path too).
- **One-primary-per-cluster** → `clusterHasPrimary` guard rejects a second PRIMARY instance.
- **UpdateUser was unreachable** → `Users.Patch` now routes to `patchUser` → `UpdateUser`.

**Low — fixed:**
- `FailoverInstance` now requires a PRIMARY; `RestartInstance` correctly stays unrestricted.
- `:promote`/`:failover`/`:restart` now 405 on a non-POST and 404 on an unknown verb (a `GET …:promote` no longer triggers a state change).
- `DescribeClusters`/`DescribeInstances` list-all now use `SortedValues()` (deterministic, consistent with the rest of the package).
- Banner-comment blocks removed.

Full lifecycle (cluster → instances → users → databases → backup → restore → secondary → promote → failover → restart → delete) is implemented, wired (metrics/cost/discovery, no CloudSQL regression), and round-trips through the real `alloydb/v1` SDK. Non-blocking follow-up: `server/gcp/alloydb` coverage is ~70% (consistent with the other GCP server handlers; broad wire-level fault-injection would lift it). LGTM.

PR #306, opened 2026-07-31
feat(databricks): remaining Microsoft.Databricks ARM resources (#209)

### LINKED ISSUE #209: feat: Azure Databricks remaining ARM resources SDK-compat
Finish the Microsoft.Databricks ARM surface beyond Workspaces (built in #164).

Operations:
- Access Connectors: createOrUpdate, get, delete, update, list by RG, list by subscription
- Private Endpoint Connections: create, get, list, delete
- Private Link Resources: get, list
- VNet Peering: createOrUpdate, get, list, delete
- Outbound Network Dependencies Endpoints: list
- Operations: list

Satyam-Trivedi-ZS:
## Summary

Finishes the `Microsoft.Databricks` ARM control-plane surface beyond Workspaces (built in #164), closing #209. Every resource is reachable over the **real `armdatabricks` SDK** pointed at the emulator.

| Resource | Path | Operations |
|---|---|---|
| **Access Connectors** | `accessConnectors/{name}` | createOrUpdate, get, update, delete, list-by-RG, list-by-subscription |
| **Private Endpoint Connections** | `workspaces/{w}/privateEndpointConnections/{name}` | create, get, list, delete |
| **Private Link Resources** | `workspaces/{w}/privateLinkResources/{groupId}` | get, list |
| **VNet Peering** | `workspaces/{w}/virtualNetworkPeerings/{name}` | createOrUpdate, get, list, delete |
| **Outbound Network Dependencies** | `workspaces/{w}/outboundNetworkDependenciesEndpoints` | list |
| **Operations** | `/providers/Microsoft.Databricks/operations` | list |

## Implementation

Built across all four layers, following the existing workspace/bedrock/rds pattern:

- **driver** — `services/databricks/driver/arm_resources.go`: new types + an `ARMResources` interface composed into `Databricks`.
- **provider** — `providers/azure/databricks/arm_*.go`: in-memory `Mock`, memstore-backed with copy-on-write reads/writes; deterministic principal/tenant GUID synthesis for system-assigned identities; synthesized per-workspace private-link and outbound-dependency catalogs; static operations catalog.
- **service** — `services/databricks/arm_resources.go`: portable API wrapped with the same `do()` cross-cutting pipeline (recorder/metrics/rate-limit/inject/latency) as the workspace ops.
- **server** — `server/azure/databricks/*`: the ARM HTTP handler's routing is extended for the new top-level `accessConnectors` type, the four workspace sub-resource collections, and the subscription-less `operations` path (special-cased because `azurearm.ParsePath` requires a `/subscriptions` prefix). Faithful ARM wire types match what the SDK emits/expects.

## Fidelity — store-and-echo

The ARM resources round-trip faithfully over the SDK (access connectors and peerings persist and list/describe; a system-assigned access-connector identity gets synthesized principal/tenant IDs; a created peering springs to `Connected`/`Succeeded`), but the underlying Azure networking side effects are **not** simulated — a private-endpoint connection stores its approval state without a real endpoint on the platform side, private-link and outbound-dependency endpoints are a synthesized workspace-scoped catalog rather than a live probe, and a VNet peering does not actually peer networks. This is disclosed in `docs/services.md` §21, consistent with the repo's accept-and-echo convention.

## Testing

Verified against the real `armdatabricks@v1.1.0` SDK (paths, models, and response shapes taken from the vendored client — no assumptions).

- **21 SDK round-trip tests** driving the real `armdatabricks` clients (`AccessConnectorsClient`, `PrivateEndpointConnectionsClient`, `PrivateLinkResourcesClient`, `VNetPeeringClient`, `OutboundNetworkDependenciesEndpointsClient`, `OperationsClient`) against an `httptest` server — the genuine user path.
- **9 provider-level unit tests** against the `Mock` directly.
- Edge cases covered: not-found, missing parent workspace (404), empty list, PATCH tag/identity semantics, `None`/system-assigned identity, rejected PEC status, name-set-independent listing, and the no-subscription operations path.

The e2e tests caught a real wire-shape bug during development — the Outbound list response must be a bare JSON array (the SDK unmarshals straight into a slice), not a `{"value":[...]}` envelope — now fixed and verified against the SDK's deserializer.

**Gates:** `gofmt` · `go build ./...` · `go vet ./...` · full `go test ./...` · `go test -race` (databricks) · `golangci-lint` (0 issues) · `go mod tidy` (unchanged) — all green.

Closes #209.

---

### Note — one bundled unrelated fix (disclosed)

This branch also carries a **one-line, unrelated fix** in `server/aws/bedrock/streaming.go` (commit `190caea`): the `converse-stream` handler now drains the request body (`io.Copy(io.Discard, r.Body)`) before starting the chunked event-stream response.

Why it's here: the required `Test` CI job was intermittently failing under `-race` on `TestSDKConverseStream` with `use of closed network connection` — a pre-existing, load-sensitive flake from #298 (an undrained request body makes net/http tear the connection down non-gracefully once streaming has started, racing the SDK client's read). It is **not** part of the Databricks change; I bundled it only to get this PR's CI reliably green (I have no admin rights to re-run the job). It is a correct, self-contained fix. Happy to split it into its own PR and rebase this one to drop it if you'd prefer a clean single-purpose diff — just say the word.

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -1514,6 +1514,32 @@ Azure-only. The control plane backs the real `armdatabricks` SDK; the data plane
 | `ListWorkspacesByResourceGroup` | `(ctx, resourceGroup) ([]Workspace, error)` |
 | `ListWorkspaces` | `(ctx) ([]Workspace, error)` |
 
+### Extended ARM resources (control plane)
+
+The rest of the `Microsoft.Databricks` ARM surface beyond workspaces
+(`services/databricks/driver/arm_resources.go`), reachable over the real
+`armdatabricks` SDK:
+
+| Resource | Operations |
+|----------|------------|
+| **Access Connectors** (`accessConnectors`) | `CreateOrUpdateAccessConnector`, `GetAccessConnector`, `UpdateAccessConnector`, `DeleteAccessConnector`, `ListAccessConnectorsByResourceGroup`, `ListAccessConnectors` |
+| **Private Endpoint Connections** (`workspaces/{w}/privateEndpointConnections`) | `PutPrivateEndpointConnection`, `GetPrivateEndpointConnection`, `DeletePrivateEndpointConnection`, `ListPrivateEndpointConnections` |
+| **Private Link Resources** (`workspaces/{w}/privateLinkResources`) | `GetPrivateLinkResource`, `ListPrivateLinkResources` |
+| **VNet Peering** (`workspaces/{w}/virtualNetworkPeerings`) | `CreateOrUpdateVNetPeering`, `GetVNetPeering`, `DeleteVNetPeering`, `ListVNetPeerings` |
+| **Outbound Network Dependencies** (`workspaces/{w}/outboundNetworkDependenciesEndpoints`) | `ListOutboundNetworkDependencies` |
+| **Operations** (`/providers/Microsoft.Databricks/operations`) | `ListOperations` |
+
+*Modeled store-and-echo:* the ARM resources round-trip faithfully over the SDK
+(access connectors and peerings persist and are listed/described; a
+system-assigned access-connector identity gets synthesized principal/tenant
+IDs; a created peering springs to `Connected`/`Succeeded`), but the underlying
+Azure networking side effects are **not** simulated — a private-endpoint
+connection stores its approval state without a real private endpoint on the
+platform side, private-link resources and outbound-dependency endpoints are a
+synthesized (workspace-scoped) catalog rather than a live probe, and a VNet
+peering does not actually peer networks. The provider operations list is a
+static catalog of the RBAC operations the namespace exposes.
+
 ### Instance Pool Operations
 
 | Operation | Signature |
--- a/docs/services.md
+++ b/docs/services.md
@@ -1621,7 +1621,7 @@ static catalog of the RBAC operations the namespace exposes.
 | `SetPermissions` | `(ctx, objectType, objectID, acl) (*ObjectPermissions, error)` |
 | `UpdatePermissions` | `(ctx, objectType, objectID, acl) (*ObjectPermissions, error)` |
 
-**Total: 52 operations**
+**Total: 70 operations**
 
 ---
 
@@ -1906,7 +1906,7 @@ still sees success.
 | Resource Discovery (engine + AWS + Azure + GCP handlers) | 26 |
 | Generative AI — AWS Bedrock (control plane + runtime) | 65 |
 | Generative AI — AWS Bedrock Agent (control plane + runtime) | 32 |
-| Databricks — Azure (control + data plane) | 52 |
+| Databricks — Azure (control + data plane) | 70 |
 | Machine Learning — AWS SageMaker (control plane + runtime) | 121 |
 | Machine Learning — Azure AI (CognitiveServices + MachineLearningServices + data plane) | 92 |
 | Machine Learning — GCP Vertex AI (Go API/driver) | 128 |
--- a/providers/azure/databricks/arm_accessconnectors.go
+++ b/providers/azure/databricks/arm_accessconnectors.go
@@ -15,6 +15,12 @@ import (
 
 const accessConnectorsType = "accessConnectors"
 
+// emulatorTenantID is the single Azure AD directory (tenant) that all
+// system-assigned identities in this emulator belong to. Real Azure has one
+// tenant per directory, so this is a fixed emulator-wide value rather than a
+// per-resource synthesized GUID.
+const emulatorTenantID = "11111111-1111-1111-1111-111111111111"
+
 // CreateOrUpdateAccessConnector creates or updates an access connector,
 // completing provisioning synchronously (store-and-echo).
 func (m *Mock) CreateOrUpdateAccessConnector(
@@ -37,7 +43,7 @@ func (m *Mock) CreateOrUpdateAccessConnector(
 		// immutable in real Azure, so it is left untouched.
 		updated := *existing
 		updated.Tags = copyMap(cfg.Tags)
-		updated.Identity = resolveIdentity(cfg.Identity, cfg.Name)
+		updated.Identity = resolveIdentity(cfg.Identity, cfg.ResourceGroup, cfg.Name)
 		m.accessConnectors.Set(k, &updated)
 
 		return cloneAccessConnector(&updated), nil
@@ -49,7 +55,7 @@ func (m *Mock) CreateOrUpdateAccessConnector(
 		ResourceGroup:     cfg.ResourceGroup,
 		Location:          cfg.Location,
 		Tags:              copyMap(cfg.Tags),
-		Identity:          resolveIdentity(cfg.Identity, cfg.Name),
+		Identity:          resolveIdentity(cfg.Identity, cfg.ResourceGroup, cfg.Name),
 		ProvisioningState: driver.StateSucceeded,
 		CreatedAt:         m.opts.Clock.Now().UTC().Format(time.RFC3339),
 	}
@@ -89,7 +95,7 @@ func (m *Mock) UpdateAccessConnector(
 	}
 
 	if identity != nil {
-		updated.Identity = resolveIdentity(identity, name)
+		updated.Identity = resolveIdentity(identity, resourceGroup, name)
 	}
 
 	m.accessConnectors.Set(k, &updated)
@@ -144,7 +150,7 @@ func sortAccessConnectors(in []driver.AccessConnector) {
 // resolveIdentity normalizes an incoming managed identity: for a system-assigned
 // identity it synthesizes deterministic principal/tenant GUIDs (as Azure does on
 // assignment); a nil or "None" identity resolves to nil.
-func resolveIdentity(in *driver.ManagedIdentity, name string) *driver.ManagedIdentity {
+func resolveIdentity(in *driver.ManagedIdentity, resourceGroup, name string) *driver.ManagedIdentity {
 	if in == nil || in.Type == "" || strings.EqualFold(in.Type, "None") {
 		return nil
 	}
@@ -155,8 +161,12 @@ func resolveIdentity(in *driver.ManagedIdentity, name string) *driver.ManagedIde
 	}
 
 	if strings.Contains(strings.ToLower(in.Type), "systemassigned") {
-		out.PrincipalID = synthGUID("principal/" + name)
-		out.TenantID = synthGUID("tenant/" + name)
+		// PrincipalID is per-resource: keying on (resource group, name) means two
+		// connectors with the same name in different RGs get distinct principals,
+		// while the value stays stable across gets/restarts for the same resource.
+		out.PrincipalID = synthGUID("principal/" + resourceGroup + "/" + name)
+		// TenantID is the emulator's single directory (one tenant per directory).
+		out.TenantID = emulatorTenantID
 	}
 
 	return out
--- a/providers/azure/databricks/arm_accessconnectors.go
+++ b/providers/azure/databricks/arm_accessconnectors.go
@@ -1,0 +1,198 @@
+package databricks
+
+import (
+	"context"
+	"fmt"
+	"hash/fnv"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/stackshy/cloudemu/v2/errors"
+	"github.com/stackshy/cloudemu/v2/internal/idgen"
+	"github.com/stackshy/cloudemu/v2/services/databricks/driver"
+)
+
+const accessConnectorsType = "accessConnectors"
+
+// CreateOrUpdateAccessConnector creates or updates an access connector,
+// completing provisioning synchronously (store-and-echo).
+func (m *Mock) CreateOrUpdateAccessConnector(
+	_ context.Context, cfg driver.AccessConnectorConfig,
+) (*driver.AccessConnector, error) {
+	switch {
+	case cfg.Name == "":
+		return nil, errors.New(errors.InvalidArgument, "access connector name is required")
+	case cfg.ResourceGroup == "":
+		return nil, errors.New(errors.InvalidArgument, "resource group is required")
+	case cfg.Location == "":
+		return nil, errors.New(errors.InvalidArgument, "location is required")
+	}
+
+	k := key(cfg.ResourceGroup, cfg.Name)
+
+	if existing, ok := m.accessConnectors.Get(k); ok {
+		// ARM PUT is create-or-update: apply the mutable fields to a copy and
+		// swap it in, preserving identity fields (ID, created time). Location is
+		// immutable in real Azure, so it is left untouched.
+		updated := *existing
+		updated.Tags = copyMap(cfg.Tags)
+		updated.Identity = resolveIdentity(cfg.Identity, cfg.Name)
+		m.accessConnectors.Set(k, &updated)
+
+		return cloneAccessConnector(&updated), nil
+	}
+
+	ac := &driver.AccessConnector{
+		ID:                idgen.AzureID(m.opts.AccountID, cfg.ResourceGroup, providerNamespace, accessConnectorsType, cfg.Name),
+		Name:              cfg.Name,
+		ResourceGroup:     cfg.ResourceGroup,
+		Location:          cfg.Location,
+		Tags:              copyMap(cfg.Tags),
+		Identity:          resolveIdentity(cfg.Identity, cfg.Name),
+		ProvisioningState: driver.StateSucceeded,
+		CreatedAt:         m.opts.Clock.Now().UTC().Format(time.RFC3339),
+	}
+
+	m.accessConnectors.Set(k, ac)
+
+	return cloneAccessConnector(ac), nil
+}
+
+// GetAccessConnector returns an access connector by resource group and name.
+func (m *Mock) GetAccessConnector(_ context.Context, resourceGroup, name string) (*driver.AccessConnector, error) {
+	ac, ok := m.accessConnectors.Get(key(resourceGroup, name))
+	if !ok {
+		return nil, errors.Newf(errors.NotFound, "access connector %q not found", name)
+	}
+
+	return cloneAccessConnector(ac), nil
+}
+
+// UpdateAccessConnector applies a PATCH (tags and/or identity) to an access
+// connector.
+func (m *Mock) UpdateAccessConnector(
+	_ context.Context, resourceGroup, name string, tags map[string]string, identity *driver.ManagedIdentity,
+) (*driver.AccessConnector, error) {
+	k := key(resourceGroup, name)
+
+	ac, ok := m.accessConnectors.Get(k)
+	if !ok {
+		return nil, errors.Newf(errors.NotFound, "access connector %q not found", name)
+	}
+
+	// Mutate a copy and swap it in so concurrent readers never observe a torn
+	// update. A nil tags/identity leaves that field unchanged (PATCH semantics).
+	updated := *ac
+	if tags != nil {
+		updated.Tags = copyMap(tags)
+	}
+
+	if identity != nil {
+		updated.Identity = resolveIdentity(identity, name)
+	}
+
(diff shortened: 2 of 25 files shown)


thzgajendra (commented):
## Review — remaining Microsoft.Databricks ARM resources (#209)

Deep 8-lens pass (wire-shape fidelity, routing, provider copy-on-write, layering, per-resource ARM semantics, edge cases, tests, docs/boundary/CI), verified against source and the vendored `armdatabricks@v1.1.0` SDK. **This is a clean, faithful PR that closes #209 exactly** (all 6 resource families, every requested op). Verified sound:

- **Wire-shape fidelity — PASS on all 6 resources.** Cross-checked every list envelope and field against the SDK's deserializers: accessConnectors/PEC/PLR/peering/operations use `{value,nextLink}`; the **Outbound list correctly emits a bare JSON array** (the SDK unmarshals into `OutboundEnvironmentEndpointArray`, not a `{value:[]}` envelope) — the known trap is handled and genuinely guarded by `TestSDKOutboundList`. Identity/PEC/peering/groupID property names, casing, and nested blocks all match.
- **Copy-on-write — PASS, no aliasing.** Every reference-type field (tags map, identity + userAssigned slice, groupIDs, address spaces) is deep-copied on both read and write; a caller mutating its input or a returned object can't corrupt the store. The shared-backing-array class that bit a sibling PR is absent here. GUID synthesis (`synthGUID`, FNV) is stable across gets and process restarts.
- **Layering — PASS.** Clean 4-layer separation; all 20 service methods go through the `do()` pipeline; server does only decode→map→call→map→write; wire DTOs (`arm_types.go`) are distinct from driver types with explicit mapping. The subscription-less operations path is a localized, well-documented special-case that short-circuits *before* the shared parser rather than hacking it.
- **Per-resource semantics — faithful.** createOrUpdate idempotency, PATCH tag-replace (correct for Databricks `AccessConnectorUpdate`), identity None↔SystemAssigned transitions, PEC approval states incl. Rejected, peering springs to Connected/Succeeded, PLR groupIds (`databricks_ui_api`/`browser_authentication`) — all correct.
- **Tests — genuine and thorough.** 21 SDK round-trips drive the *real* armdatabricks clients against an httptest server (the true user path, exercising SDK marshalling), plus 9 provider unit tests; every op + listed edge case has a real assertion; provider coverage 91.1%; `go test` + `-race` green; `golangci-lint` (v2.4.0) 0 issues; build/vet clean.
- **Docs boundary-honesty — exemplary.** services.md §21 discloses every accept-and-echo boundary (PEC stores approval without a real endpoint; PLR/outbound are synthesized catalogs; peering doesn't peer; identity GUIDs synthesized), each backed by code. Nothing overstated, no undisclosed echoed surface.

### Findings (neither blocks; two worth addressing)

**1. [Medium] Routing matches provider/type case-sensitively, diverging from ARM and from this file's own convention.** `handler.go:69-70` (and the `switch` at :94-95, and sub-resource dispatch at :141) compare `rp.Provider == providerName` / `rp.ResourceType == accessConnectorsType` with `==`, but ARM treats the provider namespace and resource-type segments **case-insensitively** — and the same file's `isOperationsPath` (:76) and the shared `parseResourceGroup` already use `strings.EqualFold`. So `.../providers/microsoft.databricks/accessconnectors` 404s. The armdatabricks SDK emits canonical casing, so the happy path works, but it's a latent 404 and internally inconsistent. Use `EqualFold` for Provider/ResourceType/SubResource to match the operations-path convention.

**2. [Medium] Delete-of-missing returns 404 instead of ARM's idempotent 204.** `DeleteAccessConnector`/`DeletePrivateEndpointConnection`/`DeleteVNetPeering` return `errors.NotFound`→404 when the resource is absent. Real ARM DELETE is idempotent (204 on missing), and this repo already implements that in `server/azure/resourcegroups/handler.go:159-162` ("deleting a group that is already gone is the caller's desired end state" → `StatusNoContent`). An SDK `BeginDelete`→`PollUntilDone` on an already-deleted resource (teardown retry, delete-then-delete) will fail with a 404 rather than succeed. No test covers delete-of-missing at the HTTP layer. Recommend treating a NotFound-on-delete as 204 to match ARM + the repo precedent (or document the divergence deliberately).

### Low / cosmetic
- **PUT createOrUpdate always returns 200, never 201** on first create (`arm_accessconnectors_ops.go:59`) — ARM returns 201 on create; the SDK poller tolerates both.
- **Delete success returns 200, not 204** — matches the pre-existing workspace delete, but the rest of the azure handlers use 204.
- **`synthGUID` is keyed on name only** — two access connectors with the same name in different resource groups get identical principalId/tenantId, and `tenantId` varies per-name rather than one-tenant-per-emulator. Cosmetic for store-and-echo, but a per-(RG,name) key + a single fixed tenant would be more realistic.
- **List determinism uses a manual `sort.SliceStable(byID)`** rather than the repo's `memstore.SortedValues()` helper — deterministic and actually *more* correct than the #164 workspace baseline (bare `All()`), but now three determinism strategies coexist in the package; worth unifying.
- **`getPEC`/`deletePEC`/`getPeering`/`deletePeering` skip `requireWorkspace`** — still 404s (the composite sub-key can't exist), only the error *message* differs from the put/list paths. Cosmetic.
- **Coverage gap:** the PATCH-identity=`None` transition on an existing SystemAssigned connector is unit-tested but not SDK-round-tripped (behavior is correct).

**Bottom line:** correct, well-tested, faithfully scoped, and honestly documented — one of the cleaner SDK-compat PRs. I'd fix **#1** (case-insensitive routing) and **#2** (idempotent delete) before merge; the rest are minor/cosmetic. Deferring the merge decision to you.

thzgajendra (commented):
## Follow-up — docs op-count (correcting my own review)

On a closer conformance check against cloudemu's conventions, one **docs** item I should have flagged in my first pass:

**`docs/services.md` §21 "Total: 52 operations" (line 1624) is stale.** This PR adds 18 driver operations (Access Connectors 6 + PEC 4 + Private Link 2 + VNet Peering 4 + Outbound 1 + Operations 1), and the "Extended ARM resources" table is correctly added — but the section's aggregate total line was not bumped and still reads **52**; it should be **70**. The diff touches no `Total:` line. Minor, but the op-count is a documented per-section invariant in services.md, so worth correcting.

For the record, **architecture and tests are conformant** to cloudemu conventions:
- **Architecture** — the "all 3 providers" rule correctly doesn't apply (Databricks is Azure-only; no `providers/aws|gcp/databricks`), the `ARMResources` interface is composed into the `Databricks` driver and implemented only by the Azure `Mock`, consistent with the #164 workspace impl. `memstore.Store`, `idgen.AzureID`, `sync.RWMutex`, the cloudemu `errors` package, copy-on-write, and the `do()` pipeline are all used per convention; no `time.Now()` so the `FakeClock` rule isn't triggered; `golangci-lint` is clean so the line/complexity/function-length limits hold.
- **Tests** — follow the de-facto cloudemu SDK-compat pattern (`server/azure/<svc>/sdk_roundtrip_test.go` driving the real SDK), matching servicebus/keyvault/cache/mysqlflex/azuresearch/etc. CONTRIBUTING.md's "add integration tests to `cloudemu_test.go`" line predates that pattern — the reference workspace PR (#164) didn't use it either, so #306 is consistent with actual practice.

Net: docs need the one-line total bump (52 → 70); architecture and tests are conformant.

Satyam-Trivedi-ZS:
Thanks @thzgajendra — thorough pass, and the idempotent-delete precedent pointer was spot on. Addressed both Mediums, two of the Lows, and the docs count in `a530e8c`:

**1 — Case-insensitive routing `[Fixed]`**
`Matches`, the `ServeHTTP` resource-type dispatch, and `serveWorkspaceChild` now use `strings.EqualFold` for the provider namespace, resource type, and sub-resource — matching the `isOperationsPath` / `parseResourceGroup` convention. So `.../providers/microsoft.databricks/accessconnectors/...` routes instead of 404-ing; canonical casing behaves exactly as before. Proven by a new raw-HTTP test (`arm_routing_test.go`) that bypasses the SDK's canonical casing and asserts lowercased → 200, canonical → 200, unknown sub-resource → 404.

**2 — Idempotent DELETE `[Fixed]`**
`DeleteAccessConnector`, `DeletePrivateEndpointConnection`, and `DeleteVNetPeering` now return **204** on success and **204** on a driver `NotFound` (idempotent — the resource being gone is the caller's desired end state), following the `resourcegroups/handler.go` precedent; only genuine errors go to `WriteCErr`. Delete-then-delete / teardown retries now succeed. New SDK round-trip tests (`arm_delete_idempotent_test.go`) cover delete-of-missing for all three. Pre-existing workspace delete (#164) left untouched as out-of-scope.

**Docs op-count `[Fixed]`** (from the follow-up) — §21 `Total:` bumped 52 → 70, and I also caught the paired invariant in the coverage-summary table (line ~1909, same `Databricks — Azure | 52` → `70`), so both stay consistent.

**Low — synthGUID keying `[Fixed]`** — system-assigned `principalId` is now keyed on `(resourceGroup, name)` (same-named connectors in different RGs get distinct principals, still deterministic per (rg,name)); `tenantId` is a single fixed emulator-wide directory GUID rather than per-name. Realism win at no cost.

**Low — coverage gap `[Fixed]`** — the PATCH-identity=`None` transition on a SystemAssigned connector is now SDK-round-tripped (`TestSDKAccessConnectorPatchIdentityToNone`), not just unit-tested.

Deliberately **not** changed (happy to if you'd prefer):
- **PUT returns 200, not 201 on first create** — the SDK poller tolerates both, and 200-inline matches the existing workspace/#164 pattern in this handler; changing it means plumbing a create-vs-update signal up from the provider for a cosmetic status difference. Left consistent with the surrounding handler.
- **List determinism via manual `sort.SliceStable(byID)`** — as you noted it's actually more correct than the `#164` bare-`All()` baseline; `memstore.SortedValues()` sorts by store key (rg/name or rg/ws/type/name) rather than by ARM ID, so switching would change ordering for a marginal consistency gain. Left as-is; unifying all three strategies is better as its own cleanup.
- **`getPEC`/`getPeering` skipping `requireWorkspace`** — cosmetic (the composite sub-key can't exist without the workspace, so it still 404s; only the error message differs), and with delete now idempotent the delete paths don't need it.

`go build ./...`, `go vet`, `gofmt`, full `go test ./...`, `go test -race` (databricks), and `golangci-lint` (0 issues) all green. Thanks again.

thzgajendra (commented):
## Re-review of `a530e8c5` — verified against source, tests green

Both Mediums plus the docs count and two Lows are addressed correctly; the three declined Lows have sound rationale. Verified each:

- **#1 Case-insensitive routing — ✅ fixed, complete.** `strings.EqualFold` now covers all three sites: `Matches` (provider + resource type), the `ServeHTTP` type dispatch, and `serveWorkspaceChild` (all four sub-resources), matching the file's own `isOperationsPath`/`parseResourceGroup` convention. `TestARMRoutingCaseInsensitive` (raw HTTP, bypassing the SDK's canonical casing) genuinely proves lowercased→200, canonical→200, unknown provider→not-routed, unknown sub-resource→404. Passes.
- **#2 Idempotent DELETE — ✅ fixed, and cleaner than I asked.** All three deletes now `if err != nil && !cerrors.IsNotFound(err)` → `WriteCErr`, else `204` — so success→204 **and** NotFound→204, per the `resourcegroups` precedent. This also resolves my Low about delete-success returning 200 (now 204, consistent with the rest of the azure handlers). `TestSDK{AccessConnector,PrivateEndpoint,VNetPeering}DeleteIdempotent` cover delete-of-missing for all three. Pass.
- **Docs op-count — ✅ fixed in both places.** §21 `Total: 52 → 70` **and** the coverage-summary table `Databricks — Azure | 52 → 70` (line ~1907) — nice catch on the paired invariant, which I hadn't flagged.
- **Low: synthGUID keying — ✅** now keyed on `(resourceGroup, name)` with a single emulator-wide `tenantId`. Realism win.
- **Low: PATCH-identity=None coverage — ✅** now SDK-round-tripped (`TestSDKAccessConnectorPatchIdentityToNone`), not just unit-tested. Pass.

The three deliberately-declined Lows are all reasonable: **PUT 200-not-201** (SDK poller tolerates both, matches the #164 workspace pattern, and plumbing a create-vs-update signal for a cosmetic status isn't worth it), **manual `sort.SliceStable(byID)`** (`memstore.SortedValues()` sorts by store key not ARM ID, so switching changes ordering — better as a separate cleanup unifying all three strategies), and **`getPEC`/`getPeering` skipping `requireWorkspace`** (cosmetic message-only difference; still 404s). Agreed on all three.

`go test` (databricks) + `-race` green; the new routing/delete/patch tests pass. Correct, well-tested, conventions-conformant, and honestly documented — LGTM from my side (deferring the merge decision to you).

NitinKumar004 (approved):
## Deep review — Azure Databricks ARM resources (closes #209) — approve

Reviewed at `190caea` (isolated worktree, honest diff vs merge-base, full gate matrix, three adversarial lenses verified against the vendored `armdatabricks@v1.1.0` SDK). **Strong, complete, well-tested PR — approving.** Every #209 resource is delivered across all four layers and round-trips over the real SDK.

### Gates — green
build / vet / gofmt / `go mod tidy` clean · golangci-lint **0 issues** · databricks tests (all 4 layers) pass · `-race` (server + provider) clean · whole-repo `go test ./...` **0 failures** (no cascade). Databricks is legitimately Azure-only, so the mirror rule doesn't apply; docs disclosure (accept-and-echo limits) is present.

### #209 — fully delivered + SDK-verified
All six resource types and their operations (access connectors, private endpoint connections, private link resources, VNet peering, outbound network dependencies, operations) are implemented and exercised by 21 real-SDK round-trip tests + 9 provider unit tests that assert field values through the genuine `armdatabricks` clients.

### Adversarial verification — 3 lenses, all clean
- **Wire fidelity**: no defects — all six shapes (list envelopes: five `{"value":[...]}` + Outbound bare-array; field tags/nesting; identity + PEC/provisioning enum strings) match the SDK's `models_serde.go`, and the tests assert values through the client.
- **Dispatch / cascade / duplicate**: clean — the subscription-less operations path is exact-matched, `accessConnectors` is gated by the unique `Microsoft.Databricks` provider (no collision in either direction), sub-resource routing is correct (`ParsePath` populates `SubResource`), parent-workspace validation is enforced in the driver (`requireWorkspace`), and case-insensitivity is symmetric between `Matches` and `ServeHTTP`. The PR's new deletes are correctly idempotent (NotFound → 204).
- **Provider correctness**: no High/Medium — deterministic GUID synthesis (`fnv`, no `rand`/`time`), fixed-order catalogs + `SliceStable`, deep-clone reads (no pointer escape), correct PATCH/`None` identity semantics, RG-keyed referential integrity, correct `Microsoft.Databricks` ID casing.

### Before/after merge — worth addressing
- **Unrelated change bundled in**: `server/aws/bedrock/streaming.go` (+8) is a **correct** fix (drain the request body before streaming to avoid a "use of closed network connection" race), but it's unrelated to this Azure Databricks PR and isn't mentioned in the description. Per the repo convention ("only stage files that are part of the change"), consider splitting it into its own PR or at least calling it out — it also entangles the "rollback = revert" story. Not blocking, but it shouldn't ride along silently.

### Low (optional follow-ups; several pre-existing)
- `deleteWorkspace` isn't delete-idempotent (NotFound → 404) while this PR's new sibling deletes correctly return 204 — but `deleteWorkspace` is pre-existing (#164), not in this diff; aligning it would be a tidy follow-up.
- `ListWorkspaces`/`ListWorkspacesByResourceGroup` iterate a map unsorted (nondeterministic order), inconsistent with the new sorted lists (pre-existing).
- Access-connector `createOrUpdate` silently ignores a changed `Location` on an existing resource (real ARM rejects it).
- `Get`-then-`Set` is a non-atomic read-modify-write (last-writer-wins) — benign for an in-memory mock.
- Sub-resource DELETE/GET on a missing workspace skips `requireWorkspace` (status codes defensible; only the error message differs).
- `armOperation.IsDataAction` is emitted but unread by this SDK version — superfluous, harmless.

Net: the Databricks implementation is complete, wire-faithful, deterministic, and well-tested, with clean dispatch and no High/Medium defects. LGTM — please split or disclose the bedrock change, and the Lows are optional cleanup.

Satyam-Trivedi-ZS:
Thanks @NitinKumar004 — appreciate the adversarial three-lens pass and the approval. Addressed the actionable items in `4ddaedd`; rationale for each below.

**Bundled bedrock change — disclosed (and happy to split).**
You're right it shouldn't ride along silently. I've updated the PR description with an explicit "one bundled unrelated fix" note explaining what `server/aws/bedrock/streaming.go` does and why it's here: the required `Test` job was intermittently red under `-race` on `TestSDKConverseStream` (`use of closed network connection`), a pre-existing #298 flake, and I have no admin rights to re-run the job. I bundled the one-line drain-fix only to get CI reliably green. If you'd prefer a clean single-purpose diff, I'll split it into its own PR against `development` and rebase this one to drop it — just say the word.

**Location change on `createOrUpdate` — `[Fixed]`.**
`CreateOrUpdateAccessConnector` now rejects a changed `Location` on an existing connector with `InvalidArgument` (400) instead of silently ignoring it, matching real ARM. A same-location PUT stays an idempotent tags/identity update. Covered by `TestAccessConnectorRejectsLocationChange` (provider) and `TestSDKAccessConnectorRejectsLocationChange` (SDK round-trip).

**The remaining Lows — deliberately deferred, with reasons:**
- **`deleteWorkspace` not delete-idempotent** and **`ListWorkspaces*` unsorted** — both are pre-existing from #164 and *not in this diff*. Aligning them would mean editing workspace code unrelated to #209 — exactly the kind of scope-creep your bedrock note (rightly) flags. Better as a small standalone follow-up; I can open one if you'd like.
- **`Get`-then-`Set` non-atomic RMW** — agreed it's benign for an in-memory mock (single-writer SDK flows don't hit it); leaving as-is, consistent with the workspace baseline.
- **Sub-resource DELETE/GET skipping `requireWorkspace`** — cosmetic (the composite sub-key can't exist without the workspace, so it still 404s; only the error *message* differs). Same call @thzgajendra and I landed on earlier.
- **`armOperation.IsDataAction`** — kept intentionally: it's part of the real ARM operation shape (Azure operations carry `isDataAction`), so emitting it is wire-faithful and forward-compatible even though this SDK version doesn't surface it. Harmless as you noted.

Gates re-verified after the fix: `build` · `vet` · `gofmt` · `go test ./...` · `go test -race` (databricks) · `golangci-lint` (0 issues) — all green. Thanks again.

Satyam-Trivedi-ZS:
Follow-up — additional end-to-end verification + the location-change fix are pushed.

Since the approval I've done a full e2e pass of the whole feature and pushed two functional commits:

- **`fix(databricks): reject access-connector location change on update`** — `CreateOrUpdateAccessConnector` now returns `InvalidArgument` (400) on a changed `Location` for an existing connector (same-location PUT stays an idempotent tags/identity update), addressing @NitinKumar004's L3. Tests: `TestAccessConnectorRejectsLocationChange` + `TestSDKAccessConnectorRejectsLocationChange`.
- **`test(databricks): add full end-to-end ARM journey test`** — one realistic user flow against a single emulator that drives **every** `Microsoft.Databricks` ARM resource type in sequence through the real `armdatabricks` SDK clients: workspace parent → access connectors (incl. location-change rejection + idempotent delete) → private endpoint connections → private link resources → VNet peerings → outbound dependencies → the subscription-less operations list. Verifies data flows across the whole surface, not just per-resource in isolation.

The ARM surface now has **27 SDK round-trip tests + 10 provider unit tests** (+ the raw-HTTP case-insensitive routing test). Local gates all green: `gofmt` · `go build ./...` · `go vet ./...` · full `go test ./...` · `go test -race` (all databricks layers) · `golangci-lint` (0 issues) · `go mod tidy` (unchanged). The bundled bedrock drain-fix also holds 15× under `-race`.

Note: GitHub briefly didn't sync the PR head past `190caea` after those pushes (no CI ran for them despite the fork ref being updated) — pushed a small nudge commit to force re-processing, so CI is re-running against the full latest state.

PR #308, opened 2026-07-31
Add Azure Managed Cassandra Full-Parity Support

thzgajendra:
## Objective

Add full-parity support for **Azure Managed Instance for Apache Cassandra** (`Microsoft.DocumentDB/cassandraClusters`) — the next missing managed database-server surface after AWS Keyspaces (#307). Covers clusters, datacenters, the deallocate/start lifecycle, status, and invoke-command, so a real `armcosmos` `CassandraClusters`/`CassandraDataCenters` client works unchanged against a custom endpoint.

## What we found

- Managed Cassandra is a **control-plane-only** cluster service under Cosmos DB (CQL data plane is out of scope), and doesn't map to the relational/database drivers → dedicated `services/managedcassandra/driver`.
- Protocol is **ARM REST/JSON** (path-routed via `server/wire/azurearm`, hand-rolled JSON — the repo's Azure convention). Its `Begin*` operations are long-running (LRO), each with a different final-state shape the SDK poller expects.

## How we fixed it

- **Driver** (`services/managedcassandra/driver`): 15 operations — clusters (create/get/list-by-RG/list-by-subscription/update/delete/deallocate/start/invokeCommand/status) + datacenters (create/get/list/update/delete).
- **Provider** (`providers/azure/managedcassandra`): in-memory `Mock` — clusters (version, delegated subnet, auth method, repair, backups, seed/gossip/client certs, deallocate/start) and datacenters (node count, disk, SKU, availability zone, seed nodes). **Parent linkage** (the #305/#307 lesson): datacenter create validates the cluster; cluster delete cascade-deletes datacenters; deallocate/start propagate to datacenters. Node count is bounded; clone-on-read on every path.
- **Server** (`server/azure/managedcassandra`): ARM handler; LROs complete on the first response so the SDK pollers terminate — create/patch return the resource, delete → 204, deallocate/start → 202 + `Azure-AsyncOperation` (→ operationStatuses), invokeCommand → 202 + `Location` returning the command output.
- **Wiring**: registered in the Azure provider/server bundles; dedicated `managedcassandra:*` cost keys (datacenter nodes are the billable unit); `docs/services.md` section + counts (Grand Total 1328 → 1343).

## Alternatives not taken

- **Reusing the cosmos/relational driver**: rejected — managed Cassandra clusters/datacenters aren't relational or Cosmos-account resources.
- **Using armcosmos types on the wire**: rejected — the repo hand-rolls ARM JSON for full control (armcosmos is a test-only dependency for the round-trip tests).
- Timestamps aren't emitted (ARM uses RFC3339, so the AWS epoch hazard doesn't apply; the cluster/datacenter shapes carry no timestamp fields anyway).

## Docs / Test / Playground

- `docs/services.md`: Managed Cassandra section (cluster + datacenter tables) + recomputed totals.
- Tests: provider unit tests (cluster lifecycle, datacenter parent linkage + cascade delete, update/list, deallocate/start/status, invoke-command, node-count bound, clone-on-read aliasing) + server SDK round-trip tests via the real `armcosmos` clients covering all LRO shapes (create/get/list/update/delete/deallocate/start/status/invokeCommand, datacenter CRUD) and a not-found error. Cost rate-catalog test.

## Test plan

- [x] CodeQL (security-extended) — 0 findings in-package (bounded a datacenter node-count allocation)
- [x] `go mod tidy` clean
- [x] Coverage: provider 87.0%, server 73.5%

## Risk & Rollback

Additive: new driver + provider/server packages plus opt-in bundle registration (nil driver ⇒ handler not registered). No existing behavior changes. Rollback = revert the commit.

## Conclusion

Managed Cassandra reaches full parity with the established driver→provider→server layering. Remaining in the missing-database-server initiative (separate PRs): AWS Timestream, GCP Spanner & Bigtable.


providers/azure/managedcassandra/datacenters.go
```diff
@@ -0,0 +1,208 @@
...
+		return def
+	}
+
+	return v
+}
+
+// CreateOrUpdateDataCenter creates or replaces a datacenter under a cluster.
+//
+//nolint:gocritic // cfg matches the driver signature.
+func (m *Mock) CreateOrUpdateDataCenter(_ context.Context, cfg mcdriver.CreateDataCenterConfig) (*mcdriver.DataCenter, error) {
```
NitinKumar004, line 30: **[Medium] `CreateOrUpdateDataCenter` ignores the cluster's deallocated state.** The new `DataCenter` is built with `Deallocated` defaulting to `false`, and the path never inherits the parent cluster's state nor preserves an existing DC's state — unlike `CreateOrUpdateCluster`, which preserves `existing.Deallocated` (`managedcassandra.go:137-139`), and `setClusterDeallocated`, which propagates the flag to all DCs. → *`DeallocateCluster` (cluster + DCs → STOPPED) then add a new datacenter, or CreateOrUpdate-replace an existing one to change disk/YAML → the new/replaced DC comes back `Deallocated=false`, so `ClusterStatus` reports its nodes `NORMAL` while the cluster and sibling DCs are STOPPED — an internally inconsistent mixed state.* Inherit the cluster's `Deallocated` (and, on replace, the existing DC's) here.

NitinKumar004 (commented):
## Deep end-to-end lifecycle review — Azure Managed Cassandra

Reviewed the whole PR in an isolated worktree (gates, static/coverage, provider lifecycle, server/ARM-LRO wire fidelity, cross-cutting wiring), with adversarial verification. **The whole lifecycle is supported and working end-to-end — the cleanest of the managed-DB PRs — with one Medium consistency bug and a Low tail.**

**Gates: green** — `build ✓ vet ✓ test ✓ -race ✓ gofmt ✓ go mod tidy ✓`; lint clean. Coverage 86.6% provider / 73.5% server (disclosed).

**Verified solid (the hard parts):**
- **All 5 LRO shapes are correct — no poller hangs** (traced against `azcore`/`armcosmos` v3.4.0 internals): BeginCreateUpdate → body poller (`provisioningState=Succeeded`); BeginDelete → 204 terminal; **Deallocate/Start → 202 + `Azure-AsyncOperation` → `operationStatuses` returns `Succeeded`**; **InvokeCommand → 202 + `Location` → `commandOutput`**. `PollUntilDone` completes for every one.
- **ARM JSON property fidelity is exact** — every hand-rolled key matches the SDK's case-sensitive labels (incl. `delegatedManagementSubnetId`, `diskSku`, `backupStorageCustomerKeyUri`, `base64EncodedCassandraYamlFragment`), and `initialCassandraAdminPassword` is correctly omitted from GETs. No silent field drops.
- **The #305/#307 lessons are genuinely applied**: DC create validates the parent cluster, cluster delete cascades datacenters, deallocate/start propagate to datacenters, seeds refresh — all working and tested. **Clone-on-write is clean** (no repeat of the #307 write-aliasing bug). Node count is bounded (a real allocation guard). ARM routing is disjoint from Cosmos/other providers, the `fromprovider.go` change is benign (gofmt realignment only), and the docs count math (+15) is exact.

### Medium
- **DataCenter create/replace ignores the cluster's deallocated state.** `CreateOrUpdateDataCenter` never sets `Deallocated` (defaults false), while `setClusterDeallocated` propagates it to all DCs and `CreateOrUpdateCluster` preserves `existing.Deallocated`. → *`DeallocateCluster` then add or replace a datacenter → the new DC comes back `Deallocated=false` while the cluster and its sibling DCs are STOPPED; `ClusterStatus` then reports an internally inconsistent mixed NORMAL/STOPPED state.* Inherit the cluster's (and any existing DC's) deallocated state on create/update. *(inline)*

### Low
- `nodeCount==0` is rewritten to `3` on create (so the create-side 0-validation is effectively dead), but a PATCH can set 0 — inconsistent create-vs-update semantics.
- `CreateOrUpdateCluster` (PUT) rebuilds from `cfg`, dropping `GossipCertificates`/`PrometheusEndpoint` on re-PUT (latent — those fields are never populated today).
- No state guards on Deallocate/Start (idempotent no-op — reasonable for a mock); `ResourceGroup` and the DC's `ClusterName` aren't name-validated.
- Cost keys (`CreateDataCenter`/`CreateCluster`) diverge from the driver method names (`CreateOrUpdateDataCenter`/`CreateOrUpdateCluster`) → a `Record` with the real method name misses the rate (cosmetic; matches nothing in shipped code).
- Server tests: error assertions are `err != nil` only (not the typed `*azcore.ResponseError`); `BeginStart` (shares the Deallocate async path) and the malformed-body / DC-on-missing-cluster 400 paths are untested (correct in code).
- Coverage below the 90% pillar (disclosed).

**Verdict: comment** — a strong PR; the single Medium (deallocated-state inconsistency) is the one real should-fix, everything else is Low polish/test-breadth.

providers/azure/managedcassandra/datacenters.go
thzgajendra: Fixed in 92ce3ee. `CreateOrUpdateDataCenter` now reads the parent cluster and sets the new/replaced datacenter's `Deallocated` from it, so a DC added to (or replaced in) a stopped cluster is STOPPED like its siblings and `ClusterStatus` reports a consistent all-STOPPED/all-NORMAL state. Covered by TestDataCenterInheritsClusterDeallocatedState (add-to-deallocated + start-brings-back).

thzgajendra:
Thanks for the review, @NitinKumar004 — fixed in 92ce3ee.

**Medium** — datacenter now inherits the cluster's deallocated state on create/replace (inline reply; test added).

**Low**
- `CreateOrUpdateCluster` now preserves the service-computed `GossipCertificates`/`PrometheusEndpoint` across a re-PUT (alongside `Deallocated`).
- `managedcassandra` cost keys renamed to the driver method names (`CreateOrUpdateCluster`/`CreateOrUpdateDataCenter`) so a recorder keyed on the real op name hits the rate.
- Server tests strengthened: typed `*azcore.ResponseError` assertions (404 get-missing-cluster, 400 dc-under-missing-cluster) and a `BeginStart` round-trip (the previously-untested half of the async path). Coverage: provider 85.9%, server 74.7%.

**Accepted as-is (with reasoning):**
- **`nodeCount==0` → default on create vs PATCH-0**: `0`-means-default is a deliberate create convention; PATCH-0 is unreachable through the SDK (the server maps `NodeCount` only when `> 0`), so there's no real inconsistency on the wire.
- **No state guards on Deallocate/Start**: kept idempotent (no-op if already in the target state) — matches how the mock treats other lifecycle transitions.
- **ResourceGroup / DC `ClusterName` name validation**: the DC path already fails closed on a missing parent cluster; RG isn't a resource the mock owns.

Full local gate green before push: build / vet / `go test ./...` / `golangci-lint` 0 issues / `go mod tidy` clean / CodeQL security-extended 0 findings in-package.

thzgajendra:
Follow-up (7b73ebc) — closed the two remaining Low items I'd initially deferred:

- **nodeCount create-vs-PATCH inconsistency**: `UpdateDataCenter` now treats a `0` node count as "unchanged", consistent with create (where `0` means "use the default"). No more divergent handling.
- **malformed-body path**: added a server test asserting a malformed request body → `400`.

Now only two items remain intentionally unchanged (both reasonable for a mock, and you flagged the first as such): idempotent Deallocate/Start (no state guards), and `ResourceGroup` name validation (the datacenter path already fails closed on a missing parent cluster; RG isn't a resource the mock owns).

NitinKumar004 (approved):

Verified `92ce3ee` against the earlier review; the Medium is resolved and two Lows were cleaned up on top, with new tests. Gates green (`build ✓ vet ✓ test ✓ -race ✓ gofmt ✓ go mod tidy ✓`; lint clean on the Managed Cassandra files).

- **Datacenter create/replace now inherits the cluster's deallocated state** — `CreateOrUpdateDataCenter` sets `Deallocated: cluster.Deallocated`, so a datacenter added to (or replaced in) a deallocated cluster is reported STOPPED rather than a spurious NORMAL, and `ClusterStatus` stays internally consistent.
- **Cost keys renamed** to match the driver methods (`CreateOrUpdateDataCenter`/`CreateOrUpdateCluster`) — a `Record` with the real method name now hits the rate.
- **Re-PUT no longer drops service-computed fields** — `CreateOrUpdateCluster` preserves `GossipCertificates`/`PrometheusEndpoint` across an update.

Plus +99 lines of new provider + SDK round-trip tests (including the deallocated-datacenter scenario).

The full lifecycle (clusters → datacenters → deallocate/start with datacenter propagation → status → invoke-command) is implemented, wired (cost, ARM routing disjoint from Cosmos/other providers), and round-trips through the real `armcosmos` client — all five LRO poller shapes verified correct (no hangs), ARM JSON property fidelity exact, and parent-linkage/cascade/clone-on-write genuinely applied. Non-blocking follow-ups: a few minor validation nits (nodeCount-0 create-vs-update, explicit deallocate/start state guards, RG/cluster-name validation) and server coverage ~75% (below the 90% pillar, consistent with the codebase). LGTM.

PR #309, opened 2026-08-01
Add GCP Bigtable Admin Full-Parity Support

thzgajendra:
## Objective

Add full-parity support for **Google Cloud Bigtable Admin** (`bigtableadmin.googleapis.com/v2`) — the next missing managed database-server surface after Azure Managed Cassandra (#308). Covers instances, clusters, tables (with column families), app profiles, backups, per-resource IAM, and long-running operations, so a real `google.golang.org/api/bigtableadmin/v2` client works unchanged against a custom endpoint.

## What we found

- Bigtable is a **control-plane-only** wide-column store here (the data plane is out of scope); it doesn't map to the relational/NoSQL drivers → dedicated `services/bigtable/driver`.
- Protocol is **GCP REST/JSON** under `/v2/{name}` with `:verb` custom methods and Google **LRO `Operation`** objects (mirrors cloudsql/gke).

## How we fixed it

- **Driver** (`services/bigtable/driver`): 38 operations — instances (6), clusters (6, incl. GetMemoryLayer), tables (11: CRUD + undelete, modifyColumnFamilies, dropRowRange, consistency token/check, restore, patch), app profiles (5), backups (6, incl. copy), operations (get), and per-resource IAM (get/set/test on instances, tables, backups).
- **Provider** (`providers/gcp/bigtable`): in-memory `Mock` keyed by full GCP resource names — instance→cluster/table/app-profile **parent linkage** with cascade delete, initial-cluster creation, recursive column-family **GC rules** (maxNumVersions/maxAge/union/intersection), soft-delete + undelete tables, backup create/copy and restore-into-new-table, an IAM policy store, bounded serve-node counts, clone-on-read everywhere.
- **Server** (`server/gcp/bigtable`): REST handler with a parity/verb router. It uses the **`bigtableadmin/v2` SDK types as the wire format** (exact fidelity, no hand-rolled structs); LRO RPCs return `Operation{done:true}` with the resource inline and `operations.get` returns a done Operation, so SDK LRO waits terminate.
- **Wiring**: registered in the GCP provider/server bundles; dedicated `bigtable:*` cost keys (cluster nodes are billable); `docs/services.md` (master table + section + counts → Grand Total 1381), `README.md`, and `docs/sdk-server.md` updated.

## Alternatives not taken

- **Reusing the database/relational driver**: rejected — Bigtable's instance/cluster/table/backup model is distinct.
- **Hand-rolling wire JSON** (as the AWS/Azure handlers do): here I used the SDK's own types because the surface is large and the SDK types guarantee round-trip fidelity; `bigtableadmin/v2` is already a dependency for the tests.
- Timestamps use RFC3339 (GCP convention) — no epoch hazard.

## Docs / Test / Playground

- `docs/services.md`: Bigtable section (per-resource tables) + master-table row + summary + recomputed totals.
- Tests: provider unit tests (instance/cluster lifecycle + parent validation, column families + consistency, deletion-protection + undelete, backups/restore + cascade, app profiles + IAM, clone-on-read aliasing, plus get/update/delete/copy/operation coverage) and server SDK round-trip tests via the real `bigtableadmin/v2` client across all resources, LRO shapes, IAM verbs, and a not-found. Cost rate-catalog test.

## Test plan

- [x] CodeQL (security-extended) — 0 findings in-package
- [x] `go mod tidy` clean
- [x] Coverage: provider 81.0%, server 67.9%

## Risk & Rollback

Additive: new driver + provider/server packages plus opt-in bundle registration (nil driver ⇒ handler not registered). No existing behavior changes. Rollback = revert the commit.

## Conclusion

Bigtable reaches full parity with the established driver→provider→server layering. Remaining in the missing-database-server initiative (separate PRs): AWS Timestream, GCP Spanner.


providers/gcp/bigtable/clusters.go
```diff
@@ -0,0 +1,170 @@
...
+		serve = defaultServeNodes
+	}
+
+	m.clusters.Set(cfg.Name, btdriver.Cluster{
+		Name:               cfg.Name,
+		Location:           cfg.Location,
+		ServeNodes:         serve,
+		DefaultStorageType: orDefault(cfg.DefaultStorageType, defaultStorageType),
+		State:              btdriver.StateReady,
+		Autoscaling:        cfg.Autoscaling,
```
NitinKumar004: **[Medium] Cluster `Autoscaling` stored by the caller's pointer (write-aliasing).** `Autoscaling: cfg.Autoscaling` stores the caller's `*Autoscaling` directly (same at `UpdateCluster`, `:126` `c.Autoscaling = autoscaling`). `memstore.Set` copies the `Cluster` value shallowly, so the pointer keeps aliasing the caller's struct. `cloneCluster` deep-copies it on *read*, which masks it, but the *stored* value is corruptible — and this is the lone uncloned write path (Labels/app-profile/IAM all clone). → *`CreateCluster({Autoscaling: as})` then `as.MaxServeNodes = 9999` silently changes the stored cluster.* Clone on write: `a := *cfg.Autoscaling; c.Autoscaling = &a`.

providers/gcp/bigtable/clusters.go
```diff
@@ -0,0 +1,170 @@
...
+	m.clusters.Set(name, c)
+
+	op := m.newOp("update-cluster", name)
+	out := cloneCluster(&c)
+
+	return &out, op, nil
+}
+
+// DeleteCluster removes a cluster and its backups.
+func (m *Mock) DeleteCluster(_ context.Context, name string) error {
```
NitinKumar004, line 158: **[Medium] The last cluster of an instance can be deleted.** `DeleteCluster` checks only `m.clusters.Has(name)` — no surviving-cluster count. `CreateInstance` enforces ≥1 cluster, but nothing preserves the invariant afterward. → *Deleting the only cluster leaves the instance with zero clusters — an invalid state the real Admin API rejects with FailedPrecondition — and its app-profiles/tables are orphaned but still queryable.* Reject deleting the last cluster of an instance.

providers/gcp/bigtable/bigtable.go
```diff
@@ -0,0 +1,360 @@
...
+	m.policies[resource] = stored
+
+	out := clonePolicy(&stored)
+
+	return &out, nil
+}
+
+// TestIamPermissions echoes back the subset of permissions the caller holds. The
+// mock grants all requested permissions.
+func (*Mock) TestIamPermissions(_ context.Context, _ string, permissions []string) ([]string, error) {
```
NitinKumar004: **[Medium] `TestIamPermissions` echoes the input.** It returns `append([]string(nil), permissions...)` and ignores the resource + stored policy (the comment says "grants all requested permissions"). SetIamPolicy *does* store real bindings, so the data to compute a real subset is present and unused. → *`SetIamPolicy` with only `roles/bigtable.viewer`, then `TestIamPermissions([...mutateRows])` reports mutateRows granted — a false positive for any least-privilege/deny test.* Intersect the requested permissions with those the stored policy actually confers.

providers/gcp/bigtable/bigtable.go
```diff
@@ -0,0 +1,360 @@
...
+	i := *in
+	i.Labels = copyLabels(i.Labels)
+
+	return i
+}
+
+// CreateInstance creates an instance and its initial clusters.
+//
+//nolint:gocritic // cfg matches the driver signature.
+func (m *Mock) CreateInstance(_ context.Context, cfg btdriver.CreateInstanceConfig) (*btdriver.Instance, *btdriver.Operation, error) {
```
NitinKumar004, line 99: **[Medium] `CreateInstance` is non-atomic.** The instance is `Set` first, then each initial cluster is created in a loop; on a cluster error (e.g. `validateServeNodes` on `ServeNodes:5000`, or a parent mismatch) the function returns without rolling back the already-stored instance/clusters. → *The create fails `InvalidArgument`, but `GetInstance`/`ListInstances` now return a live cluster-less instance the caller believes was never created; a retry then fails `AlreadyExists`.* Validate all clusters (and roll back / stage the writes) so the create is all-or-nothing.

providers/gcp/bigtable/tables.go
```diff
@@ -0,0 +1,306 @@
...
+
+	if t, ok := m.tables.Get(name); !ok || t.Deleted {
+		return false, cerrors.Newf(cerrors.NotFound, "table %q not found", name)
+	}
+
+	return token != "", nil
+}
+
+// RestoreTable creates a new table from a backup (LRO).
+func (m *Mock) RestoreTable(_ context.Context, parent, tableID, backup string) (*btdriver.Table, *btdriver.Operation, error) {
```
NitinKumar004, line 281: **[Medium] `RestoreTable` loses all column families.** The restored table is built as `Table{Name, Granularity, SourceBackup}` with nil `ColumnFamilies`. Root cause: the `Backup` struct carries no schema, so `CreateBackup` never captures the source table's column families and restore can't rebuild them. → *A restored table has an empty schema; real Bigtable restores the source table's column families.* Snapshot the source table's `ColumnFamilies` into `Backup` at create time and copy them into the restored table. (The test doesn't catch this because its source table has an empty schema.)

docs/services.md
```diff
@@ -1142,6 +1143,85 @@ deallocate/start propagate to all datacenters.
...
+### Operations & IAM
+
+`GetOperation` (LRO poll) plus per-resource IAM on instances, tables, and
+backups: `GetIamPolicy`, `SetIamPolicy`, `TestIamPermissions`.
+
+Modeling: instances own clusters/tables/app-profiles (parent linkage, cascade
+delete); backups live under a cluster and restore into a new table; serve-node
+counts are bounded; clone-on-read on every path.
+
+**Total: 44 operations**
```
NitinKumar004: **[Medium] Op-count is wrong (44 vs 38).** The driver interface has **38** methods and the §11d per-resource tables enumerate exactly 38 (Instances 6, Clusters 6, Tables 11, App Profiles 5, Backups 6, Operations+IAM 4) — but this says "Total: 44 operations", the master-table row says 44, and the Grand Total was bumped +44. Correct values: **38 operations; Grand Total 1343 + 38 = 1381** (not 1387). The PR body repeats the 44. Please correct the section total, the master-table row, and the Grand Total.

NitinKumar004 (commented):
## Deep end-to-end lifecycle review — GCP Bigtable Admin (full parity)

Reviewed the whole PR in an isolated worktree (gates, static/coverage, provider lifecycle × 2, server/REST-LRO/convert fidelity, cross-cutting wiring), with adversarial verification. **The API surface is wired end-to-end and works through the real `bigtableadmin/v2` client — the wire layer is excellent — but the provider logic has several real bugs and the docs op-count is wrong.**

**Gates: green** — `build ✓ vet ✓ test ✓ -race ✓ gofmt ✓ go mod tidy ✓`; lint clean. Coverage 81% provider / 67.9% server (disclosed).

**Verified solid (the hard parts):** using the SDK's own types as the wire format paid off — **all LRO shapes are correct** (LRO-vs-sync matched against the SDK's `Do()` return types; no poller hang), **`convert.go` drops no load-bearing field**, and routing is correct. The **recursive GC-rule is safe** (I checked for a stack-overflow crash — it's double-bounded by the 1 MB `MaxBytesReader` and Go's `encoding/json` `maxNestingDepth=10000`, which 400s before the recursion). Cascade delete, the serve-node bound, determinism, and clone-on-read all genuinely work.

### Medium
- **Cluster `Autoscaling` is stored by the caller's pointer (write-aliasing).** `putClusterLocked`/`UpdateCluster` store `cfg.Autoscaling` directly; every other write path here clones (Labels, app-profile clusters, IAM bindings). → *A caller mutating its `Autoscaling` struct after create silently changes the stored cluster.* This is the #307 write-aliasing lesson, missed on this one path. *(inline)*
- **The last cluster of an instance can be deleted.** `DeleteCluster` checks only existence. → *Deleting the only cluster leaves a zero-cluster instance (real Bigtable rejects with FailedPrecondition); app-profiles/tables are orphaned.* *(inline)*
- **`TestIamPermissions` echoes the input instead of computing the granted subset.** It returns all requested permissions and ignores the stored policy (the comment admits "grants all"). → *A least-privilege / deny assertion via `TestIamPermissions` is a false positive.* *(inline)*
- **`CreateInstance` is non-atomic.** The instance is stored, then clusters are created in a loop; a cluster error returns without rolling back. → *An initial cluster with `ServeNodes:5000` fails `InvalidArgument` but leaves a live cluster-less instance; a retry then fails `AlreadyExists`.* *(inline)*
- **`RestoreTable` loses all column families.** It builds a table with nil `ColumnFamilies`, because the `Backup` struct stores no schema, so `CreateBackup` never captures it. → *A restored table has no schema; real Bigtable restores the source table's column families.* Capture the source schema in `Backup` and rebuild it on restore. *(inline)*
- **Docs op-count is wrong.** The driver has **38** methods and §11d enumerates 38, but the section total says **"44 operations"** and the Grand Total was bumped +44 (should be **+38 → 1381, not 1387**); the PR body repeats 44. *(inline)*

### Low
- App-profile routing policy is never validated (exactly-one required); IAM has no etag/version concurrency (constant `"ACAB"` etag); duplicate initial-cluster names silently collapse; `CheckConsistency` accepts any non-empty token; `CreateBackup`'s source-table check rebinds a cross-instance name / allows a soft-deleted table; GC rules get no input validation.
- Server: the over-broad `Matches` (`Contains(p,"/instances")`) could hijack a Cloud Logging path for a log literally named `instances` (a segment-anchored check would be disjoint); `Instance.CreateTime` is dropped in `toWireInstance` (output-only); `operations.get` omits `Response`; `operations.list/cancel/delete` are unrouted; PATCH handlers ignore `updateMask` (mitigated by `orKeep`); the cost comment ("node-hour proxy for a 3-node cluster") is misleading (the rate is per-cluster flat).
- Test gaps: the nested GC-rule tree and cluster-autoscaling aliasing are untested.

**Verdict: comment.** The wire layer is genuinely strong; the six Mediums are the real work — five provider-logic bugs plus the docs miscount.

providers/gcp/bigtable/clusters.go
thzgajendra: Fixed in `25aabcc`. Added `cloneAutoscaling` and clone the caller's `*Autoscaling` on write in both `putClusterLocked` and `UpdateCluster` (`a := *cfg.Autoscaling; …`), so the stored value is no longer aliasable. New test `TestUpdateClusterAutoscalingDoesNotAlias` mutates the caller's struct post-write and asserts the store is unchanged.

providers/gcp/bigtable/clusters.go
thzgajendra: Fixed in `25aabcc`. `DeleteCluster` now counts surviving clusters via a new `clusterCount` helper and rejects deleting the last one with `FailedPrecondition`. Covered by `TestDeleteLastClusterRejected` (reject sole cluster, then allow once a second exists).

providers/gcp/bigtable/bigtable.go
thzgajendra: Fixed in `25aabcc`. Added a `rolePermissions` role→permission table (admin = all) and `roleGrants`; `TestIamPermissions` now returns only the requested perms the stored policy actually confers (empty when no policy). New test `TestTestIamPermissionsIntersectsPolicy` asserts a viewer gets `tables.get` but not `tables.mutateRows`, and no-policy grants nothing.

providers/gcp/bigtable/bigtable.go
thzgajendra: Fixed in `25aabcc`. `CreateInstance` now validates every initial cluster (serve-node bounds + duplicate-name rejection) before any store write, so a bad cluster leaves nothing behind. New test `TestCreateInstanceIsAtomicOnBadCluster` verifies the instance does not exist after a failed create.

providers/gcp/bigtable/tables.go
thzgajendra: Fixed in `25aabcc`. `Backup` now carries `ColumnFamilies`; `CreateBackup` snapshots the (validated, live, same-instance) source table's families, `CopyBackup` carries them, and `RestoreTable` rebuilds them via `cloneColumnFamilies`. New test `TestRestorePreservesColumnFamilies` uses a non-empty source schema and asserts the restored table keeps it (also fixed the source-table same-instance / not-soft-deleted check).

docs/services.md
thzgajendra: Fixed in `25aabcc`. Corrected the section total (44 → 38 operations), the master-table row (44 → 38), and the Grand Total (1387 → 1381). Will update the PR body to match.

thzgajendra:
## Review fixes — `25aabcc`

Thanks for the deep review, @NitinKumar004. All 6 Mediums fixed, plus the actionable Lows and the two test gaps.

### Mediums (all fixed)
| # | Issue | Fix |
|---|-------|-----|
| 1 | Autoscaling write-aliasing | `cloneAutoscaling` on write in `putClusterLocked` + `UpdateCluster` |
| 2 | Last cluster deletable | `DeleteCluster` rejects the sole cluster (`FailedPrecondition`) via `clusterCount` |
| 3 | `TestIamPermissions` echoes input | intersects requested perms with the stored policy (`rolePermissions` table) |
| 4 | `CreateInstance` non-atomic | validates all clusters + rejects duplicate names before any write |
| 5 | `RestoreTable` loses schema | `Backup.ColumnFamilies` snapshotted on create/copy, rebuilt on restore |
| 6 | Docs op-count 44 | corrected to **38**; Grand Total **1387 → 1381**; PR body updated |

### Lows fixed
- App-profile routing now requires exactly one policy.
- `CreateBackup` validates the source table is in the same instance and not soft-deleted.
- Server `Matches` is segment-anchored (won't hijack a `/instancesFoo` path).
- `toWireInstance` emits `CreateTime`.
- Cost comment reworded (flat per-cluster proxy, not literal node-hour).

### Lows deliberately left (mock-by-design / out of scope)
Constant `"ACAB"` etag concurrency, `CheckConsistency` accepting any token, GC-rule input validation, `operations.list/cancel/delete` routing, PATCH `updateMask` (mitigated by `orKeep`), `operations.get` omitting `Response` (the resource is already returned inline by the synchronous LRO). Happy to take any of these if you'd prefer.

### Tests added (to the existing package test file)
Autoscaling aliasing, nested GC-rule deep-clone, IAM deny, last-cluster delete, create-instance atomicity, restore-preserves-column-families, app-profile routing, backup source-table validation.

**Gate:** build ✓ · vet ✓ · test ✓ (incl. server/provider) · `go mod tidy` clean ✓ · lint 0 issues ✓ · CodeQL 0 new alerts ✓

NitinKumar004 (approved):

Verified `25aabcc` against the earlier review; all six Mediums are resolved and several Lows cleaned up on top, with new tests. Gates green (`build ✓ vet ✓ test ✓ -race ✓ gofmt ✓ go mod tidy ✓`; lint clean on the Bigtable files).

**Medium — fixed:**
- **Cluster `Autoscaling` write-aliasing** → `cloneAutoscaling` deep-copies on create and update, so a caller mutating its struct can't corrupt the store.
- **Last-cluster delete** → `DeleteCluster` now returns `FailedPrecondition` for the last cluster of an instance.
- **`TestIamPermissions` echo** → now intersects the requested permissions with the resource's policy bindings (`roleGrants`), returning only the granted subset.
- **Non-atomic `CreateInstance`** → validates every initial cluster up front, so the create is all-or-nothing (no orphaned cluster-less instance).
- **`RestoreTable` losing schema** → `Backup` now carries `ColumnFamilies`, `CreateBackup` captures the source table's families (and validates the source is in the instance), and restore rebuilds them — a full backup→restore schema round-trip.
- **Docs op-count** → corrected to "38 operations" and Grand Total 1381.

**Low — also fixed:** over-broad handler `Matches` is now segment-anchored (won't hijack a Cloud Logging path); `Instance.CreateTime` is emitted (RFC3339); the cost comment is clarified; cross-instance backup source tables are rejected.

Plus +215 lines of new provider tests. The full lifecycle (instances → clusters → tables + column families → app profiles → backups + restore → per-resource IAM → LRO) is implemented, wired, and round-trips through the real `bigtableadmin/v2` client; the LRO shapes and `convert.go` field fidelity were already correct. Non-blocking follow-ups: a few minor validation nits (app-profile routing policy, IAM etag concurrency, GC-rule input validation, `operations.list/cancel/delete`, PATCH `updateMask`) and server coverage ~68% (below the 90% pillar, consistent with the codebase). LGTM.

PR #317, opened 2026-08-03
AWS networking parity: Transit Gateway, VPN, DHCP, prefix lists, egress-only IGW, endpoint services, Client VPN, Network Firewall

thzgajendra:
## Objective
Networking parity — **PR 1 of 3** (AWS → Azure → GCP). Add the AWS-native networking resources cloudemu was missing so real `aws-sdk-go-v2` EC2 and `networkfirewall` clients round-trip.

## What we found
cloudemu already has broad **core** networking parity across all three clouds via the shared portable `Networking` driver (VPC, subnets, security groups + rules, route tables, NACLs, IGW, NAT GW, EIP, peering, flow logs, VPC endpoints). What was missing everywhere were the advanced, provider-specific resources.

## How we fixed it (AWS-only optional capabilities)
Per the agreed design, the new AWS resources are **optional capability interfaces** on the networking driver — discovered by type assertion like the existing `NetworkInterfaces`/`VPCAttributes` — implemented only by `providers/aws/vpc` and served by the EC2 handler. **No Azure/GCP stubs**; each provider PR adds its own specifics.

EC2 query protocol (each behind its own capability interface):
- **Transit Gateway** — gateways, VPC attachments, route tables
- **VPN** — customer gateways, VPN gateways (+attach/detach), VPN connections
- **DHCP option sets** (+associate to VPC)
- **Managed prefix lists** (+entries)
- **Egress-only internet gateways** (IPv6)
- **VPC endpoint services** (PrivateLink provider side)
- **Client VPN** — endpoints (+target-network association)

**Network Firewall** is a distinct AWS service (its own AWS JSON 1.0 API), so it gets a standalone vertical: `services/networkfirewall` driver, `providers/aws/networkfirewall` mock, `server/aws/networkfirewall` handler (target prefix `NetworkFirewall_20201112.`), wired into the AWS provider + server `Drivers`. Covers Firewall, FirewallPolicy, RuleGroup.

## Alternatives not taken
- **Extending the shared portable `Networking` interface** — rejected; it would force meaningless Azure/GCP implementations for AWS-only resources. Optional capability interfaces keep the AWS PR self-contained (the established repo pattern).

## Docs / Test / Playground
- Real-SDK round-trip tests: `server/aws/networking_parity_sdk_test.go` drives the real EC2 client across every new family; `server/aws/networkfirewall/sdk_roundtrip_test.go` drives the real `networkfirewall` client.
- Provider unit tests for the vpc capability families + the networkfirewall mock (validation + error paths); vpc provider coverage **84.7%**.
- `docs/services.md`: AWS-specific networking capability table + Network Firewall master-table row/summary; Grand Total 1415 → **1493** (58 AWS-specific networking + 20 Network Firewall).
- Concurrency tests (`-race`): `TestFirewallConcurrentAccess`, `TestNetworkingConcurrentReadWrite`.

## Review hardening (commit `9e7b428`)
Addresses the deep review on the depth work:
- **Thread safety** — `networkfirewall.Mock` gained a `sync.RWMutex` (was unsynchronized); all vpc capability readers take `RLock` and in-place mutators take `Lock`. The `-race` suite was a false green (sequential), so genuinely concurrent tests were added.
- **Wire fidelity** — `ModifyVpcEndpointServicePermissions` returns `<return>` (SDK-verified); VPN response carries `customerGatewayConfiguration`; Client VPN `AuthenticationOptions` captured/echoed; RuleGroup ARN honours STATELESS; corrected `UnsuccessfulItem` shape.
- **Referential integrity** — FK checks on VPN/TGW create+modify, PrefixList `MaxEntries`, Network Firewall policy references; in-use guards on `DeleteTransitGateway` / `DeleteFirewallPolicy`.
- **Coverage** — added the `Delete*`/`Describe*`/list halves + an SDK error-path assertion per service.

## Test plan
- [x] `go build ./... && go vet ./...`
- [x] `go test ./...` (incl. EC2 + Network Firewall real-SDK round-trips) and `-race` on the vpc + networkfirewall packages
- [x] `go mod tidy` clean; `golangci-lint` 0 new issues in touched files; CodeQL 0 new alerts (only pre-existing baseline)

## Risk & Rollback
Additive — new capability interfaces are optional (type-asserted), and Network Firewall is a new `Drivers.NetworkFirewall` field (nil ⇒ omitted). No existing behavior changes. Rollback = revert the commits on this branch.

## Follow-ups
- Azure networking-specifics PR (2 of 3), then GCP (3 of 3).
- Cost keys for billable resources (Transit Gateway attachment-hours, VPN connection-hours, Network Firewall endpoint-hours) — deferred; the cost catalog is a standalone simulator.

### DIFF
--- a/providers/aws/networkfirewall/networkfirewall.go
+++ b/providers/aws/networkfirewall/networkfirewall.go
@@ -1,0 +1,338 @@
+// Package networkfirewall provides an in-memory mock of AWS Network Firewall.
+package networkfirewall
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/stackshy/cloudemu/v2/config"
+	cerrors "github.com/stackshy/cloudemu/v2/errors"
+	"github.com/stackshy/cloudemu/v2/internal/idgen"
+	"github.com/stackshy/cloudemu/v2/internal/memstore"
+	nfdriver "github.com/stackshy/cloudemu/v2/services/networkfirewall/driver"
+)
+
+var _ nfdriver.NetworkFirewall = (*Mock)(nil)
+
+// Mock is the in-memory AWS Network Firewall implementation. Resources are
+// keyed by name (unique per account/region, as in the real service).
+type Mock struct {
+	firewalls  *memstore.Store[*nfdriver.Firewall]
+	policies   *memstore.Store[*nfdriver.FirewallPolicy]
+	ruleGroups *memstore.Store[*nfdriver.RuleGroup]
+	opts       *config.Options
+}
+
+// New creates a new Network Firewall mock.
+func New(opts *config.Options) *Mock {
+	return &Mock{
+		firewalls:  memstore.New[*nfdriver.Firewall](),
+		policies:   memstore.New[*nfdriver.FirewallPolicy](),
+		ruleGroups: memstore.New[*nfdriver.RuleGroup](),
+		opts:       opts,
+	}
+}
+
+func (m *Mock) arn(kind, name string) string {
+	return idgen.AWSARN("network-firewall", m.opts.Region, m.opts.AccountID, kind+"/"+name)
+}
+
+func copyTags(src map[string]string) map[string]string {
+	if len(src) == 0 {
+		return nil
+	}
+
+	out := make(map[string]string, len(src))
+	for k, v := range src {
+		out[k] = v
+	}
+
+	return out
+}
+
+func cloneStrings(s []string) []string {
+	if len(s) == 0 {
+		return nil
+	}
+
+	return append([]string(nil), s...)
+}
+
+// ---- Firewalls ----
+
+//nolint:gocritic // cfg matches the driver signature.
+func (m *Mock) CreateFirewall(_ context.Context, cfg nfdriver.CreateFirewallConfig) (*nfdriver.Firewall, error) {
+	if cfg.Name == "" {
+		return nil, cerrors.New(cerrors.InvalidArgument, "FirewallName is required")
+	}
+
+	if m.firewalls.Has(cfg.Name) {
+		return nil, cerrors.Newf(cerrors.AlreadyExists, "firewall %q already exists", cfg.Name)
+	}
+
+	fw := &nfdriver.Firewall{
+		Name:             cfg.Name,
+		ARN:              m.arn("firewall", cfg.Name),
+		PolicyARN:        cfg.PolicyARN,
+		VPCID:            cfg.VPCID,
+		SubnetIDs:        cloneStrings(cfg.SubnetIDs),
+		Description:      cfg.Description,
+		DeleteProtection: cfg.DeleteProtection,
+		Status:           "READY",
+		Tags:             copyTags(cfg.Tags),
+	}
+	m.firewalls.Set(cfg.Name, fw)
+
+	out := cloneFirewall(fw)
+
+	return &out, nil
+}
+
+func (m *Mock) DescribeFirewall(_ context.Context, name, arn string) (*nfdriver.Firewall, error) {
+	fw, ok := m.lookupFirewall(name, arn)
+	if !ok {
+		return nil, cerrors.Newf(cerrors.NotFound, "firewall %q not found", nonEmpty(name, arn))
+	}
+
+	out := cloneFirewall(fw)
+
+	return &out, nil
+}
+
+func (m *Mock) DeleteFirewall(_ context.Context, name, arn string) (*nfdriver.Firewall, error) {
+	fw, ok := m.lookupFirewall(name, arn)
+	if !ok {
+		return nil, cerrors.Newf(cerrors.NotFound, "firewall %q not found", nonEmpty(name, arn))
+	}
+
+	if fw.DeleteProtection {
+		return nil, cerrors.Newf(cerrors.FailedPrecondition, "firewall %q has delete protection enabled", fw.Name)
+	}
+
+	fw.Status = "DELETING"
+	m.firewalls.Delete(fw.Name)
+
+	out := cloneFirewall(fw)
+
+	return &out, nil
+}
+
+func (m *Mock) ListFirewalls(_ context.Context) ([]nfdriver.Firewall, error) {
+	all := m.firewalls.SortedValues()
+	out := make([]nfdriver.Firewall, 0, len(all))
+
+	for _, f := range all {
+		out = append(out, cloneFirewall(f))
+	}
+
+	return out, nil
+}
+
+func (m *Mock) lookupFirewall(name, arn string) (*nfdriver.Firewall, bool) {
+	if name != "" {
+		return m.firewalls.Get(name)
+	}
+
+	for _, f := range m.firewalls.SortedValues() {
+		if f.ARN == arn {
+			return f, true
+		}
+	}
+
+	return nil, false
+}
+
+func cloneFirewall(f *nfdriver.Firewall) nfdriver.Firewall {
+	out := *f
+	out.SubnetIDs = cloneStrings(f.SubnetIDs)
+	out.Tags = copyTags(f.Tags)
+
+	return out
+}
+
+// ---- Firewall Policies ----
+
+//nolint:gocritic // cfg matches the driver signature.
+func (m *Mock) CreateFirewallPolicy(_ context.Context, cfg nfdriver.CreateFirewallPolicyConfig) (*nfdriver.FirewallPolicy, error) {
+	if cfg.Name == "" {
+		return nil, cerrors.New(cerrors.InvalidArgument, "FirewallPolicyName is required")
+	}
+
+	if m.policies.Has(cfg.Name) {
+		return nil, cerrors.Newf(cerrors.AlreadyExists, "firewall policy %q already exists", cfg.Name)
+	}
+
+	p := &nfdriver.FirewallPolicy{
+		Name:                            cfg.Name,
+		ARN:                             m.arn("firewall-policy", cfg.Name),
+		ID:                              idgen.GenerateID(""),
+		Description:                     cfg.Description,
+		StatelessDefaultActions:         cloneStrings(cfg.StatelessDefaultActions),
+		StatelessFragmentDefaultActions: cloneStrings(cfg.StatelessFragmentDefaultActions),
+		Tags:                            copyTags(cfg.Tags),
+	}
+	m.policies.Set(cfg.Name, p)
+
+	out := cloneFirewallPolicy(p)
+
+	return &out, nil
+}
+
+func (m *Mock) DescribeFirewallPolicy(_ context.Context, name, arn string) (*nfdriver.FirewallPolicy, error) {
+	p, ok := m.lookupPolicy(name, arn)
+	if !ok {
+		return nil, cerrors.Newf(cerrors.NotFound, "firewall policy %q not found", nonEmpty(name, arn))
+	}
+
+	out := cloneFirewallPolicy(p)
+
+	return &out, nil
+}
+
+func (m *Mock) DeleteFirewallPolicy(_ context.Context, name, arn string) (*nfdriver.FirewallPolicy, error) {
+	p, ok := m.lookupPolicy(name, arn)
+	if !ok {
+		return nil, cerrors.Newf(cerrors.NotFound, "firewall policy %q not found", nonEmpty(name, arn))
+	}
+
+	m.policies.Delete(p.Name)
+
+	out := cloneFirewallPolicy(p)
+
+	return &out, nil
+}
+
+func (m *Mock) ListFirewallPolicies(_ context.Context) ([]nfdriver.FirewallPolicy, error) {
+	all := m.policies.SortedValues()
+	out := make([]nfdriver.FirewallPolicy, 0, len(all))
+
+	for _, p := range all {
+		out = append(out, cloneFirewallPolicy(p))
+	}
+
+	return out, nil
+}
+
+func (m *Mock) lookupPolicy(name, arn string) (*nfdriver.FirewallPolicy, bool) {
+	if name != "" {
+		return m.policies.Get(name)
+	}
+
+	for _, p := range m.policies.SortedValues() {
+		if p.ARN == arn {
+			return p, true
+		}
+	}
+
+	return nil, false
+}
+
+func cloneFirewallPolicy(p *nfdriver.FirewallPolicy) nfdriver.FirewallPolicy {
+	out := *p
+	out.StatelessDefaultActions = cloneStrings(p.StatelessDefaultActions)
+	out.StatelessFragmentDefaultActions = cloneStrings(p.StatelessFragmentDefaultActions)
+	out.Tags = copyTags(p.Tags)
+
+	return out
+}
+
+// ---- Rule Groups ----
+
+//nolint:gocritic // cfg matches the driver signature.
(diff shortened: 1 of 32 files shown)


providers/aws/networkfirewall/networkfirewall.go
```diff
@@ -0,0 +1,511 @@
...
+
+var _ nfdriver.NetworkFirewall = (*Mock)(nil)
+
+// Mock is the in-memory AWS Network Firewall implementation. Resources are
+// keyed by name (unique per account/region, as in the real service).
+type Mock struct {
+	firewalls  *memstore.Store[*nfdriver.Firewall]
+	policies   *memstore.Store[*nfdriver.FirewallPolicy]
+	ruleGroups *memstore.Store[*nfdriver.RuleGroup]
+	logging    map[string][]string
```
NitinKumar004, line 34: **High — the Mock has no mutex.** `logging` is a bare `map[string][]string` written in `UpdateLoggingConfiguration` and read in `DescribeLoggingConfiguration` with zero synchronization (memstore doesn't cover this field) — two concurrent updates trip `fatal error: concurrent map writes`, crashing the process. The pointer-valued `firewalls`/`policies`/`ruleGroups` stores also hand back shared `*Firewall`/`*FirewallPolicy` pointers whose fields (`AssociateFirewallPolicy` → `fw.PolicyARN`, `TagResource`, `UpdateFirewallDeleteProtection`, subnet slices) are mutated with no lock, racing `DescribeFirewall`/`cloneFirewall`. Add a `sync.RWMutex` and guard the compound read-modify-write paths + the `logging` map (the sibling `vpc.Mock` documents exactly this).

providers/aws/vpc/vpn.go
```diff
@@ -0,0 +1,263 @@
...
+func (m *Mock) DeleteVPNGateway(_ context.Context, id string) error {
+	if !m.vpnGateways.Delete(id) {
+		return errors.Newf(errors.NotFound, "vpn gateway %q not found", id)
+	}
+
+	return nil
+}
+
+// DescribeVPNGateways returns VPN gateways matching ids.
+func (m *Mock) DescribeVPNGateways(_ context.Context, ids []string) ([]driver.VPNGateway, error) {
```
NitinKumar004, line 88: **High — reader takes no lock while the mutator does.** `DescribeVPNGateways` reads the shared `*driver.VPNGateway` (via `describeResources` → `cloneVPNGateway`'s `*v` deref) with no `m.mu.RLock`, while `AttachVPNGateway`/`DetachVPNGateway` mutate `vgw.AttachedVPCID`/`AttachmentState` in-place under `m.mu.Lock`. Since the store is pointer-valued and the server calls these concurrently (one goroutine per request), that's a data race. The existing `DescribeVPCs` correctly `RLock`s — this pattern is missing from all six new capability files (vpn/transitgateway/prefixlist/clientvpn/dhcpoptions/egressonlyigw have 0 `RLock`). Add `m.mu.RLock()`/`RUnlock()` to every new Describe/Get/Search.

server/aws/ec2/endpoint_service.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		writeEndpointServiceErr(w, err)
+		return
+	}
+
+	awsquery.WriteXMLResponse(w, struct {
+		XMLName     xml.Name `xml:"ModifyVpcEndpointServicePermissionsResponse"`
+		Xmlns       string   `xml:"xmlns,attr"`
+		Req         string   `xml:"requestId"`
+		ReturnValue bool     `xml:"returnValue"`
```
NitinKumar004: **Medium — wrong response element name.** The real ec2 SDK deserializes this action's result from `<return>` (`aws-sdk-go-v2` deserializer: `EqualFold("return", ...)`), not `<returnValue>`. As written, the SDK skips the element and `resp.ReturnValue` is always `nil` — callers reading `aws.ToBool(resp.ReturnValue)` silently get `false`, or panic on `*resp.ReturnValue`. Fix: `xml:"return"`.

providers/aws/vpc/transitgateway.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+	m.transitGateways.Set(tgw.ID, tgw)
+
+	out := cloneTGW(tgw)
+
+	return &out, nil
+}
+
+// DeleteTransitGateway deletes a transit gateway.
+func (m *Mock) DeleteTransitGateway(_ context.Context, id string) (*driver.TransitGateway, error) {
```
NitinKumar004, line 37: **Medium — delete leaves dangling references.** `DeleteTransitGateway` removes the TGW even while live VPC attachments / route tables / routes still reference it (their `TransitGatewayID` then resolves to nothing), unlike `DeleteVPC` one file over which guards on attached ENIs. Same family: `CreateVPNConnection`/`ModifyVPNConnection` don't validate `VpnGatewayId`/`TransitGatewayId`, `CreateTransitGatewayRoute` doesn't validate `attachmentId`, and Network Firewall doesn't check policy/rule-group references. Consider FK-existence checks on create/associate and an in-use guard on delete (or document the permissiveness).

NitinKumar004 (changes requested):
## Deep review — request changes

Reviewed end-to-end at `59f25d5` (isolated worktree, real build/test/-race, wire-fidelity check against the vendored aws-sdk-go-v2, adversarial pass). **The architecture here is genuinely strong** — the AWS-only resources are 7 *optional* capability interfaces discovered by type assertion (like `NetworkInterfaces`/`VPCAttributes`), so the portable `Networking` driver stays minimal; factory wiring is clean; docs framing is correct; no aliasing bugs; ID prefixes correct; and the wire format is overwhelmingly faithful (list `.N` flattening, `Set`-suffix, singular Client-VPN wrappers all match the real SDK). Nice work.

Two thread-safety defects block, plus a cluster of Medium wire/coverage/integrity items.

### Gate matrix
build / vet / `go test` (new pkgs) pass; `go mod tidy` clean (deps: +1 justified — networkfirewall SDK). **`-race` is green but it's a false green** — the test suite is entirely sequential, so the races below never fire during tests even though the code races under concurrent use. golangci-lint has ~10 PR-introduced Low issues (mostly *unused* `//nolint` directives + 2 unused-receiver methods + 1 wsl) — the local gate (per CLAUDE.md; CI doesn't run golangci-lint) should be 0.

### High — thread safety (inline)
1. **`networkfirewall.Mock` has no mutex at all.** `logging map[string][]string` is written/read with zero synchronization → two concurrent `UpdateLoggingConfiguration` calls trip Go's `fatal error: concurrent map writes` (a real process crash); the pointer-struct field mutations (`AssociateFirewallPolicy`, `TagResource`, …) are also unlocked. The server serves each request in its own goroutine, so this is reachable.
2. **The 15 new vpc capability Describe/Get readers omit `m.mu.RLock`** while their mutators hold `m.mu.Lock`. The stores are pointer-valued (`Store[*driver.VPNGateway]`), so e.g. `AttachVPNGateway` mutating `vgw.AttachedVPCID` in-place races a concurrent `DescribeVPNGateways` clone-read. The existing `DescribeVPCs` *does* `RLock` — 6 of the 7 new capability files have **0** `RLock`.

### Medium
- **Wire: `ModifyVpcEndpointServicePermissions` returns `<returnValue>`; the real SDK reads `<return>`** (inline) — verified against the vendored ec2 deserializer, so `resp.ReturnValue` decodes as nil for every caller. One-line fix: `xml:"return"`.
- **Wire: `CreateVpnConnection`/`DescribeVpnConnections` omit `customerGatewayConfiguration`** — the IPsec tunnel-config blob real responses always carry (Terraform's `aws_vpn_connection` parses it) is silently absent; the response struct has no field for it.
- **Referential-integrity / validation gaps** (inline on TGW as the representative): `DeleteTransitGateway` leaves dangling attachments/route-tables (vs `DeleteVPC`'s ENI guard); `CreateVPNConnection`/`ModifyVPNConnection` don't validate `VpnGatewayId`/`TransitGatewayId`; `CreateTransitGatewayRoute` doesn't validate `attachmentId`; `PrefixList.MaxEntries` isn't enforced; Network Firewall doesn't check policy/rule-group references (delete a referenced policy, or associate a nonexistent one, both succeed). More permissive than real AWS and inconsistent with the package's own `DeleteVPC`.
- **Test coverage: 16 capability `Delete*`/`Describe*` methods + 2 Network Firewall list ops have zero coverage** — the cleanup / read-back-many halves of the CRUD surface ship with no regression protection for a parity PR.
- **SDK round-trip tests assert only happy paths** — the EC2 error-XML and NF JSON error encodings for the new resources are never verified against a real SDK client.
- **Docs op-count off by 4**: `docs/services.md` states "AWS-specific total: 54" but the capability interfaces total **58** (15+14+4+5+3+5+12), so the grand total should be 1493, not 1489.

### Low
RuleGroup ARN hardcodes `stateful-rulegroup` even for STATELESS groups; `CreateClientVpnEndpoint` silently drops the required `AuthenticationOptions`; missing attached-state delete guards (VPN/customer gateway) and DHCP association is unobservable (structural); `DeleteVpcEndpointServiceConfigurations.Unsucc` shape is wrong (dead code today); the new provider tests bypass the package's existing hand-rolled helpers; and the lint hygiene above.

Requesting changes for the two High thread-safety issues; the Mediums are worth batching alongside. The design is the right one — this is about hardening it.

providers/aws/networkfirewall/networkfirewall.go
thzgajendra: Fixed in 9e7b428. Added a `sync.RWMutex` to `networkfirewall.Mock`; every method now locks (mutators `Lock`, readers `RLock`), so the `logging` map writes and the pointer-field mutations (associate policy/subnets, tag, delete-protection) are serialized. Added `TestFirewallConcurrentAccess` which drives 20 goroutines of interleaved logging writes + describes + tags — passes under `-race`.

providers/aws/vpc/vpn.go
thzgajendra: Fixed in 9e7b428. Added `m.mu.RLock()` to every new capability reader (vpn, transit gateway, prefix list, client vpn, dhcp options, egress-only IGW, endpoint-service perms) and `Lock` to the in-place mutators that were missing it (the transit-gateway and prefix-list deletes that set `.State`). Added `TestNetworkingConcurrentReadWrite` (Attach/Detach vs Describe) — passes under `-race`.

server/aws/ec2/endpoint_service.go
thzgajendra: Good catch — fixed in 9e7b428. Changed the tag to `xml:"return"`. Confirmed against the vendored deserializer (`awsEc2query_deserializeOpDocumentModifyVpcEndpointServicePermissionsOutput` reads `EqualFold("return", ...)`).

providers/aws/vpc/transitgateway.go
thzgajendra: Fixed in 9e7b428. Added the full FK/in-use batch: `DeleteTransitGateway` now guards on live attachments/route tables; `CreateVpnConnection`/`ModifyVpnConnection` validate the gateway IDs; `CreateTransitGatewayRoute` validates `attachmentId`; `PrefixList.MaxEntries` is enforced on create+modify; and Network Firewall validates policy references on associate/create with an in-use guard on `DeleteFirewallPolicy`. (Rule-group references aren't modeled on the policy object yet, so that specific in-use check is out of scope here.)

thzgajendra:
Thanks for the thorough review — all points addressed in `9e7b428` (gate green locally: gofmt/build/vet/`go test`/tidy/lint clean; CodeQL shows only the pre-existing baseline alerts, none in changed files).

**High — thread safety**
- `networkfirewall.Mock`: added `sync.RWMutex`; all methods lock. + `TestFirewallConcurrentAccess` under `-race`.
- vpc readers: `RLock` on all new Describe/Get/Search; `Lock` on the in-place delete mutators. + `TestNetworkingConcurrentReadWrite` under `-race`.

**Medium**
- Endpoint-service `<returnValue>` → `<return>` (verified vs the SDK deserializer).
- Added `customerGatewayConfiguration` to the VPN connection response (minimal but structurally-valid tunnel-config document).
- Referential integrity: FK checks on VPN/TGW create+modify, PrefixList `MaxEntries`, NF policy references; in-use guards on `DeleteTransitGateway` / `DeleteFirewallPolicy`.
- Coverage: added the `Delete*`/`Describe*`/list read-back halves for the capabilities and an SDK error-path assertion for both EC2 (query-XML) and Network Firewall (JSON) services.
- Docs: corrected the op counts — AWS-specific **58**, Network Firewall **20**, grand total **1493**.

**Low**
- `CreateClientVpnEndpoint` now captures/echoes `AuthenticationOptions` and requires at least one.
- RuleGroup ARN uses `stateless-rulegroup` for STATELESS groups.
- Corrected the `UnsuccessfulItem` shape on `DeleteVpcEndpointServiceConfigurations`.
- Lint hygiene back to 0 PR-introduced findings (removed orphaned `//nolint` directives, fixed unused-receivers + wsl + goconst + hugeParam).

One deliberate non-change: rule-group *reference* in-use checks aren't implemented because the `FirewallPolicy` model doesn't yet carry rule-group references — happy to add that if you'd like it modeled.

NitinKumar004 (approved):
## Re-review of the fix commit — approve

Verified `9e7b428` ("address review — thread safety, FK validation, wire fidelity") against the prior round. This cleanly resolves everything that was blocking.

### Both High thread-safety issues — fixed and proven under concurrency
- `networkfirewall.Mock` now carries a `sync.RWMutex`; every method locks, and the `logging` map is guarded (write under `Lock`, read under `RLock`).
- All six vpc capability files now `RLock` their Describe/Get/Search readers (was 0 before), matching the existing `DescribeVPCs` discipline.
- I re-ran the exact previously-racy paths under a **concurrent** `-race` test — 30–40 goroutines each hammering `UpdateLoggingConfiguration`+`DescribeLoggingConfiguration`+`AssociateFirewallPolicy`+`DescribeFirewall`, and `AttachVPNGateway`+`DescribeVPNGateways` — clean, no race, no crash. A genuine green this time.

### Mediums — all fixed
- Wire: `ModifyVpcEndpointServicePermissions` now renders `<return>` (SDK-decodable); `CreateVpnConnection`/`DescribeVpnConnections` now carry `customerGatewayConfiguration`.
- FK validation: `DeleteTransitGateway` guards live attachments/route-tables; `CreateTransitGatewayRoute` validates `attachmentId`; `CreateVpnConnection`/`ModifyVpnConnection` validate the gateway ids; Network Firewall validates the policy on create/associate and blocks deleting a policy still in use.
- Coverage: the previously-untested `Delete*`/`Describe*` methods are now exercised, and SDK-level negative paths are asserted (NF `ResourceNotFoundException`, VPN query-error-XML).
- Docs op-count corrected to 58 (grand total 1493); PrefixList `MaxEntries` enforced on create + modify; STATELESS rule-group ARN fixed.

Gates: build / test / gofmt / `go mod tidy` clean.

### One optional nit (Low, non-blocking)
Local golangci-lint reports 4 `dupl` hits on the parallel per-resource marshaling / dispatch shape (`handler.go`, `transit_gateway.go`/`vpn.go`, `depth.go`). It's duplicate by design — worth a `//nolint:dupl // parallel per-resource marshaling` on those blocks to keep the local gate at 0. (The `routeVPCResource` gocyclo is pre-existing.)

Solid work addressing the review. Approving.

PR #318, opened 2026-08-04
AWS Networking (Stage B): VPC IPAM (Full Parity + Metrics)

thzgajendra:
## Objective
Add AWS IP Address Manager (IPAM) to the emulator with **full parity** — the entire control-plane surface plus the `AWS/IPAM` CloudWatch metrics — as the first Stage B AWS networking service. (Stage A merged in #317.)

## Scope — the whole service (~69 operations + metrics)
New optional EC2 capabilities, each type-asserted on the vpc driver (the Stage-A pattern, keeping the portable `Networking` interface clean):

- **Core lifecycle** — IPAM / Scope / Pool CRUD+Modify; pool CIDR Provision/Deprovision/Get; allocation Allocate/Release/Get/Modify. Creating an IPAM auto-creates the public + private default scopes **and** a default resource discovery + association (as in AWS).
- **Resource CIDRs + address history** — `GetIpamResourceCidrs`, `ModifyIpamResourceCidr`, `GetIpamAddressHistory`.
- **Resource Discovery** — CRUD + Associate/Disassociate/DescribeAssociations + `GetIpamDiscoveredAccounts` / `…ResourceCidrs` / `…PublicAddresses`.
- **BYOASN** — Provision/Deprovision/Associate/Disassociate/Describe.
- **BYOIP / public IP** — Move-to-IPAM/Provision/Deprovision/Describe/Advertise/Withdraw.
- **Prefix-List Resolver** — resolvers + targets + versions/rules/version-entries.
- **External Resource Verification Token** — Create/Delete/Describe.
- **Policy + Organization admin** — Create/Delete/Describe/Enable/Disable/GetEnabled/AllocationRules/OrgTargets + Enable/DisableIpamOrganizationAdminAccount.
- **Metrics (`AWS/IPAM`)** — the vpc driver exposes an optional `IPAMMetrics` capability; the CloudWatch handler surfaces it via `ListMetrics` + `GetMetricStatistics`: `TotalActiveIpCount`; pool `PercentAllocated/Assigned/Available` + `Compliant/NoncompliantResourceCidrs`; scope `Managed/Unmanaged/Overlapping/Compliant/NoncompliantResourceCidrs`; public-IP insight counts; and resource utilization `VpcIPUsage` / `SubnetIPUsage`. Values are computed **live** from IPAM + VPC/subnet/EIP state.

## How it's modeled (the honest part)
An in-memory single-account emulator has no live network to scan or Organization to manage, so the cross-account/discovery/metrics features are **derived from the emulator's own state**: discovered resources come from the stored VPCs/subnets/EIPs, discovered accounts / org targets resolve to the configured account, and prefix-list-resolver rules derive from the IPAM's pools. All control-plane objects (BYOASN/BYOIP, tokens, policies, resolvers) are full state machines.

## Quality bar (carried from the #317 review)
- **Thread-safe from the start** — readers `RLock`, mutators `Lock`; concurrent `-race` tests.
- **Referential integrity** — FK checks on create/associate (scope→ipam, pool→scope, resolver-target→resolver, allocate/provision→pool, policy/token/RD→ipam) and in-use guards on delete (ipam with pools/non-default scopes/RD-associations, scope with pools, pool with allocations, resolver with targets, advertised BYOIP not deprovisionable).
- **Wire fidelity** — every response element name verified against the vendored `aws-sdk-go-v2` deserializers.

## Docs / Test / Playground
- Real aws-sdk-go-v2 EC2 round-trips: `TestEC2IPAMParitySDK` (core) + `TestEC2IPAMFullSDK` (resource-cidrs/discovery/byoip+byoasn/resolver+token/policy+org-admin).
- CloudWatch SDK metrics test: `TestIPAMMetricsSDK` (ListMetrics + GetMetricStatistics on `AWS/IPAM`).
- Provider unit + `-race` tests: `TestIPAMLifecycle`, `TestIPAMFullProvider`, `TestIPAMConcurrentAccess`.
- `docs/services.md`: full IPAM capability row + AWS/IPAM metrics section; AWS-specific networking 58 → **127**; Grand Total → **1562**.

## Test plan
- [x] `go build ./... && go vet ./...`
- [x] `go test ./...` + `-race` on the vpc package
- [x] `go mod tidy` clean; `golangci-lint` 0 findings in new files; CodeQL 0 new alerts (only pre-existing baseline)

## Risk & Rollback
Additive — every IPAM group is a new optional capability discovered by type assertion; CloudWatch gains an optional metrics source that is nil-safe. No existing behavior changes. Rollback = revert the two commits on this branch.

## Follow-ups (remaining Stage B services, one PR each)
Route 53 Resolver, Global Accelerator, Direct Connect, VPC Lattice, Cloud WAN/Network Manager, Traffic Mirroring, Reachability/Network Access Analyzer, VPC Block Public Access.

---

## Review round 1 — addressed (e2ef707)

- **ListMetrics** (blocker): empty namespace now returns all real metrics (true namespaces) merged with AWS/IPAM; exact `AWS/IPAM` stays IPAM-only. Added AWS-local `ListMetricsDetailed`.
- **Provision/AllocateIpamPoolCidr** (blocker): netmask-only now derives a concrete aligned CIDR (sequential carve) instead of storing an empty one; errors if no space.
- **ModifyIpamResourceCidr** (blocker): scope-move/unmonitor now persist via a per-resource override map (write-locked).
- **Resolver target** (medium): validates `prefixListID`.
- **BYOASN associate/disassociate** (medium): validate CIDR-provisioned / ASN-exists.

Regression tests extended (`networking_parity_test.go`, `TestIPAMMetricsSDK`). Build/vet/test/tidy green; CodeQL unchanged (8 baseline).

---

## Review round 2 — addressed (a7046e4)

All three verified against the real `aws-sdk-go-v2` `ec2@v1.317.1` types before fixing:

- **ModifyIpamPolicyAllocationRules** (medium, wire): the output type has `IpamPolicyDocument`, not a `Return` field — the old `<return>true>` deserialized to `nil`. Now emits `<ipamPolicyDocument>` with `allocationRuleSet>item>sourceIpamPoolId` + `ipamPolicyId`/`locale`/`resourceType`.
- **GetIpamPolicyAllocationRules** (medium, wire + model): the driver's `AllocationRules []string` couldn't carry the real shape, so it's now structured — `[]IpamAllocationRule{SourceIpamPoolID}` plus `Locale`/`ResourceType` on the policy. Response emits `ipamPolicyDocumentSet>item` (correct nested shape) + `nextToken`; request parses `AllocationRule.N.SourceIpamPoolId` + `Locale` + `ResourceType`.
- **Cross-pool CIDR overlap** (medium, correctness): netmask-only provisioning now excludes CIDRs provisioned by **any** pool sharing the carve base, so two pools no longer both receive `10.0.0.0/16`.

New coverage: real-SDK Modify/Get round-trip in `TestNetworkingParitySDK`; two-pool non-overlap in `TestIPAMProvisionNoCrossPoolOverlap`. Build/vet/`go test ./...`/tidy green; `golangci-lint --new-from-rev` = 0 new; CodeQL unchanged (8 baseline).

### DIFF
--- a/providers/aws/vpc/ipam.go
+++ b/providers/aws/vpc/ipam.go
@@ -1,0 +1,538 @@
+package vpc
+
+import (
+	"context"
+
+	"github.com/stackshy/cloudemu/v2/errors"
+	"github.com/stackshy/cloudemu/v2/internal/idgen"
+	"github.com/stackshy/cloudemu/v2/services/networking/driver"
+)
+
+const ipamStateDeleteComplete = "delete-complete"
+
+// ipamARN builds an IPAM-family ARN. IPAM ARNs carry no region segment
+// (arn:aws:ec2::<account>:<resource>), matching the real service.
+func (m *Mock) ipamARN(resource string) string {
+	return idgen.AWSARN("ec2", "", m.opts.AccountID, resource)
+}
+
+// CreateIpam creates an IPAM plus its public and private default scopes.
+func (m *Mock) CreateIpam(_ context.Context, cfg driver.IpamConfig) (*driver.Ipam, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	id := idgen.GenerateID("ipam-")
+	ipamARN := m.ipamARN("ipam/" + id)
+
+	pub := m.newDefaultScope(ipamARN, "public")
+	priv := m.newDefaultScope(ipamARN, "private")
+
+	ipam := &driver.Ipam{
+		ID:                    id,
+		ARN:                   ipamARN,
+		Region:                m.opts.Region,
+		PublicDefaultScopeID:  pub.ID,
+		PrivateDefaultScopeID: priv.ID,
+		ScopeCount:            2,
+		OperatingRegions:      cfg.OperatingRegions,
+		Description:           cfg.Description,
+		Tier:                  orDefaultStr(cfg.Tier, "advanced"),
+		State:                 "create-complete",
+		Tags:                  copyTags(cfg.Tags),
+	}
+	m.ipams.Set(id, ipam)
+
+	out := cloneIpam(ipam)
+
+	return &out, nil
+}
+
+// newDefaultScope creates and stores a default scope for an IPAM. Caller holds mu.
+func (m *Mock) newDefaultScope(ipamARN, scopeType string) *driver.IpamScope {
+	id := idgen.GenerateID("ipam-scope-")
+	scope := &driver.IpamScope{
+		ID:        id,
+		ARN:       m.ipamARN("ipam-scope/" + id),
+		IpamARN:   ipamARN,
+		ScopeType: scopeType,
+		IsDefault: true,
+		State:     "create-complete",
+	}
+	m.ipamScopes.Set(id, scope)
+
+	return scope
+}
+
+// DescribeIpams returns IPAMs matching ids (all if empty).
+func (m *Mock) DescribeIpams(_ context.Context, ids []string) ([]driver.Ipam, error) {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	return describeResources(m.ipams, ids, cloneIpam), nil
+}
+
+// ModifyIpam updates an IPAM's description.
+func (m *Mock) ModifyIpam(_ context.Context, id, description string) (*driver.Ipam, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	ipam, ok := m.ipams.Get(id)
+	if !ok {
+		return nil, errors.Newf(errors.NotFound, "ipam %q not found", id)
+	}
+
+	ipam.Description = description
+
+	out := cloneIpam(ipam)
+
+	return &out, nil
+}
+
+// DeleteIpam deletes an IPAM and its default scopes. Non-default scopes or
+// pools referencing the IPAM block deletion.
+func (m *Mock) DeleteIpam(_ context.Context, id string) (*driver.Ipam, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	ipam, ok := m.ipams.Get(id)
+	if !ok {
+		return nil, errors.Newf(errors.NotFound, "ipam %q not found", id)
+	}
+
+	for _, s := range m.ipamScopes.SortedValues() {
+		if s.IpamARN == ipam.ARN && !s.IsDefault {
+			return nil, errors.Newf(errors.FailedPrecondition, "ipam %q has non-default scopes", id)
+		}
+	}
+
+	if m.poolsInScopes(ipam.ARN) {
+		return nil, errors.Newf(errors.FailedPrecondition, "ipam %q has pools", id)
+	}
+
+	for _, s := range m.ipamScopes.SortedValues() {
+		if s.IpamARN == ipam.ARN {
+			m.ipamScopes.Delete(s.ID)
+		}
+	}
+
+	ipam.State = ipamStateDeleteComplete
+
+	m.ipams.Delete(id)
+
+	out := cloneIpam(ipam)
+
+	return &out, nil
+}
+
+// CreateIpamScope creates a non-default (private) scope within an IPAM.
+func (m *Mock) CreateIpamScope(_ context.Context, cfg driver.IpamScopeConfig) (*driver.IpamScope, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	ipam, ok := m.ipams.Get(cfg.IpamID)
+	if !ok {
+		return nil, errors.Newf(errors.InvalidArgument, "ipam %q not found", cfg.IpamID)
+	}
+
+	id := idgen.GenerateID("ipam-scope-")
+	scope := &driver.IpamScope{
+		ID:          id,
+		ARN:         m.ipamARN("ipam-scope/" + id),
+		IpamARN:     ipam.ARN,
+		ScopeType:   "private",
+		IsDefault:   false,
+		Description: cfg.Description,
+		State:       "create-complete",
+		Tags:        copyTags(cfg.Tags),
+	}
+	m.ipamScopes.Set(id, scope)
+
+	ipam.ScopeCount++
+
+	out := cloneIpamScope(scope)
+
+	return &out, nil
+}
+
+// DescribeIpamScopes returns scopes matching ids (all if empty).
+func (m *Mock) DescribeIpamScopes(_ context.Context, ids []string) ([]driver.IpamScope, error) {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	return describeResources(m.ipamScopes, ids, cloneIpamScope), nil
+}
+
+// ModifyIpamScope updates a scope's description.
+func (m *Mock) ModifyIpamScope(_ context.Context, id, description string) (*driver.IpamScope, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	scope, ok := m.ipamScopes.Get(id)
+	if !ok {
+		return nil, errors.Newf(errors.NotFound, "ipam scope %q not found", id)
+	}
+
+	scope.Description = description
+
+	out := cloneIpamScope(scope)
+
+	return &out, nil
+}
+
+// DeleteIpamScope deletes a non-default scope with no pools.
+func (m *Mock) DeleteIpamScope(_ context.Context, id string) (*driver.IpamScope, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	scope, ok := m.ipamScopes.Get(id)
+	if !ok {
+		return nil, errors.Newf(errors.NotFound, "ipam scope %q not found", id)
+	}
+
+	if scope.IsDefault {
+		return nil, errors.Newf(errors.FailedPrecondition, "cannot delete default scope %q", id)
+	}
+
+	for _, p := range m.ipamPools.SortedValues() {
+		if p.IpamScopeARN == scope.ARN {
+			return nil, errors.Newf(errors.FailedPrecondition, "ipam scope %q has pools", id)
+		}
+	}
+
+	scope.State = ipamStateDeleteComplete
+
+	m.ipamScopes.Delete(id)
+
+	if ipam := m.ipamByARN(scope.IpamARN); ipam != nil {
+		ipam.ScopeCount--
+	}
+
+	out := cloneIpamScope(scope)
+
+	return &out, nil
+}
+
+// CreateIpamPool creates a CIDR pool in a scope.
+//
+//nolint:gocritic // cfg matches the driver signature.
+func (m *Mock) CreateIpamPool(_ context.Context, cfg driver.IpamPoolConfig) (*driver.IpamPool, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	scope, ok := m.ipamScopes.Get(cfg.IpamScopeID)
+	if !ok {
+		return nil, errors.Newf(errors.InvalidArgument, "ipam scope %q not found", cfg.IpamScopeID)
+	}
+
+	id := idgen.GenerateID("ipam-pool-")
+	pool := &driver.IpamPool{
+		ID:                             id,
+		ARN:                            m.ipamARN("ipam-pool/" + id),
+		IpamScopeARN:                   scope.ARN,
+		IpamScopeType:                  scope.ScopeType,
+		AddressFamily:                  orDefaultStr(cfg.AddressFamily, "ipv4"),
+		Locale:                         orDefaultStr(cfg.Locale, "None"),
+		PoolDepth:                      1,
+		Description:                    cfg.Description,
+		State:                          "create-complete",
+		AllocationMinNetmaskLength:     cfg.AllocationMinNetmaskLength,
+		AllocationMaxNetmaskLength:     cfg.AllocationMaxNetmaskLength,
+		AllocationDefaultNetmaskLength: cfg.AllocationDefaultNetmaskLength,
+		Tags:                           copyTags(cfg.Tags),
+	}
+	m.ipamPools.Set(id, pool)
+
+	scope.PoolCount++
+
+	out := cloneIpamPool(pool)
+
+	return &out, nil
+}
+
+// DescribeIpamPools returns pools matching ids (all if empty).
+func (m *Mock) DescribeIpamPools(_ context.Context, ids []string) ([]driver.IpamPool, error) {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	return describeResources(m.ipamPools, ids, cloneIpamPool), nil
+}
+
+// ModifyIpamPool updates a pool's description.
+func (m *Mock) ModifyIpamPool(_ context.Context, id, description string) (*driver.IpamPool, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	pool, ok := m.ipamPools.Get(id)
+	if !ok {
+		return nil, errors.Newf(errors.NotFound, "ipam pool %q not found", id)
+	}
+
+	pool.Description = description
+
+	out := cloneIpamPool(pool)
+
+	return &out, nil
+}
+
+// DeleteIpamPool deletes a pool with no live allocations.
+func (m *Mock) DeleteIpamPool(_ context.Context, id string) (*driver.IpamPool, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	pool, ok := m.ipamPools.Get(id)
+	if !ok {
+		return nil, errors.Newf(errors.NotFound, "ipam pool %q not found", id)
+	}
+
+	for _, a := range m.ipamAllocations.SortedValues() {
+		if m.ipamPoolByAllocation[a.ID] == id {
+			return nil, errors.Newf(errors.FailedPrecondition, "ipam pool %q has allocations", id)
+		}
+	}
+
+	for _, c := range m.ipamPoolCidrs.SortedValues() {
+		if m.ipamPoolByCidr[c.ID] == id {
+			m.ipamPoolCidrs.Delete(c.ID)
+			delete(m.ipamPoolByCidr, c.ID)
+		}
+	}
+
+	pool.State = ipamStateDeleteComplete
+
+	m.ipamPools.Delete(id)
+
+	if scope := m.scopeByARN(pool.IpamScopeARN); scope != nil {
+		scope.PoolCount--
+	}
+
+	out := cloneIpamPool(pool)
+
+	return &out, nil
+}
+
+// ProvisionIpamPoolCidr adds a CIDR to a pool's supply.
+func (m *Mock) ProvisionIpamPoolCidr(_ context.Context, poolID, cidr string, netmaskLength int) (*driver.IpamPoolCidr, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if !m.ipamPools.Has(poolID) {
+		return nil, errors.Newf(errors.InvalidArgument, "ipam pool %q not found", poolID)
+	}
+
+	if cidr == "" && netmaskLength == 0 {
+		return nil, errors.New(errors.InvalidArgument, "a cidr or netmaskLength is required")
+	}
+
+	id := idgen.GenerateID("ipam-pool-cidr-")
+	pc := &driver.IpamPoolCidr{
+		ID:            id,
+		CIDR:          cidr,
+		NetmaskLength: netmaskLength,
+		State:         "provisioned",
+	}
+	m.ipamPoolCidrs.Set(id, pc)
+	m.ipamPoolByCidr[id] = poolID
+
+	out := *pc
+
+	return &out, nil
+}
+
+// DeprovisionIpamPoolCidr removes a provisioned CIDR from a pool.
+func (m *Mock) DeprovisionIpamPoolCidr(_ context.Context, poolID, cidr string) (*driver.IpamPoolCidr, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if !m.ipamPools.Has(poolID) {
+		return nil, errors.Newf(errors.InvalidArgument, "ipam pool %q not found", poolID)
+	}
+
+	for _, c := range m.ipamPoolCidrs.SortedValues() {
+		if m.ipamPoolByCidr[c.ID] != poolID || c.CIDR != cidr {
+			continue
+		}
+
+		c.State = "deprovisioned"
+
+		m.ipamPoolCidrs.Delete(c.ID)
+		delete(m.ipamPoolByCidr, c.ID)
+
+		out := *c
+
+		return &out, nil
+	}
+
+	return nil, errors.Newf(errors.NotFound, "cidr %q not provisioned in pool %q", cidr, poolID)
+}
+
+// GetIpamPoolCidrs returns the CIDRs provisioned into a pool.
+func (m *Mock) GetIpamPoolCidrs(_ context.Context, poolID string) ([]driver.IpamPoolCidr, error) {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	if !m.ipamPools.Has(poolID) {
+		return nil, errors.Newf(errors.InvalidArgument, "ipam pool %q not found", poolID)
+	}
+
+	var out []driver.IpamPoolCidr
+
+	for _, c := range m.ipamPoolCidrs.SortedValues() {
+		if m.ipamPoolByCidr[c.ID] == poolID {
+			out = append(out, *c)
+		}
+	}
+
+	return out, nil
+}
+
+// AllocateIpamPoolCidr hands out a CIDR from a pool.
+func (m *Mock) AllocateIpamPoolCidr(_ context.Context, cfg driver.AllocateIpamPoolCidrConfig) (*driver.IpamPoolAllocation, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if !m.ipamPools.Has(cfg.IpamPoolID) {
+		return nil, errors.Newf(errors.InvalidArgument, "ipam pool %q not found", cfg.IpamPoolID)
+	}
+
+	if cfg.CIDR == "" && cfg.NetmaskLength == 0 {
+		return nil, errors.New(errors.InvalidArgument, "a cidr or netmaskLength is required")
+	}
+
+	id := idgen.GenerateID("ipam-pool-alloc-")
+	alloc := &driver.IpamPoolAllocation{
+		ID:           id,
+		CIDR:         cfg.CIDR,
+		ResourceType: "custom",
+		Description:  cfg.Description,
+		Tags:         copyTags(cfg.Tags),
+	}
+	m.ipamAllocations.Set(id, alloc)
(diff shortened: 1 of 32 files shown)


server/aws/cloudwatch/ops.go
```diff
@@ -154,6 +212,11 @@ func (h *Handler) listMetrics(w http.ResponseWriter, r *http.Request, body []byt
 		return
 	}
 
+	if h.ipam != nil && (in.Namespace == netdriver.IpamMetricNamespace || in.Namespace == "") {
```
Satyam-Trivedi-ZS: 🔴 **Blocker**

When `in.Namespace == ""` (the standard "list all metrics" call) and `h.ipam != nil`, this branch returns only the synthetic AWS/IPAM metrics and never calls `h.monitoring.ListMetrics` below it — so any client listing metrics across all namespaces loses every real metric whenever the VPC driver implements IPAMMetrics (which aws.go's `New()` wires up unconditionally via a type assertion). This shortcut should only trigger for `in.Namespace == netdriver.IpamMetricNamespace`; for the empty-namespace case, merge both metric sets instead of replacing one with the other.

providers/aws/vpc/ipam_resolver.go
```diff
@@ -0,0 +1,286 @@
...
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if !m.ipamResolvers.Has(resolverID) {
+		return nil, errors.Newf(errors.InvalidArgument, "ipam prefix list resolver %q not found", resolverID)
+	}
+
+	id := idgen.GenerateID("ipam-pl-res-target-")
+	t := &driver.IpamPrefixListResolverTarget{
+		ID: id, ARN: m.ipamARN("ipam-prefix-list-resolver-target/" + id), ResolverID: resolverID,
```
Satyam-Trivedi-ZS, line 105: 🟠 **Medium**

CreateIpamPrefixListResolver validates its referenced ipamID (`m.ipams.Get(ipamID)`), but CreateIpamPrefixListResolverTarget only checks the resolverID exists and never validates prefixListID against the managed-prefix-list store (m.prefixLists). A caller can create a resolver target pointing at a prefix list that never existed or was already deleted, and nothing later re-validates it. Consider mirroring the ipamID check: look up `m.prefixLists.Get(prefixListID)` and return InvalidArgument if not found.

providers/aws/vpc/ipam_byoip.go
```diff
@@ -0,0 +1,211 @@
...
+
+// DisassociateIpamByoasn removes a BYOASN↔CIDR association.
+func (m *Mock) DisassociateIpamByoasn(_ context.Context, asn, cidr string) (*driver.AsnAssociation, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	assoc := driver.AsnAssociation{Asn: asn, CIDR: cidr, State: "disassociated"}
+
+	if bc, ok := m.ipamByoipCidrs.Get(cidr); ok {
+		kept := bc.AsnAssociations[:0:0]
```
Satyam-Trivedi-ZS, line 97: 🟠 **Medium**

AssociateIpamByoasn validates the ASN exists but not the CIDR — if `cidr` isn't in m.ipamByoipCidrs, the function still returns a success AsnAssociation without recording anything. DisassociateIpamByoasn doesn't validate the ASN at all and always succeeds. Worth tightening for consistency with the ASN check that does exist on the associate path, though not a blocker for a mock with generally permissive validation.

providers/aws/vpc/ipam.go
```diff
@@ -0,0 +1,566 @@
...
+	if !m.ipamPools.Has(poolID) {
+		return nil, errors.Newf(errors.InvalidArgument, "ipam pool %q not found", poolID)
+	}
+
+	if cidr == "" && netmaskLength == 0 {
+		return nil, errors.New(errors.InvalidArgument, "a cidr or netmaskLength is required")
+	}
+
+	id := idgen.GenerateID("ipam-pool-cidr-")
+	pc := &driver.IpamPoolCidr{
```
Satyam-Trivedi-ZS, line 368: 🔴 **Blocker**

AWS's real usage pattern for both provisioning a CIDR into a pool and allocating a CIDR from a pool is to pass only NetmaskLength (e.g. "give me a /24 out of this pool") and let the service pick the actual block from the pool's available space. Here the guard only rejects the case where BOTH cidr and netmaskLength are empty/zero; when netmaskLength is supplied and cidr is not, `pc := &driver.IpamPoolCidr{CIDR: cidr, ...}` stores CIDR as the empty string forever — there is no path anywhere in this file that computes a real CIDR from netmaskLength plus the pool's known parent CIDR/available space. The same pattern repeats in AllocateIpamPoolCidr (line 430, `CIDR: cfg.CIDR`). The wire handler then serializes `Cidr: c.CIDR` verbatim (server/aws/ec2/ipam.go around line 436), so any SDK caller that provisions/allocates by netmask-only gets back a pool-cidr/allocation object with an empty Cidr field — silently wrong, not an error. It also corrupts the downstream 'AWS/IPAM' metrics: cidrSize("") in providers/aws/vpc/ipam_metrics.go (lines 172, 184) returns 0 for any such record, so PercentAllocated/PercentAvailable become permanently wrong (denominator/numerator silently drop these entries) for the exact provisioning style AWS users actually use.

providers/aws/vpc/ipam_resources.go
```diff
@@ -0,0 +1,162 @@
...
+			continue
+		}
+
+		if destScopeID != "" {
+			c.IpamScopeID = destScopeID
+		} else {
+			c.IpamScopeID = currentScopeID
+		}
+
+		if !monitored {
```
Satyam-Trivedi-ZS, line 162: 🔴 **Blocker**

ModifyIpamResourceCidr computes `cidrs := m.ipamResourceCidrs()`, which freshly re-derives the resource-cidr list from the stored VPCs/subnets every call (each entry hardcoded to ComplianceStatus:"compliant", ManagementState:"managed", OverlapStatus:"nonoverlapping" in ipamResourceCidrs()). Modify then mutates only the local loop variable `c` (a value copy pulled out of that freshly-built slice) and returns it — it writes nothing back into `m` (no side map keyed by resourceID for IpamScopeID/ManagementState overrides), and it even takes `m.mu.RLock()` (a read lock) confirming no mutation of shared state was intended. The very next call to GetIpamResourceCidrs or GetIpamAddressHistory (which both also call ipamResourceCidrs() fresh) or the AWS/IPAM scope-level metrics (Managed/UnmanagedResourceCidrs, Compliant/NoncompliantResourceCidrs) will show the resource back in its default managed/compliant state, silently discarding the requested scope move or unmonitor action. The PR's own body claims stateful, in-use-guarded behavior for these APIs; this one echoes success but doesn't stick.

Satyam-Trivedi-ZS (changes requested):
## 🔭 Review — AWS Networking (Stage B): VPC IPAM (Full Parity + Metrics)

> 🟠 **Changes requested** — IPAM pool allocation, resource-CIDR modification, and metrics filtering all skip a persistence/derivation step, so callers get silently wrong data instead of an error.

`🔴 3 Blocker` · `🟠 2 Medium` · `🔵 0 Question` · `⚪ 0 Nit` · 29 files (+5003/-12)

**Ready to merge:** ❌ not yet · **Confidence:** Medium _(large diff — spot-verified, not exhaustively)_ · **Risk:** High

This PR lands broad, well-organized IPAM parity (pools, scopes, discoveries, BYOASN, resolver targets, metrics) and the emulator boots and passes its existing suites cleanly. The common thread across the blockers is the same shape of bug repeated across surfaces: a value is accepted from the caller but never actually derived or persisted, so a subsequent read (metrics, Describe, Get) silently returns stale or wrong state instead of failing. Fixing those persistence/derivation gaps — and re-scoping the CloudWatch namespace filter — should be enough to merge.

**✅ Solid**
- Broad, consistent IPAM surface (pools, scopes, discoveries, BYOASN, resolver targets) added without touching the portable `Networking` interface — kept as an AWS-only capability, consistent with the Stage-A precedent.
- Race-clean: `go test -race` passes across `providers/aws/vpc` and all 30 `server/aws/...` packages, and `go vet`/`go build` are clean.
- IPAM metrics wiring is genuinely new functionality, not a stub — it's just insufficiently scoped (see architecture note).

**🔴 Blockers (3)**
- `server/aws/cloudwatch/ops.go:215` — `ListMetrics` with an empty namespace routes into the IPAM shortcut and returns early, so once IPAM is wired in a "list all metrics" call silently drops every non-IPAM namespace. Scope the shortcut to an exact `IpamMetricNamespace` match, or merge both metric sets when namespace is empty.
- `providers/aws/vpc/ipam.go:355` (also `AllocateIpamPoolCidr` ~:396/:430) — `ProvisionIpamPoolCidr`/`AllocateIpamPoolCidr` accept NetmaskLength-only (the real AWS pattern) but store/return `CIDR: cfg.CIDR` verbatim → empty CIDR; downstream `cidrSize("")→0` permanently corrupts the `AWS/IPAM` PercentAllocated/Available metrics. Derive a free CIDR of the requested size from the pool's provisioned space when CIDR is empty.
- `providers/aws/vpc/ipam_resources.go:122` — `ModifyIpamResourceCidr` mutates a value copy from a freshly-rebuilt slice under an `RLock` and persists nothing, so the scope-move/unmonitor silently reverts on the next Get/Describe/metrics read.
**🟠 Medium & 🔵 questions**
- 🟠 `providers/aws/vpc/ipam_resolver.go:101` — CreateIpamPrefixListResolverTarget doesn't validate prefixListID exists.
- 🟠 `providers/aws/vpc/ipam_byoip.go:91` — AssociateIpamByoasn/DisassociateIpamByoasn succeed silently when the target CIDR isn't provisioned.

**🏛 Follows ARCHITECTURE.md**
- This PR touches only `providers/aws/vpc`, `server/aws/ec2`, `server/aws/cloudwatch`, and `services/networking/driver` — consistent with the repo's Stage-A pattern of exposing new AWS-only surface as an optional capability type-asserted on the vpc driver rather than widening the portable Networking interface. `docs/architecture.md` doesn't spell out an explicit optional-capability convention in writing, so this is inferred from the Stage-A precedent (PR #317) rather than a written contract.
- ⚠️ Diverges — IPAM objects (scopes, pools, allocations, discoveries) are stored in flat maps alongside VPCs/subnets/EIPs with no per-IPAM ownership link; metrics like TotalActiveIpCount/VpcIPUsage are computed globally across all VPCs rather than scoped to the VPCs actually associated with a given IPAM's resource discovery (see inline `server/aws/cloudwatch/ops.go:215`). Worth confirming whether the single-account emulator model intentionally elides this scoping or whether multi-IPAM test scenarios would expose incorrect per-IPAM metric values.

**🔍 Load-bearing claims**
- ✅ **solves the linked issue (full IPAM parity + metrics), no scope creep** — adversarially attacked, no counterexample found.
- ❌ **Allocate/ProvisionIpamPoolCidr derive a CIDR from NetmaskLength alone** — contradicted; NetmaskLength-only calls never produce a stored CIDR.
- ❌ **ModifyIpamResourceCidr's scope-move/unmonitor persists** — contradicted; a later Get/Describe reverts to the default state.

**🧪 Verified at runtime**
- ✅ `providers/aws/vpc` — PASS (race, 1.8s)
- ✅ `server/aws/...` (30 packages incl. ec2, cloudwatch) — PASS (race, cloudwatch has no test files, not a failure)
- ✅ `go build ./...` / `go vet` — clean

**🧪 Testing**
- No missing/weak-test gaps surfaced (test findings never block a merge — none were raised).

📉 +4 lower-priority item(s) not posted — re-run with postPolicy:"all" to see them.

—
_Reviewed at `11d58787faa5e1f388bd02635e5a8e70bc8790ae` · CI: passing_

server/aws/cloudwatch/ops.go
thzgajendra: Thanks — fixed in e2ef707. An exact `AWS/IPAM` request stays IPAM-only; an empty ("list all") namespace now returns every real metric **with its true namespace** and merges the AWS/IPAM set. Since `Monitoring.ListMetrics` only lists names within one namespace, I added an AWS-local `ListMetricsDetailed` capability (type-asserted, not on the shared interface) so list-all carries correct namespaces. Regression added to `TestIPAMMetricsSDK` (asserts both `MyApp` and `AWS/IPAM` appear).

providers/aws/vpc/ipam.go
thzgajendra: Thanks — fixed in e2ef707. Netmask-only Provision/Allocate now derive a concrete, aligned CIDR (sequential carve) from the pool's provisioned supply (allocate) or a default private base (provision), skipping used ranges — never storing an empty CIDR. If no block of the requested size is free, it returns InvalidArgument instead of silently succeeding. This keeps `cidrSize`/PercentAllocated correct. New assertions in `TestIPAMFullProvider` verify the derived CIDR is valid, the right size, and non-overlapping.

providers/aws/vpc/ipam_resources.go
thzgajendra: Thanks — fixed in e2ef707. ModifyIpamResourceCidr now takes a write lock and persists the scope-move/unmonitor in a per-resourceID override map that `ipamResourceCidrs()` applies, so subsequent Get/Describe/metrics reads reflect the change instead of reverting. Regression added (unmonitor, then re-read asserts ManagementState=unmanaged).

providers/aws/vpc/ipam_resolver.go
thzgajendra: Thanks — fixed in e2ef707. CreateIpamPrefixListResolverTarget now validates `prefixListID` against the managed-prefix-list store and returns InvalidArgument if absent, mirroring the resolver's ipamID check. Negative case added to the test.

providers/aws/vpc/ipam_byoip.go
thzgajendra: Thanks — fixed in e2ef707. AssociateIpamByoasn now rejects a CIDR that isn't provisioned (InvalidArgument), and DisassociateIpamByoasn now validates the ASN. Both the provisioned-associate success and the two negative paths are covered in the test.

server/aws/ec2/ipam_policy.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+func (*Handler) modifyIpamPolicyAllocationRules(w http.ResponseWriter, r *http.Request, ip netdriver.IPAMPolicy) {
+	err := ip.ModifyIpamPolicyAllocationRules(r.Context(), r.Form.Get("IpamPolicyId"), awsquery.ListStrings(r.Form, "AddAllocationRule"))
+	if err != nil {
+		writeIPAMErr(w, err)
+		return
+	}
+
+	writeReturnTrue(w, "ModifyIpamPolicyAllocationRulesResponse")
```
NitinKumar004: **Medium (wire) — wrong response element.** This emits `<return>true</return>`, but in `ec2@v1.317.1` the `ModifyIpamPolicyAllocationRulesOutput` type has a single field `IpamPolicyDocument *types.IpamPolicyDocument` and its deserializer reads the element `ipamPolicyDocument` — there is no `Return` field. A real aws-sdk-go-v2 caller gets **`out.IpamPolicyDocument == nil`** every time (no error, since unknown elements are ignored). Emit an `<ipamPolicyDocument>` element carrying the modified policy document instead of `writeReturnTrue`.

server/aws/ec2/ipam_policy.go
```diff
@@ -0,0 +1,254 @@
...
+//nolint:dupl // parallel single-field list marshaling
+func (*Handler) getIpamPolicyAllocationRules(w http.ResponseWriter, r *http.Request, ip netdriver.IPAMPolicy) {
+	rules, err := ip.GetIpamPolicyAllocationRules(r.Context(), r.Form.Get("IpamPolicyId"))
+	if err != nil {
+		writeIPAMErr(w, err)
+		return
+	}
+
+	type docXML struct {
+		Document string `xml:"document"`
```
NitinKumar004: **Medium (wire) — payload silently dropped.** The `ipamPolicyDocumentSet>item` wrapper is correct, but the inner `<document>` element is wrong: `types.IpamPolicyDocument` has members `allocationRuleSet` (`AllocationRules []IpamPolicyAllocationRule`), `ipamPolicyId`, `locale`, and `resourceType` — no `document`. So each returned item deserializes to an **empty** `IpamPolicyDocument`. Note this isn't just a tag fix: the driver method returns `[]string` (`services/networking/driver/aws_ipam_policy.go:31`), which structurally can't carry the real shape — the driver model needs to return the rule fields. Also add `nextToken` to match the output type.

providers/aws/vpc/ipam_cidr.go
```diff
@@ -0,0 +1,147 @@
...
+
+	return "", false
+}
+
+// deriveProvisionCIDR synthesizes a /netmask block for a top-level pool from
+// the default private base, skipping CIDRs already provisioned into the pool.
+// Caller holds mu.
+func (m *Mock) deriveProvisionCIDR(poolID string, netmask int) (string, bool) {
+	s, e, _ := ipv4Range(defaultIPv4Base)
+	used := usedRanges(m.poolProvisionedCIDRs(poolID))
```
NitinKumar004: **Medium (correctness) — cross-pool CIDR overlap.** `used` is built from only *this* pool's provisioned CIDRs, but the carve source is the shared `defaultIPv4Base`. Two different top-level pools each calling `ProvisionIpamPoolCidr(netmask=16)` both receive `10.0.0.0/16` — overlapping supply, which defeats IPAM's core non-overlap guarantee. `used` should exclude CIDRs already provisioned by *any* pool drawing from the same base (scan all pools in the base, not just `poolID`). No test exercises >1 pool, so it slipped through.

NitinKumar004 (changes requested):
## Deep review — AWS IPAM (Stage B) — request changes

Reviewed the whole PR at `e2ef707` (isolated worktree, whole-repo `go test`, `-race`, four adversarial lenses, and element-name verification against the vendored `aws-sdk-go-v2/service/ec2@v1.317.1` deserializers). **The core of this PR is genuinely strong** — clean dispatch, solid concurrency, and the core IPAM lifecycle round-trips against the real SDK. I'm requesting changes for two confirmed wire-shape bugs that return silently-wrong results to a real SDK caller, plus a CIDR-overlap correctness gap — all in the untested subset.

### Gates — all green
build / vet / gofmt / `go mod tidy` clean · `go test ./...` **224 ok, 0 fail** (no cascade) · `-race` on `providers/aws/vpc` clean. golangci-lint: 2 trivial PR-introduced Lows (below); the rest is pre-existing baseline.

### What's solid (verified, not just claimed)
- **Dispatch**: 69 distinct actions, **zero duplicate `case`**, zero collision with VPC/EC2 actions, nil-safe capability gating, correct registration order; `Byoasn`/`Token` sub-routes are reachable (not dead).
- **SDK end-to-end**: `TestEC2IPAMParitySDK`, `TestEC2IPAMFullSDK` (all 5 groups), `TestIPAMMetricsSDK`, `TestIPAMConcurrentAccess` all pass.
- **Concurrency**: single-RWMutex discipline is correct — no field mutation under RLock, no re-entrant deadlock, the round-1 `ipamResourceOverrides` map is symmetric-locked, no guarded-state pointer escape.
- **Metrics safety**: every percentage division is guarded (no NaN/Inf on the wire); the optional `IPAMMetrics` source is nil-safe on every path.
- **Determinism**: no `time.Now()`/`rand` in IPAM logic; **error envelopes** use proper query-XML codes; **mirror rule** legitimately scoped (IPAM is AWS-only, type-asserted per the #317 pattern); **docs** synced.

### Blockers (Medium — should fix before merge)
1. **`ModifyIpamPolicyAllocationRules` returns the wrong element.** Emits `<return>true</return>`, but the real output type is `IpamPolicyDocument` (deserializer reads `ipamPolicyDocument`). A real SDK caller **always gets `out.IpamPolicyDocument == nil`**. (inline)
2. **`GetIpamPolicyAllocationRules` silently drops its payload.** Emits `<item><document>…`, but `types.IpamPolicyDocument` has `AllocationRules`/`IpamPolicyId`/`Locale`/`ResourceType` and **no `document` member** → every returned item deserializes **empty**. The driver signature `[]string` (`services/networking/driver/aws_ipam_policy.go:31`) structurally can't carry the real shape, so this needs a driver-model change, not just a tag fix. (inline)
3. **Cross-pool CIDR overlap on netmask-only provision.** `deriveProvisionCIDR` computes `used` from only the *current* pool's CIDRs against the shared base, so two top-level pools each doing `ProvisionIpamPoolCidr(netmask=16)` **both get `10.0.0.0/16`** — IPAM's entire purpose is non-overlap. (inline)
4. **Test coverage / claim accuracy.** The two SDK tests round-trip only **~29 of 69 ops**; ~40 (all `Modify*`, resolver rules/versions/entries, resolver targets, token Describe/Delete, and both policy ops above) have **zero SDK coverage**. That gap is exactly what let #1–#2 through, and it disproves the PR body's "every response element name verified against vendored deserializers." Extending the SDK round-trips to the Modify/policy/resolver-detail ops would catch these and guard the rest.

### Low (non-blocking)
- IPv6 pool derivation is IPv4-only (`ipam_cidr.go:96-121`) — IPv6 netmask-only allocate always errors; a small-netmask v6 provision returns an IPv4 CIDR (realistic /48–/64 just error). No IPv6 test.
- Nondeterministic dimension order — `server/aws/cloudwatch/ops.go:300` builds `Dimensions` from a map without sorting (repo sorts elsewhere via `SortedValues`); dimensions are a set and no test asserts order, so low impact, but inconsistent with the determinism discipline.
- Duplicate ListMetrics rows if a client stores a real `AWS/IPAM` metric via `PutMetricData` (`ops.go:232`).
- Metrics fidelity: `ipamTopLevelMetrics`/`ipamScopeMetrics` assign one global aggregate to every IPAM/scope.
- Lint hygiene (PR-introduced, CLAUDE.md local 0-issue gate): unused `//nolint:gosec` at `ipam_cidr.go:56`; `prealloc` at `ipam_metrics.go:19`.

Net: excellent groundwork for Stage B — the wiring, locking, and core parity are solid. Fixing the two policy-rules wire shapes (with the driver-model change #2 needs), closing the netmask-only cross-pool overlap, and widening SDK coverage to the untested ops would get this to a clean approve.

server/aws/ec2/ipam_policy.go
thzgajendra: Fixed in a7046e4. The response now emits an `<ipamPolicyDocument>` element (matching `ModifyIpamPolicyAllocationRulesOutput.IpamPolicyDocument`) carrying `allocationRuleSet>item>sourceIpamPoolId` + `ipamPolicyId`/`locale`/`resourceType`, instead of `<return>true>`. A real aws-sdk-go-v2 caller now gets a non-nil `IpamPolicyDocument`. Verified with an SDK round-trip test asserting the returned document + rule.

server/aws/ec2/ipam_policy.go
thzgajendra: Fixed in a7046e4. You're right that `[]string` couldn't carry the shape — the driver model now stores structured rules: `IpamPolicy.AllocationRules []IpamAllocationRule{SourceIpamPoolID}` plus `Locale`/`ResourceType`. The response emits `ipamPolicyDocumentSet>item` with `allocationRuleSet>item>sourceIpamPoolId` + `ipamPolicyId` and a top-level `nextToken`; the request parses `AllocationRule.N.SourceIpamPoolId` + `Locale` + `ResourceType`. SDK test asserts `IpamPolicyDocuments[0].AllocationRules[0].SourceIpamPoolId` and `IpamPolicyId` round-trip.

providers/aws/vpc/ipam_cidr.go
thzgajendra: Fixed in a7046e4. `deriveProvisionCIDR` now builds `used` from **all** pools' provisioned CIDRs (new `allProvisionedCIDRs` helper), not just the current pool, so two top-level pools each asking for a netmask-only /16 get distinct, non-overlapping blocks. Added a two-pool regression test asserting no overlap.

NitinKumar004 (approved):
## Deep re-review — AWS IPAM (Stage B) — approve

Re-reviewed at `f92f7f5`, the round addressing my request-changes (`a7046e4` + the cloudwatch `New` refactor + a development merge). **All three blockers are fixed, backed by real aws-sdk-go-v2 regression tests, and independently re-verified against the vendored `ec2@v1.317.1` structs.** Exhaustive end-to-end matrix is green and two fresh adversarial passes over the new code found nothing. Approving.

### Test matrix — all green
| Pass | Result |
|---|---|
| build / vet / gofmt / `go mod tidy` | clean |
| `go test ./...` (whole repo) | 225 ok, 0 fail |
| `go test -race ./...` (whole repo, clean run) | 225 ok, 0 data races, 0 fail |
| `-count=3` flakiness (IPAM/cloudwatch) | no flakes (deterministic) |
| e2e boot smoke (standalone binary) | builds, AWS serves 200, `/_cloudemu/reset` 200 |
| coverage on fix code | `deriveProvisionCIDR` / `allProvisionedCIDRs` 100% |

### Blockers — all resolved + SDK-verified
1. **`ModifyIpamPolicyAllocationRules`** now emits `<ipamPolicyDocument>` (the SDK output field; deserializer reads `ipamPolicyDocument`) with the full `ipamPolicyId`/`locale`/`resourceType`/`allocationRuleSet` shape — not the dropped `<return>true</return>`. Test asserts the non-nil document + `SourceIpamPoolId`.
2. **`GetIpamPolicyAllocationRules`** — the driver model was widened (`AllocationRules []IpamAllocationRule{SourceIpamPoolID}` + `Locale`/`ResourceType`, methods return `*IpamPolicy`) so it can carry the real shape; the wire now emits `ipamPolicyDocumentSet>item` with the correct members + `nextToken`. Test asserts the create→modify→get round-trip returns the rule and policy id.
3. **Cross-pool CIDR overlap** — `deriveProvisionCIDR` now carves against `allProvisionedCIDRs()` (every pool) instead of just the current pool; `TestIPAMProvisionNoCrossPoolOverlap` asserts two netmask-only /16 provisions get distinct, non-overlapping blocks.

Adversarial verification of the new code came back clean on both lenses: the policy allocation-rule flow persists correctly (pointer-store, in-place mutation under `Lock`; Get returns a deep clone under `RLock`; missing→404, empty→valid doc; `parseAllocationRules` survives empty/non-contiguous/missing-field inputs), and the cross-pool CIDR fix holds for out-of-base CIDRs, exhaustion (clean `InvalidArgument`), deprovision reuse, allocation-scoping (still pool-scoped), and concurrency. The cloudwatch `New` refactor (single-arg + `SetIPAMMetrics` setter) is clean and correctly wired at `aws.go:224` (nil-safe).

### Optional follow-ups (all Low, non-blocking)
- Lint hygiene: `ipam_cidr.go:56` unused `//nolint:gosec`; `ipam_metrics.go:19` prealloc; a policy/resolver `dupl`.
- `server/aws/cloudwatch/ops.go:300` builds metric `Dimensions` from a map without sorting (repo sorts elsewhere via `SortedValues`) — harmless (dimensions are a set) but inconsistent with the determinism discipline.
- IPv6 pool derivation is still IPv4-only.
- `ModifyIpamPolicyAllocationRules` doesn't validate `SourceIpamPoolId` refers to a real pool (real AWS 400s; the emulator faithfully stores what's sent — consistent with the codebase's lenient references).
- Provider-package coverage of the policy methods is 0% (they're exercised via the server SDK roundtrip, matching the surrounding policy methods' convention).

Solid iteration — the wire shapes, driver model, and cross-pool carve are all correct now, with real regression coverage. LGTM.

Issue #323, opened 2026-08-05
Azure Resource Graph: emit fields needed for full downstream cost-discovery E2E (disk IOPS/throughput, SQL DB databases, AKS agent-pool attrs, VMSS)


NitinKumar004:
## Context

A downstream project drives its **real** Azure discoverer (real `armresourcegraph` SDK) against cloudemu's Resource Graph endpoint to E2E-verify cost calculation. PR #316 (`feat/315-arg-resource-fields`) added the `properties`/`sku` projection, which unblocked disk `diskSizeGB`+`sku.name`, VM SKU, and SQL MI `storageSizeInGB`. 🎉

To cover the remaining cost cases, the Azure ARG emulation needs to be able to **carry a few more `properties.*` fields**. Today they can't be seeded because the per-type walkers don't populate them (the `properties` bag is open in type but closed in population).

### 1. Managed disk — provisioned IOPS / throughput
- `properties.diskIOPSReadWrite`, `properties.diskMBpsReadWrite`
- The wire walker (`services/resourcediscovery/walkers.go` `walkVolumes`) **already** projects `putInt("diskIOPSReadWrite", v.IOPS)` / `("diskMBpsReadWrite", v.Throughput)` — but `VolumeConfig` (the `CreateVolume` input, `services/compute/driver/driver.go`) has **no IOPS/Throughput fields**, so they're always 0 and dropped. Needed to price Premium SSD v2 / Ultra disks (capacity + IOPS + throughput).
- **Fix:** add `IOPS`, `Throughput` to `VolumeConfig` and carry them into the volume model.

### 2. SQL Managed Instance — `storageAccountType`
- `properties.storageAccountType` (backup redundancy: `GeoRedundant`/`ZoneRedundant`/…). `storageSizeInGB` already works via `ManagedInstanceConfig.StorageGB`. Add `StorageAccountType` to the MI config + projection.

### 3. SQL Database (`microsoft.sql/servers/databases`) — not discovered at all
- No walker enumerates individual databases (only logical servers / flex / MI). Needed: `properties.zoneRedundant` (bool), `properties.currentSku`, `sku.name`.
- **Fix:** a `servers/databases` walker + `portableToAzureTypeMap` entry.

### 4. AKS agent pool — sku + properties
[...]

Issue #315, opened 2026-08-03
Azure Resource Graph: emit resource-type-specific fields (VM SKU, managed disks, DB tier) for real-SDK discovery/cost E2E


NitinKumar004:
## Summary

The Azure Resource Graph emulator (`server/azure/resourcegraph` + `resourcediscovery`) returns a **generic cross-cloud inventory** rather than Azure-type-specific ARG rows. A real `azure-sdk-for-go` consumer that reads resource-type-specific fields (VM SKU, disk SKU/size, standalone managed disks, DB tier, etc.) cannot drive an end-to-end discovery/cost test against cloudemu today, because those fields aren't projected into the ARG response.

This blocks a real use case: pointing a production Azure **discoverer + cost engine** at cloudemu to verify SKU/tier/size-sensitive logic with **no real account** — which is otherwise cloudemu's strongest value proposition.

## Evidence (real-SDK round trip)

Driving the real `armresourcegraph` client against `httptest.NewTLSServer(azureserver.New(...))` with a VM and a `CreateVolume` disk seeded:

```go
cloudP := cloudemu.NewAzure()
cloudP.VirtualMachines.RunInstances(ctx, computedriver.InstanceConfig{InstanceType: "Standard_D2s_v3"}, 1)
cloudP.VirtualMachines.CreateVolume(ctx, computedriver.VolumeConfig{Size: 4, VolumeType: "Premium_LRS"})
// ...azureserver.New(Drivers{VirtualMachines, Disks, ResourceDiscovery, SubscriptionID})...
```

Querying `Resources | where type in~ (...)` via the real SDK returns:

| Expected (real Azure ARG) | cloudemu today |
|---|---|
| VM row with `sku.name` = `Standard_D2s_v3` (or `properties.hardwareProfile.vmSize`) | VM row present, but **SKU field empty** — consumer reads `InstanceType = ""` |
| `microsoft.compute/disks` row with `sku.name` + `properties.diskSizeGB` + `managedBy` | **No disk row emitted at all** (`CreateVolume` volume not surfaced in ARG) |

## Requested fidelity (fields real Azure ARG returns; needed for discovery/cost E2E)

**1. Compute VMs** — project into the ARG row:
- `sku.name` **or** `properties.hardwareProfile.vmSize` = the VM size (e.g. `Standard_D2s_v3`)
- `properties.storageProfile.osDisk.osType` (Linux/Windows)
- `properties.priority` = `Spot` (interruptible pricing)
[...]

PR #327, opened 2026-08-05
feat(resourcediscovery): Azure Resource Graph cost-field projection across all supported types (#323)

NitinKumar004:
## Summary

Extends the Azure Resource Graph emulation (built in #315/#316) to project the **`properties` and `sku` cost fields a real `armresourcegraph` discoverer prices on**, across every supported Azure resource type. Addresses #323.

A downstream project drives its real Azure discoverer against cloudemu's Resource Graph endpoint to E2E-verify cost calculation. Previously only disk size, VM SKU, and SQL-MI storage flowed through. This PR closes the field gap so disk IOPS/throughput, VM/VMSS Spot + Hybrid Benefit, SQL zone-redundancy, AKS Spot node pools, storage redundancy, Cosmos serverless, App Service plan tier, and more all round-trip over the real SDK.

## What's projected now

| Type | Fields |
|---|---|
| `microsoft.compute/virtualmachines` | `properties.priority` (Spot), `licenseType`, `osType`, `sku.name`, `zones` |
| `microsoft.compute/disks` | `properties.diskIOPSReadWrite`, `diskMBpsReadWrite`, `diskSizeGB`, `tier`, `sku.name`/`sku.tier` |
| `microsoft.compute/virtualmachinescalesets` *(new)* | `sku.name`/`sku.capacity`, nested `properties.virtualMachineProfile.{priority,licenseType,storageProfile.osDisk.osType}` |
| `microsoft.network/publicipaddresses` | `sku.name` (Basic/Standard), `properties.publicIPAllocationMethod` |
| `microsoft.network/virtualnetworks` / `subnets` | `properties.addressSpace.addressPrefixes` / `addressPrefix` |
| `microsoft.storage/storageaccounts` | `sku.name` (redundancy), `kind`, `properties.accessTier` |
| `microsoft.documentdb/databaseaccounts` | `kind`, `properties.databaseAccountOfferType`, `capabilities` (serverless), `enableFreeTier` |
| `microsoft.sql/managedinstances` | `properties.storageSizeInGB`, `storageAccountType` |
| `microsoft.sql/servers/databases` *(new)* | `sku.name`, `properties.currentSku`, `zoneRedundant` |
| `microsoft.dbformysql`/`dbforpostgresql/flexibleservers` | `sku.name`/`sku.tier`, `properties.version`, nested `storage.storageSizeGB` + `highAvailability.mode` |
| `microsoft.sql/servers` | `properties.version` |
| `microsoft.containerservice/managedclusters` | `sku.tier`, `properties.powerState.code`, `kubernetesVersion` |
| `.../managedclusters/agentpools` | `sku.name` (vmSize), `properties.scaleSetPriority` (Spot), `count`, `mode`/`osType` |
| `microsoft.databricks/workspaces` | `sku.name`/`sku.tier`, `properties.workspaceId`/`provisioningState` |
| `microsoft.web/serverfarms` *(new)* | `sku.name`/`sku.tier`/`sku.capacity`, `kind` |

## How it's built (per the existing architecture)

- **Foundation:** added `SKUTier`/`SKUCapacity` slots to the generic `Attributes`/`Resource` set so `resourceToWire` can render `sku.tier`/`sku.capacity`; fixed two type-map correctness bugs (NetworkInterface and ElasticIP emitted the wrong Azure type strings).
- **Additive field seeding:** widened the portable driver configs where the field can be varied (`VolumeConfig.IOPS/Throughput/Tier`, `InstanceConfig.OSType/Priority/LicenseType/Zones`, `ManagedInstanceConfig.StorageAccountType`, `ElasticIPConfig.SKU/AllocationMethod`, AKS `Tier`/`ScaleSetPriority`), and the walkers/adapters project them.
- **Optional type-asserted capabilities** (the `NetworkInterfaces` pattern) for enrichment that doesn't fit the portable config: `BucketAttributes` (storage account sku/kind/accessTier) and `TableAttributes` (Cosmos account kind/offer/capabilities/free-tier). S3/GCS/DynamoDB don't implement them and contribute nothing.
- **New provider-projected discovery capabilities** (the relational-DB / Kubernetes walker pattern) for the net-new resources: `ScaleSets` (VMSS), `AppServicePlans` (serverfarms), and the `Databases` CRUD capability implemented on Azure SQL for logical databases.
- The `NodeGroups []string` cluster projection became `[]DiscoveredNodeGroup{Name, Attrs}` so agent pools can carry per-pool attributes; the EKS/GKE/AKS adapters were all updated (mirror rule).

All additions are additive and nil-safe — AWS/GCP engines skip the Azure-only walkers, and unimplemented optional capabilities are silently absent.

## Tests

- **Real-SDK ARG round-trips** (`server/azure/resourcegraph/arg_cost_*_test.go`) drive the live `armresourcegraph` client against an httptest server and assert every projected `properties`/`sku` field per type.
- **Provider-unit tests** for each new mock field / capability (VMSS, App Service plan, SQL databases, storage/Cosmos attributes, AKS agent-pool priority, public-IP sku).
- `go build ./...`, `go vet ./...`, `gofmt`, full `go test ./...` (225 packages) and `golangci-lint` on the changed files are all green.

## Docs

`docs/services.md` §19 gains the full ARG cost-field projection table and a note on the capability patterns used.


providers/azure/azure.go
```diff
@@ -197,7 +250,30 @@ func (d sqlDiscovery) DiscoverDatabases(
...
+
+		for j := range dbs {
+			db := &dbs[j]
+			out = append(out, resourcediscovery.DiscoveredDatabase{
+				Name: db.Name, Type: resourcediscovery.TypeSQLDatabase, ARN: db.ARN,
+				Attrs: resourcediscovery.Attributes{
+					SKU:     db.SKUName,
+					SKUTier: db.SKUTier,
+					Properties: nonEmptyProps(map[string]any{
+						"zoneRedundant": db.ZoneRedundant,
```
thzgajendra, line 271: `nonEmptyProps` (line 442) only strips empty `string` and `int` — `bool` and nested `map[string]any` fall through its type switch untouched. So `zoneRedundant: false` and the `currentSku` map are emitted for **every** SQL database, including ones created without zone-redundancy or an explicit SKU.

Azure SQL genuinely surfaces `properties.zoneRedundant` as a bool, so emitting `false` may actually be faithful to what the real `armresourcegraph` discoverer returns — in which case the inconsistency is the *other* way (pruning empty strings/ints that real ARG might include). Either way the helper's doc says "drops zero-valued entries" but its behavior only does so for two of the types it receives.

Worth a quick check against the live discoverer (fidelity is the point of this PR): if defaults should be omitted, add `case bool { if !val { delete } }` + empty-map handling; if ARG-faithfulness is intended, the string/int pruning of real properties is what needs a second look.

services/resourcediscovery/walkers.go
```diff
@@ -272,12 +303,26 @@ func (e *Engine) walkStorage(ctx context.Context) ([]Resource, error) {
...
 			ID:     b.Name,
 			ARN:    e.storageBucketARN(b.Name),
 			Region: region, Tags: tags,
-		})
+		}
+
+		// Optional capability: providers whose buckets carry storage-account
+		// attributes (Azure) project SKU/kind/access-tier for cost discovery.
+		if attrer, ok := e.drivers.Storage.(storagedriver.BucketAttributes); ok {
+			if a, aErr := attrer.BucketAttributes(ctx, b.Name); aErr == nil {
```
thzgajendra: Swallowed error: if `BucketAttributes` ever returns a non-nil `aErr`, the sku/kind/accessTier are silently dropped and the storage account is emitted with **no priceable SKU** — exactly the "cost field silently absent" failure this PR is closing. Today's impl can't error, so it's latent, but the invariant is load-bearing and undiscoverable if a future/alternate impl breaks it. Suggest logging `aErr` rather than discarding it. Same pattern at line 362 for `TableAttributes`.

server/azure/resourcegraph/arg_cost_fields_test.go
```diff
@@ -0,0 +1,494 @@
...
+
+	client := argCostClient(t, cloudP)
+
+	row := queryOne(t, client, "microsoft.databricks/workspaces")
+	assert.Equal(t, "dbx-cost", row["name"])
+
+	sku := rowSKU(t, row)
+	assert.Equal(t, "premium", sku["name"])
+
+	if tier, ok := sku["tier"]; ok {
```
thzgajendra: This assertion is masked: if Databricks `sku.tier` projection regresses to absent, `ok` is false and the inner `assert` is skipped — the test still passes green. Since the input sets `SKUTier: "premium"`, make it unconditional so a broken projection actually fails:
```go
assert.Equal(t, "premium", rowSKU(t, row)["tier"])
```

thzgajendra (commented):
## Review — ARG cost-field projection (in-depth, multi-dimension pass)

Reviewed across correctness, nil-safety, type design, tests, comments/docs, simplification, and architecture-consistency, plus a manual lead pass on the ARG core.

**Overall: strong and well-architected.** The KQL parser + generic `resourceToWire` rendering (no per-type branching) are clean, the change is genuinely additive, and both advertised bug fixes are real — the base branch fell through to invalid `networking/networkinterface` / `networking/elasticip` type strings; the new `portableToAzureTypeMap` entries are correct Azure ARG types. The three claimed patterns (NetworkInterfaces optional-capability, relational-DB/K8s walker, and the NodeGroups mirror rule) are applied consistently, and AWS/GCP output is verifiably unaffected (`NodeGroupsFromNames` shim + zero-value additive fields).

**No confirmed correctness blocker.** Findings by severity:

### Should address
- **Test gap (silent-regression risk):** `microsoft.sql/servers` → `properties.version` is projected (`providers/azure/azure.go:254`) but never asserted at the ARG level — `TestARGCostFields_SQLDatabase` only queries the `.../databases` row. Add a `microsoft.sql/servers` round-trip asserting `version`. It's also missing from the `docs/services.md` §19 table.
- **Masked test assertion** (`arg_cost_fields_test.go:345`) — see inline; passes even if the projection breaks.
- **Doc error:** §19 lists VM `osType` at `properties.osType`, but the code nests it at `properties.storageProfile.osDisk.osType` (which matches real ARG). Fix the documented path. §19 also under-documents the SQL-MI row (missing `sku.name`, `vCores`, `tier`, `licenseType`).

### Consider
- `nonEmptyProps` bool / nested-map handling — see inline on `azure.go`.
- Swallowed `BucketAttributes` / `TableAttributes` errors — see inline on `walkers.go:316`.
- `flexTier` empty-tier for unlisted families — see inline.
- App Service plans bucketed under `ServiceServerless` — see inline.
- Postgres-flex tests assert fewer fields than MySQL-flex (missing `sku.tier`, `version`) — worth symmetry.

### Nits
- Stringly-typed `OSType` / `Priority` driver configs could be named enum types with consts (small closed domains).
- No compile-time `var _ BucketAttributes = (*Mock)(nil)` assertions for the optional capabilities — a typo'd method signature would fail the type-assert silently with no build error.

Nice PR — the field-flow (config → walker → adapter → wire → ARG projection) connects end-to-end for the types I traced, and the units/JSON paths are correct throughout (GB/IOPS/MBps, `sku` vs `properties`).

providers/azure/azure.go
```diff
@@ -238,6 +315,101 @@ func (d sqlDiscovery) DiscoverDatabases(
...
+	}
+
+	return out, nil
+}
+
+// flexTier derives the Azure Flexible Server SKU tier (Burstable /
+// GeneralPurpose / MemoryOptimized) from the SKU name, which encodes it as a
+// prefix in both the current ("Standard_B1ms") and legacy ("B_Gen5_1") naming.
+// Empty when the name doesn't match a known family.
+func flexTier(skuName string) string {
```
thzgajendra, line 403: `flexTier` returns `""` for SKU families outside the listed prefixes (e.g. `Standard_F*` compute-optimized), so `sku.tier` is silently dropped for those servers and no test catches it. Reasonable as best-effort (a pricing consumer can fall back to `sku.name`), but worth a short comment noting the mapping is known-incomplete, or a sensible default, so a maintainer doesn't read the empty tier as a bug later.

services/resourcediscovery/walkers.go
```diff
@@ -459,6 +544,69 @@ func (e *Engine) walkRelationalDB(ctx context.Context) ([]Resource, error) {
...
+	for i := range plans {
+		p := plans[i]
+
+		region := p.Region
+		if region == "" {
+			region = e.region
+		}
+
+		r := Resource{
+			Provider: e.provider, Service: ServiceServerless, Type: TypeAppServicePlan,
```
thzgajendra: App Service plans are bucketed under `ServiceServerless` (same discriminator as Functions). ARG type-queries still resolve correctly because `Type` (`TypeAppServicePlan` → `microsoft.web/serverfarms`) disambiguates, so this isn't a functional bug. But serverfarms aren't serverless — following the `ServiceRelationalDB` precedent, a dedicated service constant for net-new families would be semantically cleaner if anything ever groups by `Service`.

NitinKumar004:
Thanks for the thorough pass — all findings addressed in `add1050`. Gate: `go build ./...`, `go vet`, `gofmt`, and `go test ./...` all clean.

### Should address
- **Masked Databricks assertion** — now unconditional (`assert.Equal(t, "premium", rowSKU(t, row)["tier"])`), so a dropped `sku.tier` projection fails.
- **`microsoft.sql/servers` `version` test gap** — added `TestARGCostFields_SQLServer` (create → ARG query → assert `properties.version`), plus the §19 row.
- **§19 doc fixes** — VM `osType` corrected to `properties.storageProfile.osDisk.osType`; SQL-MI row expanded. Note: SQL-MI tier surfaces at `properties.tier` (not `sku.tier`) — the mock never sets `Attrs.SKUTier` for that type, so the doc + test reflect the actual path.

### Consider
- **`nonEmptyProps`** now prunes empty/nested-empty maps and nil (kills the empty `currentSku` map); bools including `false` are intentionally kept as faithful to real ARG, and the doc comment now states exactly that.
- **Swallowed `BucketAttributes`/`TableAttributes` errors** — the `walk*` funcs return `error`, so both sites now propagate rather than discard.
- **`flexTier`** — added a comment that the prefix list is known-incomplete and `""` is an intentional best-effort fallback (`sku.name` covers the consumer).
- **App Service plans** — moved to a dedicated `ServiceAppService` discriminator (following the `ServiceRelationalDB` precedent); the emitted wire type is unchanged (`microsoft.web/serverfarms`).
- **Postgres-flex** tests brought to parity with MySQL-flex (`sku.tier`, `version`).

### Nits
- Added compile-time `var _ driver.BucketAttributes/TableAttributes = (*Mock)(nil)` assertions so a signature typo fails the build instead of the runtime type-assert.

thzgajendra (commented):
## Re-review — fixes in `add1050fd` verified ✅

Re-reviewed the `fix(resourcediscovery): address ARG cost-field review findings` commit against my prior review. **Every finding addressed correctly.** Built + `go vet` + tested the affected packages green locally (`resourcegraph`, `resourcediscovery`, `blobstorage`, `cosmosdb`).

| Finding | Status |
|---|---|
| Swallowed `BucketAttributes`/`TableAttributes` errors (walkers.go 316/362) | ✅ Now propagate via `fmt.Errorf`, with a "load-bearing error" comment |
| Masked Databricks `sku.tier` assertion | ✅ Now unconditional |
| `microsoft.sql/servers` → `version` untested | ✅ New `TestARGCostFields_SQLServer` + §19 doc row |
| Postgres-flex test asymmetry | ✅ New `TestARGCostFields_PostgresFlex` (sku.tier/version/storage/HA) |
| Doc: VM `osType` path + SQL-MI under-documented | ✅ §19 corrected; MI `sku.name`/`vCores`/`tier`/`licenseType` now also asserted in the test |
| `nonEmptyProps` bool / nested-map | ✅ Recursive-map + nil pruning added; `false` bools intentionally preserved (faithful to real ARG) and now documented — the right call |
| App Service plans bucketed under `ServiceServerless` | ✅ Dedicated `ServiceAppService` / `portableAppService` applied across kql.go + handler.go + walkers.go |
| `flexTier` empty tier for unlisted families (nit) | ✅ Documented as intentional best-effort |
| Compile-time capability assertions (nit) | ✅ `var _ BucketAttributes/TableAttributes = (*Mock)(nil)` added to both mocks |

The one change with regression risk — renaming the App Service plan service discriminator — is sound: walker dispatch is driver-gated with post-filter `q.matches`, the query-parse and emit sides both use `appservice` consistently, and `TestARGCostFields_AppServicePlan` (`microsoft.web/serverfarms` round-trip) passes.

No new issues found. Clean, thorough turnaround — LGTM from my side (leaving the merge call to you).

thzgajendra (approved):

PR #331, opened 2026-08-05
Add AWS VPC Lattice SDK-Compat Parity (73 Operations)

thzgajendra:
## Objective
Add full `aws-sdk-go-v2/service/vpclattice` SDK-compat parity to the emulator: **all 73 operations, no stubs**, so real VPC Lattice clients work end-to-end against the in-memory driver.

## What we found
The emulator had no VPC Lattice service, and — more significantly — **no REST-JSON service** at all. Every existing AWS service uses either AWS JSON 1.1 (`X-Amz-Target` header) or `awsquery`. VPC Lattice speaks **REST-JSON** (`awsRestjson1`): operations are selected by HTTP method + URL path (`POST /services`, `GET /services/{id}/listeners/{id}`, `PATCH /servicenetworks/{id}`, …), with URI path parameters that may be bare IDs or full ARNs.

**Blast radius:** purely additive. New packages under `services/vpclattice/`, `providers/aws/vpclattice/`, `server/aws/vpclattice/`, wired into the existing provider and server bundles. No existing service is touched.

## How we fixed it
Standard 4-layer pattern (driver → provider → server), following the Bedrock precedent for path-based REST routing:
- **Routing:** the handler builds a `map[string]segmentHandler` in `New()`; `Matches` gates on path-root **+ method + segment shape** (not just the first segment) so path-style S3 requests on like-named buckets fall through to the S3 catch-all, and `ServeHTTP` dispatches by first segment + method. Reusable `routeCollection`/`routeByID` helpers keep each resource group short. Identifiers accept ID-or-ARN; the three ARN-in-path surfaces (`/authpolicy`, `/resourcepolicy`, `/tags`) reconstruct the full ARN from the remaining path segments.
- **Union fidelity:** union-typed fields — a listener's `defaultAction`, a rule's `match`/`action`, a target group's `config`, a resource configuration's `resourceConfigurationDefinition` — are stored as raw JSON and echoed back verbatim, so any variant round-trips without modeling every shape.
- **Tags & lifecycle:** create-time tags are persisted (create-with-tags → `ListTagsForResource` round-trips); deletes block on live service-network associations (`ConflictException`) and cascade contained children (service→listeners→rules); association counts recompute on read and skip targets that no longer exist; `UpdateResourceConfiguration.AllowAssociationToShared` is `*bool` so a partial update never resets it.
- **Cross-resource behavior:** target registration dedups by id+port; `UpdateTargetGroup` merges the health-check into the stored config.
- Wire shapes (camelCase members, RFC3339 timestamps) verified against the vendored SDK serializers/deserializers.

## Alternatives not taken
- Did **not** use `net/http.ServeMux` path patterns — its single-segment `{id}` matching mishandles ARNs-with-slashes (esp. `/tags/{resourceArn}`); manual segment parsing (the Bedrock precedent) is robust to them.
- One residual routing ambiguity is unavoidable for two REST services on a single endpoint: an identical verb+path such as `GET /services` (Lattice list-services vs. S3 list-bucket-"services"). The method+shape gate removes every other collision.
- Did not model async status machines — resources are created in terminal `ACTIVE`/`PENDING` status, matching the other emulated services.

## Docs / Tests
- **Docs:** `docs/services.md` — new "## 26. Application Networking" section (per-family op table + accepted-but-not-simulated notes), master-table row 26, summary count (+73). (Coexists with the Route 53 Resolver service that landed on `development`; combined Grand Total 1707.)
- **Tests:** real-SDK round-trip lifecycle tests (one per resource group) driving the genuine `aws-sdk-go-v2` client through `httptest` — with wire-level coverage for all 73 ops and a `Matches` S3-shadow test — plus a provider unit suite covering tags-on-create, delete guards/cascade, association-count recompute, error paths, ID/ARN resolution, scoping, batch partial-failure, and clone-on-read isolation.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt` clean
- [x] `go test -race ./.../vpclattice/...` — all pass (provider 80.3%, server 68.8%)
- [x] `golangci-lint run --new-from-rev=$(git merge-base HEAD stackshy/development) ./...` — **0 issues**
- [x] `go mod tidy` stable (adds `vpclattice v1.25.5`)
- [x] Driver interface exposes exactly 73 operations (one per SDK op)

## Risk & Rollback
Low risk — additive only; no change to existing services or shared wire code. Rollback = revert this commit / drop the three new packages and their two wiring hunks.

## Conclusion
VPC Lattice reaches full 73/73 SDK-compat parity and establishes the REST-JSON routing pattern for future REST services in the emulator.


providers/aws/vpclattice/service_networks.go
```diff
@@ -0,0 +1,139 @@
...
+	sn := &driver.ServiceNetwork{
+		ID:                   id,
+		ARN:                  m.arn("servicenetwork/" + id),
+		Name:                 in.Name,
+		AuthType:             authType,
+		SharingConfigEnabled: in.SharingConfigEnabled,
+		CreatedAt:            m.now(),
+		LastUpdatedAt:        m.now(),
+	}
+	m.serviceNetworks.Set(id, sn)
```
NitinKumar004, line 68: [High] `CreateServiceNetwork` builds the record from `in.Name`/`in.AuthType`/… but never reads `in.Tags` — and **no** `Create*`/`Start*` method in the package writes to `m.tags` (repo-wide `.Tags` grep in the non-tagging sources returns nothing). So `CreateServiceNetwork(&{Name:"x", Tags:{"a":"1"}})` → `ListTagsForResource(sn.ARN)` returns `{}` instead of `{"a":"1"}`, across ~10 resource types. This breaks the standard AWS create-with-tags → list-tags flow and is untested (the tag test only exercises standalone `TagResource`). Write `in.Tags` into `m.tags` on every create.

providers/aws/vpclattice/service_networks.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	sn.LastUpdatedAt = m.now()
+
+	out := cloneServiceNetwork(sn)
+
+	return &out, nil
+}
+
+func (m *Mock) DeleteServiceNetwork(_ context.Context, identifier string) error {
```
NitinKumar004, line 118: `DeleteServiceNetwork` removes the SN even when VPC/service/resource associations still reference it — real AWS returns `ResourceInUse`. After delete, `Get*Association` still returns an association whose `ServiceNetworkID` points at a deleted network. Same class: `DeleteService` orphans its listeners/rules + SN↔service associations, `DeleteListener` orphans its rules, `DeleteTargetGroup` leaves listener/rule default-actions dangling. Either block on live dependents (`ResourceInUse`) or cascade them (as `DeleteTargetGroup` does for its targets).

providers/aws/vpclattice/service_networks.go
```diff
@@ -0,0 +1,139 @@
...
+		return nil, serviceNetworkNotFound(id)
+	}
+
+	out := cloneServiceNetwork(sn)
+	m.applyAssocCounts(&out)
+
+	return &out, nil
+}
+
+func (m *Mock) UpdateServiceNetwork(
```
NitinKumar004, line 93: `UpdateServiceNetwork` returns the record without calling `applyAssocCounts`, so its response always reports `NumberOfAssociated{Services,VPCs,Resources} = 0` regardless of real associations (Get/List do compute them). Also, `applyAssocCounts` counts associations to services that may already be deleted (see the no-cascade delete above), so a `GetServiceNetwork` can report a phantom associated service. Call `applyAssocCounts` here, and skip associations whose target no longer exists.

providers/aws/vpclattice/resourceconfigs.go
```diff
@@ -0,0 +1,121 @@
...
+
+	if in.PortRanges != nil {
+		c.PortRanges = append([]string(nil), in.PortRanges...)
+	}
+
+	if len(in.Definition) > 0 {
+		c.Definition = append([]byte(nil), in.Definition...)
+	}
+
+	c.AllowAssociationToShared = in.AllowAssociationToShared
```
NitinKumar004: `AllowAssociationToShared` is assigned unconditionally, unlike the guarded `PortRanges` (`!= nil`) and `Definition` (`len > 0`) just above. It's a plain `bool` with no "unspecified" sentinel, so a partial `UpdateResourceConfiguration` that omits the flag silently resets a previously-`true` value to `false`. Guard it (e.g. accept `*bool` in the input, or only apply when the caller set it).

server/aws/vpclattice/handler.go
```diff
@@ -0,0 +1,129 @@
...
+		"resourcepolicy":                        h.serveResourcePolicy,
+		"domainverifications":                   h.serveDomainVerifications,
+		"tags":                                  h.serveTags,
+	}
+
+	return h
+}
+
+// Matches claims requests whose first path segment belongs to VPC Lattice.
+func (h *Handler) Matches(r *http.Request) bool {
```
NitinKumar004, line 61: `Matches` claims any request whose **first path segment** is one of ~15 generic words (`services`, `tags`, `targetgroups`, `servicenetworks`, …), and this handler is registered **before** S3 (vpclattice at aws.go:287, S3's REST fallback at :438). Because aws-sdk-go-v2 uses **path-style** S3 against a custom endpoint, a path-style S3 op on a bucket literally named `services`/`tags`/etc. (`GET /tags`, `PUT /services/mykey`) is hijacked here and fails as a Lattice op — a real cross-service regression this PR introduces. Contrast Lambda, which avoids this by claiming a versioned prefix (`/2015-03-31/functions`). Consider gating on method+shape, or claiming a more distinctive path root.

server/aws/vpclattice/sdk_roundtrip_test.go
```diff
@@ -0,0 +1,620 @@
...
+
+	gotVpc, err := client.GetServiceNetworkVpcAssociation(ctx, &awsvpcl.GetServiceNetworkVpcAssociationInput{
+		ServiceNetworkVpcAssociationIdentifier: aws.String(vpcAID),
+	})
+	if err != nil || aws.ToString(gotVpc.VpcId) != "vpc-1" || len(gotVpc.SecurityGroupIds) != 2 {
+		t.Fatalf("GetSNVpcAssociation: %v %+v", err, gotVpc)
+	}
+
+	// SN ↔ Service
+	svcA, err := client.CreateServiceNetworkServiceAssociation(ctx, &awsvpcl.CreateServiceNetworkServiceAssociationInput{
```
NitinKumar004, line 429: The harness drives 70/73 ops but never asserts `client.GetServiceNetworkServiceAssociation` / `GetServiceNetworkResourceAssociation` (both are created here but only List/Delete are checked), and `DeleteResourceEndpointAssociation` is unreachable (no endpoint association is ever synthesized). The handlers exist, so the 73-op claim stands, but these Get paths have zero wire-level coverage. Cheap to close: add `Get*Association` asserts right after the creates.

providers/aws/vpclattice/access_logs.go
```diff
@@ -0,0 +1,107 @@
+package vpclattice
```
NitinKumar004, line 1: [STRUCTURE.md §3] Smashed multi-word filenames should be snake_case: `accesslogs.go` → `access_logs.go`, and likewise `service_networks.go`, `resource_gateways.go`, `resource_configs.go`, `domain_verifications.go`, `target_groups.go` — in BOTH `providers/aws/vpclattice/` and `server/aws/vpclattice/` (a feature keeps the same filename across layers). New services now go through the STRUCTURE.md loop, so worth doing here.

NitinKumar004 (changes requested):
## Review — AWS VPC Lattice (73 operations, REST-JSON)

Strong, genuinely-complete implementation of a **new wire family** (the repo's first REST-JSON / `awsRestjson1` service). Verified in an isolated worktree: **73 operations, exactly 1:1** with the vendored SDK (73 `api_op_*` = 73 driver methods = 73 handler routes), **no stubs** (`grep TODO|FIXME|panic|501` → none), and wire-faithful — timestamps/member keys/error envelope/`X-Amzn-Errortype` match the SDK, union fields echo byte-faithful, and ID-or-ARN path params survive `%2F` decoding. Gate is green: `build`/`vet`/`gofmt`/`go test ./...`/**`-race`**/`golangci-lint 0`. Architecture pillars all pass (memstore, `idgen` ARNs, injectable clock, canonical errors, no globals, dual-factory wiring), concurrency is sound (single mutex, cross-store counts atomic, copy-on-write verified), and it's correctly AWS-only.

Requesting changes on one **High** and a few real **Medium** correctness gaps — the 73-op parity and routing are excellent; these are the semantics to close before merge. Details inline.

**Should fix**
- **[High] Create-time `Tags` are silently dropped on every resource** — no `Create*` path writes `in.Tags`, so the standard AWS create-with-tags → `ListTagsForResource` flow returns `{}` across ~10 resource types. Untested.
- **Delete has no cascade / no active-child guard** — deleting a service network with live associations succeeds (real AWS → `ResourceInUse`); service/listener/target-group deletes orphan their children.
- **Stale / missing association counts** — counts include associations to already-deleted services, and `UpdateServiceNetwork` never recomputes them (always returns 0).
- **`UpdateResourceConfiguration` clobbers `AllowAssociationToShared`** on a partial update (unconditional bool assignment).
- **`Matches` shadows the S3 catch-all** — a path-style S3 op on a bucket named `services`/`tags`/`targetgroups`/… is hijacked by VPC Lattice (registered before S3).
- **3 of 73 ops have no round-trip test coverage.**

**Also (Low):** snake_case filenames (`access_logs.go`, `service_networks.go`, `resource_gateways.go`, `resource_configs.go`, `domain_verifications.go`, `target_groups.go` per `docs/STRUCTURE.md` §3) · in-place mutation of stored objects is safe under the current `Mutex` but would race if switched to `RWMutex` (document or clone-then-`Set`) · Register/Deregister-Targets discard the driver's failure list (always report all-success) · `Delete{Auth,Resource}Policy` return nil for a missing key · no name-uniqueness / `clientToken` idempotency on create · `UpdateTargetGroup` can't clear a health check · `AccessLogSubscription.ResourceARN` blanked from a bare ID · `ListTargets` unsorted · trailing path segments silently ignored · banner-style separator comments.

Excellent work on the REST-JSON routing and union fidelity — this is the correctness/AWS-semantics polish.

providers/aws/vpclattice/service_networks.go
thzgajendra: [High] Fixed. Every Create*/Start*/Associate* now persists its tags via a shared `writeTags(arn, tags)` helper (create-with-tags → `ListTagsForResource` now round-trips across all resource types). The association/access-log/domain-verification creates that previously ignored the tag param (`_ map[string]string`) now thread it through. Added `TestCreateWritesTags`.

providers/aws/vpclattice/service_networks.go
thzgajendra: Fixed. `DeleteServiceNetwork` now returns `ConflictException` (FailedPrecondition) when any VPC/service/resource association still references it; `DeleteService` blocks on live SN↔service associations and cascades its listeners+rules; `DeleteListener` cascades its rules. (`DeleteTargetGroup` already cascades targets.) Added `TestServiceNetworkDeleteBlockedByAssociation` and `TestServiceDeleteGuardAndCascade`.

providers/aws/vpclattice/service_networks.go
thzgajendra: Fixed. `UpdateServiceNetwork` now calls `applyAssocCounts`, and `applyAssocCounts` skips SN↔service / SN↔resource associations whose target no longer exists (so a Get can't report a phantom). Covered by `TestServiceNetworkAssocCounts` (asserts the count drops to 0 after the target resource is deleted).

providers/aws/vpclattice/resourceconfigs.go
thzgajendra: Fixed. `AllowAssociationToShared` is now `*bool` in `UpdateResourceConfigurationInput` (threaded through the server decode). A partial update that omits it (nil) leaves the stored value unchanged; an explicit `false` clears it. Added `TestUpdateResourceConfigKeepsAllowFlag`.

server/aws/vpclattice/handler.go
thzgajendra: Fixed. `Matches` now gates on path-root **+ method + segment shape**, not just the first segment. Verbs Lattice doesn't use for a root are declined (so S3 `PUT /services/{key}` falls through), and the ARN-remainder roots (`tags`/`authpolicy`/`resourcepolicy`) require the identifier segment (so S3 `GET /tags` on a bucket named `tags` falls through). Added `TestMatchesDoesNotShadowS3`. Documented the one residual ambiguity that's unavoidable for two REST services on one endpoint: identical verb+path like `GET /services` (list-services vs. S3 list-bucket-"services").

server/aws/vpclattice/sdk_roundtrip_test.go
thzgajendra: Fixed. Added wire-level asserts for `GetServiceNetworkServiceAssociation` and `GetServiceNetworkResourceAssociation` right after the creates, and a `DeleteResourceEndpointAssociation` call asserting `ResourceNotFoundException` (the endpoint-association surface isn't synthesized, so this exercises the otherwise-unreachable handler). All 73 ops now have wire-level coverage.

providers/aws/vpclattice/access_logs.go
thzgajendra: [STRUCTURE.md §3] Fixed. Renamed to snake_case in both layers: `access_logs.go`, `service_networks.go`, `resource_gateways.go`, `resource_configs.go`, `domain_verifications.go`, `target_groups.go` (via `git mv`).

thzgajendra:
Thanks — the High + all Medium items are fixed (replied inline). On the **Also (Low)** list:

- **snake_case filenames** — done, both layers.
- **Register/Deregister failure list** — the handlers now serialize the driver's `unsuccessful` list instead of a literal `[]` (`targetFailuresToWire`).
- **`ListTargets` unsorted** — now sorted by `id|port`.
- **`AccessLogSubscription.ResourceARN` blanked from a bare ID** — no longer blanked; the identifier is echoed verbatim.
- **In-place mutation under a future `RWMutex`** — noted; safe under the current single `Mutex`. Leaving the clone-then-`Set` hardening for if/when the lock is split, to avoid unrelated churn here.

Deferred, with reasoning (both genuinely Low / not standard AWS behavior):
- **`clientToken` idempotency & name-uniqueness on create** — `clientToken` isn't currently plumbed through the ~10 create inputs, so wiring idempotency is a wider change better done as a focused follow-up; and AWS does **not** enforce name-uniqueness on these resources (duplicate names are allowed), so I've intentionally not added that.
- **`Delete{Auth,Resource}Policy` returning nil for a missing key** — kept idempotent (deleting an absent policy is a no-op), which matches how these detach-style ops behave.
- **`UpdateTargetGroup` can't clear a health check** — the health check arrives as a raw-JSON blob merged into `config`, so "clear" vs "absent" isn't distinguishable without modeling the sub-shape; deferred.
- **Banner-style separator comments** — cosmetic; left as-is.

Also synced the branch with `development` (merge, not rebase) and resolved the conflicts in `docs/services.md` + the provider/server `aws.go` wiring — the Route 53 Resolver service that landed on `development` and this one now coexist (VPC Lattice is §26 / +73 ops; grand total 1707).

Full gate green after all changes + the merge: `go build ./...`, `go vet ./...`, `gofmt`, `go test -race`, and `golangci-lint --new-from-rev` = 0.

server/aws/vpclattice/handler.go
```diff
@@ -0,0 +1,158 @@
...
+	case "authpolicy", "resourcepolicy":
+		// /<root>/{resourceArn}: an identifier is always present; PUT/GET/DELETE.
+		return len(rest) >= 1 &&
+			(method == http.MethodGet || method == http.MethodPut || method == http.MethodDelete)
+	case "tags":
+		// /tags/{resourceArn}: an identifier is always present; POST/GET/DELETE.
+		return len(rest) >= 1 &&
+			(method == http.MethodGet || method == http.MethodPost || method == http.MethodDelete)
+	default:
+		return isLatticeMethod(method)
```
NitinKumar004: The `latticeClaims` fix mitigates the ambiguous roots (`tags`/`authpolicy`/`resourcepolicy` now require an identifier segment, and `PUT` is excluded), but the **default case still claims any `GET`/`POST`/`PATCH`/`DELETE`** whose first segment is a Lattice root. So a path-style S3 object op — `GET /services/mykey`, `DELETE /targetgroups/mykey`, `GET /servicenetworks/mykey` — on a bucket literally named `services`/`targetgroups`/`servicenetworks`/etc. is hijacked here (VPC Lattice is registered before S3) and fails. The narrow-but-real residual: a bucket named exactly one of the ~15 roots can't do object GET/DELETE.

To close it robustly, gate the resource-id path on a VPC Lattice **identifier shape** — e.g. in `routeByID`, only claim when the id segment is a known Lattice ID prefix (`sn-`/`svc-`/`tg-`/`listener-`/…) or a `arn:aws:vpc-lattice:` ARN; otherwise fall through so S3 handles it (an S3 key like `mykey` won't match). That preserves all real Lattice paths while letting reserved-name buckets through. Alternatively, if a full fix isn't worth it, document the reserved bucket names as a known limitation so it's an explicit boundary rather than a silent mis-route.

NitinKumar004 (changes requested):
## Re-review — VPC Lattice fix commit

Great turnaround — verified against the code, **H1, M2, M3, M4, M6 and the snake_case rename are all resolved**, most exactly right:
- **H1** — all 12 `Create*`/`Start*` now `writeTags`.
- **M2** — `DeleteServiceNetwork`/`DeleteService` guard active associations (`FailedPrecondition`); `DeleteService`/`DeleteListener` cascade their children.
- **M3** — `UpdateServiceNetwork` now calls `applyAssocCounts`.
- **M4** — `AllowAssociationToShared` is now `*bool` + nil-guarded (the right nil-sentinel fix).
- **M6** — both `Get*Association` are now driven via the real SDK client.
- **Naming** — 6 files snake_cased in both layers; the CI Structure check passes.

Gate is green (`build`/`vet`/`go test`/`-race`/`golangci-lint 0`/Structure) with ~135 lines of new tests.

## Still blocking — M5 (S3 shadowing) is only partially fixed
`latticeClaims` helps the ambiguous roots (bucket-level `GET /tags` and all `PUT`s now fall through to S3), but the **core collision remains for the default roots**: because `isLatticeMethod` includes `GET`/`DELETE`, an S3 object op like `GET /services/mykey` or `DELETE /targetgroups/mykey` on a bucket named exactly `services`/`targetgroups`/`servicenetworks`/… is still claimed by this handler and fails as a Lattice op. A user with such a bucket (registered before S3) still can't read/delete its objects. Details inline — requesting this be closed (or the residual explicitly bounded) before merge.

server/aws/vpclattice/handler.go
thzgajendra: Fixed exactly as suggested — resource-scoped routes now gate on **identifier shape**. `Matches` claims a `/<root>/{id}` path only when `{id}` is Lattice-shaped: a generated ID prefix (`sn-`/`svc-`/`tg-`/`listener-`/`rule-`/`snva-`/`snsa-`/`snra-`/`rcfg-`/`rgw-`/`als-`/`dv-`/`rea-`) or contains a `vpc-lattice` ARN (`isLatticeIdentifier`). So `GET /services/mykey`, `DELETE /targetgroups/mykey`, `GET /servicenetworks/mykey` on a like-named S3 bucket now fall through to the S3 catch-all; real Lattice ops (ids/ARNs) are unaffected. The `tags`/`authpolicy`/`resourcepolicy` roots require the same Lattice id/ARN, so `GET /tags/mykey` also falls through.

The only residual is a bare `GET /<root>` (list) colliding with an S3 list-bucket on an identically-named bucket — unavoidable for two REST services on one endpoint — now called out explicitly in `docs/services.md`.

`TestMatchesDoesNotShadowS3` extended with the object-op cases (`GET`/`DELETE /<root>/mykey` → not claimed; `/<root>/<lattice-id>` → claimed). Gate green: build/vet/gofmt/-race/golangci-lint 0. (`541d4a6`)

thzgajendra:
Thanks for the re-review. **M5 is now closed** (the one remaining blocker): resource-scoped routes gate on a VPC Lattice identifier shape, so path-style S3 object ops (`GET /services/mykey`, `DELETE /targetgroups/mykey`, `GET /tags/mykey`) on a bucket named like a Lattice root fall through to the S3 catch-all instead of being mis-claimed. Only real Lattice ids/ARNs are claimed. The single unavoidable residual — bare `GET /<root>` list vs. S3 list-bucket on an identically-named bucket — is documented in `docs/services.md`.

All other items from the reviews (H1 tags-on-create, M2 delete cascade/guards, M3 counts, M4 `*bool`, M6 association Get coverage, snake_case) were resolved in the prior commit. Gate green after this change: `go build ./...`, `go vet`, `gofmt`, `go test -race`, `golangci-lint --new-from-rev` = 0. `541d4a6`.

NitinKumar004 (approved):
## Approving — all review findings resolved

Verified the M5 fix (`541d4a69`) in an isolated worktree at the PR head; gate green (`build`/`vet`/`go test -race`/`TestMatchesDoesNotShadowS3` all pass).

**M5 (S3 shadowing) is now genuinely closed.** Resource-scoped routes are claimed only when the id segment is Lattice-shaped (a known ID prefix or a `vpc-lattice` ARN), so path-style S3 object ops on a like-named bucket (`GET /services/mykey`, `DELETE /targetgroups/mykey`) fall through to the S3 catch-all. I cross-checked `isLatticeIdentifier`'s 13 prefixes against every ID the provider actually generates — **exact match, no gaps** — so no create→get-by-returned-id round-trip can wrongly fall through. The one prefix-less route (`servicenetworkvpcendpointassociations`) is list-only, so it correctly needs none. The single remaining residual — a bare `GET /<root>` list vs. an S3 list-bucket on an identically-named bucket — is inherent to two REST services on one endpoint and is now honestly documented in `docs/services.md` and the `Matches` comment.

That closes every finding from the earlier reviews — H1 (create-time tags), M2 (delete guard/cascade), M3 (assoc counts), M4 (`AllowAssociationToShared` nil-sentinel), M5, M6 (Get*Association coverage), and the snake_case rename. Clean, complete work — thanks for the thorough turnaround. LGTM.

Issue #335, opened 2026-08-06
Roadmap: "minikube for cloud resources" — lifecycle CLI, persistence, profiles, dashboard + differentiators


thzgajendra:
## Summary

cloudemu already has the hard 90% of a local cloud emulator — 1,635 operations across 25 categories and 3 providers in one binary, a real network-topology engine, chaos, cost simulation, resource discovery, and both an in-process Go API and an SDK-wire-compatible HTTP server. What it lacks is the **product/UX layer** that would make it feel like `minikube` does for Kubernetes: a managed local instance you `start`, `stop`, inspect, and that survives restarts.

This issue tracks turning cloudemu into a first-class **"minikube for cloud resources"** — adopting minikube's lifecycle ergonomics, matching LocalStack's dev-experience features (our real competitor), and leaning into the things neither of them can do.

## Framing

- **minikube** = right UX model, wrong competitor (it only does k8s).
- **LocalStack** = the real competitor for cloud resources.
- **Goal:** minikube-grade CLI/UX + LocalStack-grade features + cloudemu's unique engine strengths.

## What already exists (baseline)

- Breadth: 1,635 ops / 25 categories / AWS+Azure+GCP (`docs/services.md`).
- Network **topology reachability engine** that actually enforces SG/ACL/route rules — `CanConnect`, `TraceRoute` (`docs/topology.md`). Neither LocalStack nor minikube has this.
- Chaos injection (`features/chaos`), cost simulation (`services/cost`), cross-cloud resource discovery (`services/resourcediscovery`), auto-metrics + alarm evaluation (`SetMonitoring`), recorder/metrics/ratelimit/error-injection (`features/`).
- Admin control plane `/_cloudemu/{reset,seed,health}`, seed fixtures (`seed/`), Docker + compose + Testcontainers (`contrib/testcontainers`), emulated k8s API with kubeconfig (`services/kubernetes`).

### vs minikube UX
| minikube | cloudemu today | gap |
|---|---|---|
| `start/stop/status/delete/pause` | `serve` (foreground) + `version/help` | no lifecycle CLI / daemon |
| survives restarts | in-memory only (#107) | no persistence |
| `profile` (multiple clusters) | single instance | no named profiles |
| `dashboard` (web UI) | JSON endpoints only | no dashboard |
| `addons` | fixed feature set | no opt-in modules |

### vs LocalStack (the real competitor)
[...]

PR #336, opened 2026-08-06
Add cloudemu lifecycle CLI: start/stop/status/logs/delete (#335 P0-a)

thzgajendra:
## Objective / Issue
First deliverable of #335 (**"minikube for cloud resources"**), **P0-a**: give cloudemu a background lifecycle so you can leave it running, instead of only the foreground `cloudemu serve`.

## What we found
`cloudemu serve` runs in the foreground only — there was no `start/stop/status`. Everything needed to build on already exists: `serve` writes an endpoints file, supports `--endpoints-file`/`--quiet`, has graceful SIGTERM shutdown, and exposes `GET /_cloudemu/health`.

## How we fixed it
New `cmd/cloudemu/lifecycle.go` + a dispatch in `main.go`:
- **`start [serve flags]`** — re-execs this same binary as `serve` **detached** (new session via `Setsid`, stdout/stderr → log file), waits for readiness, records run state, prints endpoints. Idempotent (a second `start` reports the already-running instance). All `serve` flags pass through.
- **`stop`** — SIGTERM the pid and wait for exit (reuses `serve`'s graceful shutdown), clean up state.
- **`status`** — pid-liveness + endpoints.
- **`logs [-f]`** — print/follow the daemon log.
- **`delete`** — stop + remove the run directory.

Run state (pid, log, resolved endpoints) lives under `~/.cloudemu/` by default; `--home <dir>` relocates it. `serve` itself is untouched.

**Readiness probe:** plain-HTTP `/_cloudemu/health` for the AWS/GCP endpoints; a TCP-connect probe for the self-signed HTTPS endpoints (Azure/Kubernetes) — so no TLS verification is ever disabled.

## Alternatives not taken
- No third-party daemon library — a detached re-exec of `serve` keeps it dependency-free and portable.
- **Persistence is out of scope here** — state still starts empty each `start`. That's P0-b (#107) and its own PR; documented under "Not yet included".

## Docs / Tests
- **Docs:** `docs/standalone-server.md` — new "Background mode (`start`/`stop`/`status`)" section.
- **Unit tests** (`lifecycle_test.go`, TDD-first): run-dir resolution, state round-trip, `processAlive`, HTTP health poll (ready + timeout), and `--home` flag extraction/passthrough.
- **E2E (user perspective), run locally against the real `aws` CLI v2.30:**
  - `cloudemu start` (all 4 providers) → every endpoint verified up.
  - `aws s3 mb/cp/ls`, `aws dynamodb create-table/list-tables`, `aws ec2 run-instances/describe-instances` — all succeed against `:4566`.
  - `status` (running) → idempotent `start` → `stop` (ports released) → `status` (stopped) → `delete` (run dir removed); no stray processes.

## Test plan
- [x] `go build ./...`, `go vet ./cmd/cloudemu/`, `gofmt`
- [x] `go test -race ./cmd/cloudemu/`
- [x] `golangci-lint run --new-from-rev=$(git merge-base HEAD stackshy/development) ./cmd/cloudemu/...` = **0**
- [x] Manual E2E with the real AWS CLI (above)

## Risk & Rollback
Low — purely additive (new subcommands + a new file); `serve` and all in-process APIs are unchanged. Rollback = drop `lifecycle.go` and the `main.go` dispatch case.

## Follow-ups
- **P0-b — persistence (#107):** snapshot/restore state so `stop`→`start` keeps resources.
- P1 — profiles, web dashboard, cloud snapshots (per #335).

### DIFF
--- a/cmd/cloudemu/lifecycle.go
+++ b/cmd/cloudemu/lifecycle.go
@@ -1,3 +1,5 @@
+//go:build unix
+
 package main
 
 import (
@@ -29,12 +31,6 @@ const (
 
 	dirPerm  = 0o755
 	filePerm = 0o600
-
-	cmdStart  = "start"
-	cmdStop   = "stop"
-	cmdStatus = "status"
-	cmdLogs   = "logs"
-	cmdDelete = "delete"
 )
 
 // Sentinel errors so callers (and err113) get static, wrappable failures.
@@ -121,6 +117,65 @@ func processAlive(pid int) bool {
 	return p.Signal(syscall.Signal(0)) == nil
 }
 
+// terminate stops another process (not our child) gracefully (SIGTERM), then
+// escalates to SIGKILL if it hasn't exited within stopTimeout. Used by `stop`,
+// where the target daemon was started by a previous CLI invocation and has been
+// reparented to init (so it's reaped on death and signal-0 reports it gone).
+func terminate(p *os.Process) error {
+	_ = p.Signal(syscall.SIGTERM)
+
+	if waitExit(p.Pid, stopTimeout) == nil {
+		return nil
+	}
+
+	_ = p.Kill()
+
+	return waitExit(p.Pid, stopTimeout)
+}
+
+// killChild stops a process we started (our direct child), reaping it so it
+// doesn't linger as a zombie. SIGTERM first, escalate to SIGKILL after
+// stopTimeout. Reaping via Wait is what makes this safe on the spawnServe
+// failure path, where the CLI is still the child's parent.
+func killChild(cmd *exec.Cmd) {
+	_ = cmd.Process.Signal(syscall.SIGTERM)
+
+	done := make(chan struct{})
+
+	go func() {
+		_, _ = cmd.Process.Wait()
+
+		close(done)
+	}()
+
+	select {
+	case <-done:
+	case <-time.After(stopTimeout):
+		_ = cmd.Process.Kill()
+
+		<-done
+	}
+}
+
+// daemonReachable reports whether the recorded endpoints answer — a portable
+// identity check that guards against a recycled PID (a live but unrelated
+// process) before we signal it.
+func daemonReachable(eps map[string]string) bool {
+	if u := httpHealthURL(eps); u != "" {
+		return pollHealth(u, dialTimeout) == nil
+	}
+
+	for _, k := range []string{"azure", "kubernetes"} {
+		if ep := eps[k]; ep != "" {
+			if hp, err := hostPortOf(ep); err == nil {
+				return pollTCP(hp, dialTimeout) == nil
+			}
+		}
+	}
+
+	return false
+}
+
 // pollHealth polls an HTTP health URL until it returns 200 or timeout elapses.
 // It is used only for the plain-HTTP endpoints (AWS/GCP); the self-signed HTTPS
 // endpoints use pollTCP instead so no TLS verification has to be disabled.
@@ -178,12 +233,20 @@ func pollTCP(hostPort string, timeout time.Duration) error {
 	}
 }
 
-// waitServerReady blocks until the server behind eps is accepting requests,
-// using an HTTP health probe when a plain-HTTP endpoint exists and a TCP probe
-// otherwise.
+// waitServerReady blocks until every present endpoint is accepting requests:
+// an HTTP health probe for the plain-HTTP endpoints (AWS/GCP) and a TCP probe
+// for the self-signed HTTPS endpoints (Azure/Kubernetes).
 func waitServerReady(eps map[string]string, timeout time.Duration) error {
-	if u := httpHealthURL(eps); u != "" {
-		return pollHealth(u, timeout)
+	if len(eps) == 0 {
+		return errNoEndpoints
+	}
+
+	for _, k := range []string{"aws", "gcp"} {
+		if ep := eps[k]; ep != "" {
+			if err := pollHealth(strings.TrimRight(ep, "/")+"/_cloudemu/health", timeout); err != nil {
+				return err
+			}
+		}
 	}
 
 	for _, k := range []string{"azure", "kubernetes"} {
@@ -193,11 +256,13 @@ func waitServerReady(eps map[string]string, timeout time.Duration) error {
 				return err
 			}
 
-			return pollTCP(hp, timeout)
+			if err := pollTCP(hp, timeout); err != nil {
+				return err
+			}
 		}
 	}
 
-	return errNoEndpoints
+	return nil
 }
 
 // httpHealthURL returns a plain-HTTP /_cloudemu/health URL, or "" if none.
@@ -309,7 +374,7 @@ func runStart(args []string) error {
 		return err
 	}
 
-	if s, rErr := readState(dir); rErr == nil && processAlive(s.PID) {
+	if s, rErr := readState(dir); rErr == nil && processAlive(s.PID) && daemonReachable(s.Endpoints) {
 		fmt.Printf("cloudemu already running (pid %d)\n", s.PID)
 		printEndpoints(s.Endpoints)
 
@@ -364,7 +429,7 @@ func spawnServe(dir string, serveArgs []string, epPath string) (map[string]strin
 	}
 
 	if readyErr != nil {
-		_ = cmd.Process.Signal(syscall.SIGTERM)
+		killChild(cmd) // don't leave an untracked, detached child
 
 		return nil, fmt.Errorf("cloudemu failed to start (see %s): %w", logPath(dir), readyErr)
 	}
@@ -376,7 +441,15 @@ func spawnServe(dir string, serveArgs []string, epPath string) (map[string]strin
 		Args:      serveArgs,
 	}
 
-	return eps, writeState(dir, state)
+	// A persisted state is what makes the daemon findable by stop/status; if we
+	// can't record it, kill the child rather than orphan it.
+	if err := writeState(dir, state); err != nil {
+		killChild(cmd)
+
+		return nil, fmt.Errorf("failed to persist daemon state (daemon killed): %w", err)
+	}
+
+	return eps, nil
 }
 
 // runStop signals the running daemon to shut down and waits for it to exit.
@@ -389,13 +462,19 @@ func runStop(args []string) error {
 	}
 
 	s, err := readState(dir)
-	if err != nil {
+	if errors.Is(err, os.ErrNotExist) {
 		fmt.Println("cloudemu is not running")
 
 		return nil
 	}
 
-	if !processAlive(s.PID) {
+	if err != nil {
+		return fmt.Errorf("reading state (daemon may still be running): %w", err)
+	}
+
+	// Treat a dead pid, or a live pid whose endpoints don't answer (PID reused
+	// by an unrelated process), as stale — clean up without signaling it.
+	if !processAlive(s.PID) || !daemonReachable(s.Endpoints) {
 		_ = removeState(dir)
 
 		fmt.Println("cloudemu is not running (cleaned up stale state)")
@@ -408,12 +487,8 @@ func runStop(args []string) error {
 		return err
 	}
 
-	if err := proc.Signal(syscall.SIGTERM); err != nil {
-		return err
-	}
-
-	if err := waitExit(s.PID, stopTimeout); err != nil {
-		return err
+	if err := terminate(proc); err != nil {
+		return fmt.Errorf("stopping pid %d: %w", s.PID, err)
 	}
 
 	_ = removeState(dir)
@@ -448,7 +523,17 @@ func runStatus(args []string) error {
 	}
 
 	s, err := readState(dir)
-	if err != nil || !processAlive(s.PID) {
+	if errors.Is(err, os.ErrNotExist) {
+		fmt.Println("cloudemu: stopped")
+
+		return nil
+	}
+
+	if err != nil {
+		return fmt.Errorf("reading state (daemon may still be running): %w", err)
+	}
+
+	if !processAlive(s.PID) {
 		fmt.Println("cloudemu: stopped")
 
 		return nil
@@ -479,7 +564,8 @@ func runLogs(args []string) error {
 	return tailLog(logPath(dir), follow)
 }
 
-// runDelete stops the daemon (if running) and removes its run directory.
+// runDelete stops the daemon (if running) and removes only cloudemu's own files
+// from the run directory — never a blanket RemoveAll of a user-supplied --home.
 func runDelete(args []string) error {
 	home, _ := splitHomeFlag(args)
 
@@ -492,17 +578,23 @@ func runDelete(args []string) error {
 		return err
 	}
 
-	if err := os.RemoveAll(dir); err != nil {
-		return err
+	for _, p := range []string{statePath(dir), logPath(dir), endpointsPath(dir)} {
+		if rmErr := os.Remove(p); rmErr != nil && !os.IsNotExist(rmErr) {
+			return rmErr
+		}
 	}
 
-	fmt.Printf("cloudemu: removed %s\n", dir)
+	// Remove the dir only if it's now empty (ignore "not empty" / "not exist").
+	_ = os.Remove(dir)
+
+	fmt.Printf("cloudemu: removed cloudemu files under %s\n", dir)
 
 	return nil
 }
 
 // tailLog prints the log file. When follow is set it streams appended output
-// until interrupted.
+// until interrupted, re-seeking to the start if the log is truncated (e.g. a
+// restart reopens it with O_TRUNC).
 func tailLog(path string, follow bool) error {
 	f, err := os.Open(path)
 	if err != nil {
@@ -510,7 +602,8 @@ func tailLog(path string, follow bool) error {
 	}
 	defer f.Close()
 
-	if _, err := io.Copy(os.Stdout, f); err != nil {
+	offset, err := io.Copy(os.Stdout, f)
+	if err != nil {
 		return err
 	}
 
@@ -521,9 +614,25 @@ func tailLog(path string, follow bool) error {
 	for {
 		time.Sleep(healthInterval)
 
-		if _, err := io.Copy(os.Stdout, f); err != nil {
+		fi, err := f.Stat()
+		if err != nil {
 			return err
 		}
+
+		if fi.Size() < offset {
+			if _, seekErr := f.Seek(0, io.SeekStart); seekErr != nil {
+				return seekErr
+			}
+
+			offset = 0
+		}
+
+		n, err := io.Copy(os.Stdout, f)
+		if err != nil {
+			return err
+		}
+
+		offset += n
 	}
 }
 
--- a/cmd/cloudemu/lifecycle.go
+++ b/cmd/cloudemu/lifecycle.go
@@ -1,0 +1,546 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/url"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"syscall"
+	"time"
+)
+
+const (
+	stateFileName     = "state.json"
+	logFileName       = "cloudemu.log"
+	endpointsFileName = "endpoints.json"
+
+	startupTimeout = 15 * time.Second
+	stopTimeout    = 12 * time.Second
+	healthInterval = 100 * time.Millisecond
+	dialTimeout    = time.Second
+
+	dirPerm  = 0o755
+	filePerm = 0o600
+
+	cmdStart  = "start"
+	cmdStop   = "stop"
+	cmdStatus = "status"
+	cmdLogs   = "logs"
+	cmdDelete = "delete"
+)
+
+// Sentinel errors so callers (and err113) get static, wrappable failures.
+var (
+	errTimeout     = errors.New("timed out")
+	errNoEndpoints = errors.New("no endpoints to probe")
+	errUnknownCmd  = errors.New("unknown lifecycle command")
+)
+
+// daemonState is the run-dir record describing a running standalone server.
+type daemonState struct {
+	PID       int               `json:"pid"`
+	Endpoints map[string]string `json:"endpoints"`
+	StartedAt string            `json:"startedAt"`
+	Args      []string          `json:"args"`
+}
+
+// runDir resolves the directory that holds the daemon's pid/log/endpoints.
+// An explicit --home wins; otherwise it defaults to ~/.cloudemu.
+func runDir(home string) (string, error) {
+	if home != "" {
+		return home, nil
+	}
+
+	h, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+
+	return filepath.Join(h, ".cloudemu"), nil
+}
+
+func statePath(dir string) string     { return filepath.Join(dir, stateFileName) }
+func logPath(dir string) string       { return filepath.Join(dir, logFileName) }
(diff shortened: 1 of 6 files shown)


cmd/cloudemu/lifecycle.go
```diff
@@ -0,0 +1,546 @@
...
+	}
+
+	state := daemonState{
+		PID:       cmd.Process.Pid,
+		Endpoints: eps,
+		StartedAt: time.Now().UTC().Format(time.RFC3339),
+		Args:      serveArgs,
+	}
+
+	return eps, writeState(dir, state)
```
thzgajendra: **[MED] Orphaned daemon on `writeState` failure.** The child is already up and confirmed ready by this point, so returning `(eps, writeState(...))` means a `writeState` error (disk full, perms) leaves a **running but untracked** daemon — no `state.json`, so `stop`/`status` can never find it; only `pkill` can. Roll back the child on failure:
```go
if err := writeState(dir, state); err != nil {
    _ = cmd.Process.Signal(syscall.SIGTERM)
    return nil, fmt.Errorf("failed to persist daemon state (daemon killed): %w", err)
}
return eps, nil
```

cmd/cloudemu/lifecycle.go
```diff
@@ -0,0 +1,546 @@
...
+		return nil, startErr
+	}
+
+	eps, readyErr := waitForEndpoints(epPath, startupTimeout)
+	if readyErr == nil {
+		readyErr = waitServerReady(eps, startupTimeout)
+	}
+
+	if readyErr != nil {
+		_ = cmd.Process.Signal(syscall.SIGTERM)
```
thzgajendra: **[MED] Readiness-abort doesn't wait for the child.** This fires SIGTERM and returns immediately — the child is `Setsid`-detached, so if it's slow to exit or ignores SIGTERM it lingers as an untracked live process holding the port (and no state was written). Bounded-wait + escalate:
```go
if readyErr != nil {
    _ = cmd.Process.Signal(syscall.SIGTERM)
    if waitExit(cmd.Process.Pid, stopTimeout) != nil {
        _ = cmd.Process.Kill()
    }
    return nil, fmt.Errorf("cloudemu failed to start (see %s): %w", logPath(dir), readyErr)
}
```

cmd/cloudemu/lifecycle.go
```diff
@@ -0,0 +1,546 @@
...
+	proc, err := os.FindProcess(s.PID)
+	if err != nil {
+		return err
+	}
+
+	if err := proc.Signal(syscall.SIGTERM); err != nil {
+		return err
+	}
+
+	if err := waitExit(s.PID, stopTimeout); err != nil {
```
thzgajendra: **[MED] `stop` never escalates to SIGKILL.** If the daemon ignores SIGTERM, `waitExit` returns `errTimeout` after 12s and (per the flow below) `removeState` still runs — orphaning a live daemon that the CLI now can't track. After the timeout, `proc.Signal(syscall.SIGKILL)` + a short second `waitExit` before removing state.

cmd/cloudemu/lifecycle.go
```diff
@@ -0,0 +1,546 @@
...
+// runStop signals the running daemon to shut down and waits for it to exit.
+func runStop(args []string) error {
+	home, _ := splitHomeFlag(args)
+
+	dir, err := runDir(home)
+	if err != nil {
+		return err
+	}
+
+	s, err := readState(dir)
```
thzgajendra, line 460: **[MED] `readState` conflates "absent" with "unreadable".** A missing file, a permission error, and malformed JSON all collapse to "cloudemu is not running" here — so a corrupt `state.json` on a *live* daemon reports it down and it becomes unstoppable via the CLI (and `runDelete`, which calls this, then `RemoveAll`s the state, orphaning it for good). `runStatus:451` has the same conflation. Fix:
```go
if errors.Is(err, os.ErrNotExist) { fmt.Println("cloudemu is not running"); return nil }
if err != nil { return fmt.Errorf("reading state (daemon may still be running): %w", err) }
```

cmd/cloudemu/lifecycle.go
```diff
@@ -0,0 +1,546 @@
...
+	if err := runStop(args); err != nil {
+		return err
+	}
+
+	dir, err := runDir(home)
+	if err != nil {
+		return err
+	}
+
+	if err := os.RemoveAll(dir); err != nil {
```
thzgajendra: **[MED] `RemoveAll` footgun with `--home`.** `dir` is the verbatim `--home` value, so `delete --home /some/shared/dir` wipes the entire tree, not just cloudemu's `state.json`/`cloudemu.log`/`endpoints.json`. Safer: remove the known files (`removeState` + `os.Remove(logPath)` + `os.Remove(endpointsPath)`) then `os.Remove(dir)` (succeeds only if now empty), or refuse `RemoveAll` when `home != ""`.

cmd/cloudemu/lifecycle.go
```diff
@@ -0,0 +1,546 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer logF.Close()
+
+	full := append([]string{"serve", "--endpoints-file", epPath, "--quiet"}, serveArgs...)
+	cmd := exec.CommandContext(context.Background(), exe, full...)
+	cmd.Stdout = logF
+	cmd.Stderr = logF
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setsid: true}
```
thzgajendra, line 416: **[LOW-MED] Unix-only, no build tag → doesn't compile on Windows.** `syscall.SysProcAttr{Setsid: true}` (plus `syscall.Signal(0)`/`SIGTERM`) is Unix-only; `GOOS=windows go build ./cmd/cloudemu/` fails here. CI is Linux-only so it isn't caught, but `docs/standalone-server.md` advertises bare `go install …@latest`, so a Windows user hits a cryptic error. Either split the platform bits into `lifecycle_unix.go` (`//go:build unix`) + a Windows stub that errors gracefully, or document Unix/macOS-only.

thzgajendra (commented):
## Review — lifecycle CLI (start/stop/status/logs/delete), #335 P0-a

**Solid, ship-worthy after a few fixes — no blockers.** Purely additive, well-structured, `serve` untouched. `go build`, `go vet`, `go test -race ./cmd/cloudemu/`, and `gofmt` are all green locally, and golangci-lint is clean on the new lines. The strongest issues cluster around one theme: **failure paths that leave an invisible, untracked daemon** — worth closing before merge because the whole point of the feature is a managed instance you can reliably `stop`.

### Fix before merge (all small, localized)
- **Orphan daemon cluster** (inline): `writeState` failure returns an error while the daemon keeps running untracked; the readiness-abort SIGTERMs without waiting; `stop` never escalates to SIGKILL after its 12s timeout. Each leaves a live process no `stop`/`status` can find.
- **`readState` conflates "absent" vs "unreadable"** (inline): a corrupt/permission-denied `state.json` on a *running* daemon makes `stop`/`status` report it down — and `delete` then wipes the state, orphaning it permanently. One-line fix: branch on `errors.Is(err, os.ErrNotExist)`.
- **`delete` RemoveAll footgun** (inline): `os.RemoveAll` on the verbatim `--home` deletes the whole dir, not just cloudemu's files.

### Should fix / follow-up
- **PID-reuse safety** — `processAlive` is a bare signal-0 probe; a recycled PID means `stop` could SIGTERM an unrelated process. Store & verify identity (start-time / exe path / held lock file).
- **Windows** (inline): `Setsid` is Unix-only and there are no build tags; CI is Linux-only so it isn't caught, but `docs/standalone-server.md` advertises bare `go install …@latest`, so a Windows user gets a cryptic compile error. `//go:build unix` split + a graceful stub, or a doc note.
- **Concurrent `start`** has no lock (`O_CREATE|O_EXCL` fixes it); **`logs -f`** silently stalls after a restart truncates the log (re-seek when size < offset).
- **Readiness probe checks only the FIRST endpoint** — fine as a proxy since `serve` binds all listeners before writing the endpoints file, but the PR text "all 4 providers verified up" overstates what the code checks. Probe all present endpoints if you want the claim to hold.

### Tests (coverage 9.2%)
The pure helpers are cheaply testable but skipped: `waitServerReady` (all 3 branches), `pollTCP` (twin of the tested `pollHealth`), `readEndpoints` (prune), `waitForEndpoints` (all-empty gate), `httpHealthURL`, `hostPortOf`, `runLifecycle` unknown-cmd, `waitExit` timeout, `runStop` stale-state, and `runStart` already-running (writable via a self-PID state). `spawnServe`/full `runStop` as manual-E2E-only is defensible, but the **readiness-failure child-cleanup** path is the highest-risk untested behavior — give it a harness test.

### Docs (minor, while you're in the file)
- The doc shows the **foreground** banner ("standalone server", Title-case keys); `start`/`status` print `printEndpoints` ("cloudemu — running", **lowercase** keys). Reconcile or show a real `start` sample.
- Ports table lists **Kubernetes as HTTP**; it's served **HTTPS** (and the new TCP-probe branch depends on that) — pre-existing, but you're editing the file.
- Nit: `os.FindProcess` err-check is dead code on Unix.

Great first slice of the minikube-ification roadmap — the structure (detached re-exec, no daemon lib) is the right call.

cmd/cloudemu/lifecycle.go
thzgajendra: Fixed. `spawnServe` now kills **and reaps** the child on a `writeState` failure (`killChild` — SIGTERM, then SIGKILL after the timeout, then `Wait`) and returns `failed to persist daemon state (daemon killed)`. No untracked daemon is left behind.

cmd/cloudemu/lifecycle.go
thzgajendra: Fixed. The readiness-abort path now calls `killChild(cmd)` — bounded SIGTERM→SIGKILL with a reap — instead of a fire-and-forget SIGTERM. (Reaping matters here specifically: the CLI is still the child's parent at this point, so a bare signal would leave a zombie that signal-0 still reports alive — that's the exact failure a harness test now catches, `TestKillChildReaps`.)

cmd/cloudemu/lifecycle.go
thzgajendra: Fixed. `stop` now uses `terminate` — SIGTERM, and if the daemon hasn't exited within `stopTimeout`, SIGKILL + a second bounded wait — and only removes state once the process is confirmed gone. (`terminate` is the non-child variant: the target daemon is reparented to init, so it's reaped on death and signal-0 reports it gone.)

cmd/cloudemu/lifecycle.go
thzgajendra: Fixed. Both `stop` and `status` now branch on `errors.Is(err, os.ErrNotExist)`: a missing state is "stopped/not running", but a corrupt/permission-denied `state.json` returns `reading state (daemon may still be running): …` rather than silently reporting it down. `delete` therefore no longer wipes an unreadable-but-live daemon's state.

cmd/cloudemu/lifecycle.go
thzgajendra: Fixed. `delete` now removes only cloudemu's own files (`state.json`, `cloudemu.log`, `endpoints.json`) and then `os.Remove(dir)` (which succeeds only if the dir is now empty) — never `RemoveAll` on a user-supplied `--home`.

cmd/cloudemu/lifecycle.go
thzgajendra: Fixed. Split the daemon impl behind `//go:build unix` (`lifecycle.go`, `lifecycle_test.go`) and added `lifecycle_other.go` (`//go:build !unix`) whose `runLifecycle` returns a clear `start/stop/... are only supported on Unix/macOS; run \`cloudemu serve\` instead`. The command dispatch + names live in `main.go` (all platforms). Verified: `GOOS=windows go build ./cmd/cloudemu/` now compiles.

thzgajendra:
Thanks for the thorough review — all six **fix-before-merge** items are addressed inline. On the rest:

**Should fix / follow-up**
- **PID-reuse safety** — addressed portably: `stop` and `start` now verify the recorded endpoints actually answer (`daemonReachable`) before signalling a live PID, so a recycled PID belonging to an unrelated process is treated as stale (cleaned up, not killed) rather than SIGTERM'd. Full start-time/exe identity is a reasonable further hardening but needs OS-specific process metadata; the endpoint check covers the realistic case without it.
- **Readiness probes all endpoints** — done; `waitServerReady` now probes every present endpoint (HTTP health for AWS/GCP, TCP for the self-signed HTTPS ones), so the "all providers up" claim holds.
- **`logs -f` truncation** — done; the follow loop re-seeks to 0 when the file shrinks below the read offset (a restart's `O_TRUNC`).
- **Concurrent `start` lock** — deferred as a follow-up. The `readState`+alive+reachable check closes the common window; a robust `O_CREATE|O_EXCL` lock needs careful lifecycle across the detached child (the lock must outlive the CLI process but be reclaimable after a crash), which is more than this PR should carry. Happy to do it as a small follow-up.

**Tests** — added unit coverage for the cheap pure helpers you listed (`waitServerReady` all branches, `pollTCP`, `readEndpoints` prune, `waitForEndpoints` timeout, `httpHealthURL`, `hostPortOf`, `waitExit` timeout, `runLifecycle` unknown-cmd, `runStop` stale-state, `runStart` already-running) plus the highest-risk one you flagged — the child-cleanup/reaping path (`TestKillChildReaps`). Coverage 9.2% → 24.3%. `spawnServe`/full `runServe` stay manual-E2E-only.

**Docs** — corrected the Kubernetes protocol to **HTTPS** in the ports table, the banner sample, and the endpoints-file example.

Full gate green after the changes: `go build ./...`, `go vet`, `gofmt`, `go test -race ./cmd/cloudemu/`, `GOOS=windows go build ./cmd/cloudemu/`, and `golangci-lint --new-from-rev` = 0.

cmd/cloudemu/lifecycle.go
```diff
@@ -0,0 +1,655 @@
...
+// an HTTP health probe for the plain-HTTP endpoints (AWS/GCP) and a TCP probe
+// for the self-signed HTTPS endpoints (Azure/Kubernetes).
+func waitServerReady(eps map[string]string, timeout time.Duration) error {
+	if len(eps) == 0 {
+		return errNoEndpoints
+	}
+
+	for _, k := range []string{"aws", "gcp"} {
+		if ep := eps[k]; ep != "" {
+			if err := pollHealth(strings.TrimRight(ep, "/")+"/_cloudemu/health", timeout); err != nil {
```
NitinKumar004: **M1 (Medium).** This probes `/_cloudemu/health`, which only returns 200 when the admin control plane is mounted (`server/admin/admin.go:96`; `handlerFor` gates it on `c.admin`, `serve.go:234`). Since `start` forwards every serve flag, `cloudemu start --admin=false` boots `serve` fine but this probe never sees 200 → `waitServerReady` times out (~15s) → `killChild` kills a healthy server → "failed to start." Make the aws/gcp readiness check admin-independent: TCP-accept probe like azure/kubernetes, or fall back to `pollTCP(hostPortOf(ep))` when `/_cloudemu/health` doesn't return 200. (Same admin dependency is in `daemonReachable`.)

cmd/cloudemu/lifecycle.go
```diff
@@ -0,0 +1,655 @@
...
+		return nil, err
+	}
+
+	logF, err := os.OpenFile(logPath(dir), os.O_CREATE|os.O_WRONLY|os.O_TRUNC, filePerm)
+	if err != nil {
+		return nil, err
+	}
+	defer logF.Close()
+
+	full := append([]string{"serve", "--endpoints-file", epPath, "--quiet"}, serveArgs...)
```
NitinKumar004, line 412: **L2 (Low).** You inject `--endpoints-file <epPath> --quiet` then append user args. If a user passes `cloudemu start --endpoints-file /x`, Go's flag parser is last-wins → `serve` writes to `/x`, so `waitForEndpoints(epPath)` waits on a file that's never written and times out (~15s) with a confusing "failed to start." Strip/reject a user-supplied `--endpoints-file` the way you strip `--home` (and optionally dedupe `--quiet`).

NitinKumar004 (changes requested):
## Deep review — lifecycle CLI

Nice work — this is careful, well-structured code. Verified in an isolated worktree at the PR head: **CI gates all green** (`build` / `vet` / `go test -race ./cmd/cloudemu/` `ok` / `gofmt` clean), the two-phase readiness is sound (`serve` binds every listener before writing the endpoints file, so "file exists ⇒ ports accept," then `waitServerReady` confirms), the unix/non-unix build-tag split is clean, `killChild` (reap our child) vs `terminate` (signal-0 on the reparented daemon) is the right distinction, and `delete` removes only cloudemu's own files rather than `RemoveAll`-ing a user `--home`. Docs are updated and even fix the pre-existing k8s HTTP→HTTPS port error. Mirror rule is N/A (ops/CLI feature).

Holding on one real functional bug (**M1**) plus a few low notes.

### M1 (Medium) — `cloudemu start --admin=false` kills a healthy server
The AWS/GCP readiness probe hits `/_cloudemu/health`, but that route returns 200 **only when the admin control plane is mounted** — `server/admin/admin.go:96`, and `handlerFor` mounts it only when `c.admin` is true (`serve.go:234`). `start` forwards *all* serve flags through (this PR advertises "all serve flags pass through"), so `cloudemu start --admin=false` — which includes the default all-providers run — starts `serve` fine, but the probe never sees 200, `waitServerReady` times out (~15s), and `killChild` then kills a perfectly healthy server, reporting "failed to start." Confusing failure on a documented, realistic flag (hardening the emulator by disabling the state-wiping admin plane). The same admin dependency also sits in `daemonReachable` (idempotent-start / stop's stale-check); that path is unreachable today only because `start` never manages to record an `--admin=false` daemon. Inline on the probe line. Fix: make the aws/gcp probe admin-independent (TCP-accept like azure/k8s, or fall back to TCP when `/health` isn't 200).

### L1 (Low) — gosec G703 ×3 on `--home` (version drift)
`os.MkdirAll`/`os.Remove` on a `--home`-derived path (`lifecycle.go:71,384,588`). In a CLI the user supplies their own `--home` — not a trust boundary, so it's a false-positive vuln. G703 taint analysis is newer gosec; with golangci-lint **2.11.4** it fires, which explains why your `--new-from-rev` run reported 0 (linter version drift). Not CI-blocking (CI doesn't run golangci-lint), but the local gate per CLAUDE.md wants 0 — please confirm against the repo's pinned linter version; if it fires there, a `//nolint:gosec // user-supplied --home, not a trust boundary` closes it.

### L3 (Low) — no automated test for the spawn happy path
The pure helpers are well covered, but the actual detached-serve spawn (`spawnServe`/`runStart` happy path) is only in the manual e2e — and an e2e over `--admin=false` is exactly what would have caught M1. Optional: one test that spawns the built binary and asserts start→status→stop.

No AI attribution. Requesting changes on M1; the Lows are cheap follow-ups.

cmd/cloudemu/lifecycle.go
thzgajendra: Fixed. Readiness is now an **admin-independent TCP-accept probe** for every provider — `waitServerReady` (and `daemonReachable`) call `pollTCP(hostPortOf(ep))` instead of hitting `/_cloudemu/health`. Since `serve` binds every listener before writing the endpoints file, "accepts a connection" is a sufficient readiness signal and no longer depends on the admin control plane. `pollHealth`/`httpHealthURL` are removed.

Verified E2E: `cloudemu start --admin=false --providers aws --aws-port 4577` now returns exit 0 with the server up (`/_cloudemu/health` → 404, but `status`/`stop` work); previously it timed out and `killChild` killed the healthy server. Added a regression guard `TestWaitServerReady` case: a bare TCP listener under the `aws` key (no HTTP handler, i.e. the `--admin=false` shape) is seen as ready. `2237be6`

cmd/cloudemu/lifecycle.go
thzgajendra: Fixed. `start` now strips any user-supplied `--endpoints-file` (with its value) and `--quiet` from the forwarded args before injecting its own, via `stripFlag(rest, "endpoints-file", true)` / `stripFlag(rest, "quiet", false)` — so last-wins parsing can't redirect serve's output away from the run dir or duplicate the flag. `start` owns both flags.

Verified E2E: `cloudemu start ... --endpoints-file /tmp/should-be-ignored.json --quiet` starts normally and never creates that file. Added `TestStripFlag` (space-separated and `=` forms). `2237be6`

thzgajendra:
Thanks for the deep review. Pushed `2237be6`.

**M1 (fixed)** — `start --admin=false` killing a healthy server. Readiness and the PID-reuse identity check are now provider-agnostic TCP-accept probes (`pollTCP`), not `/_cloudemu/health` hits, so they no longer depend on the admin control plane being mounted. `serve` binds every listener before writing the endpoints file, so TCP-accept is a sufficient readiness signal. `pollHealth`/`httpHealthURL` removed. E2E-verified `start --admin=false` now starts (exit 0, server up, `/health` 404), plus a `TestWaitServerReady` regression case using a bare listener (the `--admin=false` shape).

**L2 (fixed)** — a forwarded `--endpoints-file`/`--quiet`. `start` now strips user copies of both before injecting its own (`stripFlag`), so last-wins parsing can't redirect the endpoints file. E2E-verified + `TestStripFlag`.

**L3 (addressed)** — added the `--admin=false`-shaped readiness unit test (`TestWaitServerReady` bare-listener case), which is exactly what would have caught M1. I skipped the full built-binary spawn→status→stop test to keep the unit suite from shelling out and building itself, but I ran that flow by hand against the compiled binary (start `--admin=false` → status → stop, and the L2 flag case) as part of this fix.

**L1 (won't-fix, with reason)** — gosec G703 on the `--home` fs ops is a version-drift false positive. The repo's CI runs golangci-lint **advisory/report-only** (`ci.yml:170-171`, `|| echo '::warning::…'`), so it never blocks, and G703 doesn't fire on golangci-lint 2.4.0 (my local pin) — a `//nolint:gosec` would itself be flagged as an *unused directive* by nolintlint on the non-firing versions, making the gate worse rather than better. `--home` is a user-supplied CLI arg, not a trust boundary, so the finding is a genuine FP. Happy to add the `//nolint` if the repo bumps its pinned linter to a version that fires it.

Gate: `go build ./...`, `go vet`, `gofmt`, `go test -race ./cmd/cloudemu/`, and `golangci-lint --new-from-rev` = 0.

NitinKumar004 (approved):
## Approving — M1 and L2 resolved

Verified the fix (`2237be62`) in an isolated worktree at the PR head; gate green (`build` / `vet` / `go test -race ./cmd/cloudemu/` `ok` / `gofmt` clean, no dangling refs to the removed helpers).

**M1 (the real bug) is fixed correctly.** Readiness and `daemonReachable` now use a plain TCP-accept probe for every provider — the `/_cloudemu/health` HTTP dependency is gone. That's sound: `serve` binds every listener before writing the endpoints file, so "accepts a connection" is a sufficient, admin-independent ready signal, and `cloudemu start --admin=false` no longer boots a healthy server the probe can't see and kills. Nice touch adding `TestWaitServerReady` as an explicit regression guard for the no-admin case.

**L2 is fixed** — `stripFlag` drops a user-supplied `--endpoints-file`/`--quiet` before forwarding (covering `--flag value`, `--flag=value`, and single-dash forms), backed by `TestStripFlag`, so last-wins flag parsing can no longer redirect serve's output and break the handshake.

On **L1** (gosec G703): confirmed non-blocking — it's linter-version drift (newer bundled gosec taint analysis), there's no gosec config opt-in in `.golangci.yml`, and CI doesn't run golangci-lint. It's a false-positive on the user's own `--home` flag; if a pinned linter ever surfaces it, a `//nolint:gosec` closes it. Not a merge concern.

Clean, well-tested turnaround — thanks. LGTM.

PR #341, opened 2026-08-08
Add Cost Preview: Per-SKU, Per-Region Monthly Estimate (#335 P2)

thzgajendra:
## Objective / Issue
P2 differentiator of the "minikube for cloud resources" roadmap (#335). `cloudemu cost` estimates the monthly cost of what you've provisioned in the local emulator — catch "why is this so expensive?" before it hits a real bill. LocalStack has no built-in cost view; this is free and local.

## What we found
- `services/cost` is a per-*operation* usage tracker (not wired into serve), and `resourcediscovery.Resource` carries SKU/region but no price. So a cost preview must be an **inventory → estimated-monthly** calculation with its own rate model — there's no dollar figure to surface as-is.
- Each provider already exposes a `ResourceDiscovery` engine (`ListAll`) with per-resource `Service`/`Type`/`SKU`/`Region`/`Properties` — enough to price per-SKU and per-region.

- **New `services/pricing` package** — representative on-demand rate tables gathered across AWS/Azure/GCP: **97 compute SKUs**, **28 region multipliers**, **25 DB classes**, **14 flat provisioned rates** (EKS/AKS/GKE control planes, load balancers, NAT, idle IPs), **11 disk $/GB-month rates** (EBS / Azure managed disks / GCE PDs). `Monthly(provider, service, type, sku, region, props)` = per-SKU hourly × per-region multiplier × 730h for compute/DB; flat×730 for control planes / idle IPs / load balancers / NAT gateways; $/GB-month × size for block volumes (AWS/Azure/GCP); usage-based/unknown → 0, with sane defaults for unknown SKUs/regions.
- **Priced vs. not:** compute instances, managed DB instances, K8s control planes, idle IPs, and block volumes across all three providers are priced today. Load balancers and NAT gateways have rates wired and price *once they're discoverable* — that discovery lands in #365, after which they light up with no further change here. **Object storage is intentionally not priced**: it's billed per-GB *stored*, which the inventory can't observe, so buckets stay on the usage-based $0 path rather than showing a misleading flat number.
- **serve** captures each provider's `ResourceDiscovery` in `rebuild()`; `GET /_cloudemu/cost` sums the estimates → JSON, routed through the admin control plane's `extra` handler (alongside `/net/*`).
- **CLI:** `cloudemu cost [--json]` — a per-provider/service table + total.

```mermaid
sequenceDiagram
    actor U as You
    participant CLI as cloudemu cost
    participant D as discovery engines
    participant P as pricing
    U->>CLI: cloudemu cost
    CLI->>S: GET /_cloudemu/cost
    S->>D: ListAll (all providers)
    D-->>S: resources (service/type/sku/region)
    S->>P: Monthly(...) per resource
    P-->>S: per-SKU × per-region × 730
    S-->>U: table + total
```

## Alternatives not taken
- **Wiring the per-operation tracker into every handler** — cross-cutting and only reflects operations since start, not current inventory.
- **Live pricing-API mirror** — the only path to billing-accurate parity, but it's a real API-integration effort; called out as a follow-up. This PR ships representative on-demand rates instead.

## How the rate tables were built
Gathered via a parallel research pass across compute/DB/storage/network/regions per cloud. Values are **representative on-demand prices (~2 significant figures), not a live feed** — documented in the package and the docs.

## Docs / Tests / Playground
- Docs: `docs/standalone-server.md` — cost section (per-SKU/region, honest scope, follow-up).
- Tests: `pricing` (per-SKU, per-region sa-east-1>us-east-1, DB, flat control planes/IPs, load balancer + NAT, volume-by-size using the walker's real `diskSizeGB` key, Azure/GCP disks, usage-based=0, unknown-SKU fallback); `serveCost` (sums inventory, empty→$0).

## Test plan
- [x] `go test -race ./services/pricing/... ./cmd/cloudemu/...`
- [x] Full local CI: gofmt / build / vet / `go test ./...` / tidy / golangci-lint = clean (CodeQL: only pre-existing findings, none in changed files).
- [x] **Live E2E via `cloudemu start` + real `aws` CLI:** t3.micro → **$7.59/mo**, m5.xlarge → **$140.16/mo** (distinct per-SKU, summed); S3 bucket excluded (usage-based); `--json` structured; empty inventory → $0.

## Risk & Rollback
- Low: read-only endpoint, gated on `--admin`; additive. Rates are estimates, clearly labeled. Rollback = revert.

## Conclusion
`cloudemu cost` gives a per-SKU, per-region monthly estimate of your local cloud — a free cost-preview no other local emulator offers.

Follow-up (#335): live pricing-API integration for billing-accurate figures; usage-based metering (storage ops, throughput, data transfer); LB/NAT pricing activates with the discovery work in #365.

### DIFF
--- a/services/pricing/pricing.go
+++ b/services/pricing/pricing.go
@@ -1,0 +1,347 @@
+// Package pricing turns a discovered cloud resource into an approximate monthly
+// USD cost, using per-SKU compute/DB rates, per-region multipliers, and flat
+// rates for always-on provisioned resources. Values are representative on-demand
+// prices (not a live pricing-API feed) — a preview to catch cost surprises
+// early, not a billing-accurate figure.
+package pricing
+
+// HoursPerMonth turns an hourly rate into a monthly estimate (~730h).
+const HoursPerMonth = 730
+
+const (
+	defaultComputeHourly = 0.10 // fallback for an unknown compute SKU
+	defaultDBHourly      = 0.10 // fallback for an unknown DB class
+	defaultRegionMult    = 1.0  // regions absent from the table
+)
+
+// Provider identifiers, matching the resourcediscovery Provider field.
+const (
+	provAWS   = "aws"
+	provGCP   = "gcp"
+	provAzure = "azure"
+)
+
+// computeHourly entries are approximate on-demand rates gathered from public
+// pricing knowledge (not a live feed); accurate to ~2 significant figures.
+//
+//nolint:gochecknoglobals // static pricing table
+var computeHourly = map[string]float64{
+	"Standard_B1ms":    0.0208,
+	"Standard_B1s":     0.0104,
+	"Standard_B2ms":    0.0832,
+	"Standard_B2s":     0.0416,
+	"Standard_B4ms":    0.166,
+	"Standard_B8ms":    0.333,
+	"Standard_D16s_v3": 0.768,
+	"Standard_D16s_v5": 0.768,
+	"Standard_D2_v5":   0.088,
+	"Standard_D2as_v5": 0.086,
+	"Standard_D2s_v3":  0.096,
+	"Standard_D2s_v5":  0.096,
+	"Standard_D32s_v5": 1.536,
+	"Standard_D4_v5":   0.176,
+	"Standard_D4as_v5": 0.172,
+	"Standard_D4s_v3":  0.192,
+	"Standard_D4s_v5":  0.192,
+	"Standard_D8_v5":   0.352,
+	"Standard_D8as_v5": 0.344,
+	"Standard_D8s_v3":  0.384,
+	"Standard_D8s_v5":  0.384,
+	"Standard_DS1_v2":  0.073,
+	"Standard_DS2_v2":  0.146,
+	"Standard_DS3_v2":  0.293,
+	"Standard_E16s_v5": 1.008,
+	"Standard_E2as_v5": 0.113,
+	"Standard_E2s_v5":  0.126,
+	"Standard_E32s_v5": 2.016,
+	"Standard_E4as_v5": 0.226,
+	"Standard_E4s_v5":  0.252,
+	"Standard_E8as_v5": 0.452,
+	"Standard_E8s_v5":  0.504,
+	"Standard_F16s_v2": 0.677,
+	"Standard_F2s_v2":  0.0846,
+	"Standard_F4s_v2":  0.169,
+	"Standard_F8s_v2":  0.338,
+	"c2-standard-16":   0.8352,
+	"c2-standard-4":    0.2088,
+	"c2-standard-8":    0.4176,
+	"c5.2xlarge":       0.34,
+	"c5.4xlarge":       0.68,
+	"c5.large":         0.085,
+	"c5.xlarge":        0.17,
+	"c6i.2xlarge":      0.34,
+	"c6i.4xlarge":      0.68,
+	"c6i.large":        0.085,
+	"c6i.xlarge":       0.17,
+	"e2-medium":        0.033503,
+	"e2-micro":         0.008376,
+	"e2-small":         0.016751,
+	"e2-standard-2":    0.067006,
+	"e2-standard-4":    0.134012,
+	"e2-standard-8":    0.268024,
+	"m5.2xlarge":       0.384,
+	"m5.4xlarge":       0.768,
+	"m5.large":         0.096,
+	"m5.xlarge":        0.192,
+	"m6i.2xlarge":      0.384,
+	"m6i.4xlarge":      0.768,
+	"m6i.large":        0.096,
+	"m6i.xlarge":       0.192,
+	"n1-standard-1":    0.0475,
+	"n1-standard-16":   0.76,
+	"n1-standard-2":    0.095,
+	"n1-standard-4":    0.19,
+	"n1-standard-8":    0.38,
+	"n2-standard-16":   0.776944,
+	"n2-standard-2":    0.097118,
+	"n2-standard-4":    0.194236,
+	"n2-standard-8":    0.388472,
+	"r5.2xlarge":       0.504,
+	"r5.large":         0.126,
+	"r5.xlarge":        0.252,
+	"r6i.2xlarge":      0.504,
+	"r6i.large":        0.126,
+	"r6i.xlarge":       0.252,
+	"t2.2xlarge":       0.3712,
+	"t2.large":         0.0928,
+	"t2.medium":        0.0464,
+	"t2.micro":         0.0116,
+	"t2.nano":          0.0058,
+	"t2.small":         0.023,
+	"t2.xlarge":        0.1856,
+	"t3.2xlarge":       0.3328,
+	"t3.large":         0.0832,
+	"t3.medium":        0.0416,
+	"t3.micro":         0.0104,
+	"t3.nano":          0.0052,
+	"t3.small":         0.0208,
+	"t3.xlarge":        0.1664,
+	"t3a.2xlarge":      0.3008,
+	"t3a.large":        0.0752,
+	"t3a.medium":       0.0376,
+	"t3a.micro":        0.0094,
+	"t3a.nano":         0.0047,
+	"t3a.small":        0.0188,
+	"t3a.xlarge":       0.1504,
+}
+
+// dbHourly entries are approximate on-demand rates gathered from public
+// pricing knowledge (not a live feed); accurate to ~2 significant figures.
+//
+//nolint:gochecknoglobals // static pricing table
+var dbHourly = map[string]float64{
+	"GP_Gen5_16":        2.96,
+	"GP_Gen5_4":         0.74,
+	"GP_Gen5_8":         1.48,
+	"Standard_B1ms":     0.021,
+	"Standard_D2ds_v4":  0.14,
+	"Standard_D4ds_v4":  0.28,
+	"db-custom-1-3840":  0.085,
+	"db-custom-2-7680":  0.17,
+	"db-custom-4-15360": 0.34,
+	"db-f1-micro":       0.015,
+	"db-g1-small":       0.05,
+	"db.m5.2xlarge":     0.684,
+	"db.m5.large":       0.171,
+	"db.m5.xlarge":      0.342,
+	"db.r5.2xlarge":     0.96,
+	"db.r5.large":       0.24,
+	"db.r5.xlarge":      0.48,
+	"db.t3.large":       0.136,
+	"db.t3.medium":      0.068,
+	"db.t3.micro":       0.017,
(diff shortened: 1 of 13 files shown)


services/pricing/pricing.go
```diff
@@ -0,0 +1,347 @@
...
+		return "azure:public-ip-idle"
+	default:
+		return ""
+	}
+}
+
+// sizeFromProps extracts a resource size in GB from the discovery properties,
+// tolerating the several key spellings walkers use. Returns 0 when unknown.
+func sizeFromProps(props map[string]any) float64 {
+	for _, k := range []string{"SizeGiB", "SizeGB", "sizeGb", "size", "Size", "sizeGiB"} {
```
NitinKumar004: **H1 (High).** The volume walker emits size under key `"diskSizeGB"` (`services/resourcediscovery/walkers.go:136`), which isn't in this key list — so `sizeFromProps` returns 0 for every real discovered volume, `volumeMonthly` hits `size <= 0`, and every volume prices at $0 (AWS included). `pricing_test.go:64` masks this by passing a hand-built `"SizeGiB"` map the walker never produces. Add `"diskSizeGB"` here (or change the walker to emit a recognized key), and update the test to feed the walker's actual key so the wired path is exercised.

services/pricing/pricing.go
```diff
@@ -0,0 +1,347 @@
...
+	}
+
+	return 0
+}
+
+// volumeMonthly prices an AWS EBS volume by its type (SKU) and size, or 0 when
+// the size or tier is unknown.
+func volumeMonthly(provider, sku string, props map[string]any) float64 {
+	size := sizeFromProps(props)
+	if size <= 0 || provider != provAWS {
```
NitinKumar004: **M1 (Medium).** `provider != provAWS` plus the hardcoded `storageGBMonth["aws:ebs-"+sku]` lookup means Azure managed disks and GCE PDs always price at $0 — the 6 `azure:managed-disk-*`/`gcp:pd-*` rate entries are unreachable. Either branch on provider to the right disk-rate prefix, or drop the dead entries.

services/pricing/pricing.go
```diff
@@ -0,0 +1,347 @@
...
+		"relationaldb/MySqlFlexibleServer",
+		"relationaldb/PostgresFlexibleServer":
+		return hourly(dbHourly, sku, defaultDBHourly) * mult * HoursPerMonth
+	case "kubernetes/Cluster":
+		return flatHourly[controlPlaneKey(provider)] * mult * HoursPerMonth
+	case "networking/ElasticIP":
+		return flatHourly[eipKey(provider)] * mult * HoursPerMonth
+	case "compute/Volume":
+		return volumeMonthly(provider, sku, props) * mult
+	default:
```
NitinKumar004, line 400: **H2/M2 (Medium).** This `default: return 0` swallows load balancers, NAT gateways, and object-storage buckets — there's no case for any of them, so all 8 `flatHourly` LB/NAT rates and all 9 object-storage `storageGBMonth` rates are dead. The PR body advertises LB + NAT as priced; today they're $0. Add the cases (and a walker for LB/NAT if discovery doesn't emit them) or correct the body/tables so advertised coverage matches reality.

NitinKumar004 (changes requested):
## Deep review — cost preview

Verified in an isolated worktree at the PR head; gates green (`build` + `GOOS=windows` / `vet` / `go test -race ./services/pricing/... ./cmd/cloudemu/` / `gofmt`). The endpoint itself is safely wired — admin-gated, read-only, reads the discovery map under `rebuildMu` then iterates the snapshot (same pattern as `net`/`snapshot`), nil-guarded, and no panic on nil props / unknown SKU / unknown region / unknown service-type (all edge paths refuted).

The problem is the pricing model: it silently prices most resource categories at **$0**, so the "catch cost surprises" goal is undercut for several always-on cost drivers.

### H1 (High) — every discovered volume prices at $0, and the test hides it
The volume walker emits size as `{"diskSizeGB": v.Size}` (`services/resourcediscovery/walkers.go:136`), but `sizeFromProps` only recognizes `SizeGiB/SizeGB/sizeGb/size/Size/sizeGiB` — not `diskSizeGB`. So `volumeMonthly` sees size 0 and returns $0 for **every** real volume (AWS included). `pricing_test.go:64` passes only because it feeds a synthetic `"SizeGiB"` key the walker never emits — a false green. Fix: recognize `diskSizeGB` (or align the walker's key), and make the test drive the real walker key.

### H2 → treat as Medium: load balancers & NAT gateways are never priced (contradicts the body)
No walker emits a load-balancer/NAT resource and `Monthly` has no `loadbalancer/*` or `*/NATGateway` case, so all 8 `flatHourly` LB/NAT rates are dead — yet the body claims "14 flat provisioned rates (… load balancers, NAT …)" are priced. (Note: #365 makes LBs *discoverable*, but this switch still won't price them.)

### M1 (Medium) — Azure/GCP disks never priced
`volumeMonthly` early-returns unless `provider == aws` and hardcodes `storageGBMonth["aws:ebs-"+sku]`, so the 6 `azure:managed-disk-*` / `gcp:pd-*` rates can never be reached.

### M2 (Medium) — object storage never priced
Buckets are discovered as `storage/Bucket`, but `Monthly` has no such case, so all 9 S3/Blob/GCS `storageGBMonth` rates are dead.

Net: only compute instances, managed DBs, K8s control planes, and idle IPs ever yield a non-zero figure. Requesting changes: fix H1 (real bug + misleading test), and for H2/M1/M2 either wire the cases or make the rate tables and the PR/docs match what's actually priced (don't advertise coverage that returns $0). No AI attribution.

services/pricing/pricing.go
thzgajendra: Fixed in e5fcfc2cb3a0a51ca2b0b2203103c4a78294fd0f. Added `diskSizeGB` (the key the volume walker actually emits) to `sizeFromProps`, so discovered volumes price instead of returning $0. Updated `TestMonthlyVolumeBySize` to feed `diskSizeGB` so the wired discovery→pricing path is exercised, not a synthetic key. Thanks for catching the false green.

services/pricing/pricing.go
thzgajendra: Fixed in e5fcfc2cb3a0a51ca2b0b2203103c4a78294fd0f. `volumeMonthly` is now provider-aware via a `diskRateKey(provider, sku)` mapper: AWS `ebs-<sku>`, GCP `gcp:<sku>` (pd-ssd/pd-balanced/pd-standard), and Azure Premium_LRS/StandardSSD_LRS/Standard_LRS → the managed-disk rates. Unknown SKUs return "" and price at $0 rather than guessing. Added a test covering Azure and GCP disks.

services/pricing/pricing.go
thzgajendra: Fixed in e5fcfc2cb3a0a51ca2b0b2203103c4a78294fd0f, split by what discovery can actually observe:

- **LB/NAT (H2):** added `loadbalancer/LoadBalancer` and `networking/NatGateway` cases to `Monthly` (with `lbKey`/`natKey` per-provider helpers), so the flat rates are wired and price correctly. On this branch discovery doesn't yet emit LB/NAT — that lands in #365 — so they light up automatically once merged. The `Monthly` doc comment and PR body now say exactly that instead of claiming they're priced today.
- **Object storage (M2):** buckets are discovered but carry no size, and object storage is billed per-GB *stored* — which discovery can't observe — so a flat estimate would be misleading. Removed the 9 dead S3/Blob/GCS rate entries and left buckets on the usage-based $0 path, matching the other usage-based services. Body updated to not advertise object-storage pricing.

thzgajendra:
Thanks for the deep review — all three findings were real. Fixes pushed:

- **H1 (volumes priced $0):** `sizeFromProps` now recognizes `diskSizeGB`, the key the volume walker actually emits, so discovered volumes price correctly. `TestMonthlyVolumeBySize` now feeds `diskSizeGB` so the wired path is exercised instead of a synthetic key.
- **M1 (Azure/GCP disks $0):** `volumeMonthly` is provider-aware via a `diskRateKey(provider, sku)` mapper (AWS `ebs-<sku>`, GCP `gcp:<sku>`, Azure Premium/StandardSSD/Standard_LRS → managed-disk rates); unknown SKUs price at $0 rather than guessing. Added Azure + GCP disk tests.
- **H2 (LB/NAT):** added `loadbalancer/LoadBalancer` and `networking/NatGateway` cases so the flat rates are wired and price correctly; they activate once those resources are discoverable (#365). Doc comment, PR body, and docs now say that instead of claiming they're priced today.
- **M2 (object storage):** buckets carry no size and object storage is billed per-GB *stored* (not observable from inventory), so a flat estimate would mislead. Removed the 9 dead S3/Blob/GCS rate entries; buckets stay on the usage-based $0 path. Body + docs updated to not advertise object-storage pricing.

Gates green locally: `go test -race ./services/pricing/... ./cmd/cloudemu/`, full `go test ./...`, `go vet`, `gofmt`, `go mod tidy`, and `golangci-lint --new-from-rev` (0 issues).

NitinKumar004 (approved):
## Approving — pricing gaps resolved

Re-reviewed the fix commits (`e5fcfc2` pricing + `94d9ef9` docs) in an isolated worktree at the PR head; gates green (`build` + `GOOS=windows` / `vet` / `go test -race ./services/pricing/... ./cmd/cloudemu/` / `gofmt`). Every prior finding is resolved, verified end-to-end:

- **H1 (volume size-key mismatch) — fixed.** `sizeFromProps` now recognizes `diskSizeGB` (the key the walker actually emits), and the test drives that real key rather than a synthetic one. Live e2e (CreateVolume → ListAll → Monthly) prices non-zero on all three: AWS gp3 100 GB $8.00, Azure Premium_LRS 100 GB $13.50, GCP pd-ssd 50 GB $8.50.
- **M1 (Azure/GCP disks) — fixed.** `diskRateKey` now branches per provider (AWS `aws:ebs-`, GCP `gcp:`, Azure SKU→managed-disk tier); the `provider==aws` guard is gone.
- **H2 (LB/NAT) — honestly scoped.** `Monthly` now has `loadbalancer/LoadBalancer` and `networking/NatGateway` cases, and the docs + code comments state they price once those resources become discoverable (tied to #365) — the earlier overclaim is gone.
- **M2 (object storage) — fixed/documented.** `storage/Bucket` returns $0 by design (usage-based, unobservable from inventory), asserted by a test and stated in the docs.

No new correctness breaks (unknown provider/SKU/region, nil props, negative size all handled; region multiplier applied once; no false-green tests).

One tiny non-blocking cleanup for later: `aws:classic-elb` and `aws:nlb` in `flatHourly` are referenced by no key function (`lbKey` only returns `aws:alb`), so they're dead even once LB discovery lands — drop them or wire them when #365's LB discovery is available. Not a blocker. LGTM.

PR #342, opened 2026-08-08
Add OCI provider foundation: identity, OCIDs, work requests, wire layer

arunesh-j:
## Summary

- Lays the groundwork for **OCI as a fourth provider**, so its ~17 portable services can be built independently and in parallel.
- **No OCI service is implemented here.** This is only the scaffolding every one of them needs, landed once so per-service branches never have to restructure shared files.
- Additive: no existing provider's behavior changes. Full suite green (236 packages).

## Changes

**Shared primitives**
- `config` — `TenancyOCID` / `CompartmentID` / `Realm` with their `With*` options, plus `OCIRegion()`, which substitutes an OCI region when `Region` is still the AWS default. Without it every OCID would carry a region code derived from `us-east-1`. `CompartmentID` defaults to the tenancy *after* options apply, so a caller setting only a custom tenancy doesn't inherit the built-in root.
- `internal/idgen` — `OCID()` / `GlobalOCID()` and a 37-region code table. Identity resources are region-agnostic and emit the doubled dot real OCI returns (`ocid1.compartment.oc1..aaaa…`), so clients splitting on `.` still see five segments. Unknown regions fall back to a per-region code rather than a shared one, which would let two regions collide.
- `services/scope` — `Compartment` field. Matching is exact; real OCI only descends the compartment tree when `compartmentIdInSubtree` is set.

**Wire layer**
- `server/wire/ocirest` — JSON/error encoding, `opc-*` headers, pagination, and the required `compartmentId` parameter. `NotFound` and `PermissionDenied` both map to `404 NotAuthorizedOrNotFound`, since OCI does not distinguish them.
- `server/oci/workrequest` — OCI's async envelope. Registered ahead of service handlers and matched on any path containing `workRequests`, because the collapsed server cannot tell one service's version prefix from another's — the same reasoning as `server/gcp/lro`. Mutations are synchronous, so an accepted request is already `SUCCEEDED`; the envelope carries the new resource's OCID back to SDK waiters.

**Aggregators**
- `providers/oci` + `server/oci` — `Provider` and `Drivers` with all 17 service slots declared up front as *driver interfaces*, so a service that hasn't landed reads as `nil` rather than failing to compile, and a service branch fills a slot instead of editing the struct. Monitoring is cross-wired by type assertion.

**Plumbing**
- `cmd/cloudemu` — `--providers` accepts `oci`, `--oci-port` (4571; 4570 is the k8s data plane), endpoint JSON and banner. OCI is **not** in the default provider set until its services land, so the default never binds a port serving nothing.
- `docs/oci-conventions.md` — the contract each service PR follows.
- PR template gains an OCI checkbox.

`printBanner` is now table-driven: the added rows pushed it past the cyclomatic limit and `endpointSet` past `hugeParam`. The refactor also clears five pre-existing `wsl` warnings in that file.

## Provider Coverage

- [ ] AWS
- [ ] Azure
- [ ] GCP
- [x] OCI

## Checklist

- [x] Linter passes for every package touched here
- [ ] Integration tests added to `cloudemu_test.go` — deferred to the first service PR; there is no OCI service to integration-test yet

## Test Plan

`go test ./...` — exit 0, **236 packages ok, 0 failures**. The full suite matters here because `config` and `scope` are shared by every provider.

`golangci-lint run --timeout=9m` — clean for all new packages. `cmd/cloudemu/endpoints.go` now reports fewer issues than before this branch.

New unit tests:
- `internal/idgen` — region-code mapping, per-region distinctness, OCID shape for region-scoped vs. global resources, uniqueness, five-segment invariant
- `config` — OCI defaults, compartment-defaults-to-custom-tenancy, explicit compartment wins, `OCIRegion()` substitution, and that OCI options leave AWS/GCP defaults untouched
- `server/wire/ocirest` — error-code mapping for all nine canonical codes, JSON decode failure, required `compartmentId`, limit fallbacks, `opc-next-page` omitted on last page
- `server/oci/workrequest` — accept/get/list, compartment filtering, creation order, handler routing (including what `Matches` must *not* claim), sub-collections, 404 shape
- `providers/oci` + `server/oci` — identity defaults and overrides, discovery wiring, nil service slots, `DriversFrom` round-trip, 501 for unrouted paths

Manual, against the built binary:

```
$ cloudemu serve -providers aws,gcp,oci
  AWS         http://127.0.0.1:4566
  GCP         http://127.0.0.1:4569
  OCI         http://127.0.0.1:4571

GET /20160918/workRequests        -> 200 [] , Opc-Request-Id set
GET /20160918/workRequests/nope   -> 404 {"code":"NotAuthorizedOrNotFound",...}
GET /20160918/instances           -> 501  (no services yet)
```

## Related Issues

None yet — happy to open a tracking issue for the OCI Tier 1 rollout if that's preferred.

## Follow-ups

Once this merges, the ~17 portable services (Object Storage, Compute, VCN, Identity, Monitoring, DNS, Load Balancer, Queue, Cache, Vault, Logging, Notifications, OCIR, Events, Functions, NoSQL, Database) can be cut from `development` in parallel, one branch and one PR each, per `docs/oci-conventions.md`.

### DIFF
--- a/internal/idgen/ocid.go
+++ b/internal/idgen/ocid.go
@@ -56,8 +56,10 @@ var regionCodes = map[string]string{
 }
 
 // RegionCode returns the three-letter code for an OCI region name. An
-// unrecognized region falls back to its city segment, so regions this table
-// predates still get distinct codes.
+// unrecognized region falls back to the first three letters of its city
+// segment, which is not the real IATA code and can collide with another
+// unmapped city sharing that prefix. The table covers every current
+// commercial region, so the fallback only applies to ones added later.
 func RegionCode(region string) string {
 	if code, ok := regionCodes[region]; ok {
 		return code
--- a/internal/idgen/ocid.go
+++ b/internal/idgen/ocid.go
@@ -1,0 +1,104 @@
+package idgen
+
+import (
+	"fmt"
+	"strings"
+)
+
+// DefaultRealm is the OCI realm used when none is configured.
+const DefaultRealm = "oc1"
+
+// regionCodeLen is the length of the region code embedded in an OCID.
+const regionCodeLen = 3
+
+// regionCodes maps an OCI region name to the three-letter code that appears
+// inside an OCID.
+//
+//nolint:gochecknoglobals // static lookup table, never mutated
+var regionCodes = map[string]string{
+	"af-johannesburg-1": "jnb",
+	"ap-chuncheon-1":    "yny",
+	"ap-hyderabad-1":    "hyd",
+	"ap-melbourne-1":    "mel",
+	"ap-mumbai-1":       "bom",
+	"ap-osaka-1":        "kix",
+	"ap-seoul-1":        "icn",
+	"ap-singapore-1":    "sin",
+	"ap-sydney-1":       "syd",
+	"ap-tokyo-1":        "nrt",
+	"ca-montreal-1":     "yul",
+	"ca-toronto-1":      "yyz",
+	"eu-amsterdam-1":    "ams",
+	"eu-frankfurt-1":    "fra",
+	"eu-madrid-1":       "mad",
+	"eu-marseille-1":    "mrs",
+	"eu-milan-1":        "lin",
+	"eu-paris-1":        "cdg",
+	"eu-stockholm-1":    "arn",
+	"eu-zurich-1":       "zrh",
+	"il-jerusalem-1":    "mtz",
+	"me-abudhabi-1":     "auh",
+	"me-dubai-1":        "dxb",
+	"me-jeddah-1":       "jed",
+	"mx-monterrey-1":    "mty",
+	"mx-queretaro-1":    "qro",
+	"sa-bogota-1":       "bog",
+	"sa-santiago-1":     "scl",
+	"sa-saopaulo-1":     "gru",
+	"sa-valparaiso-1":   "vap",
+	"sa-vinhedo-1":      "vcp",
+	"uk-cardiff-1":      "cwl",
+	"uk-london-1":       "lhr",
+	"us-ashburn-1":      "iad",
+	"us-chicago-1":      "ord",
+	"us-phoenix-1":      "phx",
+	"us-sanjose-1":      "sjc",
+}
+
+// RegionCode returns the three-letter code for an OCI region name. An
+// unrecognized region falls back to its city segment, so regions this table
+// predates still get distinct codes.
+func RegionCode(region string) string {
+	if code, ok := regionCodes[region]; ok {
+		return code
+	}
+
+	if region == "" {
+		return ""
+	}
+
+	parts := strings.Split(strings.ToLower(region), "-")
+
+	city := parts[0]
+	if len(parts) > 1 {
+		city = parts[1]
+	}
+
+	if len(city) >= regionCodeLen {
+		return city[:regionCodeLen]
+	}
+
+	return city
+}
+
+// OCID generates an Oracle Cloud Identifier of the form
+// ocid1.<type>.<realm>.<region>.<unique>. Identity resources are global and
+// pass an empty region, producing the doubled dot real OCI emits.
+func OCID(resourceType, realm, region string) string {
+	if realm == "" {
+		realm = DefaultRealm
+	}
+
+	return fmt.Sprintf("ocid1.%s.%s.%s.%s", resourceType, realm, RegionCode(region), uniqueSuffix())
+}
+
+// GlobalOCID generates an OCID for a region-agnostic identity resource.
+func GlobalOCID(resourceType, realm string) string {
+	return OCID(resourceType, realm, "")
+}
+
+// uniqueSuffix returns the opaque trailing segment of an OCID, keeping the
+// leading "a" run of a real one and the shared monotonic counter.
+func uniqueSuffix() string {
+	return fmt.Sprintf("aaaaaaaa%016x", next())
+}
--- a/server/oci/workrequest/workrequest.go
+++ b/server/oci/workrequest/workrequest.go
@@ -1,0 +1,210 @@
+// Package workrequest provides OCI's asynchronous work request envelope.
+//
+// Real OCI returns 202 with an opc-work-request-id from most mutating calls,
+// and SDK waiters poll GET /{version}/workRequests/{id} until the status is
+// terminal. Each service publishes that endpoint under its own API version
+// prefix; CloudEmu collapses every service onto one HTTP server, so this
+// handler claims any path ending in workRequests and answers uniformly.
+//
+// Every CloudEmu mutation completes synchronously, so an accepted work request
+// is already SUCCEEDED — the envelope exists to keep SDK waiters happy and to
+// carry the created resource's OCID back to them.
+package workrequest
+
+import (
+	"net/http"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/stackshy/cloudemu/v2/config"
+	"github.com/stackshy/cloudemu/v2/internal/idgen"
+	"github.com/stackshy/cloudemu/v2/internal/memstore"
+	"github.com/stackshy/cloudemu/v2/server/wire/ocirest"
+)
+
+// Work request lifecycle states.
+const (
+	StatusAccepted   = "ACCEPTED"
+	StatusInProgress = "IN_PROGRESS"
+	StatusSucceeded  = "SUCCEEDED"
+	StatusFailed     = "FAILED"
+)
+
+// Action types a work request reports against an affected resource.
+const (
+	ActionCreated    = "CREATED"
+	ActionUpdated    = "UPDATED"
+	ActionDeleted    = "DELETED"
+	ActionInProgress = "IN_PROGRESS"
+	ActionRelated    = "RELATED"
+)
+
+const segment = "workRequests"
+
+// Resource is a resource a work request affected.
+type Resource struct {
+	EntityType string `json:"entityType"`
+	ActionType string `json:"actionType"`
+	Identifier string `json:"identifier"`
+	EntityURI  string `json:"entityUri,omitempty"`
+}
+
+// WorkRequest is OCI's asynchronous operation record.
+type WorkRequest struct {
+	ID              string     `json:"id"`
+	OperationType   string     `json:"operationType"`
+	Status          string     `json:"status"`
+	CompartmentID   string     `json:"compartmentId"`
+	Resources       []Resource `json:"resources"`
+	PercentComplete float32    `json:"percentComplete"`
+	TimeAccepted    string     `json:"timeAccepted"`
+	TimeStarted     string     `json:"timeStarted,omitempty"`
+	TimeFinished    string     `json:"timeFinished,omitempty"`
+}
+
+// Store records work requests so SDK waiters can poll them.
+type Store struct {
+	requests *memstore.Store[*WorkRequest]
+	opts     *config.Options
+	mu       sync.RWMutex
+	order    []string
+}
+
+// New creates a work request store.
+func New(opts *config.Options) *Store {
+	return &Store{
+		requests: memstore.New[*WorkRequest](),
+		opts:     opts,
+	}
+}
+
+// Accept records a completed work request and returns its OCID, which the
+// caller stamps as opc-work-request-id.
+func (s *Store) Accept(operationType, compartmentID string, resources ...Resource) string {
+	now := s.opts.Clock.Now().UTC().Format(time.RFC3339)
+	id := idgen.OCID("workrequest", s.opts.Realm, s.opts.OCIRegion())
+
+	wr := &WorkRequest{
+		ID:              id,
+		OperationType:   operationType,
+		Status:          StatusSucceeded,
+		CompartmentID:   compartmentID,
+		Resources:       resources,
+		PercentComplete: 100,
+		TimeAccepted:    now,
+		TimeStarted:     now,
+		TimeFinished:    now,
+	}
+
+	s.requests.Set(id, wr)
+
+	s.mu.Lock()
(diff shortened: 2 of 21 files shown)


server/wire/ocirest/ocirest.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+}
+
+// SetWorkRequestID stamps the work request a client polls for completion.
+func SetWorkRequestID(w http.ResponseWriter, id string) {
+	w.Header().Set(HeaderWorkRequestID, id)
+}
+
+// setRequestID echoes the caller's opc-request-id, or mints one.
+func setRequestID(w http.ResponseWriter) {
```
thzgajendra: **[nit] Comment says it echoes the caller's `opc-request-id`, but it doesn't.** This reads `w.Header().Get(...)` — the *response* writer — and always mints a fresh id; the client's request `opc-request-id` is never read. Real OCI echoes the request id back in the response, and SDKs/log-correlation rely on that. Either thread the `*http.Request` through and `w.Header().Set(HeaderRequestID, r.Header.Get(HeaderRequestID))` when present, or drop "echoes the caller's" from the comment. Low impact for an emulator, but the comment currently describes behavior the code doesn't have.

server/oci/workrequest/workrequest.go
```diff
@@ -0,0 +1,210 @@
...
+// ServeHTTP answers a single work request, its sub-collections, or a list.
+func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	id, sub, ok := parse(r.URL.Path)
+	if !ok {
+		ocirest.WriteError(w, http.StatusBadRequest, "InvalidParameter", "malformed work request path")
+		return
+	}
+
+	if id == "" {
+		ocirest.WriteJSON(w, http.StatusOK, h.store.List(ocirest.CompartmentID(r)))
```
thzgajendra: **[low] List returns work requests across ALL compartments when `compartmentId` is absent.** This uses `ocirest.CompartmentID(r)` (optional) rather than `RequireCompartmentID`. Real OCI `ListWorkRequests` requires `compartmentId` and 400s without it. For fidelity, consider requiring it here (the helper already exists and is used elsewhere in this PR), or confirm the permissive all-compartments listing is an intentional emulator convenience.

internal/idgen/ocid.go
```diff
@@ -0,0 +1,104 @@
...
+	"uk-cardiff-1":      "cwl",
+	"uk-london-1":       "lhr",
+	"us-ashburn-1":      "iad",
+	"us-chicago-1":      "ord",
+	"us-phoenix-1":      "phx",
+	"us-sanjose-1":      "sjc",
+}
+
+// RegionCode returns the three-letter code for an OCI region name. An
+// unrecognized region falls back to its city segment, so regions this table
```
thzgajendra: **[nit] The fallback still allows region-code collisions.** The comment says unknown regions get a distinct code "so regions this table predates still get distinct codes," but `city[:3]` collides for any two future regions whose city shares the first three letters, and gives a mapped city a different code than its real IATA one (e.g. an unknown `…-milano-…` → `mil` vs the table's `eu-milan-1` → `lin`). It only reduces, not eliminates, collisions. Fine as a fallback for the (currently complete) 37-region table — just worth softening the claim.

thzgajendra (commented):
## Review — OCI provider foundation (from cloudemu's perspective)

Reviewed the OCI-fidelity and cloudemu-convention aspects in depth, built the binary, ran the OCI-touched packages, and **E2E-tested against a running server** (`serve -providers aws,gcp,oci`). This is a clean, faithful, well-scoped foundation — **no blockers, no must-fix.** A handful of LOW/nits below.

### Verified — OCI fidelity is good
- **OCID format** (`idgen/ocid.go`): `ocid1.<type>.<realm>.<region>.<suffix>`, the **doubled-dot** for global/identity resources, the 5-segment invariant preserved on `split('.')`, the `aaaaaaaa` suffix prefix, and correct **IATA region codes** (iad/phx/fra/nrt/bom/lhr/icn/sin/gru/yyz/zrh/dxb…). Matches real OCI.
- **Error wire shape** (`server/wire/ocirest`): `{code,message}`, and the mappings are right — NotFound **and** PermissionDenied → `404 NotAuthorizedOrNotFound` (OCI conflates them), AlreadyExists→`409 Conflict`, FailedPrecondition→`409 IncorrectState`, Throttled→`429 TooManyRequests`, `opc-*` headers, `opc-next-page` omitted on last page.
- **Work-request envelope** (`server/oci/workrequest`): correct async model (SUCCEEDED-immediately since cloudemu mutations are synchronous), faithful `WorkRequest` shape, and `Matches` mirrors `server/gcp/lro` (claims any `workRequests` path regardless of version prefix) — the right call for the collapsed server.
- **Config** (`config/options.go`): tenancy/compartment/realm defaults, `OCIRegion()` substitution when `Region` is still the AWS default, and compartment-defaults-to-custom-tenancy (after options apply) — all correct.
- **Scope** (`services/scope`): `Compartment` exact-match, correctly noting real OCI only descends the tree with `compartmentIdInSubtree` — consistent with the Azure/GCP scope pattern.
- **Plumbing**: OCI is opt-in (not in the default set, so no port serves nothing), `--oci-port 4571`, additive provider validation + endpoint/banner, and the 17 slots are **driver interfaces** (nil until filled) reusing the shared portable driver types — the correct cloudemu pattern.

### E2E (server running locally on :4571)
```
GET /20160918/workRequests            -> 200 []            Opc-Request-Id set
GET /20160918/workRequests/<missing>  -> 404 {"code":"NotAuthorizedOrNotFound",...}  application/json
GET /20160918/workRequests/x/errors   -> 404 (id x absent)  # [] only for an existing WR
GET /20160918/instances               -> 501  (no services yet)
GET /20160918/workRequests?compartmentId=... -> 200 []
POST /20160918/instances / root /     -> 501
```
All match the PR's manual test plan. OCID/error/header fidelity confirmed on the wire. (The populated create→poll work-request path isn't E2E-reachable yet — nothing calls `Accept` until a service lands — so it's unit-tested only; noted, not a gap for a foundation PR.) OCI-touched packages: **all green**; build + vet clean.

### 🟡 Low / nits
1. **`setRequestID` doesn't echo the client's `opc-request-id`** (inline, `ocirest.go`). The comment says "echoes the caller's opc-request-id, or mints one," but it reads the *response* writer's header and always mints a fresh id — the request's `opc-request-id` is never read. Real OCI echoes the client's request id back (SDKs correlate on it). Either thread `r` and echo `r.Header.Get("opc-request-id")`, or fix the comment to "mints one."
2. **`ListWorkRequests` without `compartmentId` returns ALL** (inline, `workrequest.go`). The list path uses `CompartmentID` (optional) not `RequireCompartmentID`; real OCI `ListWorkRequests` requires `compartmentId` (400 if absent). Consider requiring it for fidelity, or confirm the permissive behavior is intended for the emulator.
3. **Region-code fallback can still collide** (inline, `idgen/ocid.go`). The comment says the `city[:3]` fallback avoids the collision a shared code would cause, but two future regions whose city shares the first three letters still collide (and an unknown variant of a mapped city gets a different code than the table's IATA one). Only affects regions not yet in the 37-entry table (which covers all current commercial regions) — worth softening the claim.
4. **`DefaultRealm` is defined twice** (`config/options.go` and `internal/idgen/ocid.go`), both `"oc1"` — one referencing the other would prevent drift.
5. **`Limit` has no upper cap** (`ocirest.go`) — a client can request an arbitrarily large page; real OCI caps (~1000). Fine to leave to services, just flagging.
6. **Dead branch:** the `!ok` → `400 malformed work request path` in `workrequest.ServeHTTP` is unreachable because `Matches` uses the same `parse` (a path `parse` rejects → 501, confirmed E2E on `.../workRequests/a/b/c`). Harmless.

**Bottom line:** high-fidelity, additive OCI scaffolding that follows cloudemu's existing provider/wire/scope conventions and behaves correctly on the wire. The nits are cosmetic/fidelity-polish; none block the foundation. Nice groundwork for the per-service PRs.

server/wire/ocirest/ocirest.go
arunesh-j: Fixed in 16606f3. Threaded the `*http.Request` through `WriteJSON`/`WriteError`/`WriteDriverError` so the caller's `opc-request-id` is echoed and one is minted only when none was sent. Chose that over an `EchoRequestID(w, r)` helper handlers call by convention, because 17 service branches would each have to remember it — this way the compiler enforces it, and it's free to change now before any service exists.

Verified on the wire:
```
-H 'opc-request-id: my-trace-id-123'  -> Opc-Request-Id: my-trace-id-123
(no header sent)                      -> Opc-Request-Id: cloudemu00000003
-H 'opc-request-id: err-trace-456' (404) -> Opc-Request-Id: err-trace-456
```

server/oci/workrequest/workrequest.go
arunesh-j: Fixed in 16606f3. Not intentional — you're right that real OCI `ListWorkRequests` requires `compartmentId`, so it now uses `RequireCompartmentID` and 400s without it.

```
GET /20160918/workRequests                    -> 400 {"code":"InvalidParameter","message":"compartmentId is required"}
GET /20160918/workRequests?compartmentId=...  -> 200 []
```

Also added the rule to `docs/oci-conventions.md` so the per-service PRs don't reintroduce the permissive form on their own list endpoints.

internal/idgen/ocid.go
arunesh-j: Fixed in 16606f3 — the claim was wrong, so I softened it rather than changing the code:

```go
// RegionCode returns the three-letter code for an OCI region name. An
// unrecognized region falls back to the first three letters of its city
// segment, which is not the real IATA code and can collide with another
// unmapped city sharing that prefix. The table covers every current
// commercial region, so the fallback only applies to ones added later.
```

Kept the fallback as-is: the alternative is failing on an unknown region, which would be worse for an emulator. `TestRegionCodeIsDistinctPerRegion` still guards the table itself against collisions, which is the case that actually matters today.

arunesh-j:
Thanks for the thorough review — the E2E pass was genuinely useful.

Addressed 5 of the 6 nits in 16606f3, pushed back on one:

| # | Nit | Outcome |
|---|---|---|
| 1 | `opc-request-id` not echoed | **Fixed** — write helpers take `*http.Request`, echo the caller's id, mint only when absent |
| 2 | `ListWorkRequests` without `compartmentId` | **Fixed** — now requires it and 400s, matching real OCI |
| 3 | Region-code fallback overclaims | **Fixed** — comment softened; fallback behavior kept deliberately |
| 4 | `DefaultRealm` defined twice | **Fixed** — `config.DefaultRealm` aliased from `idgen.DefaultRealm` |
| 5 | `Limit` uncapped | **Fixed** — added `MaxLimit = 1000` |
| 6 | Dead `!ok` branch | **Kept** — see below |

On #6: it's unreachable *through* `server.Server`, which calls `Matches` first, but `ServeHTTP` is exported and the package doc explicitly invites mounting handlers directly. On a plain `http.ServeMux` the branch is live. Deleting defensive error handling from an exported handler to remove a branch that's only conditionally dead seemed like the wrong trade, so I documented why it's there instead.

For #1 I picked threading the request over an `EchoRequestID(w, r)` helper called by convention — 17 service branches would each have to remember the latter, whereas the signature change is compiler-enforced and costs nothing now, before any service exists. `docs/oci-conventions.md` is updated for the new signatures and the list rule.

Verification: `go test ./...` 236 packages, exit 0. `golangci-lint` clean for the OCI packages. All five behavior changes confirmed against a running server on :4571.

Still needs a maintainer to click **Approve and run workflows** — CI is sitting at `action_required` since this is a fork PR.

thzgajendra (commented):
## Re-review — commit `16606f3` ("echo opc-request-id, require compartmentId on list")

Verified each fix in source, rebuilt, ran the OCI packages, and **E2E-tested the two behavior changes against a running server**. Everything I raised — the 3 inline findings AND both summary-only nits — is resolved. Nothing outstanding.

| Finding | Resolution (verified) |
|---|---|
| **[nit] `setRequestID` didn't echo the client's `opc-request-id`** | Threaded `*http.Request` through `WriteJSON`/`WriteError`/`WriteDriverError`/`DecodeJSON`/`RequireCompartmentID`; `setRequestID(w, r)` now echoes `r.Header.Get(opc-request-id)` when present, else mints. Comment corrected. All call sites updated (build clean, no stragglers). **E2E:** a request with `opc-request-id: my-custom-req-42` echoes it back; a request with none gets a minted `cloudemu…` id. ✅ |
| **[low] `ListWorkRequests` returned all compartments without `compartmentId`** | List path now uses `RequireCompartmentID`. **E2E:** `GET /workRequests` (no compartmentId) → `400 {"code":"InvalidParameter","message":"compartmentId is required"}`; with `compartmentId` → `200 []`. Matches real OCI. ✅ |
| **[nit] Region-code fallback comment overstated collision-avoidance** | Comment softened to reflect that `city[:3]` only reduces, not eliminates, collisions. ✅ |
| **[nit] `DefaultRealm` duplicated across packages** | `config.DefaultRealm` now aliases `idgen.DefaultRealm`, so the OCID format owner is the single source — they can't drift. ✅ |
| **[nit] `Limit` uncapped** | Added `MaxLimit = 1000` and `return min(n, MaxLimit)`, matching real OCI's page cap. ✅ |

Regression check (E2E): missing work request → `404 NotAuthorizedOrNotFound`, unrouted service → `501` — both intact. `go build ./...` clean; `ocirest`/`server/oci`/`workrequest`/`config`/`idgen` all green (and the prior full run was 236 packages, 0 failures).

Great, responsive follow-up — the OCI foundation is solid and faithful. No blockers, nothing left from my side; leaving the merge call to the maintainers.

thzgajendra (approved):

PR #366, opened 2026-08-08
Add OCI Monitoring: metrics and alarms

### LINKED ISSUE #377: OCI Monitoring: metrics and alarms
Implement **OCI Monitoring** against the existing portable `monitoring` driver.

## Scope

**Metrics** — `PostMetricData`, `ListMetrics`, `SummarizeMetricsData`, compartment-scoped series with bucketed aggregation, and OCI's metric query language.

**Alarms** — `CreateAlarm`, `GetAlarm`, `ListAlarms`, `UpdateAlarm`, `DeleteAlarm`, `GetAlarmHistory`, `ListAlarmsStatus`, with evaluation on metric post and state history.

Wire surface is the `/20180401` REST API.

## Design notes

The portable `Monitoring` interface carries neither a compartment nor an alarm OCID, so a faithful OCI surface needs an optional capability discovered by type assertion, per `docs/oci-conventions.md`.

Alarm mutations are synchronous in real OCI Monitoring, so no work request is recorded.

## Acceptance

- Implements `driver.Monitoring`, with the OCI surface as an optional capability
- OCIDs from `idgen.OCID`

arunesh-j:
## Summary

- First OCI service on the foundation merged in #342.
- Implements the portable `monitoring` driver, plus the OCI surface the portable model cannot express.
- **Merge order: this is 1 of 3.** Follow with #identity, then #vcn — they share wiring lines in `providers/oci/oci.go` and `server/oci/oci.go`.

## Changes

- **`providers/oci/monitoring/`** — `Mock` over `memstore` implementing `driver.Monitoring`: compartment-scoped metric series with bucketed aggregation, OCID-keyed alarm lifecycle, alarm evaluation on metric post, state history, and a parser for OCI's metric query language in its single-metric threshold form (`CpuUtilization[1m].mean() > 80`).
- **`server/oci/monitoring/`** — the `/20180401` REST surface: PostMetricData, ListMetrics, SummarizeMetricsData, and the alarm lifecycle (Create/Get/List/Update/Delete/History/Status).
- **`services/monitoring/driver/driver.go`** — an **additive** optional `OCIMonitoring` capability discovered by type assertion, as `docs/oci-conventions.md` directs and following `storage/driver.BucketAttributes`. The portable `Monitoring` interface carries neither a compartment nor an alarm OCID, so a faithful OCI surface is impossible through it alone. Every line is an insertion; nothing existing changed and no other driver implements it.
- Wiring: one line in `providers/oci/oci.go`, one registration in `server/oci/oci.go`, one section in `docs/services.md`.

## Provider Coverage

- [ ] AWS
- [ ] Azure
- [ ] GCP
- [x] OCI

## Checklist

- [ ] Integration tests added to `cloudemu_test.go` — unit + handler tests instead; happy to add if you'd prefer

## Test Plan

`go test ./...` — exit 0, **238 packages** (236 before, +2 new). `golangci-lint` — 0 issues for `providers/oci`, `server/oci`, `services/monitoring`.

Verified against a running server (`serve -providers oci`):

```
PostMetricData                     -> 200 {"failedMetricsCount":0,...}  opc-request-id echoed
ListMetrics                        -> 200 [CpuUtilization ...]
ListMetrics (no compartmentId)     -> 400 {"code":"InvalidParameter",...}
SummarizeMetricsData               -> 200 aggregatedDatapoints
CreateAlarm                        -> 200 ocid1.alarm.oc1.iad.aaaa...
ListAlarmsStatus                   -> 200 status FIRING
CreateAlarm (duplicate)            -> 409 {"code":"Conflict",...}
UpdateAlarm (partial)              -> 200 severity WARNING
GetAlarmHistory                    -> 200 entries
compartment isolation              -> each compartment sees only its own alarms
DeleteAlarm, then GetAlarm         -> 204, then 404 NotAuthorizedOrNotFound
/20160918/instances                -> 501  (Matches does not over-claim)
```

## Notes for review

Two things I'd value your opinion on:

1. **Alarm `displayName` uniqueness is enforced per compartment** (duplicate → 409). Real OCI does not require it to be unique, but the portable driver is name-keyed (`GetAlarmHistory(ctx, alarmName, limit)`), so allowing duplicates leaves the portable API unable to address one of two same-named alarms. Fidelity vs. portable-API coherence — happy to flip it.
2. **Where OCI-only capabilities should live.** This PR puts them in `services/monitoring/driver/driver.go`; the identity PR uses a separate `services/iam/driver/oci.go`; the vcn PR uses a consumer-side interface in `server/oci/vcn`. All three are defensible, but we should settle on one before the next wave of services. I lean toward the consumer-side interface (zero shared-file surface) — your call.

## Related Issues

Closes #377
Part of #376

Follows #342.


providers/oci/monitoring/metrics.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	return strings.Join([]string{compartmentID, d.Namespace, resourceGroup, d.MetricName, strings.Join(parts, ",")}, "|")
+}
+
+// aggregate buckets samples into fixed periods and reduces each with stat.
+func aggregate(points []metricPoint, start, end time.Time, period int, stat string) (stamps []time.Time, values []float64) {
+	sort.Slice(points, func(i, j int) bool { return points[i].timestamp.Before(points[j].timestamp) })
+
+	step := time.Duration(period) * time.Second
```
NitinKumar004: **High (DoS hang).** `period := int(sel.interval.Seconds())` truncates any sub-second resolution to 0, so `step` here is 0 and the `for bucket := start; bucket.Before(end); bucket = bucket.Add(step)` loop never advances — an infinite loop. `SummarizeMetricsData` passes `Resolution` straight from the wire, so `POST …/summarizeMetricsData` with `CpuUtilization[500ms].mean()` hangs a serving goroutine. `formatSelector` already guards `period<=0` → default; do the same here (guard `period<=0` in `aggregate`).

providers/oci/monitoring/alarms.go
```diff
@@ -0,0 +1,258 @@
...
+	for _, rec := range m.alarms.SortedValues() {
+		if rec.spec.MetricCompartmentID == compartmentID {
+			m.evaluateAlarm(rec)
+		}
+	}
+}
+
+// evaluateAlarm compares an alarm's query against the samples in its window and
+// records any status change.
+func (m *Mock) evaluateAlarm(rec *alarmRecord) {
```
NitinKumar004, line 188: **High (data race).** `evaluateAlarm` reads `rec.spec.{Query,IsEnabled,Namespace,ResourceGroup}` (and `evaluateAlarms` reads `rec.spec.MetricCompartmentID`) with no lock, while `UpdateOCIAlarm` writes `rec.spec` under `m.mu`. `PostMetricData → evaluateAlarms` runs outside the lock (correct for deadlock avoidance), so a concurrent `PUT /alarms/{id}` races these reads — reproduced with `-race`. Snapshot the needed spec fields under `m.mu` (copy), then evaluate against the copy. Add a concurrent `UpdateOCIAlarm`+`PostMetricData` `-race` test — the suite is currently all-sequential, which is why this is green.

providers/oci/monitoring/monitoring.go
```diff
@@ -0,0 +1,304 @@
...
+	if err != nil {
+		return err
+	}
+
+	return m.DeleteOCIAlarm(ctx, a.id)
+}
+
+// DescribeAlarms returns alarms matching the given display names, or every
+// alarm in the default compartment when names is empty.
+func (m *Mock) DescribeAlarms(_ context.Context, names []string) ([]driver.AlarmInfo, error) {
```
NitinKumar004, line 208: **Medium (compartment isolation).** `DescribeAlarms` iterates all alarms filtering only by display name — it never checks `a.spec.CompartmentID` against `opts.CompartmentID`, unlike the by-name siblings (`findByDisplayName`). The doc comment says "every alarm in the default compartment" but the code returns alarms from every compartment, so a portable `DescribeAlarms(nil)` leaks other compartments' alarms. Filter by the default compartment to match the doc + siblings.

NitinKumar004 (changes requested):
## Deep review — OCI Monitoring (first OCI service)

Reviewed in an isolated worktree at the PR head via a multi-agent pass (structure/parity + correctness/concurrency), every High independently re-verified. **Gate green** (build/vet/`go test -race`/whole-suite/gofmt) — but the `-race` run is green only because no test exercises the concurrent or sub-second paths, which is where the two blockers live.

**Structural conformance is exemplary** — implements the existing portable `Monitoring` driver with the OCI surface as an optional `OCIMonitoring` capability (type-asserted, like `BucketAttributes`); `o.OCIRegion()` + `idgen.OCID`; `scope`/`RequireCompartmentID` on every list; `ocirest` throughout with `404 NotAuthorizedOrNotFound` collapse; the 3 sanctioned wiring edits only; docs updated. Nice work.

Holding on two Highs + three Mediums.

### High
1. **Sub-second resolution hangs a serving goroutine (DoS).** `aggregate` computes `step := time.Duration(period)*time.Second` where `period := int(sel.interval.Seconds())`; a `[500ms]` resolution truncates to `period=0` → `step=0` → the `for … bucket.Add(step)` loop never advances. `formatSelector` guards `period<=0` (query.go:149) but `aggregate` doesn't, and `SummarizeMetricsData` passes `Resolution` straight from the wire — so one `POST …/summarizeMetricsData` with `[500ms]` hangs. Inline on metrics.go.
2. **Data race on `rec.spec` in alarm evaluation.** `evaluateAlarm`/`evaluateAlarms` read `rec.spec.{Query,IsEnabled,Namespace,ResourceGroup,MetricCompartmentID}` with no lock, while `UpdateOCIAlarm` writes `rec.spec` under `m.mu`. `PostMetricData` correctly evaluates outside the lock (deadlock avoidance) but then touches shared mutable spec unlocked — reproduced under `-race`. Inline on alarms.go. (This is one instance of a series-wide pattern — see the note below.)

### Medium
3. **Alarms drop `Dimensions`** — `CreateAlarm`/`formatQuery` never encode `cfg.Dimensions` and `evaluateAlarm`'s filter omits them, so an alarm scoped to one resource evaluates across *every* series of that metric name in the compartment. AWS filters by dimensions (`cloudwatch.go` `matchDimensions`). Cross-provider behavior mismatch.
4. **Half-open eval window** — `valuesIn` uses `< to`, so a datapoint at exactly `now` is excluded; under `config.FakeClock` a `PutMetricData` with no explicit timestamp posts at `now` and is excluded from its own evaluation → the alarm won't fire where AWS's would. Inline on monitoring.go / see alarms.go:153.
5. **`DescribeAlarms` is not compartment-scoped** — it filters only by name, never by `spec.CompartmentID`, contradicting its own doc ("default compartment") and its siblings (`findByDisplayName` filters by `opts.CompartmentID`). Portable-layer isolation leak. Inline on monitoring.go.

### Low
No pagination tokens (results truncate past 100 — and this is the first OCI service, so it sets the pattern for the series); unknown aggregation silently becomes `mean`; out-of-range threshold silently never fires; `OKActions`/`InsufficientDataActions` dropped and `Severity` hard-coded.

**Series note:** all three OCI PRs share a "mutate a store-returned pointer outside any lock" data race. Here it's High; worth a shared fix pattern (evaluate against a locked copy of the spec, or hold `m.mu` for the field reads). No AI attribution.

arunesh-j:
Thanks — both Highs were real, and your note that `-race` was green only because nothing exercised the concurrent paths was the useful catch. Fixed in `6dd71a3`.

### High
**1. Sub-second resolution hang.** Root cause was the `int(seconds)` representation, not just the missing guard, so aggregation moved to `time.Duration` end to end: `aggregate(…, step time.Duration, …)`, `resolutionLabel(step)`, and a new `resolutionOf(interval, resolution)` that prefers an explicit `Resolution` and rejects anything below `minResolution = time.Second`. `aggregate` also guards `step <= 0` independently, so the spin is unreachable by any path.

Confirmed on the wire with your exact payload:
```
POST /20180401/metrics/actions/summarizeMetricsData  {"resolution":"500ms"}
-> {"code":"InvalidParameter","message":"resolution 500ms is finer than the 1s minimum"}
   [HTTP 400] elapsed=0.007s      # server still responsive afterwards
```
Tests: `TestSummarizeSubSecondResolution`, `TestAggregateNeverSpinsOnZeroStep`, `TestSummarizeRejectsSubSecondResolution`. Restoring the `int` truncation makes the wire test hang until `go test -timeout` kills it.

**2. Data race on `rec.spec`.** Added `specOf(rec)`, which copies the spec under `RLock`; evaluation runs against the copy, so `PostMetricData`'s deadlock avoidance is untouched and the lock is not widened. The same unlocked-read pattern was fixed in `lookupByName`, `findByDisplayName`, `UpdateOCIAlarm` and `DescribeAlarms`. Test: `TestConcurrentUpdateAndPost` (200×200) — reverting just the two `specOf` calls reproduces the race at `alarms.go:130` vs `alarms.go:91`.

### Medium
3. **Dimensions** — `splitSelector` now parses MQL's `{k = "v"}` predicate, `formatSelector` emits it sorted for determinism, and evaluation filters on it. `GetMetricData` keeps passing dimensions structurally rather than through MQL, since a value containing a quote would not survive the round-trip.
4. **Half-open window** — `valuesIn` grew an `inclusiveEnd`; alarm evaluation passes true, aggregation buckets pass false (adjacent buckets abut, so an inclusive end would double-count a boundary sample).
5. **`DescribeAlarms`** — now filters on `place.Compartment`.

### Low — all fixed
Pagination via an opaque offset cursor over `ocirest.Page`/`Limit`/`SetNextPage` on all three list endpoints (`getAlarmHistory` left alone — its `limit` is last-N, not a cursor); unknown aggregation and unknown/NaN threshold now fail with `InvalidArgument` instead of silently becoming `mean`/`>`; `OKActions`/`InsufficientDataActions` folded into OCI's single destination list, deduped and order-stable; hard-coded `Severity` replaced by a single default in `validateSpec`.

### One judgement call for your eye
Rejecting an unknown `ComparisonOperator` in `CreateAlarm` is a behavior change on the portable path — previously it silently became `>`. It is contained to this provider (AWS's `server/aws/cloudwatch/query.go` passes the raw form value to the AWS mock, not this one), and a permanently-dead alarm seemed worse than a 400. Happy to restore the fallback if you disagree.

**displayName uniqueness** left as-is per the PR note, now with a two-line comment recording that real OCI permits duplicates and that the portable driver's name-keyed addressing is why this mock does not.

Verification: `go build ./...` clean · `go test ./...` exit 0, 238 packages · `go test -race ./providers/oci/... ./server/oci/...` clean · `golangci-lint` 0 issues. `docs/services.md` updated for pagination, the dimension predicate and the resolution floor.

Tracking issue: #377.

NitinKumar004 (approved):

Re-reviewed the fix commit (`6dd71a3`) via a 2-agent pass (fix-verify + regression-hunt); gate green (build / vet / `go test -race -count=2` both Monitoring packages / gofmt / golangci-lint 0). Every prior finding is fixed and I re-verified both Highs myself:

- **High (sub-second resolution hang) — fixed.** Double-guarded: `resolutionOf` rejects a finer-than-1s step with `InvalidArgument` before `aggregate`, and `aggregate` itself guards `step<=0`. A `[500ms]` summarize now returns an error in <1s instead of spinning.
- **High (`rec.spec` data race) — fixed.** All alarm-eval reads go through `specOf`, which copies the spec under `m.mu.RLock`; `evaluate` works on the copy. `evaluateAlarms` does not hold the lock across `specOf` (no re-entrant RLock), and `PostMetricData` calls eval outside the lock — no deadlock. `-race` clean.
- **Mediums — fixed:** alarms now carry `Dimensions` and filter the series on eval (undimensioned alarms still match, backward-compatible); the eval window is inclusive at `now` (fires under FakeClock); `DescribeAlarms` is compartment-scoped.
- **Lows — fixed:** pagination tokens wired; unknown aggregation → parse error (not silent mean); OK/InsufficientData actions folded into OCI destinations.

Determinism, OCI-conventions, and STRUCTURE all still hold. Only two new Lows remain (non-blocking): a dimension value containing a `,` makes an alarm silently un-fireable (`splitDimensions` splits before honoring quotes); alarm dimension matching is subset rather than exact (a defensible OCI-MQL-filter choice). LGTM.

NitinKumar004:

arunesh-j:
Rebased on `development` and ran `go generate ./...` — OCI Monitoring now appears in `docs/coverage/`, and `docs/coverage/oci/README.md` no longer says "foundation only".

```
    OCI credited with "Monitoring", but oci.New() constructs no services
```

```
oci credited with "Monitoring", but oci.New() leaves that field nil
```

providers/oci/monitoring/alarms.go
```diff
@@ -0,0 +1,281 @@
...
+	cerrors "github.com/stackshy/cloudemu/v2/errors"
+	"github.com/stackshy/cloudemu/v2/internal/idgen"
+	"github.com/stackshy/cloudemu/v2/services/monitoring/driver"
+	"github.com/stackshy/cloudemu/v2/services/scope"
+)
+
+// CreateOCIAlarm creates a compartment-scoped alarm and returns it.
+//
+//nolint:gocritic // hugeParam: spec is the alarm's full definition.
+func (m *Mock) CreateOCIAlarm(_ context.Context, spec driver.OCIAlarmSpec) (*driver.OCIAlarm, error) {
```
NitinKumar004, line 19: **Medium (check-then-set race).** `CreateOCIAlarm` enforces per-compartment display-name uniqueness via `lookupByName(...)` then `m.alarms.Set(...)` with no lock spanning the two, and the store is keyed by unique OCID so `SetIfAbsent` can't help. Two concurrent same-name creates both see "absent" and both insert (reproduced: 116/200 runs → two alarms sharing a name); `findByDisplayName`/`DeleteAlarm`/`SetAlarmState` then resolve non-deterministically. Guard with a name→record index (or an inline scan reading `rec.spec.DisplayName` under the write lock — note `lookupByName` can't be reused under `Lock` since it calls `specOf`/RLock).

providers/oci/monitoring/alarms.go
```diff
@@ -0,0 +1,281 @@
...
+// evaluateAlarm re-evaluates one alarm against its current definition.
+func (m *Mock) evaluateAlarm(rec *alarmRecord) {
+	spec := m.specOf(rec)
+	m.evaluate(rec, &spec)
+}
+
+// evaluate compares a spec's query against the samples in its window and records
+// any status change. The spec is passed in rather than read off rec because
+// PostMetricData evaluates outside m.mu, where a concurrent update races it.
+func (m *Mock) evaluate(rec *alarmRecord, spec *driver.OCIAlarmSpec) {
```
NitinKumar004, line 196: **Medium (fidelity).** `evaluate` transitions to FIRING as soon as the single `[interval]` window breaches — but `spec.PendingDuration` (and the portable `EvaluationPeriods`) is stored/echoed and never read here. Real OCI only fires after the breach persists the full pending duration (independent of the aggregation interval). A client that sets `pendingDuration: PT5M`, posts one spike, and asserts "not yet firing" passes against this mock and fails against OCI. Evaluate over the pending window, or reject/disclose the field rather than accept-and-ignore. (Same class: alarm `suppression`/`overrides` at server/oci/monitoring/types.go:94-95 are accepted then silently dropped — never stored, never echoed — so a client's maintenance-window suppression has no effect.)

providers/oci/monitoring/query.go
```diff
@@ -0,0 +1,294 @@
...
+		stat:       stat,
+		interval:   alarmDuration(strings.TrimSpace(q[openIdx+1 : closeIdx])),
+		dimensions: dimensions,
+	}, tail[end+1:], true
+}
+
+// splitDimensions reads MQL's optional dimension predicate — the
+// {k = "v", ...} that scopes a selector to one series — and the rest after it.
+// Only the equality conjunction is understood; anything richer fails to parse.
+func splitDimensions(tail string) (dimensions map[string]string, rest string, ok bool) {
```
NitinKumar004: **Medium (fidelity).** `splitDimensions` does `strings.Cut(pair, "=")`, so a valid OCI predicate using a non-equality operator — `{resourceId != "x"}`, `{name =~ "prod-.*"}`, `!~` — cuts inside the operator, yields a bogus key, matches no series, and returns an empty datapoint list with HTTP 200 (a false "no data", not an error). OCI dimension predicates support `=`/`!=`/`=~`/`!~`; parse the operator, or reject unsupported ones explicitly. (Plain comma-AND equality is parsed correctly.)

NitinKumar004 (commented):
## Additional findings

Gates green (build / `-race` / docs-generate clean — `docs/coverage/oci/monitoring.md` matches the driver). The previously-raised Highs/Mediums remain fixed, and blast radius is contained (the `OCIMonitoring` capability is type-asserted only in the OCI mock + handler, `wireMonitoring` is nil-safe, and registration can't swallow other services). A closer look surfaced the items below — none are crashes, so this is a comment, not a block:

- **Medium** — `CreateOCIAlarm` check-then-set name-uniqueness race (inline).
- **Medium** — `pendingDuration`/`EvaluationPeriods` ignored; alarm fires on the first breaching datapoint (inline).
- **Medium** — MQL `!=`/`=~`/`!~` dimension predicate misparse → silent empty result (inline).
- **Medium** — alarm `suppression`/`overrides` accepted then silently dropped (folded into the pendingDuration comment).
- **Low** — resolution floor is 1s vs OCI's 60s (accepts queries OCI 400s); PostMetricData batch/format limits unenforced.
- **Low (generator, not this PR)** — because `OCIMonitoring` correctly lives in the shared `services/monitoring/driver`, coveragegen renders its ops into `docs/coverage/aws/cloudwatch.md` / `azure/monitor.md` / `gcp/cloudmonitoring.md`, promising `PostMetricData`/`CreateOCIAlarm` for providers that don't implement it (same shape as the existing `cache/SubnetGroups` leak) — worth a coveragegen fix so shared optional capabilities render per-implementing-provider.

Worth addressing the four Mediums as a follow-up.

arunesh-j:
Addressed in `4351247` — four Mediums, both Lows, plus the stale docs.

### Medium 1 — `CreateOCIAlarm` check-then-set race
`insertAlarm` holds `m.mu.Lock()` across the uniqueness scan and the insert. Test `TestConcurrentCreateHasOneWinner`: 200 rounds × 8 gated goroutines on one display name, asserting one winner and one stored alarm. Reverting gives `expected: 1, actual: 2`.

Worth noting for anyone writing a similar test: **2 goroutines did not reproduce it** — startup serialised them. The gate plus 8 creators is what makes it deterministic.

### Medium 2 — `pendingDuration` / `EvaluationPeriods`
`alarmRecord` gained `breachSince`; the alarm only transitions to FIRING once `now - breachSince >= pendingOf(spec.PendingDuration)`, clearing on recovery and on `UpdateOCIAlarm` (a new condition starts a new window). `pendingOf` parses OCI's ISO-8601 form (`PT5M`, `PT1H30M`, `P1D`). Portable `CreateAlarm` maps `EvaluationPeriods` → `PT<(N-1)×Period>S`.

**One deliberate deviation, documented:** an *unset* `pendingDuration` fires immediately rather than applying OCI's `PT1M` default. Evaluation here is driven by `PostMetricData`, not a timer, so a `PT1M` default would never elapse on its own. Tests run on `FakeClock`.

### Medium 3 — MQL predicates
Implemented `=` and `!=`; **rejected** `=~`/`!~` with `InvalidArgument` naming the operator. `splitPredicate` now finds the earliest operator (longest at that position), so a value containing `!=` is not mistaken for one and `=~` is not read as `=`.

The split is deliberate: `=`/`!=` are exact string tests with no ambiguity, whereas `=~`/`!~` are pattern matches whose flavour (glob / RE2 / OCI's own) I would have to guess — and guessing wrong replaces a silent-empty answer with a silent-*wrong* one, which is worse. The parse chain now returns errors rather than `bool`, so the wire surfaces the specific message, and `validateSpec` rejects an alarm carrying a refused predicate at create instead of storing one that can never fire.

### Medium 4 — `suppression` / `overrides`
**Rejected rather than persisted.** `acceptable()` 400s create and update naming the unsupported field. Persisting and echoing would still leave a caller's maintenance window with no effect on firing — accept-and-ignore with extra steps, which is the pattern you asked us to eliminate on #367.

### Lows
- **Resolution floor** raised 1s → `time.Minute` to match real OCI. Sub-second still rejected by the same check, now reporting the actual duration.
- **PostMetricData limits** enforced: 50 entries per request; namespace non-empty/≤255/letter-then-alnum-underscore/not `oci_`|`oracle_`; metric name non-empty/≤255; dimension keys and values non-empty/≤256 with keys free of periods and spaces. **Left out and stated in `docs/services.md`**: per-request datapoint cap, `metadata` limits, and the ingestion time window — I could not pin those to documented numbers, and a guessed cap that rejects valid traffic is worse than none.

### Docs
`AGENTS.md` and `docs/architecture.md` both stated OCI "emulates nothing yet", has no `<service>/` subpackages, and that "every OCI cell is `—`" — all false once this merges. Both now describe OCI as in progress with services arriving incrementally and name `docs/coverage/` as the source of truth, enumerating nothing in prose so they stay accurate as the rest land. The section anchor was renamed and its inbound link updated; no stale references remain.

### One correction
The review's framing of finding 2 says a client asserting "not yet firing" *passes against this mock and fails against OCI*. It was the reverse — the mock fired early, so such a test failed against the mock and passed against OCI. The defect and the fix are unaffected.

Verification: `go build ./...` clean · `go test ./...` exit 0, 254 packages · `-race` clean · `golangci-lint` 0 issues · `go generate ./...` leaves `docs/coverage/` unchanged.

Not touched, as instructed: the coveragegen leak of `OCIMonitoring` ops into the AWS/Azure/GCP pages.

NitinKumar004 (commented):
Re-reviewed after the fix commit. All four second-round findings are resolved and verified; gate is green (`go build`, `go test -race -count=2` on provider + server, run twice).

**Verified fixed**
- **Create name-uniqueness race** — `insertAlarm` now performs the duplicate scan and `Set` under a single `m.mu.Lock()`, so the check-then-set is atomic. Reproduced with 64 concurrent same-name `CreateOCIAlarm`: exactly one succeeds, 63 `AlreadyExists`, race-clean.
- **pendingDuration** — honored via a `breachSince` window (ISO-8601 parse; resets on recovery); alarm stays `OK` until the window elapses.
- **MQL `!=`** — correct for both the dimension predicate and the threshold operator (`NotEqualToThreshold`).
- **suppression/overrides** — rejected at the wire with 400 rather than silently stored.

Lock pairing (RLock reads / Lock writes) is correct, no field is read after the lock releases, and `evaluate*` is never called while holding the write lock — no aliasing or deadlock.

**Follow-ups (pre-existing fidelity gaps in this surface, not regressions from the fix):**
- **[High] No create-time query validation.** `validateSpec` never parses the MQL query, so `CreateOCIAlarm` with an invalid query returns 200 `ACTIVE`; `evaluate` then fails to parse and the alarm is silently inert forever (real OCI returns 400 `InvalidParameter`). This is the worst failure mode — looks configured, never fires. Recommend validating query (and `severity`/`pendingDuration`/`resolution` enums/format) at create time.
- **[Medium] `retrieveDimensionStates` unimplemented** — the per-alarm dimension-state action returns 404 where real OCI returns per-dimension FIRING/OK.
- **[Low]** hard delete (no transient `DELETING`/`DELETED`); `UpdateOCIAlarm` doesn't re-check display-name uniqueness within a compartment.

arunesh-j:
Rebased onto `development` (now carrying OCI Identity from #367) and addressed the latest findings in `2e4424b`.

### High — create-time query validation
`validateSpec` now parses the MQL query and rejects anything `evaluate` could not read, on **create and update**. The partial `checkDimensions` scan it supersedes is gone. Validated in the same pass, after `applyDefaults` so an omitted field still defaults rather than failing: `severity` against OCI's four values, `pendingDuration` (empty legal, unreadable 400), and `resolution` against the 1m floor.

The revert proof is itself the clearest statement of the bug:
```
--- FAIL: TestAlarmRejectsUnparseableQuery/CpuUtilization.mean()_>_80
    expected: 3 (InvalidArgument)  actual: 0 (OK)
--- FAIL: .../CpuUtilization[1m].p99()_>_80
    expected: 3  actual: 2 (AlreadyExists)
```
That `AlreadyExists` on the later case is the "looks configured, never fires" state you described — the first bad alarm had been stored under the name.

### Medium — `retrieveDimensionStates`: disclosed, not implemented
Now answers `501` naming the action rather than a generic `404 unknown monitoring path`.

Implementing it honestly is not a small change. This mock evaluates an alarm by pooling samples from **every** series the query selects into one slice and computing a single statistic; real OCI evaluates per stream and fires if any breaches. There is no per-dimension state in here to report. Synthesising one would let it disagree with the alarm's own status — `items: [{A, OK}, {B, OK}]` on an alarm reporting FIRING because the pooled mean breached — which is a subtler instance of the accept-and-ignore class this PR set out to remove. Making it real means reworking `evaluate` to be per-stream, which changes *when alarms fire*. I'd support that, but as its own change with its own tests, not folded into a review response.

### Low
- **Update display-name uniqueness** — fixed. The scan is now `checkNameFree(rec, name)` (skipping the record itself), called under the same `m.mu.Lock()` that writes `rec.spec`, so update is atomic the way create is. Covers the conflict, that a refused update leaves name and query untouched, that an alarm may keep its own name, and that another compartment may hold it.
- **Hard delete** — not changed, and I'd push back. `DELETING`/`DELETED` are transient states no caller of this emulator can observe: every mutation completes synchronously (`docs/oci-conventions.md` states this), so a state existing only between a delete and its own completion is never returned by any GET. `204` then `404` is what a client sees against real OCI too. Happy to add it if you see an observable difference I'm missing.

### Correction on the coverage drift
I told you earlier that the `docs/coverage/azure/databricks.md` drift did not reproduce. **That was wrong** — I tested three sequential runs, which is where it is stable.

It reproduces reliably under load: 25/25 identical when idle, but **8/12 vs 4/12 split** with `go test ./...` running concurrently. The generator alternates between `driver.DataPlane` (46 ops) and `driver.Databricks` (24 ops) for the same package. Same root cause as the `networking` case — a package declaring several interfaces, picked by method set — but non-deterministic, which means `go generate` is not idempotent and a `generate && git diff --exit-code` CI check is a coin flip on a busy runner. Evidence added to #383.

Verification: `go build ./...` clean · `go test ./...` exit 0, 260 packages · `-race` clean · `golangci-lint` 0 issues.

NitinKumar004 (approved):
Re-reviewed after the "Validate OCI alarms at create time and disclose retrieveDimensionStates" commit. Both previously-flagged items are resolved; gate is green (whole-module `go build`, `go test -race -count=2` on provider + server).

**Verified fixed**
- **Create-time validation** — `validateSpec` now runs the full `parseQuery` → `applyDefaults` → `validateFields`: an unsupported/garbage query is rejected `InvalidArgument` at create (wire 400) instead of being stored as an inert `ACTIVE` alarm; severity enum, pendingDuration ISO-8601, and resolution ≥ 1m are validated. `parseQuery` is the same parser `evaluate` uses, so parse/evaluate stay symmetric, and the removed `checkDimensions` guard is subsumed (strictly stronger). The common path is not over-rejected — `applyDefaults` runs before `validateFields`, so a minimal create (omitting resolution/pending/severity/metricCompartment/destinations) still succeeds via defaults, and `parsePending("")` is accepted.
- **Update parity** — `UpdateOCIAlarm` applies the same validation and re-checks display-name uniqueness (excluding self), so an update can't create a duplicate that create refuses.
- **retrieveDimensionStates** — routed at `/alarms/{id}/actions/retrieveDimensionStates` and disclosed as a typed 501 with an honest message (wrong action → 404, non-POST → 405), rather than a bare 404 or fabricated per-dimension data. Server-layer only; no driver-interface change, so no cross-provider impact.

**Non-blocking follow-up:** rich MQL (absence/grouping/compound) is now rejected 400 where real OCI returns 200 — more honest than the prior behavior (stored ACTIVE, never fired), since the mock can't evaluate those forms. A one-line doc note on the single-metric-threshold limitation would help users of absence/grouping alarms.

PR #367, opened 2026-08-08
Add OCI Identity: users, groups, policies, compartments

### LINKED ISSUE #378: OCI Identity: users, groups, policies, compartments
Implement **OCI Identity** against the existing portable `iam` driver, plus the compartment tree every other OCI service scopes against.

## Scope

Users, groups, user-group memberships, policies and **compartments** — full CRUD on the `/20160918` REST surface. Compartment listing honours `compartmentIdInSubtree`.

## Design notes

**OCI policies are English-like statement strings, not JSON documents:**

```
Allow group Admins to manage all-resources in compartment dev
```

`services/iam`'s JSON-document model does not fit and should not be forced. This needs an OCI statement parser and evaluator covering `Allow|Endorse|Admit|Define`, subjects by name / comma-list / OCID, verb subsumption (`manage > use > read > inspect`), resource families, and compartment-path locations. Evaluation is allow-only — OCI has no Deny.

**Compartments nest.** The tree belongs to this service. `services/scope` deliberately matches exactly, so ancestry must be resolved here rather than by changing the shared type.

Identity resources are region-agnostic, so OCIDs come from `idgen.GlobalOCID` and carry the empty region segment (`ocid1.user.oc1..aaaa…`).

Portable operations with no OCI meaning — policy attach/detach, instance profiles — should return `Unimplemented` naming the OCI equivalent rather than pretending to work.

## Acceptance

- Implements `driver.IAM`, with OCI-only behaviour as optional capabilities
- Compartment tree with sibling-name uniqueness and empty-only delete

arunesh-j:
## Summary

- Second OCI service. Implements the portable `iam` driver and the **compartment tree** every other OCI service scopes against.
- **Merge order: this is 2 of 3.** Merge #366 (Monitoring) first, then this, then vcn. Expect small conflicts in `providers/oci/oci.go`, `server/oci/oci.go` and `docs/services.md` — one import and one assignment each.

## Changes

- **`providers/oci/identity/`** — `Mock` over `memstore` implementing `driver.IAM`; the compartment tree (sibling-name uniqueness, empty-only delete, ancestry by walking `ParentID` to the tenancy); an OCI policy statement parser and evaluator.
- **`server/oci/identity/`** — the `/20160918` REST surface for users, groups, userGroupMemberships, policies and compartments. Memberships have no PUT, matching OCI. Compartment listing honours `compartmentIdInSubtree`; compartment create/delete stamp `opc-work-request-id`.
- **`services/iam/driver/oci.go`** — three optional capabilities discovered by type assertion: `Compartments`, `OCIIdentity`, `StatementPolicies`. **New file**, so the shared package gains no conflict surface.

## The policy language

OCI policies are English-like statement strings, not JSON documents:

```
Allow group Admins to manage all-resources in compartment dev
```

`services/iam`'s JSON-document model does not fit, so it was not reused. The parser handles `Allow|Endorse|Admit|Define`, subjects `any-user`/`group`/`dynamic-group`/`service` (by name, comma-list, or `id <ocid>`), verb subsumption (`manage > use > read > inspect`), resource families (`object-family` → buckets/objects/…), and locations `tenancy` / `compartment <name>` / `compartment <a>:<b>` / `compartment id <ocid>`. A `where` clause is captured but not evaluated. Evaluation is allow-only, as OCI has no Deny.

Portable operations with no OCI meaning return `Unimplemented` naming the OCI equivalent, rather than pretending: policy attach/detach and `ListAttached*` (OCI uses statements, not attachment) and all instance-profile operations (dynamic group matching rules).

## Compartments

Every resource records `scope.Scope{Compartment: …}` at create; lists filter with exact `Scope.Matches` — **`services/scope` is untouched**. Ancestry is resolved only inside this service, with `compartmentIdInSubtree` and policy-location resolution as the only callers. All identity OCIDs come from `idgen.GlobalOCID`, so they carry the empty region segment real OCI emits (`ocid1.user.oc1..aaaa…`, 5 segments).

## Provider Coverage

- [ ] AWS
- [ ] Azure
- [ ] GCP
- [x] OCI

## Checklist

- [ ] Integration tests added to `cloudemu_test.go` — unit + handler tests instead

## Test Plan

`go test ./...` — exit 0, 238 packages. `golangci-lint` — 0 issues for `providers/oci`, `server/oci`, `services/iam`.

Verified against a running server:

```
CreateCompartment              -> 200 + Opc-Work-Request-Id, ocid1.compartment.oc1..aaaa...
ListCompartments               -> ['dev','dev2'] direct
  with compartmentIdInSubtree  -> ['dev','dev2','team']
ListUsers                      -> ['alice'] in tenancy vs ['bob'] in dev2
CreatePolicy                   -> both statements round-trip verbatim
missing compartmentId          -> 400 InvalidParameter
unknown OCID                   -> 404 NotAuthorizedOrNotFound
duplicate name                 -> 409 Conflict
"do whatever you like"         -> 400 (statement rejected)
delete non-empty compartment   -> 409 IncorrectState
work request poll              -> SUCCEEDED, operationType CREATE_COMPARTMENT
opc-request-id                 -> echoed
/20160918/instances            -> 501  (Matches does not over-claim)
```

A test runs the **GCP** IAM mock through this handler to confirm a driver lacking the capabilities gets 501 rather than a wrong shape.

## Notes for review

- Users and groups accept any `compartmentId` rather than being confined to the tenancy as real OCI requires — otherwise compartment filtering is untestable for them. Flagged in a one-line comment.
- No policy-simulation endpoint: real OCI Identity has none, so evaluation is reachable only through the driver and the `StatementPolicies` capability.
- `dynamicGroups`, `authTokens` and `tagNamespaces` are deliberately absent from `Matches`, so a later service can claim them.
- **Capability placement** differs across the three PRs in this wave (see #366). This one uses a new file in the driver package. Worth settling on one before the next wave.

## Related Issues

Closes #378
Part of #376

Follows #342. Merge after #366.


providers/oci/identity/identity.go
```diff
@@ -0,0 +1,409 @@
...
+// authToken backs the portable driver's access keys.
+type authToken struct {
+	ID          string
+	UserName    string
+	Token       string
+	TimeCreated string
+}
+
+// Mock is an in-memory mock implementation of the OCI Identity service.
+type Mock struct {
```
NitinKumar004, line 86: **High (data race).** This `Mock` has no mutex. `UpdateStatementPolicy` (policy.go:104), `CreatePolicyVersion`/`DeletePolicyVersion`/`SetDefaultPolicyVersion`, `updatePrincipal`, and `UpdateCompartment` all `Get` the stored pointer and mutate its fields (`p.parsed`, `p.Statements`, `p.versions`, `p.Description`, `c.Name`) outside any lock. Meanwhile `Evaluate` iterates `&p.parsed[i]` and `List/GetPolicyVersion` iterate `p.versions` unlocked — so a concurrent `PUT /policies/{id}` during a `CheckPermission` is a torn read → misevaluated authz or panic. The sibling `providers/gcp/iam/iam.go` guards the identical paths with `mu sync.RWMutex` (line 9, Lock/RLock throughout). Add the same here and a concurrent Update+Evaluate `-race` test (the suite is all-sequential today, which is why `-race` is green).

providers/oci/identity/statement.go
```diff
@@ -0,0 +1,313 @@
...
+	if len(tokens) > 0 && strings.EqualFold(tokens[0], keywordID) {
+		s.LocationByID = true
+		tokens = tokens[1:]
+	}
+
+	if len(tokens) == 0 {
+		return cerrors.Newf(cerrors.InvalidArgument, "policy statement %q: names no compartment", text)
+	}
+
+	s.Location = tokens[0]
```
NitinKumar004, line 205: **Medium (over-grant).** `s.Location = tokens[0]` keeps only the first token, so `in compartment dev : team` (spaces around the colon) becomes `Location="dev"` and grants the whole `dev` subtree rather than `dev:team`. Real OCI uses `dev:team` (handled at `resolvePath`), but the spaced variant is plausible hand-written text and should return `InvalidArgument`, not silently broaden the grant. Reject trailing location tokens (or rejoin+parse the `:` path).

NitinKumar004 (changes requested):
## Deep review — OCI Identity

Reviewed in an isolated worktree at the PR head via a multi-agent pass (structure/parity + correctness/concurrency), the High re-verified. **Gate green** (build/vet/`go test -race`/whole-suite/gofmt).

**This is excellent, contract-faithful work.** All structural checks pass: implements the full portable `IAM` driver (**all 40 methods**, honest `Unimplemented` naming the OCI equivalent for attach/detach/instance-profiles), three optional capabilities in a new `services/iam/driver/oci.go` (additive), `GlobalOCID` for every region-agnostic identity resource, compartment tree (case-insensitive sibling uniqueness, empty-only delete, ancestry-to-tenancy), `ocirest` + `404 NotAuthorizedOrNotFound`. The statement parser/evaluator was adversarially checked and is **correct** — allow-only with unmatched→deny, verb subsumption, resource families, and it never grants outside the attaching compartment's subtree.

Two things to fix.

### High — data race (authz correctness under concurrency)
The `Mock` has **no mutex** — it relies only on the per-`memstore` map lock, but the update paths mutate the store-returned pointer *after* the lock is released: `UpdateStatementPolicy` reassigns `p.parsed`/`p.Statements`/`p.VersionDate` and `addRevision`; `CreatePolicyVersion`/`DeletePolicyVersion`/`SetDefaultPolicyVersion` reassign `p.versions`/`p.parsed`; `updatePrincipal`/`UpdateCompartment` mutate their pointees. Concurrent readers — `Evaluate` iterates `&p.parsed[i]`, `List/GetPolicyVersion` iterate `p.versions` — touch the same fields unlocked, so a concurrent policy update during `CheckPermission` is a torn read that can **misevaluate authorization or panic**. The direct sibling for this same portable interface, `providers/gcp/iam/iam.go`, carries `mu sync.RWMutex` and locks exactly these paths — so this is a deviation from the established pattern, not house style. Add the mutex (mirror `gcp/iam`) + a concurrent Update+Evaluate `-race` test. Inline on the Mock struct.

### Medium — policy nested-path with spaces over-grants
`parseLocation` sets `s.Location = tokens[0]` and drops the rest, so `... in compartment dev : team` (spaces around the colon) parses to `Location="dev"` and `covers(dev, target)` grants the **entire `dev` subtree** instead of `dev:team` — a broadening misgrant on plausible hand-written policy text. The canonical `dev:team` (no spaces) is handled correctly; the spaced form should error, not over-grant. Inline on statement.go.

### Low
STRUCTURE §3 filename singular/plural across layers (`compartment.go` vs `compartments.go`, `policy.go`/`statement.go` vs `policies.go`); no visited-set/depth cap in the ancestry walk (not reachable today — no reparent op — but a future move-compartment op turns it into an infinite loop); user names use exact `==` while compartments/policies use `EqualFold` (inconsistent case handling). No AI attribution.

arunesh-j:
Thanks — the race was real and the `gcp/iam` precedent was the right thing to point at. Fixed in `e24eca0`.

### High — data race
Added `mu sync.RWMutex` to `Mock`, mirroring `providers/gcp/iam/iam.go`. All 53 exported driver methods now lock (`Lock` on mutators, `RLock` on readers).

`RWMutex` is not reentrant and several portable methods delegated to exported OCI ones (`DeleteUser`→`DeleteOCIUser`, `CheckPermission`→`Evaluate`, `CreatePolicy`→`CreateStatementPolicy`, …), so those were split into unlocked cores — `deleteUser`, `deleteGroup`, `createMembership`, `deleteMembership`, `createPolicy`, `statementPolicy`, `deletePolicy`, `evaluate` — each marked "Callers hold m.mu". I swept the package for any locking method that calls another locking method and it comes back empty, so there is no reentrant path.

Test: `providers/oci/identity/race_test.go` — `TestPolicyUpdateConcurrentWithEvaluate`, `TestPolicyVersionWritesConcurrentWithReads`, `TestPrincipalAndCompartmentUpdatesConcurrentWithReads`. Stripping the `m.mu.*Lock()` calls makes all three fail with `WARNING: DATA RACE` at exactly the sites you named — `UpdateStatementPolicy` (policy.go:133) writing while `evaluate` reads `p.parsed`, and `clearDefaults`/`addRevision` writing `p.versions`.

### Medium — spaced nested path over-grant
`parseLocation` now returns `InvalidArgument` when more than one token remains after `in compartment [id]`, instead of keeping `tokens[0]`. Catches `dev : team`, `dev :team`, `dev: team` and trailing garbage; canonical `dev:team` is unaffected.

Worth recording that this was a live misgrant, not just a parse bug — a scratch test with the `dev : team` statement granted `manage buckets` on the **sibling** compartment `other`. Tests: four rows in `TestParseStatementRejects` plus `TestSpacedNestedPathIsRejectedRatherThanBroadened`, which asserts the canonical form grants `team` but neither `dev` nor `other`.

### Low
- **Filenames** — STRUCTURE.md §3's rule is that a feature uses the same filename across layers, so I renamed the wire files to singular (`compartments.go`→`compartment.go`, `memberships.go`→`membership.go`, `policies.go`→`policy.go`, `principals.go`→`principal.go`) rather than pluralising the provider. Both directories are now uniformly singular and match the repo's dominant feature-file convention (`nat_gateway.go`, `route_table.go`, `health_check.go`). Pure `git mv`.
- **Ancestry walk** — `covers()` is bounded by `m.compartments.Len()`, exactly enough to reach the tenancy from the deepest leaf. Allocation-free, which matters since it runs per statement per policy inside `Evaluate`.
- **Case handling** — `findByName` and `dynamicGroupNamed` now use `EqualFold`. This is the correct direction rather than merely the consistent one: OCI and AWS both make principal names unique per tenancy without distinguishing case, and `grantsSubject` already matched group names with `intersectsFold`. Before this you could create both `Admins` and `admins`, a statement saying `group admins` matched both, and `GetGroup("admins")` failed. Covered by `TestPrincipalNamesIgnoreCase`.

Verification: `go build ./...` clean · `go test ./...` exit 0, 238 packages · `go test -race ./providers/oci/... ./server/oci/...` clean · `golangci-lint` 0 issues.

Tracking issue: #378.

NitinKumar004 (approved):

Re-reviewed the fix commit (`e24eca08`) via a 2-agent pass (fix-verify + regression-hunt); gate green (build / vet / `go test -race -count=2` both Identity packages / gofmt / golangci-lint 0). Both prior findings fixed:

- **High (data race) — fixed, proven load-bearing.** `Mock` now carries `mu sync.RWMutex` (mirroring `gcp/iam`); every writer takes `Lock` and every reader — including `Evaluate`/`CheckPermission` (the authz path), `List*`/`GetPolicyVersion`, and the `to*Info` converters — takes `RLock`, with reentrancy avoided via locked entry points delegating to unlocked cores. Readers return copies (`copyTags`/`copyStrings`), so no pointer escapes the lock. An agent neutered the locks and `-race` fired, then passed again with them restored — the guard is load-bearing. No deadlock/lock-ordering issue.
- **Medium (nested-path over-grant) — fixed.** `parseLocation` now rejects a multi-token location (`in compartment dev : team`) with `InvalidArgument` instead of truncating to `dev`; canonical `dev:team` still resolves. Verified: the spaced form no longer grants the `dev` subtree.
- **Lows — fixed:** the ancestry walk is now depth-bounded by compartment count (no infinite loop if a reparent op is ever added); name uniqueness is uniformly case-insensitive (`EqualFold`) across users/groups/policies/compartments; the earlier singular/plural filename mismatch is resolved (server files renamed to match).

Parser/evaluator correctness (allow-only, verb subsumption, never-outside-compartment) still holds. Three trivial Lows remain (tight `covers` loop bound, `DeletePolicy` the lone un-converted delegate, one misleading error string). LGTM.

NitinKumar004:

arunesh-j:
Rebased on `development` and ran `go generate ./...` — OCI Identity now appears in `docs/coverage/`, and `docs/coverage/oci/README.md` no longer says "foundation only".

```
    OCI credited with "Identity", but oci.New() constructs no services
```

```
oci credited with "Identity", but oci.New() leaves that field nil
```

providers/oci/identity/statement.go
```diff
@@ -0,0 +1,321 @@
...
+	case subjectDynamicGroup:
+		return intersectsFold(s.Subjects, req.DynamicGroups)
+	default:
+		return false
+	}
+}
+
+// grantsAccess reports whether the statement's verb and resource type cover the
+// request. The location is resolved separately, against the compartment tree.
+func (s *statement) grantsAccess(req *driver.AccessRequest) bool {
```
NitinKumar004: **Medium (fidelity, evaluation-path).** `grantsAccess` never reads `s.Condition`, so a statement carrying a `where` clause — `Allow group G to manage all-resources in tenancy where request.permission='BUCKET_INSPECT'` — grants full unconditional `manage`. `where` is pervasive in real OCI (request.*/target.* conditions, tag conditions, `any{}`/`all{}`). This is off the SDK wire (statements round-trip verbatim; there's no OCI policy-eval REST endpoint), so it only affects the portable `CheckPermission`/`Evaluate` — hence Medium — but silence is the defect: either evaluate conditions or reject a `where`-bearing statement as unsupported (mirroring unsupported.go's honesty), rather than accept-and-ignore. `statement_test.go` even asserts `Condition` is captured, which gives false confidence it's honored.

providers/oci/identity/statement.go
```diff
@@ -0,0 +1,321 @@
...
+		return rankManage
+	default:
+		return rankUnknown
+	}
+}
+
+// resourceFamilies maps an OCI resource family onto the types it covers.
+func resourceFamilies() map[string][]string {
+	return map[string][]string{
+		"object-family":          {"buckets", "objects", "object-versions"},
```
NitinKumar004: **Medium (fidelity, evaluation-path).** Resource-family coverage is a hard-coded 7-family subset; valid real OCI families absent here — `functions-family`, `stream-family`, `resource-manager-family`, `logging-family`, `email-family`, load-balancers, nosql-tables — match only their literal self, so `Allow group G to manage functions-family in tenancy` evaluates to a wrong **deny** for a `functions` request. Same wire-confinement as the `where` item, so Medium; document the supported family set or extend it.

NitinKumar004 (commented):
## Additional findings

Gates green; the data-race fix (Mock `RWMutex`) and nested-path over-grant fix are intact and verified — all creates are atomic (lock spans scan+insert, one-winner confirmed per kind under `-race`), no reentrancy/deadlock, `covers()` ancestry sound, blast radius contained (the new capabilities are additive/type-asserted with zero cross-provider reach), and docs-generate is clean (`identity.md` matches the driver).

Two real-cloud fidelity gaps, both **off the SDK wire** (statements round-trip verbatim; OCI has no policy-evaluation REST endpoint), so they bite only the portable `CheckPermission` — comment, not block:
- **Medium** — policy `where` conditions parsed + stored but silently ignored → unconditional grant (inline).
- **Medium** — resource-family coverage is a 7-family subset → valid families wrongly deny (inline).

Recommendation: make both *disclose-or-reject* rather than accept-and-ignore (consistent with `unsupported.go`). Lows: no name-constraint enforcement on compartments/users/groups (real OCI rejects spaces/length); `MoveCompartment`/quotas unimplemented + undisclosed; `occupantOf` doesn't scan dynamic groups (practically unreachable today).

arunesh-j:
Addressed in `b3f12dc`. Both Mediums fixed, all three Lows closed.

### Medium 1 — `where` conditions
**Disclosed at evaluation rather than rejected at parse**, which is a deliberate departure from the obvious reading of your suggestion.

Rejecting at parse would convert an off-wire evaluation gap into an on-wire *create* failure. `where` is pervasive in real OCI, and you verified yourself that statements round-trip verbatim — a 400 on `CreatePolicy` for a valid production policy is a worse fidelity break than the one being fixed, and it would damage a surface you confirmed as correct.

So parse and storage are unchanged and the statement still round-trips byte-for-byte, but it can no longer produce a grant. `evaluate` calls a new `applies()` per statement: one that reaches the request on subject, verb, resource **and** location but carries a `Condition` returns `Unimplemented` naming the `where` clause instead of `true`. Location resolves *before* the disclosure decision, so a condition on a statement that never reaches the request stays silent, and an unconditional grant elsewhere still wins — the deferred error surfaces only when nothing granted.

Tests: `TestEvaluateDisclosesAWhereCondition` (your exact case) and `TestEvaluateIgnoresAWhereConditionItDoesNotReach`. Neutering the condition branch fails the first.

### Medium 2 — resource families
**Both broadened and made explicit.** I checked the OCI policy reference: the aggregate family list is long, service-scoped and still growing, and the docs do not enumerate members. A closed list would be wrong the day it shipped, and guessing at members would reintroduce the same silent-deny bug one level down.

- `coversResourceType` returns a tri-state — granted / denied / unknown-family. An unrecognised `-family` token is undecidable and returns `Unimplemented` naming it, never a deny.
- The table grew 7 → 11 families with fuller member lists (`compute-management-family`, `functions-family`, `stream-family`, `email-family`, plus expansions to the original seven: `boot-volumes`, `vlans`, `network-security-groups`, `cluster-work-requests`).
- `logging-family` and `resource-manager-family` are deliberately **not** guessed at — they surface as an explicit error rather than a wrong deny.

Tests: `TestEvaluateGrantsTheBroadenedFamilies` (6 pairs), `TestEvaluateDisclosesAnUnmodeledFamily`, plus three `TestStatementGrantsAccess` cases. Reverting either half fails the corresponding tests.

### Lows
- **Name constraints** — `validateName`: non-empty, ≤100 chars, letters/digits/`-`/`_`/`.`, applied to compartments (create and rename), users, groups, dynamic groups and policies.
- **`MoveCompartment` / quotas** — disclosed where they are addressable, on the wire: `POST /compartments/{id}/actions/moveCompartment` and `/20181025/quotas…` are claimed by `Matches` and answer `501` naming the operation. I did not invent driver methods — `Compartments` has no `MoveCompartment` and quotas are a separate OCI API, so stubs would be unreferenced and would force a driver-interface change.
- **`occupantOf`** — now scans dynamic groups; `TestDeleteCompartmentSeesADynamicGroup` asserts `FailedPrecondition` naming it.

Verification: `go build ./...` clean · `go test ./...` exit 0, 254 packages · `-race` clean · `golangci-lint` 0 issues · `go generate ./...` leaves `docs/coverage/` unchanged. I re-ran the reentrancy sweep after the new methods — still no locked-method-calls-locked-method path.

NitinKumar004 (approved):
Re-reviewed after the two fix commits. All three findings are resolved and verified; gate is green (`go build`, `go test -race -count=2` on provider + server, run twice). I could not construct a policy that still over-grants or still races.

**Verified fixed**
- **`where`/condition over-grant** — `parseStatement` now captures the condition and `unresolved()` returns `Unimplemented` when it can't be evaluated, checked *after* subject/verb/resource/location all match. A conditional statement can no longer silently authorize; a plain grant elsewhere still wins, and a non-matching conditional returns a clean `(false, nil)`.
- **Resource-family / verb subset** — `use instance-family` grants `use` on `instances` but not `manage`; unknown `*-family` types are disclosed via `unresolved`, not granted.
- **Data race** — `Mock.mu` serializes the in-place `p.parsed` rewrite against `evaluate`. Reproduced with 32×2 concurrent `CreatePolicy`/`Evaluate` under `-race`: clean.

Malformed statements reject with `InvalidArgument` → 400 at the wire; allow-only semantics (no Deny) are faithful to OCI.

**Follow-ups (non-blocking):**
- **[Low] Service-principal subjects silently under-grant.** `service <name>` parses but `grantsSubject` has no case for it, so it falls through to never-grant with **no disclosure** — the one inconsistency with the otherwise-disclosed model (conditions/unknown families all signal `Unimplemented`). Consider disclosing it the same way.
- **[Low]** `Endorse`/`Admit`/`Define` are accepted without syntax validation (stored where real OCI returns 400); safe direction (never grants), but over-accepts malformed input.

PR #368, opened 2026-08-08
Add OCI VCN: networks, subnets, gateways, security

### LINKED ISSUE #379: OCI VCN: networks, subnets, gateways, security
Implement **OCI VCN** (Core Networking) against the existing portable `networking` driver.

## Scope

VCNs, subnets, network security groups, security lists, route tables, internet / NAT / service gateways, DHCP options, VNICs, private and public IPs — create/list/get/update/delete on the `/20160918` Core Networking surface, plus NSG `securityRules` actions and `actions/changeCompartment`.

## Design notes

OCI's model does not line up with the portable driver everywhere, and the mapping should be honest rather than forced:

- OCI has **both** NSGs and security lists. NSGs map to security groups; a security list is a VCN-owned, subnet-attached rule collection with no direct portable analog.
- OCI has no deny rules, so any ACL mapping is allow-only.
- A subnet has exactly one route table, so associating replaces rather than adds.
- An internet gateway is attached at create.
- A NAT gateway hangs off the VCN, not a subnet.
- A public IP is assigned to a private IP, not to an instance.

Where OCI has a concept the driver cannot express — DRG, peering gateways — leaving it out and documenting the gap is preferred to bending an unrelated driver method.

`ListPrivateIps` takes no `compartmentId` in real OCI; it filters by `vnicId`/`subnetId`. Requiring one would reject every genuine SDK client.

## Acceptance

- Implements `driver.Networking` and the relevant optional capabilities
- Compartment recorded at create; every list requires `compartmentId` except where the real API does not accept it
- OCIDs from `idgen.OCID` per resource type
- Deleting a VCN with attached resources fails, as real OCI does

arunesh-j:
## Summary

- Third OCI service. Implements the portable `networking` driver over OCI's Core Networking API.
- **Merge order: this is 3 of 3.** Merge #366 (Monitoring), then #367 (Identity), then this. It will need a rebase after #367 — both touch `providers/oci/oci_test.go`.
- Also retires `TestUnimplementedServicesAreNil` (see below).

## Changes

- **`providers/oci/vcn/`** — `Mock` over `memstore` implementing `driver.Networking` plus the optional `VPCAttributes`, `NetworkInterfaces` and `NetworkInterfaceCreator` capabilities: VCNs, subnets, NSGs, security lists, route tables, internet/NAT/service gateways, DHCP options, VNICs, private and public IPs, local peering, flow logs.
- **`server/oci/vcn/`** — the `/20160918` Core Networking surface: create/list/get/update/delete per collection, NSG `securityRules` with add/remove/update actions, and a generic `actions/changeCompartment`.

## Concept mapping

Where OCI and the portable model differ:

| OCI | Portable | Note |
|---|---|---|
| VCN / Subnet / NSG | VPC / Subnet / SecurityGroup | direct |
| **Security list** | **NetworkACL** | VCN-owned, subnet-attached rule collection — the closest structural equivalent. OCI has no deny rules, so `Action` is always allow and rule numbers are positional |
| Route table + subnet attachment | RouteTable + Association | a subnet has exactly one table, so associating replaces |
| Internet gateway | InternetGateway | OCI attaches at create; handler calls Create then Attach |
| NAT gateway | NATGateway | OCI hangs it off the VCN, so config accepts a VCN **or** subnet OCID and resolves the VCN |
| Service gateway | VPCEndpoint | `ServiceName` = service OCID, type Gateway |
| Public IP | ElasticIP | assignment targets a private IP, not an instance |
| VNIC | `NetworkInterfaces` capability | creating one also mints its primary private IP |
| Local peering / flow logs | PeeringConnection / FlowLog | driver-only, no REST surface |

OCI-only behaviour is reached through an `Extras` interface asserted in the handler, so a driver without it gets 501 rather than a wrong answer: compartment scope, creation timestamps, a VCN's default route table/security list/DHCP options, whole-rule-set replacement, DHCP options, private IPs, and the richer VNIC view. Display names and OCI-only attributes ride in `cloudemu:`-prefixed tags stripped on the way out — the same approach `server/azure/vnet` uses for ARM names.

## One deliberate deviation, flagged

**`ListPrivateIps` requires `vnicId` or `subnetId`, not `compartmentId`.** `docs/oci-conventions.md` says every list endpoint must require `compartmentId`, but real OCI's `ListPrivateIps` takes no such parameter — enforcing it would 400 every genuine SDK client. Results are confined to a single subnet or VNIC, so they cannot span compartments. Every other list endpoint requires it. Happy to amend the conventions doc to record the exception.

## Retiring a foundation test

`TestUnimplementedServicesAreNil` enumerated service slots by name, so **every** service branch had to edit the same four lines — Identity and this PR already collided on it. Its value expired once services began landing. Replaced with a per-service wiring test (`TestVCNIsWired`), which is additive and conflict-free for the remaining services.

## Left out, deliberately

DRG and its attachments, local/remote peering gateways and remote peering connections (no honest driver shape without inventing a gateway resource), `GET /20160918/services` (the service-gateway catalogue), VCN CIDR add/modify/remove, and `definedTags` (echoed as `{}`; CloudEmu models no tag namespaces).

## Provider Coverage

- [ ] AWS
- [ ] Azure
- [ ] GCP
- [x] OCI

## Checklist

- [ ] Integration tests added to `cloudemu_test.go` — unit + handler tests instead

## Test Plan

`go test ./...` — exit 0, 238 packages. `golangci-lint` — 0 issues for `providers/oci`, `server/oci`.

Verified against a running server:

```
CreateVcn      -> 200 ocid1.vcn.oc1.iad.aaaa...  with defaultRouteTableId,
                  defaultSecurityListId, defaultDhcpOptionsId, vcnDomainName
CreateSubnet   -> 200 ocid1.subnet...  inherits VCN defaults, virtualRouterIp 10.0.1.1
ListVcns / ListSubnets (by compartmentId, vcnId)  -> 200
NSG + addSecurityRules / securityRules            -> rule id 4946249076a296b47024
internet + NAT gateways, custom DHCP options      -> 200
reserved public IP                                -> 129.146.72.149
changeCompartment       -> 202 + opc-work-request-id
cross-compartment list  -> empty
missing compartmentId   -> 400
unknown OCID            -> 404 NotAuthorizedOrNotFound
delete VCN with subnet  -> 409 IncorrectState
/20160918/instances     -> 501  (Matches does not over-claim)
```

## Related Issues

Closes #379
Part of #376

Follows #342. Merge after #366 and #367.


providers/oci/vcn/vcn.go
```diff
@@ -0,0 +1,485 @@
...
+	m.forget(v.DefaultSecurityListID)
+	m.forget(v.DefaultDHCPOptionsID)
+	m.vcns.Delete(id)
+	m.forget(id)
+
+	return nil
+}
+
+// vcnIsEmpty reports whether anything still references the VCN.
+func (m *Mock) vcnIsEmpty(id string) error {
```
NitinKumar004, line 256: **High (dependency-violation bypass).** `vcnIsEmpty` checks only subnets, IGWs, and NAT gateways. It misses service gateways, NSGs, and non-default route tables / security lists / DHCP options / peerings — all of which carry `VCNID`. So `DeleteVPC` succeeds with, say, a service gateway or NSG still attached, leaving orphaned records with dangling VCN references (and later child creates reject against the now-deleted VCN). Real OCI blocks until every attached resource is gone. Extend `vcnIsEmpty` to scan `serviceGWs`, `nsgs`, and non-default `routeTables`/`securityLists`/`dhcpOptions`/`peerings`, and extend `TestDeleteVPCRefusesWhileOccupied` to cover them.

providers/oci/vcn/security_list.go
```diff
@@ -0,0 +1,180 @@
+package vcn
```
NitinKumar004, line 1: **Medium (STRUCTURE §3).** Filename should be `security_list.go` (snake_case, full words — cf. the doc's `eip.go→elastic_ip.go`). Same for `flowlog.go`→`flow_log.go` and `publicip.go`→`public_ip.go`. Also cross-layer mismatches break the "same filename across layers" rule: provider `route_table.go` vs server `routetable.go` (pick one — the snake_case form), and provider `publicip.go`/`vnic.go` vs server `ip.go`. STRUCTURE §7's CI check is intended to fail non-snake_case filenames, so this may block on merge.

providers/oci/vcn/vnic.go
```diff
@@ -0,0 +1,378 @@
...
+			}
+		}
+	}
+
+	return m.addPrivateIP(s, vnicID, address, name, hostname, false), nil
+}
+
+// addPrivateIP stores a private IP, allocating the next free address in the
+// subnet when the caller names none.
+func (m *Mock) addPrivateIP(s *subnetData, vnicID, address, name, hostname string, isPrimary bool) *PrivateIP {
```
NitinKumar004: **Medium (duplicate IP).** `addPrivateIP` computes the address from `firstHostOffset + countPrivateIPs(subnet)`. Because the count shrinks when an IP is deleted, allocating `.3` then `.4`, deleting `.3`, then auto-allocating again re-mints `.4` — a duplicate on the same subnet. The auto path never checks for collisions (only the explicit-address path dedups). Allocate from the max-seen or a used-address set instead of the live count.

NitinKumar004 (changes requested):
## Deep review — OCI VCN

Reviewed in an isolated worktree at the PR head via a multi-agent pass (structure/parity + correctness/concurrency), the High re-verified. **Gate green** (build/vet/`go test -race`/whole-suite/gofmt).

**Strong implementation** — full portable `Networking` driver with `VPCAttributes`/`NetworkInterfaces`/`NetworkInterfaceCreator` capabilities and a handler-side `Extras` type assertion that 501s cleanly; the OCI↔portable concept mappings (security list→allow-only NetworkACL, route table one-per-subnet associate-replaces, IGW create-then-attach, NAT VCN-or-subnet resolution, service gateway→VPCEndpoint) are faithful and honestly documented; `Matches` is narrow and negatively tested. Findings below.

### High — `DeleteVPC` orphans attached resources
`vcnIsEmpty` only checks subnets, internet gateways, and NAT gateways — it does **not** check service gateways, NSGs, or non-default route tables / security lists / DHCP options / peerings. So `DeleteVPC` succeeds while those still carry the VCN's OCID, leaving live records with dangling references (and `DeleteVPC` also deletes only the *default* RT/SL/DHCP, leaving non-default siblings dangling). Real OCI returns a dependency error until all attached resources are removed. `TestDeleteVPCRefusesWhileOccupied` covers only subnet/IGW/NAT — the exact gap. Inline on vcn.go.

### Medium
- **STRUCTURE §3 filename violations** — non-snake_case `flowlog.go`/`publicip.go`/`securitylist.go` (should be `flow_log.go`/`public_ip.go`/`security_list.go`, per the doc's own `eip.go→elastic_ip.go`), and cross-layer mismatches: provider `route_table.go` vs server `routetable.go` (self-inconsistent within this PR), provider `publicip.go`/`vnic.go` vs server `ip.go`. STRUCTURE §7's CI check is meant to fail non-snake_case filenames. Inline on securitylist.go.
- **Auto private-IP duplication** — `addPrivateIP` derives the address from `firstHostOffset + countPrivateIPs(subnet)`; since the count shrinks on delete, allocating `.3`,`.4`, deleting `.3`, then allocating again re-mints `.4`. Track the used-set/max, not the count. Inline on vnic.go.
- **Public IP double-assign** — `AssociateAddress` neither checks the target private-IP OCID exists nor enforces 1:1, so two public IPs can point at the same (even nonexistent) private IP; `DisassociateAddress` then clears only the first (map-order dependent).
- **IGW create-then-attach** partial failure leaks an orphaned gateway (Create persists a detached IGW; if Attach fails on a bad `vcnId` the record is already stored).
- **Default route table not projected as a `Main` association** → `topology.findRouteTableForSubnet` falls back to first-RT-by-OCID, so `TraceRoute` binds the wrong table once a VCN has 2+ route tables (latent until OCI compute lands; AWS sets `Main:true`, and OCI's `IsDefault` makes this a cheap fix).
- **Data race on rule/route/gateway/VNIC/IP mutation** (Get-then-mutate outside lock). This matches the existing `aws/vpc` house pattern (inherited, not a regression), but note this PR is internally inconsistent — its own `Replace*`/tag methods use `store.Update` correctly.

### Low
Duplicate security-list rule numbers can't be individually addressed; content-derived NSG rule IDs collapse identical rules; no subnet-bounds check in IP allocation (wraparound); retiring `TestUnimplementedServicesAreNil` dropped the `ObjectStorage/Compute/Identity`-stay-nil invariant (VCN retirement itself is legit — it's genuinely wired). No AI attribution.

NitinKumar004:
@arunesh-j  please create issue ticket first

arunesh-j:
Thanks — the `DeleteVPC` gap was real and the filename issues were a fair catch. Fixed in `a0d1471`.

### High — `DeleteVPC` orphans attached resources
`vcnIsEmpty` now checks subnets, IGWs, NAT gateways **plus service gateways, NSGs, non-default route tables / security lists / DHCP options, and peerings (either end)**. `DeleteVPC` still removes only the three defaults, which is now correct precisely because a non-default sibling blocks the delete.
Tests: `TestDeleteVPCRefusesWhileOccupied` gained 6 cases, and `TestDeleteVPCTakesItsDefaultsWithIt` pins the other half. Reverting the fix fails 6/6 new subcases.

### Medium
1. **Filenames.** `git mv` throughout: provider `flowlog→flow_log`, `publicip→public_ip`, `securitylist→security_list`; server `routetable→route_table`, `securitylist→security_list`. I also split server `ip.go` into `vnic.go` + `public_ip.go` and `gateway.go` into `gateway.go` + `service_gateway.go`, so every shared feature now has **one filename across both layers** (`dhcp, gateway, nsg, public_ip, route_table, security_list, service_gateway, subnet, vcn, vnic`). Remaining single-layer files are genuinely single-layer.
2. **Private-IP duplication.** `freeAddress` takes the lowest address no private IP holds — a used-set, not a count — and stops at the broadcast address, which also closes the Low "no subnet-bounds check". Tests `TestAutoPrivateIPNeverRepeatsAnAddress` and `TestPrivateIPAllocationStopsAtTheSubnetEdge`; the count-based version fails with `address 10.0.1.4 was handed out twice`.
3. **Public-IP double-assign.** `AssociateAddress` validates the private IP exists and holds no address yet. Test `TestPublicIPAssignsOneToOne`.
4. **IGW leak.** The handler deletes the gateway if `Attach` fails, and `place` moved after the attach. Covered by `TestGatewayOperations/internet_gateway_against_an_unknown_VCN`.
5. **Main association.** `toRouteTableInfo` projects an implicit `Main:true` for the default table, so `topology.findRouteTableForSubnet` binds correctly. Test `TestDefaultRouteTableIsTheMainAssociation`.
6. **Data race.** Added a `mutate` helper; every rule/route/gateway/VNIC/IP mutation now runs inside `store.Update`, and rules/routes swap fresh slices instead of appending into shared backing arrays. `aws/vpc` untouched, as you noted it is inherited there. Two side benefits: `UpdateVNIC` validates NSG membership before writing anything (previously a partial write), and `AttachInternetGateway` no longer has a Get/check/write window.

**Worth flagging:** the first pass at #3 deadlocked — `publicIPFor` takes the store's read lock and it was called inside that same store's `Update` write lock, and `sync.RWMutex` is not reentrant. It hung the suite. The 1:1 check is now hoisted above the `Update`, and I ran a static audit over every `mutate` closure confirming none re-enters its own store.

### Low
- **Duplicate security-list rule numbers** — `AddNetworkACLRule` rejects a duplicate `(number, direction)`. The wire layer already numbers positionally per direction, so this only tightens the portable path.
- **NSG rule-ID collapse** — fixed at the source rather than the ID: the driver refuses an exact duplicate rule, so content-derived handles stay unique. I deliberately did **not** switch to index-derived IDs — they would be unique but would shift under every removal, which is worse for a client holding a handle.
- **Nil-slot invariant** — added `TestServiceSlotsAreNilOrLive`, which reflects over `Provider`'s interface fields and asserts none holds a typed nil. That keeps what the retired test was really guarding (a typed nil passes `!= nil` and then panics) without naming a service, so no future branch has to edit it.

Verification: `go build ./...` clean · `go test ./...` exit 0, 238 packages · `go test -race ./providers/oci/... ./server/oci/...` clean · `golangci-lint` 0 issues · `gofmt -l` empty.

Tracking issue: #379. Merge order is still #366 → #367 → this; I will rebase as each lands.

NitinKumar004 (commented):
## Re-review — original findings fixed, but the hardening introduced a new regression

Re-reviewed the fix commit (`a0d1471d`) via a 2-agent pass (fix-verify + regression-hunt); gate green (build / `go test -race -count=2` both VCN packages / golangci-lint 0). All four findings from the original review are **fixed** — nice work:
- **High (DeleteVPC orphans children) — fixed.** `vcnIsEmpty` now blocks on every user-created child (service gateways, NSGs, non-default route tables/security lists/DHCP/peerings) while correctly still allowing a VCN that holds only its own defaults to be deleted (defaults excluded via `IsDefault`, then cascaded).
- **Medium (auto private-IP duplication) — fixed.** `freeAddress` now allocates the lowest free host from a used-set (freed addresses reused), bounded to exclude network/router/broadcast — no wraparound, `/30`–`/32` → `ResourceExhausted`.
- **Public-IP double-assign / IGW create-then-attach leak / default-RT `Main` association — fixed.** 
- **STRUCTURE §3 filenames — fixed.** All snake_case and consistent across provider↔server layers.

Not approving yet — the `AssociateAddress` hardening introduced a new regression on the sibling paths it didn't also guard:

### Medium (new) — `UpdatePublicIp` reassign corrupts state on a failed associate
`reassign` (server/oci/vcn/public_ip.go) disassociates the current binding **first**, then associates the new target with no rollback. Now that `AssociateAddress` can fail (target already holds a public IP / doesn't exist), reassigning a bound public IP onto an occupied private IP returns `FailedPrecondition` but has **already detached** the public IP from its original private IP — leaving it unbound (reproduced empirically). Validate the target is free before disassociating, or re-associate the original on failure — exactly the rollback pattern this same commit added for the IGW create-then-attach path.

### Low (new) — `createPublicIp` leaks the allocated address on associate failure
`createPublicIP` allocates + places the public IP, then `AssociateAddress`; on failure it returns the error but leaves the allocated IP behind (listable orphan). Add the symmetric `ReleaseAddress` rollback (mirroring `DeleteInternetGateway` in the IGW path).

### Low (pre-existing, systemic — noted, not a blocker)
Cross-store check-then-act races remain (`freeAddress`+`Set`; `publicIPFor`+`mutate`), and stores hold pointers that `Describe*`/`toXInfo` read field-by-field after the store lock releases (e.g. `DescribeVPCs` vs `ModifyVPCAttribute` — `-race`-reproducible). Both agents confirm this is the same pattern used across the codebase and pre-dates this commit (not a regression). Worth a follow-up (a `Mock`-level lock or copy-under-lock in the read paths, as #366/#367 now do), but out of scope for this fix.

Add the reassign/create rollback (the Medium + its Low twin) and I'll approve. No AI attribution.

arunesh-j:
Good catch — the 1:1 hardening did exactly what you describe: it gave two sibling paths a way to mutate and *then* fail, leaving state worse than before the call. Both fixed in `0aa8276`.

### Medium (new) — `UpdatePublicIp` reassign corrupts state
`reassign` now records the previous binding and restores it when the associate is refused:

```go
if _, err := h.net.AssociateAddress(ctx, info.AllocationID, privateIPID); err != nil {
    if previous != "" {
        _, _ = h.net.AssociateAddress(ctx, info.AllocationID, previous)
    }
    return err
}
```

I chose rollback over validate-before-detach deliberately: a pre-check is another cross-store check-then-act, which is the systemic pattern you flagged below, so adding one more felt like the wrong direction. Rollback also matches the IGW path you pointed at.

### Low (new) — `createPublicIp` leaks the allocated address
`createPublicIP` now releases the address when the associate is refused, symmetric with `DeleteInternetGateway` in the IGW create-then-attach path.

### Tests
`TestPublicIPRollsBackOnFailedAssign`, two subtests over the wire. Both were verified to fail with their own fix reverted:

```
--- FAIL: .../a_rejected_reassign_keeps_the_original_binding
        a rejected move must not leave the address detached
--- FAIL: .../a_rejected_create_releases_the_allocated_address
        a rejected create must not leave an orphan address
```

Restored, both pass. I also added a `primaryPrivateIP` fixture helper, since asserting on the binding needs a second occupied target.

### On the pre-existing systemic finding
Agreed, and thanks for separating it out. The cross-store check-then-act windows and the read-after-unlock in `Describe*`/`toXInfo` are the same shape as `aws/vpc`, so fixing them here would leave this provider inconsistent with its siblings rather than fixing the class. #366 and #367 now do copy-under-lock; the honest fix is a `Mock`-level lock applied across the providers in one pass. Happy to open a follow-up issue against #376 for it if you'd like it tracked rather than remembered.

Verification: `go build ./...` clean · `go test ./...` exit 0, 238 packages · `go test -race ./providers/oci/... ./server/oci/...` clean · `golangci-lint` 0 issues · `gofmt -l` empty.

NitinKumar004:

arunesh-j:
Rebased on `development` and ran `go generate ./...` — OCI VCN now appears in `docs/coverage/`, and `docs/coverage/oci/README.md` no longer says "foundation only".

```
    OCI credited with "VCN", but oci.New() constructs no services
```

```
oci credited with "VCN", but oci.New() leaves that field nil
```

arunesh-j:
One caveat on the generated page for this PR, which I did **not** hand-correct because the file carries a do-not-edit header.

`docs/coverage/oci/vcn.md` says:

```
OCI's `networking` service · portable interface `driver.NetworkInterfaces`
## Operations (3)
```

That understates the branch considerably — it implements the full `driver.Networking` surface plus `VPCAttributes`, `NetworkInterfaces` and `NetworkInterfaceCreator`.

**It is not this branch's bug.** It reproduces on `development` at `ef9d73a`:

```
| `networking` | [VPC](./aws/vpc.md) | — | — | — | 3 |
```

AWS VPC is credited with the same 3 operations, and Azure `vnet` and GCP `vpc` are credited with nothing at all. A networking mock satisfies both `driver.Networking` and the optional `driver.NetworkInterfaces`, and `implementedService`'s method-set match appears to pick the narrower one.

Filed as #383 with a suggested fix. Happy to take it in a follow-up, or in this PR if you would rather the OCI page land accurate — just say which.

providers/oci/vcn/vcn.go
```diff
@@ -0,0 +1,544 @@
...
+	}
+
+	return DefaultResources{
+		RouteTableID:   v.DefaultRouteTableID,
+		SecurityListID: v.DefaultSecurityListID,
+		DHCPOptionsID:  v.DefaultDHCPOptionsID,
+	}
+}
+
+func toVPCInfo(v *vcnData) driver.VPCInfo {
```
NitinKumar004, line 462: **Medium (data race — the last of the OCI trio still open).** `toVPCInfo` reads `v.DNSSupport`/`v.DNSHostnames` off the shared `*vcnData` pointer that `describeResources`→`store.Get` returns, **after the store lock releases**, concurrent with `ModifyVPCAttribute`'s in-place `Update` write — reproduced under `-race` (same for `DescribeSubnets`/`toSubnetInfo` vs `UpdateSubnetTags`, and `DescribeSecurityGroups`/`toNSGInfo` vs `AddIngressRule`). The prior `mutate`/`Update` change closed only the writer-writer race. Its two sibling OCI PRs fixed exactly this class — #366 with `specOf` snapshots, #367 with a Mock-level `sync.RWMutex` (+ a retained race_test) — and CLAUDE.md mandates "all mock implementations use sync.RWMutex." VCN has no Mock-level mutex. Apply the same Mock-`RWMutex` remedy and add a concurrent Describe+Mutate `-race` test (the suite's green `-count=2` only reflects the absence of such a test).

providers/oci/vcn/public_ip.go
```diff
@@ -0,0 +1,150 @@
...
+// DescribeAddresses returns public IPs matching the given OCIDs, or all if empty.
+func (m *Mock) DescribeAddresses(_ context.Context, ids []string) ([]driver.ElasticIP, error) {
+	return describeResources(m.publicIPs, ids, toPublicIPInfo), nil
+}
+
+// AssociateAddress assigns a public IP to a private IP. The returned handle is
+// the private IP's OCID, which is what OCI clears the assignment by. A private
+// IP holds at most one public IP, and DisassociateAddress takes that handle,
+// so a second address on the same private IP would be unaddressable.
+func (m *Mock) AssociateAddress(_ context.Context, allocationID, instanceID string) (string, error) {
```
NitinKumar004, line 88: **Medium (check-then-act).** `AssociateAddress` reads `publicIPFor(instanceID)` (scans `publicIPs.All()`) then `mutate`s a *different* key — two concurrent associates onto the same private IP both pass the `publicIPFor` guard and each mutate a different public-IP record, breaking the 1:1 invariant the code explicitly enforces. Cross-store check-then-act; a Mock-level mutex (above) would also close it. (CreateSubnet overlap-check + Set has the same TOCTOU shape, Low.)

NitinKumar004 (commented):
## Additional findings — rollback fixed; one consistency gap remains

The public-IP rollback (the last blocking item) is fixed and verified — `createPublicIP` releases the correct allocation on a refused associate (no double-release, no scope leak) and `reassign` restores the previous binding; probes pass, `-race` clean, docs-generate clean. Much is genuinely faithful to real OCI Core (default RT/SL/DHCP auto-created + blocked from standalone delete, subnet CIDR containment/overlap rejection, delete-ordering dependencies, and the synthetic `Main` association making TraceRoute resolve the default route table correctly).

Remaining (comment — the blocker is cleared):
- **Medium (consistency)** — the shared-pointer reader-writer race is still open; the other two OCI services fixed this same class (Monitoring via `specOf` snapshots, Identity via a Mock-level `sync.RWMutex` + a retained race test), and CLAUDE.md requires "all mock implementations use sync.RWMutex." VCN has no Mock-level mutex. Inline, with the recommended remedy.
- **Medium** — `AssociateAddress` 1:1 check-then-act race; route target/CIDR not validated (a route to a detached gateway is accepted, so TraceRoute would treat it as a reachable hop).
- **Low** — default security list omits OCI's two default ICMP rules (unrepresentable in `NetworkACLRule`); `prohibitPublicIpOnVnic` round-tripped but not enforced; topology `CanConnect` evaluates NSGs but not security lists — a latent AWS-centric engine gap that will false-DENY OCI instances once an OCI compute provider lands (track for that follow-up, not this PR).

Adding the Mock-`RWMutex` would close both races and match the other two OCI services.

arunesh-j:
Addressed in `bdae5f1`. VCN is now in line with its two siblings.

### Medium 1 — Mock-level `RWMutex`
Added `mu sync.RWMutex` as the first `Mock` field, matching #367. All 80 exported methods lock — `RLock` on the 20 readers (`Describe*`, `Scope`, `Created`, `Defaults`, `VNICsInNSG`, `GetFlowLogRecords`), `Lock` on the other 60. Every unexported helper stays lock-free.

**Deadlock sweep.** Unlike Identity, VCN needed *no* core splits — it has no exported→exported call at all. I verified independently rather than taking that on trust:
- no unexported method takes the lock — empty
- no exported method calls another exported method — empty
- a static pass over all `mutate`/`store.Update` closures for same-store re-entry — clean

Three closures read *other* stores (`AttachInternetGateway`, `UpdateVNIC`, the new route validation); none re-enters its own. The historical bug in this package — `publicIPFor` inside `publicIPs.Update` — is not present. The full suite completes rather than hangs.

Test: `TestDescribeConcurrentWithMutation`. Stripping every `m.mu.` line produces 3 `WARNING: DATA RACE` reports on `vcnData.DNSSupport`, `subnetData.Tags` and `nsgData.IngressRules`.

### Medium 2 — `AssociateAddress` check-then-act
Closed by the mutex; the `publicIPFor` guard and the write are now one critical section. `TestConcurrentAssociateAddressKeepsOneToOne` races 16 public IPs onto one private IP and asserts exactly one winner — it fails **without** `-race` too (`expected: 1, actual: 2`).

### Medium 3 — route target and CIDR validation
New `validateRoute` on `CreateRoute` and `ReplaceRoutes`: rejects an unparseable destination, a target OCID naming nothing, a **detached** internet gateway, and a gateway belonging to another VCN. `ReplaceRoutes` moved to `mutate` so the whole set is validated before any of it is stored. `local`/empty targets still pass.

Worth noting the existing `TestRouteTablesAndAssociations` was routing to a detached IGW — exactly the bug you described — so it now attaches first.

### Lows
- **Default ICMP rules** — not forced, per your allowance. `driver.NetworkACLRule` has no ICMP type/code field, so OCI's type-3/code-4 defaults are genuinely unrepresentable; widening them to "all ICMP" would allow traffic real OCI drops, and a false-ALLOW is worse than the omission. Recorded in the `defaultSecurityRules` doc comment.
- **`prohibitPublicIpOnVnic`** — now enforced on both create and reassign, with no orphan address left behind on refusal. No interface change.
- **Topology `CanConnect` vs security lists** — left for the OCI compute PR, as you directed.

### One correction
The review says the `Mock`-`RWMutex` is required by CLAUDE.md. There is no CLAUDE.md in this repo — `AGENTS.md` is the agent-facing doc, and it has no concurrency or mutex requirement. I have made the change regardless because the consistency argument stands on its own, but the convention is worth writing into `AGENTS.md` if it is going to be cited.

I also could not reproduce a `docs/coverage/azure/databricks.md` drift: `development` and this branch both regenerate byte-identical across repeated runs, and the branch's copy matches `development` exactly.

Verification: `go build ./...` clean · `go test ./...` exit 0, 254 packages, completes · `go test -race ./providers/oci/... ./server/oci/...` all ok · `golangci-lint` 0 issues.

NitinKumar004 (commented):
Re-reviewed after the two fix commits. Both findings are resolved and verified; gate is green (`go build`, `go test -race -count=2` on provider + server, run twice).

**Verified fixed**
- **Reader-writer / aliasing race** — the Mock-level `sync.RWMutex` is acquired by every exported method, and the lock spans the full read-modify-write. Critically, the aliasing hazard is genuinely closed because `Describe/List/Get` return *value* structs with deep-copied maps and slices (`copyTags`, `make+copy`) — the stored pointer is never handed out. Reproduced by walking returned `Rules`/`Routes`/`Tags` outside the lock while 32 goroutines mutate the same resources: clean under `-race`. `AssociateAddress`'s 1:1 guard is atomic (one winner under contention).
- **Public-IP partial-failure rollback** — an allocate→assign refusal calls `ReleaseAddress` (frees the record and its scope entry); no double-free, retry works. The same rollback discipline is applied to `reassign` and IGW-attach.

No re-entrancy or deadlock: no exported method re-enters `m.mu`, and no cross-service call is made under the lock; a 24-goroutine mixed-op stress run is clean.

**Follow-ups (pre-existing fidelity gaps in this surface, not regressions from the fix):**
- **[High] Ephemeral public-IP semantics.** `scope` is hardcoded `REGION` regardless of lifetime (ephemeral should report `AVAILABILITY_DOMAIN`); an ephemeral can be created unassigned and unassigned via update (real OCI refuses both), and delete of an assigned ephemeral is refused (real OCI allows it — the only way to unassign one).
- **[Medium] Multi-CIDR VCNs** — only `cidrBlocks[0]` is kept, so a subnet in the second block is wrongly rejected; `add-vcn-cidr` has no route.
- **[Medium] LPG / DRG / remote-peering** are modeled in the provider but not served on the wire (unhandled collections), so cross-VCN connectivity is unavailable over the wire.
- **[Low]** NSG rules are CIDR-only (ignore `sourceType`/`destinationType`); no transient lifecycle states.

arunesh-j:
Rebased onto `development` (now carrying OCI Identity from #367) and addressed the latest findings in `f10f752`.

### High — ephemeral public-IP semantics
All three fixed in `server/oci/vcn/public_ip.go`; the lifetime rules are OCI-specific, so they live in the wire layer and the portable driver keeps EC2's semantics.

1. **Scope** — `publicIPScope(lifetime)` returns `AVAILABILITY_DOMAIN` for `EPHEMERAL`, `REGION` otherwise.
2. **Always attached** — create refuses an ephemeral with no `privateIpId`; `moveAssignment` refuses any change of `privateIpId` on an ephemeral, covering unassign *and* move.

   This surfaced a **latent bug you did not ask about**: `publicIPRequest.PrivateIPID` was a plain `string`, so a tag-only PUT — which omits the field — was indistinguishable from an explicit `""` and silently unassigned the address. It is now `*string`, matching OCI's omitted-means-no-change semantics, and that case is covered.
3. **Delete unassigns** — `deletePublicIP` reads, disassociates when assigned, then releases, re-associating if the release fails.

   **I extended this to reserved addresses too, which the finding did not ask for.** OCI documents `DeletePublicIp` as "unassigns and deletes… (either ephemeral or reserved)", so keeping the refusal for reserved only would have been an inconsistency I could not justify. That changed the existing `"an assigned address cannot be released"` assertion in `TestVNICAndIPOperations`. The provider's `ReleaseAddress` still refuses an assigned address — that is the portable contract AWS shares. Flagging it explicitly since it widens the diff beyond the finding.

Test: `TestEphemeralPublicIPLifetime`, four subtests, each revert-verified.

### Medium 1 — multi-CIDR VCNs
`CIDRBlock` → `CIDRBlocks []string`; `VPCInfo.CIDRBlock` reports the primary. New Extras capability (`VCNCIDRs`/`AddVCNCIDR`/`RemoveVCNCIDR`), containment checked against every block, peering overlap compares both sides in full. Wire: `cidrBlocks` on create (mutually exclusive with `cidrBlock`; a refused block rolls the VCN back), and `actions/{addVcnCidr,removeVcnCidr}` returning 202 + `opc-work-request-id`. `modifyVcnCidr` is deliberately **not** served — add+remove cover it — and that is stated in the package doc rather than left as a silent 404.

### Medium 2 — LPG / DRG / remote peering
**Local peering gateways are now served**: CRUD, `actions/connect`, `NEW`/`PEERED`/`REVOKED`, `peerAdvertisedCidr` from the peer VCN's blocks, teardown of one end revoking the other, and `vcnIsEmpty` blocking on them. The connection the provider already modelled is the portable projection of a connected pair, so nothing was invented.

**DRG, DRG attachments and remote peering connections are not served.** The driver models no DRG and I would not invent one for the wire. Instead the handler now *claims* those three collections and answers `501` naming the gap and pointing at `localPeeringGateways` — disclosed rather than an unhandled path.

### Low
- **Entity types fixed** — `entityTypeOf` derives `CIDR_BLOCK` / `SERVICE_CIDR_BLOCK` / `NETWORK_SECURITY_GROUP` from the stored value, so a service label or NSG OCID round-trips. Projection only; I did not add a type field to the shared `SecurityRule` for one provider's cosmetics.
- **Transient lifecycle states — not implemented, and I'd push back.** Every CloudEmu mutation is synchronous (`docs/oci-conventions.md` states it), so a `PROVISIONING` no waiter can ever observe leaving is a lie with a timer attached. Documented in the handler package doc.

### Deadlock sweep
115 `*Mock` methods, 87 taking `m.mu`. `ConnectLocalPeeringGateways` was the one place re-entrancy could appear, so `CreatePeeringConnection` was split into a locking wrapper plus an unlocked `createPeering`.

I verified independently rather than on trust: no locking method calls another locking method, no unexported method takes the lock, and a static pass over every `mutate`/`store.Update` closure found no same-store re-entry. Two `-race` tests pin it — `TestConcurrentCIDRAndGatewayMutation` (a re-entrant call hangs here rather than merely failing) and `TestConcurrentConnectPeersOnce`.

Verification: `go build ./...` clean · `go test ./...` exit 0, 260 packages, completes · `go test -race -count=2` all ok · `golangci-lint` 0 issues (two stale `//nolint:dupl` directives removed).

On the generator nondeterminism you may see when regenerating: confirmed and quantified in #383 — 25/25 identical when idle, 8/12 vs 4/12 under concurrent load.

NitinKumar004 (approved):
Re-reviewed after the fix commit. All three fidelity items from the earlier review are genuinely resolved against OCI semantics, with real provider- and wire-level tests driving the paths. Gate green (build/vet, golangci-lint 0 issues, `go test -race -count=2` on provider + server), docs-generate clean.

**Verified fixed**
- **Ephemeral public-IP semantics** — `scope` is now lifetime-derived (`AVAILABILITY_DOMAIN` for EPHEMERAL, `REGION` for RESERVED); creating an EPHEMERAL with no privateIpId is rejected (and leaves no orphan address); updating to unassign/move an ephemeral is rejected; deleting an assigned ephemeral is now allowed (unassign-then-release, with rollback on a refused release). RESERVED reassign/rollback intact.
- **Multi-CIDR VCNs** — both blocks are kept and echoed (`CidrBlocks` + deprecated `CidrBlock`=first); a subnet whose CIDR falls in the second block is accepted; `add-vcn-cidr`/`remove-vcn-cidr` are served with work requests, overlap-across-all-blocks rejection, a keep-at-least-one-block guard, and a subnet-in-use guard on removal. Single-CIDR VCNs are unchanged.
- **Local Peering Gateways** — full CRUD + `connect` on the wire, with a faithful `NEW → PEERED → REVOKED` state machine (self-peer, already-peered, and overlapping-CIDR connects all rejected; deleting one end revokes the far end), and `PeerAdvertisedCIDRs` resolved from the peer VCN. DRG / remote-peering are honestly disclosed as 501 rather than faked.

Concurrency invariants hold across the new code: the Mock-level `RWMutex` and deep-copy discipline extend to the LPG store and the new `CidrBlocks` slice (verified with a concurrent Describe+mutate `-race` probe), new creates and `AddVCNCIDR` are atomic, and `ConnectLocalPeeringGateways` avoids re-entrancy via an unlocked helper.

**Optional follow-ups (all Low)**
- `modifyVcnCidr` is a real OCI action but isn't served (documented; add/remove cover the common cases).
- A `REVOKED` local peering gateway can't be reconnected without delete+recreate, whereas real OCI lets you re-establish it.
- The rule that an ephemeral public IP may only sit on a VNIC's primary private IP isn't modeled (pre-existing).

PR #370, opened 2026-08-09
Add AWS EFS Emulation with Full Parity (#295 Workstream B)

thzgajendra:
## Objective / Issue

Second net-new service of **Workstream B (#295)**: emulate **AWS EFS** (Elastic File System) with SDK-wire fidelity so real `aws-sdk-go-v2/service/efs` clients and the `aws efs` CLI work against cloudemu by only changing the endpoint. Full parity across file systems, mount targets, access points, lifecycle/backup/replication configuration, and account preferences — **30 operations**.

## What we found

cloudemu had no EFS. It needed a full 4-layer vertical (driver → in-memory provider → REST-JSON wire handler → registration). Unlike the JSON 1.1 services, EFS is **REST-JSON** (path + method routing under the `/2015-02-01/` API-version prefix) — the same family as the existing VPC Lattice handler, which this mirrors.

## How we fixed it

New EFS vertical following the existing service pattern: `services/efs/driver/`, `providers/aws/efs/`, `server/aws/efs/`, registered via `Drivers.EFS` in `server/aws/aws.go` (before the S3 catch-all) and `p.EFS` in `providers/aws/aws.go`. Built in phases, each committed after building green with real SDK roundtrip tests + `aws efs` CLI E2E:

| Family | Operations |
|--------|-----------|
| File systems | Create/Delete/Describe/UpdateFileSystem (idempotent creation token, throughput modes) |
| File system policy | Put/Describe/DeleteFileSystemPolicy |
| Mount targets | Create/Delete/DescribeMountTargets, Describe/ModifyMountTargetSecurityGroups |
| Access points | Create/Delete/DescribeAccessPoints (PosixUser, RootDirectory) |
| Lifecycle | Put/DescribeLifecycleConfiguration |
| Backup | Put/DescribeBackupPolicy |
| Replication | Create/Delete/DescribeReplicationConfigurations |
| Account preferences | Put/DescribeAccountPreferences |
| Tags | TagResource/UntagResource/ListTagsForResource + legacy CreateTags/DeleteTags/DescribeTags |

**Routing:** `Matches` gates on the `/2015-02-01/` version prefix, so EFS never shadows the S3 catch-all (no real bucket path begins with that prefix). Timestamps are emitted as epoch-seconds numbers to match the wire.

**Behavioral fidelity:** creating a file system is idempotent on the creation token; a file system with mount targets can't be deleted until they're removed (`NumberOfMountTargets` tracks this); one mount target per subnet; access points and file systems share a tag store with the `Name` tag mirrored onto the resource name; replication assigns a destination file-system id per destination.

## Alternatives not taken

- **JSON 1.1 dispatch (X-Amz-Target)** — wrong protocol for EFS; it's REST-JSON. We route on path + method under the version prefix.
- **Modeling cross-region replicas as separate queryable endpoints** — a single-process emulator has one regional endpoint, so replication returns wire-complete metadata (including an assigned destination FS id) but the replica isn't separately queryable. Documented.

## Docs / Tests / Playground

- Docs: `docs/services.md` (service row 28 + a detail section; op-count total updated to 1782) and `docs/architecture.md`.
- Tests: real `aws-sdk-go-v2/service/efs` SDK roundtrip tests per phase (file-system lifecycle, policy, tags, mount targets + security groups, access points, lifecycle/backup/replication config, account preferences) + provider unit tests.
- E2E: every phase verified against a live `cloudemu start` server via the real `aws efs` CLI.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go mod tidy` (adds `aws-sdk-go-v2/service/efs`)
- [x] CodeQL (0 findings in EFS files; the 9 reported are pre-existing on development)
- [x] `aws efs` CLI E2E on a running server for every phase

## Risk & Rollback

Low risk: entirely additive — a new service package plus two registration lines and one struct field; nothing existing changes behavior. `Drivers.EFS` is nil-skipped when unset. Rollback = revert the branch or drop any per-phase commit.

## Conclusion

AWS EFS is fully emulated — the second net-new Workstream B service after KMS. Branch is synced to the latest development (KMS included) with no conflicts. Follow-ups (separate issues): resource-discovery surfacing of EFS file systems; EFS on Azure Files / GCP Filestore if cross-provider parity is wanted.


providers/aws/efs/file_systems.go
```diff
@@ -0,0 +1,301 @@
...
+)
+
+// nameTag is the tag key EFS mirrors onto FileSystem.Name.
+const nameTag = "Name"
+
+// CreateFileSystem creates a new file system. A repeated creation token returns
+// the existing file system (idempotent create), matching EFS.
+//
+//nolint:gocritic // in is the public CreateFileSystem input, taken by value to match the driver API
+func (m *Mock) CreateFileSystem(_ context.Context, in driver.CreateFileSystemInput) (*driver.FileSystem, error) {
```
NitinKumar004, line 21: **M1 (idempotency race).** The token scan over `fileSystems.All()` below and the `Set(id,…)` at the end of this function aren't atomic, and the store is keyed by a random `fs-` id, not the token. Two concurrent `CreateFileSystem` calls with the same `CreationToken` both scan a set that lacks the other's entry, both pass, and both `Set` → two file systems sharing one token, violating EFS idempotency. Guard with a token→fsID index (`SetIfAbsent`) or a mutex spanning check+create. (Also: the doc comment says it "returns the existing file system" but the code returns `AlreadyExists` — the behavior is the correct real-EFS one; fix the comment.)

providers/aws/efs/file_systems.go
```diff
@@ -0,0 +1,301 @@
...
+		backup:    driver.BackupDisabled,
+	})
+
+	out := fs
+
+	return &out, nil
+}
+
+// DeleteFileSystem deletes a file system. It must have no mount targets.
+func (m *Mock) DeleteFileSystem(_ context.Context, fileSystemID string) error {
```
NitinKumar004, line 89: **M3 (orphan + leak).** This guards only on mount targets. Deleting a file system that still has access points succeeds but never removes the `apIndex`/`mtIndex` entries (unbounded leak across create/delete churn) and silently drops policy/lifecycle/replication config. There's also a TOCTOU window — the MT count is read under `RLock` and released before `Delete`, so a concurrent `CreateMountTarget` (which still resolves the not-yet-deleted FS) can attach a live MT to an FS that then gets deleted. Delete the index entries on FS delete, and hold the lock across the check+delete (and decide whether to block while access points exist, per EFS).

server/aws/efs/errors.go
```diff
@@ -0,0 +1,92 @@
...
+func writeError(w http.ResponseWriter, status int, errType, msg string) {
+	w.Header().Set("Content-Type", contentTypeJSON)
+	w.Header().Set("X-Amzn-Errortype", errType)
+	w.WriteHeader(status)
+
+	_ = json.NewEncoder(w).Encode(errorBody{Type: errType, ErrorCode: errType, Message: msg})
+}
+
+// writeErr maps a canonical cloudemu error to the closest EFS exception.
+func writeErr(w http.ResponseWriter, err error) {
```
NitinKumar004: **M2 (typed-error fidelity).** This collapses every `NotFound→FileSystemNotFound` and `FailedPrecondition→FileSystemInUse`, but EFS has per-resource typed exceptions: `MountTargetNotFound`, `MountTargetConflict`, `AccessPointNotFound`, `PolicyNotFound`, `ReplicationNotFound` (all separate in aws-sdk-go-v2). So `errors.As(err, &*efstypes.MountTargetNotFound{})` after e.g. `DeleteMountTarget` on a bogus id never matches — it gets `FileSystemNotFound`. This single-exception mapper is fine for vpclattice (one `ResourceNotFoundException`) but not EFS. Pass the resource kind / a specific error code from the call site so the right `X-Amzn-Errortype` is emitted (same fix you applied on #369).

NitinKumar004 (changes requested):
## Deep review — AWS EFS emulation

Reviewed in an isolated worktree at the PR head via a 2-agent fan-out (structure/wire/parity + correctness/state/concurrency), every substantive finding re-verified. **Gate green** — build / vet / `go test -race` (both EFS packages) / `go mod tidy` clean / whole-suite / gofmt.

**Strongly built, credited up front:**
- Clean 3-layer split with compile-time `var _ driver.EFS = (*Mock)(nil)`; IDs/ARNs via `idgen`.
- **Routing is right** — `Matches` gates on the `/2015-02-01/` prefix + a known root, registered *before* the S3 catch-all, so it can't shadow S3 and can't miss nested paths (`…/mount-targets/{id}/security-groups`, `…/file-systems/{id}/policy`). This is the #331 shadow class handled correctly.
- Timestamps as epoch-seconds; correct 201/204/200 codes; all 30 ops routed; real `aws-sdk-go-v2/service/efs` roundtrip tests; determinism (`Clock`) and concurrency clean (except M1); docs op-count correct (1782).

Four Mediums to fix (all categories you already fixed cleanly on the KMS PR).

### M1 — Idempotency race in `CreateFileSystem`
The creation-token scan over `fileSystems.All()` and the later `Set(randomID, …)` aren't atomic and aren't token-keyed, so two concurrent `CreateFileSystem` calls with the same `CreationToken` both pass the scan and both create → **two file systems sharing one token**, violating EFS idempotency. Not caught by `-race` because no concurrent same-token test exists. Fix: a token→fsID index set atomically (e.g. a `SetIfAbsent`-backed token index) or a mutex around check+create. Inline.

### M2 — Error types collapse to file-system exceptions
`writeErr` maps every `NotFound→FileSystemNotFound`, `AlreadyExists→FileSystemAlreadyExists`, `FailedPrecondition→FileSystemInUse`. EFS's real API has per-resource typed exceptions — `MountTargetNotFound`, `MountTargetConflict`, `AccessPointNotFound`, `PolicyNotFound`, `ReplicationNotFound` — all separate in `aws-sdk-go-v2/service/efs`. So a client's `errors.As(err, &*efstypes.MountTargetNotFound{})` after deleting a bogus mount target never matches (it gets `FileSystemNotFound`). HTTP status is correct, so this is SDK-fidelity, not a crash — but it's the same class you fixed on #369, and the root cause is that this single-exception mapper was copied from vpclattice (where one `ResourceNotFoundException` is correct) to EFS's ~8-exception API. Fix: pass the resource kind / a specific code at the call site so the wire selects the right `X-Amzn-Errortype`. Inline.

### M3 — `DeleteFileSystem` orphans access points & leaks index entries
It guards only on mount targets. Deleting a file system that has access points succeeds but never removes the `apIndex`/`mtIndex` entries (an unbounded leak across create/delete churn) and silently drops policy/lifecycle/replication. There's also a TOCTOU window: the MT count is checked under `RLock` then released before `Delete`, so a concurrent `CreateMountTarget` can attach a live MT to an FS that's then deleted. Fix: delete the `apIndex`/`mtIndex` entries on FS delete (and decide whether to block delete while access points exist, per EFS). Inline.

### M4 — Filenames violate STRUCTURE.md §3 (snake_case, full words)
`filesystems.go` / `mounttargets.go` in all three layers should be `file_systems.go` / `mount_targets.go` (AWS's own types are two words: `FileSystem`, `MountTarget`). The sibling this PR mirrors — `server/aws/vpclattice/` — already splits every compound (`service_networks.go`, `target_groups.go`, `access_logs.go`), and STRUCTURE §7's CI check targets non-snake_case filenames. Pure `git mv` across the layers (the same-filename-across-layers half of §3 is already satisfied).

### Low
The `CreateFileSystem` doc comment says it "returns the existing file system (idempotent)" but the code returns `AlreadyExists` — the behavior is the real-EFS one, so just fix the comment; hardcoded mount-target AZ/VPC/IP (`us-east-1a` regardless of region; same default IP and a differing VPC per MT on one FS); `notFound`/405 emit `BadRequest` errortype with the wrong status; `DescribeReplicationConfigurations` returns empty instead of `ReplicationNotFound`; access-point input pointer-aliasing (deep-copy `PosixUser`/`RootDirectory` on store); `UpdateFileSystemProtection` unimplemented; and test gaps (MT-per-subnet conflict, typed-error negatives, describe-MT-by-MT/AP id, plus `TestSDKDeleteWithMountTargetsGuarded` is mislabeled — it never creates a mount target).

Requesting changes on M1–M4; the Lows are quick follow-ups. No AI attribution.

providers/aws/efs/file_systems.go
thzgajendra: Fixed. `CreateFileSystem` now claims the creation token atomically via a dedicated token→id index (`m.tokenIndex.SetIfAbsent(token, id)`): if the token is already owned the call returns `FileSystemAlreadyExists` without creating anything, so two concurrent same-token creates can't both succeed. The `All()` scan is gone. Also corrected the doc comment — a repeated token is rejected (matching real EFS's FileSystemAlreadyExists), not silently returned.

providers/aws/efs/file_systems.go
thzgajendra: Fixed. `DeleteFileSystem` now takes the file system's write lock and holds it across the check+delete (closing the TOCTOU window against a concurrent `CreateMountTarget`), also blocks when access points still exist, and releases the token and mount-target/access-point index entries on delete so nothing is orphaned or leaked.

server/aws/efs/errors.go
thzgajendra: Fixed. Errors now carry a resource `Kind` via `driver.ResourceError`, and `writeErr` uses `errors.As` to select the precise exception: MountTargetNotFound / AccessPointNotFound / PolicyNotFound / ReplicationNotFound and MountTargetConflict / ReplicationAlreadyExists, falling back to the file-system exception for untagged errors. Added SDK tests asserting each typed exception unmarshals.

thzgajendra:
Thanks for the review — all four Mediums and the Lows are addressed in `19c7827`.

**Mediums**
- **M1 — idempotency race:** `CreateFileSystem` claims the creation token atomically via a token→id index (`SetIfAbsent`); duplicate tokens are rejected without creating. Removed the `All()` scan and fixed the misleading doc comment.
- **M2 — typed-error fidelity:** errors carry a resource `Kind` (`driver.ResourceError`); `writeErr` maps to the precise EFS exception (MountTargetNotFound / AccessPointNotFound / PolicyNotFound / ReplicationNotFound / MountTargetConflict / ReplicationAlreadyExists) with a file-system fallback.
- **M3 — orphan + leak on delete:** `DeleteFileSystem` holds the write lock across check+delete, blocks on remaining access points, and releases token/MT/AP index entries.
- **M4 — filenames:** renamed to snake_case across all layers (`file_systems.go`, `mount_targets.go`).

**Lows**
- `DescribeReplicationConfigurations` returns `ReplicationNotFound` for a filtered FS with no config.
- Deep-copy `PosixUser`/`RootDirectory` on the access-point store to stop input aliasing.
- Mount-target AZ derived from the configured region (`use1-az1` for us-east-1) instead of hardcoded.
- Renamed the mislabeled describe-all test and added a real delete-with-mount-target guard test, plus MT-per-subnet conflict, typed not-found negatives, and describe-MT-by-MT/AP-id coverage.

**Gate:** `go build/vet/test ./...` green, `go mod tidy` clean, `golangci-lint` 0 new issues, CodeQL 0 EFS findings, and AWS-CLI E2E confirms the typed exceptions, per-subnet conflict, and delete guard end-to-end.

NitinKumar004 (approved):

Re-reviewed the fix commit (`19c78279`) in an isolated worktree at the PR head; gate green (build / vet / `go test -race` both EFS packages / gofmt), every prior finding verified fixed end-to-end.

- **M1 (idempotency race) — fixed.** `CreateFileSystem` now atomically claims the creation token via `tokenIndex.SetIfAbsent` (file_systems.go:25), and `DeleteFileSystem` frees it. No more duplicate FS per token.
- **M2 (error-type collapse) — fixed, and cleanly.** New `driver.ResourceError{Kind}` tags every mount-target / access-point / policy / replication error with its resource kind, and the wire maps the kind to the right `X-Amzn-Errortype` (`MountTargetNotFound`, `AccessPointNotFound`, `MountTargetConflict`, `PolicyNotFound`, `ReplicationNotFound`). New SDK negative tests assert these via `errors.As` — exactly the typed-error coverage that was missing. Nice reusable discriminator (cleaner than per-op sentinels).
- **M3 (delete orphan/leak/TOCTOU) — fixed.** `DeleteFileSystem` holds a write lock, blocks when access points exist, and cleans the token index.
- **M4 (STRUCTURE §3) — fixed.** Renamed to `file_systems.go` / `mount_targets.go` across all three layers.

Lows addressed too: `ReplicationNotFound` on missing config, deep-copied access-point input, region-derived mount-target AZ, and the previously-missing MT-per-subnet-conflict + typed-not-found negative tests. Thorough, well-tested turnaround. LGTM.

PR #371, opened 2026-08-09
Add AWS ACM Emulation with Full Parity (#295 Workstream B)

thzgajendra:
## Objective / Issue

Third net-new service of **Workstream B (#295)**: emulate **AWS ACM** (Certificate Manager) with SDK-wire fidelity so real `aws-sdk-go-v2/service/acm` clients and the `aws acm` CLI work against cloudemu by only changing the endpoint. All **17 operations**, no stubs.

## What we found

cloudemu had no ACM. It needed a full 4-layer vertical (driver → in-memory provider → JSON 1.1 wire handler → registration). ACM uses AWS JSON 1.1 (`X-Amz-Target: CertificateManager.`) — the same family as KMS/Secrets, whose handler pattern this mirrors.

## How we fixed it

New ACM vertical: `services/acm/driver/`, `providers/aws/acm/`, `server/aws/acm/`, registered via `Drivers.ACM` in `server/aws/aws.go` and `p.ACM` in `providers/aws/aws.go`.

| Family | Operations |
|--------|-----------|
| Lifecycle | RequestCertificate, DescribeCertificate, ListCertificates, DeleteCertificate, GetCertificate |
| Import / export | ImportCertificate, ExportCertificate |
| Renewal / revocation | RenewCertificate, RevokeCertificate, ResendValidationEmail |
| Options / search | UpdateCertificateOptions, SearchCertificates |
| Tags | AddTagsToCertificate, RemoveTagsFromCertificate, ListTagsForCertificate |
| Account | GetAccountConfiguration, PutAccountConfiguration |

**Certificates are real, not stubbed** (parallel to the KMS crypto work): `RequestCertificate` generates a genuine self-signed X.509 (RSA-2048, SHA-256) for the domain + SANs and auto-issues it (the emulator can't perform real domain validation), so `GetCertificate` returns parseable PEM. `ImportCertificate` validates + stores external PEM, `ExportCertificate` returns cert/chain/key, `RenewCertificate` re-issues fresh material, `RevokeCertificate` moves to `REVOKED`.

## Alternatives not taken

- **Leaving certs in `PENDING_VALIDATION`** — would make them unusable in local dev; instead we auto-issue with real material (the local analog of a validated public cert), while still exposing the DNS CNAME validation record real clients read.
- **Private CA (PCA) issuance** — out of scope for this PR; the emulated certs are self-signed (their own chain root).

## Docs / Tests / Playground

- Docs: `docs/services.md` (service row 28 + detail section; grand total 1769) and `docs/architecture.md`.
- Tests: real `aws-sdk-go-v2/service/acm` SDK roundtrip tests (request→describe→get with X.509 parse assertion, list/delete, tags, import/export, renew, update-options, revoke, account config, typed ResourceNotFoundException) + provider unit tests.
- E2E: every operation verified against a live `cloudemu start` server via the real `aws acm` CLI, including an export→import round-trip and openssl-parseable PEM.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go mod tidy` (adds `aws-sdk-go-v2/service/acm`)
- [x] CodeQL (pending confirmation; will verify 0 findings in ACM files)
- [x] `aws acm` CLI E2E on a running server

## Risk & Rollback

Low risk: entirely additive — a new service package plus two registration lines and one struct field; nothing existing changes behavior. `Drivers.ACM` is nil-skipped when unset. Rollback = revert the branch.

## Conclusion

AWS ACM is fully emulated with real X.509 certificates — the third net-new Workstream B service after KMS and EFS.


providers/aws/acm/operations.go
```diff
@@ -0,0 +1,192 @@
...
+	if chain != "" {
+		return chain
+	}
+
+	return cert
+}
+
+// ExportCertificate returns the cert, chain, and (passphrase-protected in real
+// ACM; returned as-is here) private key PEM.
+func (m *Mock) ExportCertificate(
```
NitinKumar004, line 93: **M1 (key-export leak).** This gates only on passphrase presence and `PrivateKeyPEM != ""` — no check of `cd.cert.Type`. Since `RequestCertificate`/`RenewCertificate` always store a private key, a caller can `ExportCertificate` the key of a managed `AMAZON_ISSUED` cert, which real ACM forbids (only private-CA/exportable certs are exportable). `GetCertificate` already withholds the key, so the code distinguishes public metadata from private material — gate export on `Type == IMPORTED` (or an exportable flag).

providers/aws/acm/operations.go
```diff
@@ -0,0 +1,192 @@
...
+	"time"
+
+	"github.com/stackshy/cloudemu/v2/errors"
+	"github.com/stackshy/cloudemu/v2/services/acm/driver"
+)
+
+// ImportCertificate imports an externally-issued certificate. With an ARN it
+// re-imports (updates) the existing certificate; otherwise it creates a new
+// IMPORTED certificate. The certificate PEM is validated as parseable X.509.
+func (m *Mock) ImportCertificate(_ context.Context, in driver.ImportCertificateInput) (string, error) {
```
NitinKumar004, line 15: **M3 (import validation).** Only `PrivateKeyPEM == ""` is checked here, and `applyImported` just stores the bytes — the key is never parsed or matched to the certificate, so garbage and mismatched keys import successfully (`err=nil`). Real ACM returns `InvalidParameterException` for both. `tls.X509KeyPair(in.CertificatePEM, in.PrivateKeyPEM)` validates parse + match in one call; return `InvalidArgument` on failure. (The certificate side is already parsed, so this is an asymmetric gap.)

providers/aws/acm/certificates.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+
+		out = append(out, dv)
+	}
+
+	return out
+}
+
+// DescribeCertificate returns a certificate's metadata.
+func (m *Mock) DescribeCertificate(_ context.Context, arn string) (*driver.Certificate, error) {
```
NitinKumar004, line 115: **M2 (aliased map → data race).** `out := cd.cert; return &out` copies the struct under RLock, but `Tags` (map) plus `SubjectAlternativeNames`/`DomainValidationOptions`/`InUseBy` (slices) are reference types shared with the stored cert. Once the RLock is released, a caller reading `.Tags` concurrent with `AddTagsToCertificate` (mutates `cd.cert.Tags` under Lock) is a concurrent map read/write → race/panic. Deep-copy these on the way out (as `ListTagsForCertificate` already does with `copyTags`); same for `ListCertificates`.

NitinKumar004 (changes requested):
## Deep review — AWS ACM emulation

Reviewed in an isolated worktree at the PR head via a 2-agent fan-out (structure/wire/X.509 + state/import-export/concurrency), every finding re-verified. **Gate green** — build / vet / `go test -race` (both ACM packages) / `go mod tidy` clean / whole-suite / gofmt.

**Strongly built, credited up front:**
- **The X.509 core is genuinely solid** — RSA key + 128-bit serial via `crypto/rand`, generated PEM round-trips through `x509.ParseCertificate`, `DNSNames` = CN + deduped SANs, unique serials across issues, sane validity; `GetCertificate` returns parseable PEM.
- **Determinism boundary is exactly right** — timestamps via `Clock`, key/serial via `crypto/rand` (correct — you can't make X.509 keys deterministic).
- **Private material never crosses the wire** — the Describe/Get wire types omit the PEM fields even though the driver struct carries them. Verified, no leak there.
- `Matches` on `CertificateManager.` is disjoint from every sibling; 17-op parity confirmed (driver = handler routes = docs); real `aws-sdk-go-v2/service/acm` roundtrip tests; per-cert `RWMutex`; clean `-race`.

No High. A cluster of Mediums to fix (same categories you turned around cleanly on #369/#370).

### M1 — `ExportCertificate` leaks the private key of managed (AMAZON_ISSUED) certs
It gates only on passphrase presence + `PrivateKeyPEM != ""`, with no `Type` check — and `RequestCertificate`/`RenewCertificate` always store a private key — so a caller can export the key of a managed public cert. Real ACM forbids `ExportCertificate` on public `AMAZON_ISSUED` certs (only private-CA/exportable). The code already withholds the key in `GetCertificate`, so it knows the distinction. Inline.

### M2 — `DescribeCertificate`/`ListCertificates` alias the live `Tags` map/slices (data race)
`out := cd.cert; return &out` shallow-copies under RLock, but `Tags` (map) and `SubjectAlternativeNames`/`DomainValidationOptions`/`InUseBy` (slices) are reference types shared with the store. A caller reading `.Tags` after the lock releases, concurrent with `AddTagsToCertificate`/`RemoveTagsFromCertificate` (which mutate `cd.cert.Tags` under Lock), is a concurrent map read/write → race/panic. `ListTagsForCertificate` already deep-copies via `copyTags` — follow that on Describe/List. Inline.

### M3 — `ImportCertificate` never validates the private key
It checks only `PrivateKeyPEM == ""`; `applyImported` just stores the bytes. It never parses the key or matches it to the certificate, so garbage and mismatched keys are accepted (`err=nil`) — the most common ACM import error path. Real ACM returns `InvalidParameterException` ("Unable to parse the private key" / "certificate and private key do not match"). `tls.X509KeyPair(certPEM, keyPEM)` does both checks in one call. Inline.

### M4 — Revoke/Renew have no state-machine guards
`RevokeCertificate` unconditionally sets `Status=REVOKED` (revoke a PENDING/IMPORTED/already-REVOKED cert with no error), and `RenewCertificate` re-issues fresh serial/PEM but never resets `Status` — so Revoke-then-Renew yields a cert with brand-new material still marked `REVOKED`, an inconsistent state a real client can produce.

### M5 — Reported `KeyAlgorithm`/`SignatureAlgorithm` don't match the material
`RequestCertificate` echoes the requested `KeyAlgorithm` (e.g. `EC_prime256v1`, `RSA_4096`) into `DescribeCertificate`, but generation is always RSA-2048/SHA256WITHRSA, and `ImportCertificate` hardcodes `RSA_2048` regardless of the imported cert. A client requesting EC gets Describe→EC but Get→RSA PEM. Normalize/validate the input to what's actually produced.

### M6 — No `CertificateArn` format validation
A malformed ARN falls through `getCert` → store miss → `ResourceNotFoundException`, where real ACM returns `InvalidArnException`. (Narrower than the KMS/EFS taxonomy issues since `ValidationException` is SDK-modeled — this is only about the specific ARN-error type.)

### Low
`x509gen.go` should be snake_case/full-word (`certificate_gen.go`) per STRUCTURE §3; `ExportCertificate` returns the key unencrypted despite requiring a passphrase; no 50-tag `TooManyTagsException` limit; `ListCertificates`/`SearchCertificates` ignore `MaxItems`/`NextToken` (never paginate); the `InUseBy` delete-guard is dead code (never populated); re-import over an ARN force-flips `Type`→IMPORTED; SAN list is unbounded (real ACM caps at 100).

Requesting changes on the M1/M2/M3 cluster (+ M4–M6); the Lows are quick follow-ups. No AI attribution.

providers/aws/acm/operations.go
thzgajendra: Fixed. `ExportCertificate` now rejects public `AMAZON_ISSUED` certs with `InvalidStateException` (only IMPORTED are exportable), mirroring `GetCertificate`'s existing key-withholding. Added `TestSDKExportManagedCertRejected`; `TestSDKImportAndExport` now exports an imported cert (the allowed path).

providers/aws/acm/operations.go
thzgajendra: Fixed. `ImportCertificate` now runs `tls.X509KeyPair(cert, key)`, which both parses the key and verifies it matches the certificate — garbage or mismatched keys return `InvalidParameterException`. Added a mismatched-key negative case to `TestSDKImportAndExport`.

providers/aws/acm/certificates.go
thzgajendra: Fixed. Added `copyCert` (deep-copies `Tags` + `SubjectAlternativeNames`/`DomainValidationOptions`/`InUseBy`) and use it in both `DescribeCertificate` and `ListCertificates` (Search delegates to List), so returned values no longer alias the store under concurrent tag mutation.

thzgajendra:
Thanks for the thorough review — all six Mediums and the actionable Lows are fixed in `3019917` (branch also synced with `development`, which now includes the merged EFS PR).

**Mediums**
- **M1 — export leak:** `ExportCertificate` rejects public `AMAZON_ISSUED` certs (`InvalidStateException`); only IMPORTED are exportable.
- **M2 — aliasing race:** new `copyCert` deep-copies `Tags` + SANs/DomainValidationOptions/InUseBy in `DescribeCertificate`/`ListCertificates`.
- **M3 — import validation:** `tls.X509KeyPair` parses the key and verifies it matches the cert; garbage/mismatch → `InvalidParameterException`.
- **M4 — state guards:** `RevokeCertificate` only revokes `ISSUED` certs; `RenewCertificate` resets `Status` to `ISSUED` so revoke-then-renew is consistent.
- **M5 — algorithm fidelity:** `RequestCertificate` rejects unsupported `KeyAlgorithm` (the emulator issues RSA_2048, so Describe==Get); `ImportCertificate` derives `KeyAlgorithm`/`SignatureAlgorithm` from the parsed cert.
- **M6 — ARN validation:** malformed ARNs return `InvalidArnException` (typed via `driver.APIError` + `errors.As` in `writeErr`).

**Lows**
- Renamed `x509gen.go` → `certificate_gen.go` (STRUCTURE §3).
- 50-tag `TooManyTagsException` cap; 100-domain SAN cap.
- Removed the dead `InUseBy` delete-guard.
- Re-import over a managed cert is rejected (no silent `Type` flip).

**Deferred (Low, needs interface changes — happy to do as a follow-up):** `MaxItems`/`NextToken` pagination on List/Search, and passphrase-encrypting the exported key (currently returned unencrypted). Flag if you'd like these folded in here instead.

**New tests:** import/export via self-generated material, key-mismatch reject, managed-export reject, double-revoke reject, invalid-ARN.

**Gate:** build/vet/`go test -race` (ACM pkgs) green, `go mod tidy` clean, `golangci-lint` 0 new, CodeQL 0 ACM findings.

NitinKumar004 (approved):

Re-reviewed the fix commit (`30199177`) in an isolated worktree at the PR head; gate green (build / vet / `go test -race` both ACM packages / gofmt), every finding verified fixed with an SDK regression test.

- **M1 (managed-cert key export) — fixed.** `ExportCertificate` gates on `Type == TypeAmazonIssued` and refuses (`InvalidState`), matching real ACM and `GetCertificate`'s own key-withholding. `TestSDKExportManagedCertRejected`.
- **M2 (aliased `Tags` map → data race) — fixed.** New `copyCert` deep-copies the map + `SubjectAlternativeNames`/`InUseBy`/`DomainValidationOptions`, used by both `DescribeCertificate` and `ListCertificates`.
- **M3 (import key validation) — fixed.** `tls.X509KeyPair(cert, key)` validates both parse and cert↔key match; garbage/mismatched keys now rejected.
- **M4 (revoke/renew state) — fixed.** Revoke restricted to ISSUED; double-/non-issued-revoke rejected. `TestSDKRevokeNonIssuedRejected`.
- **M5 (algorithm mismatch) — fixed.** Unsupported request algorithms rejected (emulator issues RSA_2048); Import derives key/signature algorithm from the parsed certificate.
- **M6 (ARN validation) — fixed.** Malformed ARNs yield `InvalidArnException` (distinct from a well-formed-but-absent ARN). `TestSDKInvalidArn`.

Lows handled too: `x509gen.go`→`certificate_gen.go`, 50-tag `TooManyTagsException`, 100-domain SAN cap, dead `InUseBy` guard removed, re-import over a managed cert blocked. The X.509 core and determinism boundary were already sound. Thorough, well-tested turnaround. LGTM.

Issue #30, opened 2026-03-19
Add Azure Key Vault (Secrets) service


NitinKumar004:
### Operations
- `CreateSecret(name, value)` — store a secret
- `GetSecret(name)` — retrieve secret value
- `UpdateSecret(name, value)` — update, auto-increment version
- `DeleteSecret(name)`
- `ListSecrets()` — list names only
- `GetSecretVersion(name, version)` — retrieve specific version
- `RotateSecret(name)` — simulate rotation with callback

### Why
Azure Key Vault is the standard for secret management in Azure. Testing secret lifecycle without real Key Vault reduces risk and speeds up development.

### Acceptance Criteria
- [ ] Azure implementation in `providers/azure/keyvault/`
- [ ] Uses shared driver interface from `secretmanager/driver/driver.go`
- [ ] Thread-safe using `sync.RWMutex`
- [ ] Versioning auto-increments on each `UpdateSecret`
- [ ] `GetSecretVersion` returns correct version or `NotFound`
- [ ] `RotateSecret` triggers registered callback and increments version
- [ ] Duplicate secret name returns `AlreadyExists` error
- [ ] Operations on non-existent secret return `NotFound` error
- [ ] Added to Azure `Provider` struct and wired in factory
- [ ] Unit tests with ≥90% coverage
- [ ] Integration test added in `cloudemu_test.go`
- [ ] Passes `golangci-lint run` with zero issues
- [ ] README.md updated

PR #372, opened 2026-08-09
Add AWS Kinesis Data Streams Emulation with Full Parity (#295 Workstream B)

thzgajendra:
## Objective
Extend cloudemu's AWS coverage (roadmap #295, Workstream B) with **Kinesis Data Streams** at full SDK-wire parity, so real `aws-sdk-go-v2/service/kinesis` clients and the `aws kinesis` CLI work by only changing the endpoint.

## What we built
Kinesis across the standard four layers — `services/kinesis/driver` (interface + types), `providers/aws/kinesis` (in-memory `Mock`), `server/aws/kinesis` (AWS JSON 1.1, `X-Amz-Target: Kinesis_20131202.`), and registration in `providers/aws/aws.go` + `server/aws/aws.go`.

**Records behave like real Kinesis, not stubs:**
- Streams are partitioned into shards owning a 128-bit MD5 hash-key range; `PutRecord`/`PutRecords` route a record to the open shard covering `MD5(partitionKey)` (or an explicit hash key) and assign a per-stream **monotonic sequence number**.
- Opaque **shard iterators** honor `TRIM_HORIZON`, `LATEST`, `AT`/`AFTER_SEQUENCE_NUMBER`, and `AT_TIMESTAMP`; `GetRecords` advances the iterator and, at the end of a closed shard, returns the **child shards**.
- **Resharding** — `SplitShard`/`MergeShards`/`UpdateShardCount` close parents and create children with correct hash-key ranges and parent links, so records already written stay readable.
- Enhanced fan-out **consumers**, enhanced **monitoring**, retention, **encryption**, stream **modes** (PROVISIONED/ON_DEMAND), **tags**, **resource policies**, and account settings/limits.

**38 operations** wired. Typed exceptions (`ResourceNotFoundException`, `ResourceInUseException`, `InvalidArgumentException`, `ExpiredIteratorException`) via `errors.As`.

## Alternatives not taken
- **`SubscribeToShard`** (enhanced fan-out HTTP/2 bidirectional event stream) is intentionally deferred — it needs an HTTP/2 event-stream transport distinct from the JSON 1.1 request path. Polling via `GetShardIterator`/`GetRecords` covers the same read semantics. Flagged in docs as a follow-up.
- Iterators encode a stable per-shard record index rather than a real byte offset — records are never trimmed in the emulator, so an index is a sufficient, deterministic cursor.

## Docs / Test / Playground
- `docs/services.md`: new service #30 + detail section + op-count total (→ 1837).
- `docs/architecture.md`: server-layer catalog mentions streaming (Kinesis).

## Test plan
- Provider unit tests: monotonic sequence numbers, deterministic partition-key routing, iterator advancement, retention guards.
- SDK-roundtrip tests (real `aws-sdk-go-v2` client vs `httptest`): create→describe→put→get-iterator→get-records, PutRecords batch, consumers + tags, SplitShard increases shard count, typed error assertions.
- E2E with the `aws kinesis` CLI: create-stream, put-record, get-records (data round-trips), put-records, register/list consumers, tags, delete guarded by consumers + `--enforce-consumer-deletion`.
- Gate: `go build/vet/test ./...`, `go mod tidy` clean, `golangci-lint` 0 new issues, CodeQL 0 Kinesis findings.

## Risk & rollback
Additive — a new self-contained service package plus two registration lines; no existing services touched. Rollback = revert the commit.

## Conclusion
Kinesis Data Streams joins the AWS catalog with real shard/record/iterator behavior. Follow-up: `SubscribeToShard` enhanced fan-out over HTTP/2.


providers/aws/kinesis/streams.go
```diff
@@ -0,0 +1,380 @@
...
+
+	sd.desc.RetentionPeriodHours = hours
+
+	return nil
+}
+
+// UpdateShardCount reshards to the target count by closing all open shards and
+// creating target new open shards partitioning the full hash-key space. The old
+// shards remain readable (closed) so in-flight records aren't lost.
+func (m *Mock) UpdateShardCount(
```
NitinKumar004, line 294: **High.** `UpdateShardCount` closes the open shards but builds replacements via `buildShards`, which leaves `ParentShardID`/`AdjacentParentShardID` empty — so `childShardsOf` (records.go:295, matches by parent id) returns nothing, and draining a closed shard yields `ChildShards=[]` + empty `NextShardIterator`. A consumer following the closed-shard→child-shard contract dead-ends and reprocesses from TRIM_HORIZON. `SplitShard`/`MergeShards` link parents correctly (resharding.go:104) — thread the same parent-linking into the reshard path here.

providers/aws/kinesis/records.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	key, ok := new(big.Int).SetString(keyStr, base10)
+	if !ok {
+		return "", "", invalidArg("ExplicitHashKey %q is not a valid hash key", explicitHashKey)
+	}
+
+	shard := shardForHashKey(sd.shards, key)
+	if shard == nil {
+		return "", "", errInUse("no open shard covers the record's hash key")
```
NitinKumar004, line 219: **Medium.** An out-of-range or negative `ExplicitHashKey` parses fine, `shardForHashKey` returns nil, and this surfaces `ResourceInUseException` (via `errInUse`→FailedPrecondition). Real Kinesis returns `InvalidArgumentException`. Bounds-check the explicit hash key (0 ≤ key < 2^128) and return `invalidArg`, keeping the "no open shard" branch as a genuine internal invariant (unreachable for MD5-routed keys, which always tile the full range).

providers/aws/kinesis/records.go
```diff
@@ -0,0 +1,329 @@
...
+				HashKeyRange: ss.shard.HashKeyRange,
+			})
+		}
+	}
+
+	return out
+}
+
+// ListShards lists a stream's shards.
+func (m *Mock) ListShards(_ context.Context, in driver.ListShardsInput) (*driver.ListShardsOutput, error) {
```
NitinKumar004, line 379: **Medium (pagination truncation).** `pageShards` computes a `more` flag but this discards it and returns an empty `NextToken`, so shards beyond `MaxResults` are dropped with no way to fetch the rest. Propagate the continuation token (`ListShardsOutput.NextToken` already exists). Same issue in `ListStreams`/`ListStreamConsumers`.

NitinKumar004 (changes requested):
## Deep review — Kinesis Data Streams

3-agent-equivalent fan-out (structure/wire/domain + concurrency/state), gate green (build/vet/`go test -race`/gofmt), every High re-verified. **The recurring aliasing race is absent here** — DescribeStream/GetRecords/ListShards return value copies (EnhancedMonitoring copy-on-write, Record `Data` copied, Tags deep-copied), `-race` clean. Typed errors are correct (no collapse — `APIError` passthrough + real-SDK roundtrip asserts `ResourceNotFoundException`/`ResourceInUseException`). Hash-space partitioning, MD5 big-endian routing, ExplicitHashKey, and Split/Merge lineage+record-preservation are all probe-verified correct. Determinism clean. Nice work.

### High
**`UpdateShardCount` breaks shard lineage.** `buildShards` (streams.go:13) creates the new shards with empty `ParentShardID`/`AdjacentParentShardID`, so `childShardsOf` can't link them — after an auto-scale, draining a closed shard returns `ChildShards=[]` and an empty `NextShardIterator`. A KCL-style consumer following the documented "GetRecords returns child shards at a closed shard's end" contract dead-ends and re-processes from TRIM_HORIZON. `SplitShard`/`MergeShards` set parents correctly (resharding.go:104) — only the auto-scaling path is inconsistent. Records aren't lost, but continuity is. Inline.

### Medium
- **ExplicitHashKey out-of-range → wrong exception.** An out-of-range/negative `ExplicitHashKey` falls through `shardForHashKey`→nil→`errInUse`→`ResourceInUseException` (records.go:157), where real Kinesis returns `InvalidArgumentException`. Add bounds validation. Inline.
- **Pagination silently truncates.** `ListShards` discards the `more` flag from `pageShards` and always returns an empty `NextToken` (records.go:317); `ListStreams`/`ListStreamConsumers` ignore `limit`/token too — shards/streams beyond `MaxResults` are dropped with no continuation. Inline.
- **No record/batch size/count validation** — real Kinesis rejects >1 MiB record, >5 MiB batch, >500 records; here they're silently accepted (`maxRecKiB` is stored but never enforced).
- **Test gaps** — resharding-preserves-records, non-TRIM_HORIZON iterator types, PutRecords partial-failure, and `Matches` negatives are untested (happy-path only).

### Low
Dead `ExLimitExceeded` constant / no limit enforcement (no-dead-code rule); `formatSeq` zero-pads only 8 digits so sequence numbers collide past 1e8; `SequenceNumberForOrdering` ignored; stream lifecycle states (`CREATING`/`UPDATING`/`DELETING`) are dead (always `ACTIVE`); STRUCTURE §3 server filename drift (`records_ops.go` vs provider `records.go`, consistent with the KMS/ACM siblings).

Requesting changes on the High + the two inline Mediums. No AI attribution.

providers/aws/kinesis/streams.go
thzgajendra: Fixed. `UpdateShardCount` now snapshots the open shards' hash ranges before closing them and links each new shard to the parent(s) whose range it overlaps (`ParentShardId`/`AdjacentParentShardId`), so `childShardsOf` finds them and draining a closed shard returns its `ChildShards` with a nil iterator — same closed-shard→child contract as SplitShard/MergeShards.

providers/aws/kinesis/records.go
thzgajendra: Fixed. `appendRecord` now bounds-checks the explicit hash key to [0, 2^128) and returns `InvalidArgumentException` for out-of-range/negative values. The `no open shard` branch stays as a genuine internal invariant (unreachable for MD5-routed and bounds-checked explicit keys).

providers/aws/kinesis/records.go
thzgajendra: Fixed. `ListShards` now propagates `NextToken` (and accepts it as the resume cursor), and `ListStreams`/`ListStreamConsumers` honor the page limit with `HasMoreStreams`/`NextToken`, ordering deterministically by name so pages don't drop or duplicate entries.

thzgajendra:
Thanks for the thorough review — all three fixed in `199a727` (service is also at full parity now: `SubscribeToShard` implemented and verified end-to-end with the real SDK). **High**: reshard links parent shard ids. **Med**: out-of-range ExplicitHashKey → InvalidArgumentException. **Med**: List{Shards,Streams,StreamConsumers} propagate the continuation token. Gate: build/vet/test/tidy clean, golangci-lint 0 new, race-clean.

NitinKumar004 (commented):
## Re-review — most findings fixed; one Medium remains

Re-reviewed the fix commit in an isolated worktree; gate green (build / `go test -race` both Kinesis packages). Great progress:
- **High (UpdateShardCount lineage) — fixed.** New shards are now linked to the parent(s) whose range they overlap (`ParentShardID`/`AdjacentParentShardID` set by range overlap), so a closed shard's `ChildShards` resolve after auto-scale.
- **Medium (ExplicitHashKey) — fixed.** Now bounds-checked to `[0, 2^128)` → `InvalidArgumentException` (records.go:157).
- **Medium (pagination) — fixed.** `ListShards` resumes from and emits `NextToken`.

Not approving yet — one Medium from the review is still open:
- **No record/batch size/count validation.** `PutRecord`/`PutRecords` still accept a >1 MiB record, a >5 MiB batch, or >500 records — real Kinesis rejects these with `ValidationException`, and `maxRecKiB` is stored but never enforced. For a "full parity" service this is the remaining canonical edge case.
- Lows still open (non-blocking): dead `ExLimitExceeded` constant / no limit enforcement; `formatSeq` 8-digit zero-pad collides past 1e8 records; `records_ops.go` vs provider `records.go` filename drift.

Fix the size/count validation and I'll approve. No AI attribution.

providers/aws/kinesis/records.go
```diff
@@ -0,0 +1,402 @@
...
+			continue
+		}
+
+		start, _ := new(big.Int).SetString(ss.shard.HashKeyRange.StartingHashKey, base10)
+		end, _ := new(big.Int).SetString(ss.shard.HashKeyRange.EndingHashKey, base10)
+
+		if key.Cmp(start) >= 0 && key.Cmp(end) <= 0 {
+			return ss
+		}
+	}
```
thzgajendra, line 97: Addressed the remaining Medium in `2678387`. `PutRecord` now rejects a record over the per-record cap (the stream's configured MaxRecordSize, else 1 MiB) with `ValidationException`, and `PutRecords` rejects >500 records or an aggregate payload (data + partition keys) over 5 MiB. Also enforced MaxRecordSize (was stored but unused), widened the sequence counter to 20 digits (no collision past 1e8 records), and removed the dead `ExLimitExceeded` constant. Also merged latest development (resolved docs/registration conflicts after SES v2 landed). Added provider tests for the oversize-record and 501-record cases.

thzgajendra:
Thanks for the re-review. Fixed the remaining Medium (record/batch size & count validation → `ValidationException`, MaxRecordSize now enforced) in `2678387`, plus the noted Lows (widened sequence counter to avoid collision past 1e8 records; removed the dead `ExLimitExceeded` constant). Also merged the latest `development` and resolved the docs/registration conflicts that appeared after SES v2 merged. Gate: build/vet/test/tidy clean, golangci-lint 0 new.

NitinKumar004 (approved):

Re-reviewed the fix commit in an isolated worktree; gate green (build / `go test -race` both Kinesis packages).

- **Medium (record/batch size + count validation) — fixed.** `PutRecord`/`PutRecords` now enforce the real Kinesis limits: 1 MiB per record (honoring a configured `MaxRecordSize`), 5 MiB per batch, and 500 records per batch, each returning `ValidationException` via `validationErr`. Covered by new tests (oversize record → `ValidationException`, 501-record batch → `ValidationException`).

That closes the last open finding — the High (UpdateShardCount shard-lineage), the ExplicitHashKey bounds, and the pagination Mediums were all fixed in the prior commit, and concurrency/determinism/typed-errors were already sound. Only Lows remain (non-blocking): dead `ExLimitExceeded` constant; `formatSeq` 8-digit collision past 1e8 records; `records_ops.go` vs `records.go` filename drift. LGTM.

PR #373, opened 2026-08-09
Add AWS SES v2 Emulation with Full Parity (#295 Workstream B)

thzgajendra:
## Objective
Extend cloudemu's AWS coverage (roadmap #295, Workstream B) with **Amazon SES v2** at full SDK-wire parity, so real `aws-sdk-go-v2/service/sesv2` clients and the `aws sesv2` CLI work by only changing the endpoint.

## What we built
SES v2 across the standard four layers — `services/sesv2/driver` (interface + types, 28 operations), `providers/aws/sesv2` (in-memory `Mock`), `server/aws/sesv2` (REST-JSON, `awsRestjson1`, path routing under the `/v2/email/` version prefix), and registration in `providers/aws/aws.go` + `server/aws/aws.go`.

- **Email identities** — create/get/delete/list; identities auto-verify to SUCCESS in the emulator (no real DNS/email round-trip), with DKIM attributes.
- **Configuration sets** — create/get/delete/list.
- **Sending** — `SendEmail` validates the from-identity and returns a generated MessageId; sent messages are stored so tests can assert them.
- **Suppression list**, **email templates** (create/get/update/delete/list), **account** details, and **tags**.

The `/v2/email/` version prefix makes the `Matches` predicate unambiguous with the S3 catch-all (no real bucket path begins with `/v2/email/`), matching how EFS gates on its version prefix.

## Docs / Test / Playground
- `docs/services.md`: new service row + detail section + op-count total.
- `docs/architecture.md`: server-layer catalog mentions SES.

## Test plan
- Provider unit tests + SDK-roundtrip tests (real `aws-sdk-go-v2` client vs `httptest`): identity lifecycle, SendEmail returns a MessageId, templates, tags, and error paths (not-found, duplicate).
- Gate: `go build/vet/test ./...`, `go mod tidy` clean, `golangci-lint` 0 new issues.

## Risk & rollback
Additive — a new self-contained service package plus registration lines; no existing services touched. Rollback = revert the commit.


providers/aws/sesv2/sending.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	if at := strings.LastIndex(from, "@"); at >= 0 {
+		domain := from[at+1:]
+		if d, ok := m.identities.Get(domain); ok {
+			return verifiedForSending(d)
+		}
+	}
+
+	return cerrors.Newf(cerrors.NotFound,
```
NitinKumar004: **Medium.** Returning `cerrors.NotFound` here makes `SendEmail` from an unverified/unknown identity surface as `NotFoundException`, but real SES v2 returns `MessageRejected` ("Email address is not verified…") for this — the main SendEmail rejection path. `MessageRejected` is emitted nowhere in the service, and `sdk_roundtrip_test.go` asserts `NotFoundException`, cementing the wrong type (a consumer keying on `types.MessageRejected` never matches). Tag this failure so the wire emits `MessageRejected` (HTTP 400) and fix the test. NotFoundException stays correct for the missing-config-set case (sending.go:30).

NitinKumar004 (changes requested):
## Deep review — SES v2

2-agent fan-out (structure/wire/domain + concurrency/state), gate green (build/vet/`go test -race`/gofmt), findings re-verified. **Strongly built** — `Matches` is prefix-gated on `/v2/email/` + known roots (can't shadow S3, registered before the catch-all, nested paths route correctly); typed exceptions for the common cases via `X-Amzn-Errortype` (SDK `errors.As` verified); **no aliasing race** (Get/List deep-copy Tags/DkimTokens/etc., 250-goroutine `-race` probe clean); determinism clean (Clock + idgen); auto-verify + send-validation + suppression/templates/config-sets all correct; 28 ops parity.

### Medium — `SendEmail` from an unverified/unknown identity returns `NotFoundException`, not `MessageRejected`
`checkFromIdentity` returns `cerrors.NotFound` (sending.go:74) → the wire maps it to `NotFoundException`. Real SES v2 `SendEmail` returns **`MessageRejected`** ("Email address is not verified…") for this — the single most common SendEmail error path. `MessageRejected` is never emitted anywhere in the service, and `sdk_roundtrip_test.go` *asserts* `NotFoundException`, enshrining the wrong taxonomy — so a consumer whose bounce/retry logic keys on `types.MessageRejected` never matches. (Not a crash: `NotFoundException` is a valid, SDK-modeled SendEmail error, so `errors.As` still deserializes — hence Medium.) Fix: tag the from-identity failure so SendEmail surfaces `MessageRejected`/HTTP 400 (the per-kind mechanism EFS/WAF use), and update the test. Inline.

### Low
`configsets.go` → `configuration_sets.go` (STRUCTURE §3: snake_case, full words, no abbreviation); `notFound`/`methodNotAllowed` emit a 404/405 with `X-Amzn-Errortype: BadRequestException` (status/type mismatch, unreachable via the SDK); unknown SES roots fall through to the S3 catch-all (acceptable anti-shadow tradeoff — worth a comment); `decodeJSON` empty-body detection via `err.Error()=="EOF"` string compare rather than `errors.Is(io.EOF)`; DKIM/MAIL-FROM sub-resources and config-set/template tagging are untested; `verifiedForSending`'s `FailedPrecondition`→`ConflictException` branch is dead (nothing ever clears the auto-verify flag).

Requesting changes on the Medium (headline SendEmail error type + the test that locks it in). No AI attribution.

providers/aws/sesv2/sending.go
thzgajendra: Fixed. `SendEmail` from an unverified/unknown sending identity now returns `MessageRejected` (HTTP 400) — I added a MessageRejected-tagged `driver.APIError` carried through `writeErr` via `errors.As`, and `checkFromIdentity`/`verifiedForSending` emit it. `NotFoundException` stays for the missing-configuration-set path. Updated both the SDK-roundtrip and provider tests to assert `MessageRejected`.

thzgajendra:
Thanks — fixed in `ee8e016` (service is also at full parity now: 113/113 SES v2 operations). Unverified-sender `SendEmail` returns `MessageRejected`; NotFoundException retained for missing config sets. Gate: build/vet/test/tidy clean, golangci-lint 0 new. (Same commit also carries a small bounded-retry stabilization for a pre-existing flaky Bedrock streaming test that was intermittently reddening CI.)

NitinKumar004 (approved):

Re-reviewed the fix commit in an isolated worktree; gate green (build / `go test -race` both SES v2 packages).

- **Medium (SendEmail error type) — fixed.** `checkFromIdentity` now returns a `MessageRejected`-tagged error (`ExMessageRejected`), so sending from an unverified/unknown identity surfaces `MessageRejected` (HTTP 400) as real SES v2 does, while `NotFoundException` is retained for the missing-config-set case. The wire honors the tagged exception via `errors.As`.

The concurrency (deep-copy Get/List), determinism (Clock + idgen), routing (prefix-gated `Matches`), and domain behavior were already sound. Only Lows remain (non-blocking): `configsets.go` → `configuration_sets.go` (STRUCTURE §3); the `notFound`/`methodNotAllowed` status/type mismatch; the dead `verifiedForSending` branch; untested DKIM/MAIL-FROM sub-resources. LGTM.

PR #374, opened 2026-08-09
Add AWS WAFv2 Emulation with Full Parity (#295 Workstream B)

thzgajendra:
## Objective
Extend cloudemu's AWS coverage (roadmap #295, Workstream B) with **AWS WAFv2** at full SDK-wire parity, so real `aws-sdk-go-v2/service/wafv2` clients work by only changing the endpoint.

## What we built
WAFv2 across the standard four layers — `services/wafv2/driver` (interface + types, 27 operations), `providers/aws/wafv2` (in-memory `Mock`), `server/aws/wafv2` (AWS JSON 1.1, `X-Amz-Target: AWSWAF_20190729.`), and registration in `providers/aws/aws.go` + `server/aws/aws.go`.

- **WebACLs, IPSets, RuleGroups, RegexPatternSets** — create/get/update/delete/list.
- **WebACL resource association** — Associate/Disassociate/GetWebACLForResource/ListResourcesForWebACL.
- **Tags** — TagResource/UntagResource/ListTagsForResource.
- **LockToken optimistic concurrency** — every mutation rotates the resource's lock token; Update/Delete require the caller's token to match, else `WAFOptimisticLockException`.
- **Scope partitioning** — REGIONAL vs CLOUDFRONT namespaces are keyed separately. Rules/statements are stored verbatim so Get returns exactly what was Put.

Typed WAF exceptions (`WAFNonexistentItemException`, `WAFDuplicateItemException`, `WAFOptimisticLockException`, `WAFInvalidParameterException`) via `errors.As`.

## Docs / Test / Playground
- `docs/services.md`: new service row + detail section + op-count total.
- `docs/architecture.md`: server-layer catalog mentions WAFv2.

## Test plan
- Provider unit tests + SDK-roundtrip tests (real `aws-sdk-go-v2` client vs `httptest`): create→get→update(with lock token)→delete, IPSet CRUD, tags, and error paths (nonexistent, duplicate, stale lock token).
- Gate: `go build/vet/test ./...`, `go mod tidy` clean, `golangci-lint` 0 new issues.

## Risk & rollback
Additive — a new self-contained service package plus registration lines; no existing services touched. Rollback = revert the commit.


providers/aws/wafv2/webacl.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+	}
+
+	return nil, false
+}
+
+// CreateWebACL creates a web ACL, storing its rule configuration verbatim.
+//
+//nolint:gocritic // in is the public input, taken by value to match the driver API.
+func (m *Mock) CreateWebACL(_ context.Context, in driver.CreateWebACLInput) (*driver.WebACL, error) {
```
NitinKumar004, line 45: **Medium (duplicate-name race + scope validation).** Two things here: (1) the `*ByName` uniqueness scan and the `Set` below aren't atomic, so two concurrent creates with the same scope+name both succeed (uniqueness invariant violated under concurrency — same class across all 4 Creates); guard with an atomic name index or `SetIfAbsent`. (2) `Scope` is only checked for non-empty — a bogus value like `"FOO"` is accepted and stored in an unreachable namespace; allow-list `REGIONAL`/`CLOUDFRONT` and return `WAFInvalidParameterException` otherwise.

providers/aws/wafv2/webacl.go
```diff
@@ -0,0 +1,169 @@
...
+	wd.acl.CustomResponses = in.CustomResponses
+	wd.acl.CaptchaConfig = in.CaptchaConfig
+	wd.acl.ChallengeConfig = in.ChallengeConfig
+	wd.acl.LockToken = newLockToken()
+
+	return wd.acl.LockToken, nil
+}
+
+// DeleteWebACL removes a web ACL, enforcing the lock token.
+func (m *Mock) DeleteWebACL(_ context.Context, ref driver.Ref, lockToken string) error {
```
NitinKumar004, line 143: **Low/Medium (Delete optimistic-lock TOCTOU).** This locks, compares the lock token, `Unlock()`s, and only then calls `webACLs.Delete()`. A concurrent `Update` landing between the unlock and the delete rotates the token and rewrites the rules, yet this delete — authorized by a now-stale token — still proceeds. `Update` gets this right (compare+mutate+rotate under one lock); fold the `Delete` into the same locked section. Same shape in ipset/rulegroup/regexset.

NitinKumar004 (changes requested):
## Deep review — WAFv2

2-agent fan-out (structure/wire/domain + concurrency/state), gate green (build/vet/`go test -race`/gofmt). **This is the cleanest of the batch** — the marquee risks are all handled right: LockToken `Update` is compare-mutate-rotate under a single held lock (no TOCTOU); Get/List deep-copy (`copyWebACL`/`copyIPSet`, `json.RawMessage` via `copyBytes`) so no aliasing race; scope is a composite storage key (`scope+"/"+id`) so REGIONAL/CLOUDFRONT are isolated (a wrong-scope Get/Delete 404s); typed WAF exceptions resolve via `errors.As`-first; rules stored verbatim as `json.RawMessage` (byte-exact round-trip); determinism clean. `-race` + an adversarial 200× double-update probe both clean.

### Medium
- **Duplicate-name Create race.** `CreateWebACL` (and the other 3 Creates) scan `*ByName` then `Set` with no mock-level lock spanning the two, so two concurrent same-scope+name creates both pass the scan and both insert — the uniqueness invariant real WAF enforces is violated under concurrency (logical race, not a data race, so `-race` stays green). Guard with an atomic name index / `SetIfAbsent`. Inline.
- **No scope-value validation on Create.** A bogus `Scope` like `"FOO"` is accepted and stored under a namespace neither valid-scope List returns; real WAF → `WAFInvalidParameterException`. Allow-list REGIONAL/CLOUDFRONT.

### Low
- **`Delete*` optimistic-lock TOCTOU** — Delete locks, compares the token, `Unlock()`s, *then* calls `webACLs.Delete()` (webacl.go:136), so a concurrent Update in that window rotates the token yet the delete still proceeds. Fold the store delete into the locked section. Inline.
- **Delete of an associated WebACL isn't rejected** — real WAF → `WAFAssociatedItemException`; here it succeeds and leaves a dangling assoc. The `driver.ExUnavailableEntity` constant is defined but never used (no-dead-code) — its natural home.
- `AssociateWebACL` doesn't enforce REGIONAL-only despite the comment saying so; `Get*` ignore the `Name` field (lookup by scope+id only); no CIDR/regex validation on IPSet/RegexPatternSet; `TagResource` counts overwritten keys as additions against the 200-tag cap; `WAFInvalidParameterException` is only provider-tested, never asserted over the wire.

Requesting changes on the two Mediums + the Delete TOCTOU. No AI attribution.

providers/aws/wafv2/webacl.go
thzgajendra: Fixed both. (1) All four Create paths (WebACL/IPSet/RuleGroup/RegexPatternSet) now run the by-name uniqueness scan and the store insert under a shared `createMu`, so two concurrent same-(scope,name) creates can't both succeed. (2) Added `validateScope` — Scope is allow-listed to REGIONAL/CLOUDFRONT, anything else returns `WAFInvalidParameterException`.

providers/aws/wafv2/webacl.go
thzgajendra: Fixed. `Delete` now compares the lock token and calls the store delete under a single held item lock (`defer wd.mu.Unlock()`), matching `Update`'s compare-mutate-under-one-lock shape — the compare/unlock/delete window is gone. Applied across WebACL, IPSet, RuleGroup, and RegexPatternSet.

thzgajendra:
Thanks — both Mediums fixed in `33e911b` (service is also at full SDK parity now, 59 ops). Create uniqueness+insert is atomic under a shared mutex across all four resource types; Scope is allow-listed (`WAFInvalidParameterException`); Delete folds the lock-token compare and store delete into one held lock. Gate: build/vet/test/tidy clean, golangci-lint 0 new, race-clean.

NitinKumar004 (approved):

Re-reviewed the fix commit in an isolated worktree; gate green (build / `go test -race` both WAFv2 packages).

- **Medium (duplicate-name Create race) — fixed.** Create now takes a `createMu.Lock()` spanning the name-uniqueness check and the insert, so concurrent same-scope+name creates can't both succeed.
- **Medium (no scope validation) — fixed.** `validateScope` rejects anything but REGIONAL/CLOUDFRONT.
- **Delete optimistic-lock TOCTOU — fixed.** `DeleteWebACL` now holds `wd.mu.Lock()` (deferred unlock) across the token compare **and** the store delete, so a concurrent Update can't slip between them.

This was already the cleanest of the batch (LockToken Update atomic, deep-copy Get/List, typed exceptions, scope-composite keying, verbatim rules). Only Lows remain (non-blocking): delete-of-associated not rejected (+ dead `ExUnavailableEntity`); `Get*` ignore the `Name` field; no CIDR/regex validation; `TagResource` overcounts on overwrite; `WAFInvalidParameterException` not asserted over the wire. LGTM.

PR #380, opened 2026-08-09
Add AWS OpenSearch Service Emulation with Full Parity (#295 Workstream B)

thzgajendra:
## Objective
Extend cloudemu's AWS coverage (#295, Workstream B) with **Amazon OpenSearch Service** at full SDK-wire parity — real `aws-sdk-go-v2/service/opensearch` clients work by only changing the endpoint.

## What we built
OpenSearch across the four layers — `services/opensearch/driver`, `providers/aws/opensearch` (in-memory `Mock`), `server/aws/opensearch` (REST-JSON under the version-path prefix, gated ahead of the S3 catch-all), registered in both `aws.go` files.
- **Domains** — create/describe/describe-config/update/delete/list; CREATING→ACTIVE with an endpoint; versions/compatible-versions.
- **Tags, access policies, packages, VPC endpoints, data sources, applications, reserved instances, upgrades, cross-cluster connections**, and instance-type/limit describes.

## Correctness (applies the review-lessons checklist from #369–375)
- **Atomic create** via `memstore.SetIfAbsent` on the domain name → `ResourceAlreadyExistsException` (no Has()+Set() race).
- **Deep-copy reads** — Describe/List copy Tags + nested config/slices; `-race` no-alias test included.
- **Typed exceptions** via `driver.APIError` + `errors.As` first in `writeErr` (`ResourceNotFoundException`, `ResourceAlreadyExistsException`, `ValidationException`, …), asserted in SDK-roundtrip tests.
- **Correct error per reason** + domain-name validation before lookup; pagination tokens honored.

Read-only catalog/limit ops return plausible synthesized results (documented).

## Test plan
- Provider tests incl. `-race` concurrent-create and no-alias checks; SDK-roundtrip (real client vs httptest): create→describe→update→list→delete, tags, typed error paths via `errors.As`.
- Gate: build/vet/`go test ./...`, `go mod tidy` clean, golangci-lint 0 new.

## Risk & rollback


NitinKumar004:

After you rebase on `development`, run `go generate ./...` and commit the updated `docs/coverage/` so your service shows up there.

providers/aws/opensearch/opensearch.go
```diff
@@ -0,0 +1,197 @@
...
+	for i := range in {
+		out[i] = copyRaw(in[i])
+	}
+
+	return out
+}
+
+// paginate returns the offset window and the next token for a slice of length
+// n, honoring an opaque numeric offset token. An invalid token yields offset 0.
+func paginate(n int, page driver.Page) (start, end int, next string) {
```
NitinKumar004: **Medium (error taxonomy).** A corrupt/invalid `NextToken` is silently reset to offset 0 here rather than raising `InvalidPaginationTokenException`, so a paginating client that hands back a bad token gets HTTP 200 with page 1 again (silent re-read) instead of an error it can branch on. More broadly, of the declared exceptions only NotFound/AlreadyExists/Validation are ever emitted — `LimitExceeded`/`Conflict`/`DisabledOperation`/`InvalidPaginationToken` appear only in the status switch with no code path, and `BaseException` is dead. Emit the specific types (limit breaches → `LimitExceeded`, bad token → `InvalidPaginationToken`) so the SDK unmarshals them, and drop any that stay unused.

providers/aws/opensearch/domains.go
```diff
@@ -0,0 +1,399 @@
...
+	out.AdvancedOptions = copyTags(c.AdvancedOptions)
+	out.RawOptions = copyRaw(c.RawOptions)
+
+	return out
+}
+
+// CreateDomain provisions a domain that is immediately Active with an endpoint.
+//
+//nolint:gocritic // hugeParam: signature is fixed by the driver.OpenSearch interface (by-value input).
+func (m *Mock) CreateDomain(_ context.Context, in driver.CreateDomainInput) (*driver.DomainStatus, error) {
```
NitinKumar004, line 48: **Medium (state machine).** `CreateDomain` returns `Processing:false, DomainProcessingStatus:Active` immediately, and `UpdateDomainConfig`/`reflectConfigOnStatus` never set `Processing`/`Modifying` — so a domain is always Active, the declared `ProcessingCreating/Modifying/Upgrading` states are unused, and there's no rejection of update/delete while a domain is Processing (real OpenSearch → `ValidationException`). This also contradicts the package doc and `services.md` ("reports Processing then Active"). Either model a CREATING/Modifying phase (e.g. via the Clock) or correct the doc claims.

providers/aws/opensearch/domains.go
```diff
@@ -0,0 +1,399 @@
...
+	s.ClusterConfig = copyClusterConfig(cfg.ClusterConfig)
+	s.AccessPolicies = cfg.AccessPolicies
+	s.IPAddressType = cfg.IPAddressType
+	s.AdvancedOptions = copyTags(cfg.AdvancedOptions)
+	s.RawOptions = copyRaw(cfg.RawOptions)
+}
+
+// DeleteDomain removes a domain under the write lock and returns its final
+// (Deleted) status. Package associations referencing it are released.
+func (m *Mock) DeleteDomain(_ context.Context, name string) (*driver.DomainStatus, error) {
```
NitinKumar004, line 224: **Medium (dependency/orphan).** `DeleteDomain` releases only package associations; inbound/outbound cross-cluster connections and VPC endpoints keyed by this domain's ARN are left in their stores and keep appearing in `DescribeInbound/OutboundConnections` / `ListVpcEndpoints` after the domain is gone. Real OpenSearch blocks domain deletion while active cross-cluster connections exist. Either block the delete (ConflictException) or cascade-clean the referencing connections/VPC endpoints.

providers/aws/opensearch/tags.go
```diff
@@ -0,0 +1,73 @@
...
+
+	if i := strings.Index(arn, marker); i >= 0 {
+		return arn[i+len(marker):]
+	}
+
+	return arn
+}
+
+// AddTags adds tags to a domain identified by its ARN.
+func (m *Mock) AddTags(_ context.Context, arn string, tags map[string]string) error {
```
NitinKumar004, line 21: **Medium (partial mutation).** `AddTags` writes every tag into `dd.tags` and only *then* checks the 50-tag cap, returning `ValidationException` while leaving the over-limit tags persisted — so a client that gets the error and assumes atomicity finds the tags applied anyway. Validate the resulting count before committing (same pattern applies to `AssociatePackages`/`DissociatePackages`, which persist item-by-item with no rollback on first failure).

NitinKumar004 (commented):
## Review summary — AWS OpenSearch Service

Gates green (build / `go test` / `-race` on both packages). Strong, well-structured PR: **96/96/96** op parity (driver methods = handler routes = docs), correct restJson1 error envelope proven against the real `aws-sdk-go-v2/service/opensearch` client, `Matches` narrowly prefix-gated on `/2021-01-01/` and registered before EC2/S3, atomic `SetIfAbsent` creates across every resource kind, copy-on-write reads (no aliasing races, `-race` clean), and deterministic `Clock`/`idgen` usage. No High-severity issues — the headline domain/tags/packages/versions flow works end-to-end.

The items below are fidelity/robustness gaps worth addressing before the "full parity" banner; none block the core flow, so this is a comment.

### Medium
- **Error taxonomy is largely inert** — only `ResourceNotFound`/`ResourceAlreadyExists`/`Validation` are ever emitted; `LimitExceeded`, `Conflict`, `DisabledOperation`, `InvalidPaginationToken`, and `BaseException` are declared (or switch-mapped) but never produced. Most user-visible: a corrupt `NextToken` silently resets to page 0 (client re-reads page 1) instead of `InvalidPaginationTokenException`, and limit breaches raise `Validation` not `LimitExceeded`. Inline.
- **Domain state machine is a no-op** — `CreateDomain` returns `Active` immediately and `UpdateDomainConfig` never sets `Processing`, contradicting the package/docs claim ("reports Processing then Active") and skipping the real "can't modify mid-Processing → ValidationException" guard. Inline.
- **`DeleteDomain` orphans cross-cluster connections and VPC endpoints** — only package associations are released; inbound/outbound connections and VPC endpoints keyed by the domain ARN survive and keep listing. Real OpenSearch blocks domain deletion while active connections exist. Inline.
- **Partial mutation on rejected batch ops** — `AddTags` writes all tags then checks the 50-tag cap, leaving the over-limit tags persisted despite the error; `AssociatePackages`/`DissociatePackages` persist item-by-item with no rollback on first failure. Inline.

### Low
- Non-persisting write-path stubs beyond the "synthesized" disclaimer (`AuthorizeVpcEndpointAccess`/`RegisterCapability`/`AttachDataSource` return success but don't persist, so the paired list is empty) — worth folding into the docs disclaimer so "every client method implemented" isn't overstated.
- No name-uniqueness for packages/applications (dedup is on a random ID, so duplicate `PackageName` is allowed; real OpenSearch enforces unique names).
- Doc drift in `services.md §34` (the "Processing then Active" line; `StopServiceSoftwareUpdate` vs the driver's `Cancel`/`Rollback`; out-of-order `## 34`→`## 31` heading).
- Note: once this rebases onto `development` (which now generates capability docs from the driver interfaces), run `go generate ./...` and commit the resulting `docs/coverage/aws/opensearch.md`.

providers/aws/opensearch/opensearch.go
thzgajendra: Fixed (`0b73417`). A corrupt/out-of-range `NextToken` now raises `InvalidPaginationTokenException` (in `decodeToken`/`paginate`, propagated through every list op incl. `filterAssociations`/`listVpcEndpoints`); tag-cap breaches now raise `LimitExceededException`. Dead constants `BaseException`/`Conflict`/`DisabledOperation` and their server switch arms removed. Added `TestInvalidPaginationToken`.

providers/aws/opensearch/domains.go
thzgajendra: Fixed by correcting the claim (deterministic option). Removed the unused `ProcessingCreating/Modifying/Upgrading/Isolated/NotApplied` constants and dropped the "reports Processing then Active" line from the package doc and services.md §OpenSearch — domains are Active on create, and the docs now say so (no wall-clock nondeterminism introduced).

providers/aws/opensearch/domains.go
thzgajendra: Fixed. `DeleteDomain` now calls `cascadeDeleteDomainRefs`, which deletes VPC endpoints keyed by the domain ARN and inbound/outbound cross-cluster connections referencing the domain on either side, so they no longer linger in `ListVpcEndpoints`/`Describe*Connections`. `TestDeleteDomainCascades` covers it.

providers/aws/opensearch/tags.go
thzgajendra: Fixed. `AddTags` now computes the merged tag count and rejects with `LimitExceededException` *before* mutating, so over-cap tags aren't persisted. `AssociatePackages`/`DissociatePackages` validate every item first, so a mid-batch failure leaves no partial state. `TestAddTagsOverCap` asserts the original tags survive the rejected call.

thzgajendra:
Thanks for the review — all four Mediums + the Lows fixed in `0b73417`:
- **Error taxonomy:** corrupt NextToken → `InvalidPaginationTokenException`; limit breaches → `LimitExceededException`; removed dead `BaseException`/`Conflict`/`DisabledOperation`.
- **Domain state machine:** corrected the docs (no Processing→Active claim) and removed the unused status constants — deterministic, no wall-clock transitions.
- **DeleteDomain orphans:** cascade-cleans referencing VPC endpoints + inbound/outbound connections.
- **Partial mutation:** `AddTags` validates the merged count before committing; `Associate/DissociatePackages` validate all items first.
- **Lows:** unique package/application names via `SetIfAbsent` name index; non-persisting stubs folded into the synthesized disclaimer; services.md drift fixed (Cancel/Rollback names, `## 34` ordering); dry-run JSON via `json.Marshal`.

New tests: InvalidPaginationToken, AddTagsOverCap, DeleteDomainCascades, DuplicatePackageName, DuplicateApplicationName. Gate: build/vet/`go test -race`/`go test ./...`/tidy clean, golangci-lint 0 new, **CodeQL 0 opensearch findings** (re-ran locally).

NitinKumar004 (approved):

Re-reviewed the fix commits in an isolated worktree; gates green (build / vet / gofmt / golangci-lint 0 / `go test -race -count=2` both packages, plus throwaway concurrent Delete+List/Describe/Update tests — all clean).

- **Error taxonomy — fixed.** A corrupt/out-of-range `NextToken` now raises `InvalidPaginationTokenException`; the tag-cap now emits `LimitExceededException`; the unused `BaseException`/`ConflictException`/`DisabledOperationException` constants and their dead switch arms are removed — every remaining exception is reachable.
- **Domain state machine — fixed.** `DeleteDomain` reports `ProcessingDeleting`, and the doc now correctly states create/update land immediately `Active` (no wall-clock Processing phase) rather than claiming a transition that didn't exist; unused Processing states removed.
- **Cascade delete — fixed.** `cascadeDeleteDomainRefs` removes VPC endpoints (by `DomainARN`) and inbound/outbound cross-cluster connections referencing the deleted domain; state stays consistent, no data race (copy-on-write stores, `-race` clean).
- **Atomic mutations — fixed.** `AddTags` checks the merged tag count before any write; `AssociatePackages`/`DissociatePackages` validate every item before mutating.
- **CodeQL unsafe-quoting — fixed** correctly via `json.Marshal` (escapes embedded quotes).

Only Low fidelity edges remain (non-blocking): cascade matches connections by name only (a narrow over-delete for a coincidentally-named foreign remote domain — real OS keys by ARN); real OpenSearch *blocks* delete while active connections exist rather than cascading (defensible emulator choice — worth a doc note); `LimitExceededException` maps to 429 vs AWS's 409 (cosmetic; the SDK keys off `X-Amzn-Errortype`). Solid, thorough turnaround. LGTM.

PR #381, opened 2026-08-09
Add AWS CloudTrail Emulation with Full Parity (#295 Workstream B)

thzgajendra:
## Objective
Extend cloudemu's AWS coverage (#295, Workstream B) with **AWS CloudTrail** at full SDK-wire parity — real `aws-sdk-go-v2/service/cloudtrail` clients work by only changing the endpoint.

## What we built
CloudTrail across the four layers — `services/cloudtrail/driver`, `providers/aws/cloudtrail` (in-memory `Mock`), `server/aws/cloudtrail` (AWS JSON 1.1, `X-Amz-Target: CloudTrail_20131101.`), registered in both `aws.go` files. **60/60 SDK client methods** implemented.
- **Trails** — create/get/update/delete/describe/list, GetTrailStatus, StartLogging/StopLogging (reflected in status with real Clock times).
- **Event data stores** (CRUD + restore + termination-protection soft-delete), **channels**, **dashboards**, **event/insight selectors**, **imports**, **resource policy**, **event config**, **federation**, **org delegated-admin**, **tags**.

## Correctness (applies the review-lessons checklist from #369–375)
- **Atomic create** via `SetIfAbsent` for trails/EDS/channels/dashboards (`-race` concurrent-create tests assert one winner).
- **Deep-copy reads** — no-alias `-race` test mutates returned Tags + nested selectors.
- **Typed exceptions** via `driver.APIError` + `errors.As` first: bad name→`InvalidTrailNameException`, missing→`TrailNotFoundException`, dup→`TrailAlreadyExistsException`, bad EDS ARN→`EventDataStoreARNInvalidException` vs absent→`EventDataStoreNotFoundException`.
- Name/ARN format validated before lookup; pagination honored.

**Synthesized (documented, no real event stream):** LookupEvents/ListPublicKeys/insights return empty pages; StartQuery completes instantly with an empty result store; StartImport/StartDashboardRefresh complete synchronously.

## Test plan
- Provider tests (incl. `-race` concurrent-create + no-alias); SDK-roundtrip (real client vs httptest): trail lifecycle, duplicate/invalid-name/not-found via `errors.As`, EDS CRUD, tags, event selectors, empty LookupEvents.
- Gate: build/vet/`go test ./...`, `go mod tidy` clean, golangci-lint 0 new.

## Risk & rollback


providers/aws/cloudtrail/imports.go
```diff
@@ -0,0 +1,127 @@
...
+	imp, ok := m.imports.Get(id)
+	if !ok {
+		return nil, errImportNotFound(id)
+	}
+
+	return cloneImport(imp), nil
+}
+
+// StopImport marks an import STOPPED.
+func (m *Mock) StopImport(_ context.Context, id string) (*driver.Import, error) {
```
NitinKumar004, line 67: **High (data race).** `imports` is a `Store[*driver.Import]` with no per-item mutex (unlike `trailData`/`edsData`/`channelData`/`dashboardData`, which each embed `sync.RWMutex`). `StopImport` does `imp, _ := m.imports.Get(id)` then mutates `imp.Status`/`imp.UpdatedAt` on the shared pointer **unlocked**, racing `GetImport`/`ListImports` (and the `StartImport` resume read). Reproduced under `-race`. Wrap `Import` in a per-item-lock struct like the other four resources (or route the mutation through `memstore.Update`), and add a concurrent StopImport + GetImport `-race` test.

providers/aws/cloudtrail/queries.go
```diff
@@ -0,0 +1,141 @@
...
+	q, ok := m.queries.Get(queryID)
+	if !ok {
+		return nil, errQueryNotFound(queryID)
+	}
+
+	return &driver.QueryResults{QueryStatus: q.Status, ResultRows: []map[string]string{}}, nil
+}
+
+// CancelQuery stops a running query and returns its resulting status.
+func (m *Mock) CancelQuery(_ context.Context, _, queryID string) (string, error) {
```
NitinKumar004, line 82: **High (data race).** Same class as `StopImport`: `queries` is a `Store[*driver.Query]` with no per-item lock, and `CancelQuery` writes `q.Status` on the shared pointer unlocked, racing `DescribeQuery` (`out := *q`) and `ListQueries`. Reproduced under `-race`. Give `Query` the same per-item mutex treatment as the other resources.

providers/aws/cloudtrail/eventdatastores.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	ed.eds.Status = driver.EDSStatusPendingDeletion
+	ed.eds.UpdatedAt = m.now()
+
+	return nil
+}
+
+// RestoreEventDataStore returns a PENDING_DELETION store to ENABLED.
+func (m *Mock) RestoreEventDataStore(_ context.Context, arn string) (*driver.EventDataStore, error) {
```
NitinKumar004, line 202: **Medium.** `RestoreEventDataStore` sets `Status = ENABLED` unconditionally — restoring a never-deleted (already ENABLED) store succeeds. Real CloudTrail only restores a `PENDING_DELETION` store within the 7-day window and otherwise raises `InvalidEventDataStoreStatusException`; the `ExInactiveEventDataStore` constant is already declared for exactly this. Guard on the current status. (Related: `DeleteEventDataStore` never frees `edsNameIdx`, so a same-name recreate is blocked forever.)

NitinKumar004 (changes requested):
## Deep review — AWS CloudTrail

Reviewed in an isolated worktree at the PR head via a 2-agent fan-out (structure/wire/domain + concurrency/state/errors); gate green (build / vet / `go test -race` both packages / `go mod tidy` clean / gofmt). The PR explicitly claims the #369–375 lessons, so I verified each.

**This is strong, and it shows the lessons landed** — atomic `SetIfAbsent` create for trails/EDS/channels/dashboards (one-winner tests), deep-copied selectors (verbatim round-trip, no aliasing), per-condition typed exceptions via `driver.APIError` + `errors.As`-first (proven end-to-end by real `aws-sdk-go-v2/service/cloudtrail` roundtrip tests for `TrailAlreadyExists`/`InvalidTrailName`/`TrailNotFound`/`EventDataStoreNotFound`), ARN-shape-before-lookup splitting `EventDataStoreARNInvalid` vs `NotFound`, logging lifecycle via the real `Clock`, disjoint `Matches`, honest synthesized surfaces (documented in services.md §34), 60-op parity, clean determinism.

Two things to fix.

### High — two data races: `imports`/`queries` lack the per-item lock the other resources have
`trailData`/`edsData`/`channelData`/`dashboardData` each embed a `sync.RWMutex`, but `imports`/`queries` are `memstore.Store[*driver.Import]`/`[*driver.Query]` (bare pointers), and the Stop/Cancel ops mutate the shared pointee **unlocked**:
- `StopImport` (imports.go:52) — `imp.Status = …; imp.UpdatedAt = …`
- `CancelQuery` (queries.go:68) — `q.Status = …`

Both race a concurrent reader (`GetImport`/`ListImports`; `DescribeQuery` does `out := *q`, `ListQueries`) — reproduced under `-race`. The as-shipped suite passes only because no test exercises concurrent Stop/Cancel + read. This deviates from the PR's own pattern for the other four resources and from CLAUDE.md's "all mock implementations use sync.RWMutex." Fix: wrap `Import`/`Query` in a per-item-lock struct (or route the mutation through `memstore.Update`), and add a concurrent Stop/Cancel + Describe `-race` test. Inline on both.

### Medium — `RestoreEventDataStore` has no state guard
`RestoreEventDataStore` (eventdatastores.go:168) flips *any* status to ENABLED. Real CloudTrail restores only a `PENDING_DELETION` store (within the 7-day window) and otherwise returns `InvalidEventDataStoreStatusException` — the `ExInactiveEventDataStore` constant is already declared but unused. Guard on the current status. Inline.

### Low (quick follow-ups)
- Trail name isn't format-validated on Get/Update/Delete/GetTrailStatus (`resolveTrail` goes straight to lookup → `TrailNotFoundException` instead of `InvalidTrailNameException`; `CreateTrail` validates correctly).
- `DeleteEventDataStore` soft-deletes but never frees `edsNameIdx`, so recreating an EDS with the same name is blocked forever (no purge); pairs with the Restore guard above as the soft-delete-lifecycle cluster.
- `TagResource`/`RemoveTags` on a nonexistent resource silently succeed (real CT → `ResourceNotFoundException`).
- `StartQuery` doesn't validate the referenced EDS, and a cluster of exception constants (`ExInvalidQueryStatement`, `ExInvalidParameterCombo`, `ExResourceARNNotValid`, …) are declared but never emitted (dead taxonomy vs the no-dead-code rule).
- `writeErr` maps `IsAlreadyExists → "ResourceNotFoundException"` (handler.go copy-paste; currently masked because creates return tagged `APIError`s, but a latent bug).
- `channel.Tags`/`dashboard.Tags` aliased on read (benign today since tag ops use the separate `m.tags` map); STRUCTURE §3 filename drift (`eds_ops.go` abbreviation + plural provider / singular `_ops` wire split).

Requesting changes on the two races + the Restore state guard; the Lows are quick. No AI attribution.

NitinKumar004:

After you rebase on `development`, run `go generate ./...` and commit the updated `docs/coverage/` so your service shows up there.

providers/aws/cloudtrail/imports.go
thzgajendra: Fixed. `Import` is now wrapped in an `importData{imp, mu sync.RWMutex}` struct (like the other four resources). `StopImport` mutates under the write lock; `GetImport`/`ListImports`/`StartImport`-resume copy under the read lock. Added `TestConcurrentStopImportAndRead` (50× Stop+Get) — clean under `-race`.

providers/aws/cloudtrail/queries.go
thzgajendra: Fixed. Same treatment: `Query` wrapped in `queryData{q, mu sync.RWMutex}`; `CancelQuery` writes under the write lock, `DescribeQuery`/`GetQueryResults`/`ListQueries` read under the read lock. Added `TestConcurrentCancelQueryAndRead` — clean under `-race`.

providers/aws/cloudtrail/eventdatastores.go
thzgajendra: Fixed. `RestoreEventDataStore` now returns `InactiveEventDataStoreException` (via the previously-unused `ExInactiveEventDataStore`) unless the store is `PENDING_DELETION`. Paired the soft-delete lifecycle: `DeleteEventDataStore` frees `edsNameIdx` so a same-name store can be recreated during the window, and `RestoreEventDataStore` re-claims the name (conflict if taken). Added `TestRestoreRequiresPendingDeletion`.

thzgajendra:
Thanks for the deep review. Fixed in `bba661c`:

**High — two data races:** `imports`/`queries` now use the same per-item-lock wrapper (`importData`/`queryData`) as the other four resources; Stop/Cancel mutate under the write lock, all reads copy under the read lock. Added concurrent `-race` tests (StopImport+GetImport, CancelQuery+DescribeQuery).

**Medium — Restore state guard:** `RestoreEventDataStore` now only restores a `PENDING_DELETION` store (`InactiveEventDataStoreException` otherwise), using the previously-dead constant. Also freed `edsNameIdx` on delete + re-claim on restore so same-name recreate works (the soft-delete-lifecycle cluster).

**Low:** fixed the `writeErr` `AlreadyExists → ResourceNotFoundException` copy-paste (now `ResourceAlreadyExistsException`).

Remaining Lows (trail-name format-validation on Get/Update/Delete, tag-ops-on-nonexistent → ResourceNotFound, StartQuery EDS validation + emitting the remaining declared exception constants, filename drift) I can fold into a quick follow-up — flag if you'd like them in this PR instead.

Gate: build/vet/`go test -race` both packages / `go mod tidy` clean / golangci-lint 0 new / CodeQL 0 findings.

providers/aws/cloudtrail/eventdatastores.go
```diff
@@ -0,0 +1,309 @@
...
+
+	e.UpdatedAt = m.now()
+	out := copyEDS(e)
+
+	return &out, nil
+}
+
+// DeleteEventDataStore marks an event data store PENDING_DELETION (soft delete,
+// matching CloudTrail's 7-day wait) unless termination protection is on.
+func (m *Mock) DeleteEventDataStore(_ context.Context, arn string) error {
```
NitinKumar004, line 161: **Medium (regression).** `DeleteEventDataStore` sets `PENDING_DELETION` then unconditionally `m.edsNameIdx.Delete(ed.eds.Name)` — no already-pending idempotency guard and no owner check. Sequence: create "foo"→A; delete A (frees "foo"); create "foo"→B (live); delete A *again* (A is still in `m.eds`, still PENDING_DELETION, unguarded) → this deletes **B's** name claim; create "foo"→C succeeds → two live ENABLED stores named "foo". Guard: return early if already `PENDING_DELETION`, and only free the name when `edsNameIdx[name]` still maps to *this* ARN. (`UpdateEventDataStore` rename at :104 has the same unconditional `Delete(e.Name)` shape.)

providers/aws/cloudtrail/eventdatastores.go
```diff
@@ -0,0 +1,309 @@
...
+	// window; RestoreEventDataStore re-claims it.
+	m.edsNameIdx.Delete(ed.eds.Name)
+
+	return nil
+}
+
+// RestoreEventDataStore returns a PENDING_DELETION store to ENABLED. Real
+// CloudTrail only restores a store that is pending deletion; any other status is
+// an InactiveEventDataStoreException.
+func (m *Mock) RestoreEventDataStore(_ context.Context, arn string) (*driver.EventDataStore, error) {
```
NitinKumar004, line 202: **Medium (wrong exception type).** The new status guard is right to reject a non-`PENDING_DELETION` restore, but it returns `errEDSInactive` → `InactiveEventDataStoreException`. Real CloudTrail's `RestoreEventDataStore` throws `InvalidEventDataStoreStatusException` for this case (`InactiveEventDataStoreException` belongs to the query/ingestion-on-stopped path). Since the wire tags `X-Amzn-Errortype` from the exception name, a client doing `errors.As(err, &types.InvalidEventDataStoreStatusException{})` never matches. Add/emit the correct type (and assert it in `TestRestoreRequiresPendingDeletion`, which currently checks only `require.Error`).

providers/aws/cloudtrail/eventdatastores.go
```diff
@@ -0,0 +1,309 @@
...
+
+	out := copyEDS(&ed.eds)
+
+	return &out, nil
+}
+
+// UpdateEventDataStore applies the non-nil fields of in.
+//
+//nolint:gocritic,gocyclo // in matches the driver API (by value); one branch per optional field.
+func (m *Mock) UpdateEventDataStore(
```
NitinKumar004, line 94: **Medium (state guards).** `UpdateEventDataStore` (and `StartQuery`, ingestion start/stop, `EnableFederation`/`DisableFederation`) operate on the store regardless of status — only `RestoreEventDataStore` checks. Real CloudTrail rejects update/query/ingestion/federation against a `PENDING_DELETION` or stopped store with `InvalidEventDataStoreStatusException`/`InactiveEventDataStoreException`. Add the self-referential status guard on these ARN-in-hand operations (StartQuery's EDS-from-SQL parsing is genuinely harder and can be left).

providers/aws/cloudtrail/tagging.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (m *Mock) deleteResourceTags(resourceID string) {
+	m.tagsMu.Lock()
+	defer m.tagsMu.Unlock()
+
+	delete(m.tags, resourceID)
+}
+
+// AddTags adds or overwrites tags on a resource (trail/EDS/channel/dashboard).
+func (m *Mock) AddTags(_ context.Context, resourceID string, tags map[string]string) error {
```
NitinKumar004, line 71: **Medium (state guard).** `AddTags`/`RemoveTags` validate only `resourceID != ""` and `AddTags` auto-creates a tag map for any arbitrary ARN, so tagging a nonexistent trail/EDS returns success and a later `ListTags` reports phantom tags. Real CloudTrail returns `ResourceNotFoundException`/`CloudTrailARNInvalidException`. Resolve the ARN against the resource stores first.

NitinKumar004 (changes requested):
## Review summary — AWS CloudTrail (re-review after the fix commit)

The two data races are **fixed and verified** — `imports`/`queries` now wrap their value in a per-item `sync.RWMutex` (matching `trailData`/`edsData`); `StopImport`/`CancelQuery` mutate under `Lock` and every read path copies under `RLock`, and a concurrent `-race` run over Stop/Cancel + Get/List/Describe is clean. Nice. Much is genuinely faithful: ARN-shape-vs-existence distinction, the EDS soft-delete lifecycle, termination-protection guard, thorough trail-name validation, atomic `SetIfAbsent` name claims, selector mutual-exclusion, deep-copy reads, accurate wire field names, and honestly-disclosed synthesized read surfaces.

Holding on one regression the name-index cleanup introduced, plus a cluster of state-guard/fidelity Mediums.

### Medium
- **Regression — `DeleteEventDataStore` name-index double-free.** Now that delete frees `edsNameIdx`, doing so unconditionally (no already-`PENDING_DELETION` guard, no owner check) lets delete→recreate-same-name→re-delete strip the *new* store's name claim, producing two live ENABLED stores sharing a name — defeating the `SetIfAbsent` uniqueness this relies on. Inline.
- **Inactive-EDS state guards missing.** Only `RestoreEventDataStore` checks status; `UpdateEventDataStore`/`StartQuery`/ingestion/federation operate on a `PENDING_DELETION`/stopped store, which real CloudTrail rejects (`InvalidEventDataStoreStatusException`/`InactiveEventDataStoreException`). Inline.
- **`RestoreEventDataStore` returns the wrong exception type.** It emits `InactiveEventDataStoreException`; real CloudTrail uses `InvalidEventDataStoreStatusException` for restore-on-non-pending, so a client's `errors.As(&InvalidEventDataStoreStatusException{})` never matches. Inline.
- **Tagging a nonexistent resource silently succeeds** — `AddTags` auto-creates a tag map for any ARN (no existence check), so a later `ListTags` returns phantom tags; real CloudTrail → `ResourceNotFoundException`. Inline.

### Low
- `DeleteDashboard` ignores `TerminationProtectionEnabled` (EDS delete honors its equivalent — inconsistent).
- `CreateEventDataStore` drops the `StartIngestion=false` flag (always ENABLED).
- Dead query-exception constants (`ExInvalidQueryStatement`, `ExInvalidParameterCombo`, `ExResourceARNNotValid`, `ExInsufficientS3BucketPolcy`).
- `Dashboard` Get/List/Create/Update alias the stored `Tags` map (every other resource deep-copies).
- "Full parity" wording overstates the write path given the state-guard/validation gaps.
- Once rebased onto `development`, run `go generate ./...` and commit the generated `docs/coverage/aws/cloudtrail.md`.

providers/aws/cloudtrail/eventdatastores.go
thzgajendra: Fixed (`7b95247`). `DeleteEventDataStore` now no-ops when the store is already `PENDING_DELETION`, and frees the name only via an owner-checked `freeEDSName` (frees iff `edsNameIdx[name]` still maps to *this* ARN). The rename path in `UpdateEventDataStore` uses the same owner-checked free. Added `TestDeleteRecreateReDeleteNameIndex` asserting the create→delete→recreate→re-delete sequence leaves exactly one live store owning the name.

providers/aws/cloudtrail/eventdatastores.go
thzgajendra: Fixed. Added `ExInvalidEDSStatus = "InvalidEventDataStoreStatusException"` and `errEDSInvalidStatus`; `RestoreEventDataStore` now emits it (and so do the new state guards). Removed the now-unused `ExInactiveEventDataStore`. `TestSDKRestoreInvalidStatus` asserts via `errors.As(&types.InvalidEventDataStoreStatusException{})` and the provider test now checks the exception type, not just `require.Error`.

providers/aws/cloudtrail/eventdatastores.go
thzgajendra: Fixed. `UpdateEventDataStore`, ingestion start/stop (`setEDSStatus`), and `EnableFederation`/`DisableFederation` now reject a non-ENABLED (e.g. `PENDING_DELETION`) store with `InvalidEventDataStoreStatusException`. Left `StartQuery`'s EDS-from-SQL parsing alone as you noted. `TestInactiveEDSGuards` covers these.

providers/aws/cloudtrail/tagging.go
thzgajendra: Fixed. `AddTags`/`RemoveTags` now resolve the ARN against the trail/EDS/channel/dashboard stores first via `resourceARNExists` → `ResourceNotFoundException` when absent, and `CloudTrailARNInvalidException` (new `ExCloudTrailARNInvalid`) for a malformed ARN. `TestTagNonexistentResource` added.

thzgajendra:
Thanks for the re-review — all four Mediums + the Lows fixed in `7b95247`:
- **Regression (name-index double-free):** delete is idempotent on `PENDING_DELETION` and frees the name only when it still owns it (owner-checked); same for the rename path.
- **Inactive-EDS state guards:** Update/ingestion/federation reject non-ENABLED stores with `InvalidEventDataStoreStatusException`.
- **Restore exception type:** now `InvalidEventDataStoreStatusException` (was Inactive…); asserted via `errors.As`.
- **Tag-on-nonexistent:** resolves the ARN first → `ResourceNotFoundException`/`CloudTrailARNInvalidException`.
- **Lows:** DeleteDashboard honors termination protection; CreateEventDataStore honors `StartIngestion=false`; dead query-exception constants removed; Dashboard reads deep-copy Tags; softened the parity wording.

New tests: name-index attack, inactive-EDS guards, StartIngestion=false, tag-nonexistent, restore-invalid-status (typed). Gate: build/vet/`go test -race`/`go test ./...`/tidy clean, golangci-lint 0 new, **CodeQL 0 cloudtrail findings** (re-ran locally).

NitinKumar004 (approved):

Re-reviewed the fix commit in an isolated worktree; gates green (build / `go test -race -count=2` both packages; adversarial name-index + restore-ordering probes clean).

- **Delete name-index double-free (regression) — fixed.** An owner-checked `freeEDSName` deletes `name→arn` only when it still maps to *this* ARN, used on both the delete and rename paths (the old unconditional `Delete` is gone), plus an idempotency guard for an already-`PENDING_DELETION` store. Verified across delete→recreate→re-delete / restore-after-recreate orderings — exactly one live ENABLED store per name in every case, with a test locking it in.
- **RestoreEventDataStore exception — fixed.** Now emits `InvalidEventDataStoreStatusException` (verified the emitted type, asserted by a test).
- **Inactive-EDS state guards — fixed.** `UpdateEventDataStore`, ingestion start/stop, and federation enable/disable now reject a pending/inactive store with the correct exception. (`StartQuery` stays unguarded, consistent with the documented empty-query emulation — a by-design Low.)
- **Tagging nonexistent/malformed resource — fixed.** `AddTags` and `RemoveTags` validate ARN format (`CloudTrailARNInvalidException`) + existence (`ResourceNotFoundException`) across all four taggable types (trail/EDS/channel/dashboard).
- Lows fixed too: `DeleteDashboard` termination protection, `StartIngestion=false` → `STOPPED_INGESTION`, dead query-exception constants removed, `Dashboard` Tags now deep-copied.

No new races/deadlocks/aliasing (imports/queries per-item mutex intact, lock ordering uniform). The two original HIGH races remain fixed. LGTM.

PR #384, opened 2026-08-09
fix(coveragegen): networking resolves to full Networking driver, not NetworkInterfaces capability

### LINKED ISSUE #383: coveragegen: networking resolves to the NetworkInterfaces capability, not the Networking driver
Found while running `go generate ./...` for the OCI services (#377/#378/#379). This predates those branches — it reproduces on `development` at `ef9d73a`.

## What the docs say

`docs/coverage/README.md`:

```
| `networking` | [VPC](./aws/vpc.md) | — | — | — | 3 |
```

`docs/coverage/aws/vpc.md`:

```
AWS's `networking` service · portable interface `driver.NetworkInterfaces`

## Operations (3)
| `DeleteNetworkInterface` |
| `DescribeNetworkInterfaces` |
| `DetachNetworkInterface` |
```

## Two problems

1. **Wrong interface.** `networking` resolves to `driver.NetworkInterfaces` — the 3-method *optional capability* — instead of `driver.Networking`, which carries the ~51-operation surface (VPCs, subnets, security groups, route tables, gateways, ACLs, peering, ENIs, endpoints…). The page understates AWS networking by an order of magnitude.
2. **Azure and GCP are credited with nothing**, though `providers/azure/vnet` and `providers/gcp/vpc` both implement the same driver.

## Likely cause

`resolveField` falls through to `implementedService(mockDir, byName)` for a field typed as a provider mock. A networking mock satisfies **both** `driver.Networking` and the optional `driver.NetworkInterfaces`, so a method-set match can pick either; it appears to be selecting the narrower one. Azure/GCP presumably fail to match at all for a related reason.

The OCI branches are unaffected in mechanism — `Provider.VCN` is typed as the driver interface directly, so it resolves by import path — but they inherit the same wrong operation count and interface name on their page.

## Suggested fix

When a mock's method set satisfies several candidate interfaces for one service, prefer the service's primary driver interface over its optional capabilities. Worth a regression test asserting `networking` reports the full `driver.Networking` surface for every provider that implements it.

## Impact

`docs/coverage` is generated specifically so the docs cannot promise more than the code delivers. Here it promises considerably *less*, which is the same accuracy problem pointing the other way.

NitinKumar004:
Fixes #383.

## The bug

The coverage generator resolved the `networking` service to `driver.NetworkInterfaces` — a 3-method **optional** capability — instead of `driver.Networking`, the full ~57-operation surface. So the docs understated AWS networking by an order of magnitude, and Azure/GCP (which implement `driver.Networking` but not the AWS-only `NetworkInterfaces` capability) showed nothing at all.

## Why

`referencedInterface` picked a service's primary interface by counting **every** `driver.<Name>` mention in the portable package. The networking wrappers probe the optional capability with type assertions — `n.driver.(driver.NetworkInterfaces)` — three times, out-counting the stored `driver driver.Networking` field (referenced twice). The narrower optional interface won.

## The fix

Count a name only where it appears as a **field or parameter type** (`ast.Field.Type`). Those are the positions that mean "this package stores/consumes this interface"; type-assertion expressions aren't fields, so they no longer skew the choice. One targeted change in `internal/coveragegen`.

## Result

`networking` now resolves to `driver.Networking`:

- **57 operations** (was 3), across **VPC / VNet / VPC** — Azure and GCP are credited again.
- `NetworkInterfaces`, `NetworkInterfaceCreator`, and `VPCAttributes` now correctly appear under **Optional capabilities**.
- No other service changed — verified the regenerated diff touches only networking pages.

Added a regression test asserting `networking` reports the primary `Networking` interface with its full surface, not the optional capability.

## Verification

`go build ./...`, generator tests, lint (0 issues) all pass; `go generate ./...` is deterministic.


thzgajendra (commented):
## Review — coveragegen: networking resolves to full `Networking` driver (from cloudemu's perspective)

Reviewed the fix in depth and verified the generated output end-to-end. **Clean, correct, well-tested — no blockers.**

### The fix is right
`referencedInterface` previously used `countSelectors`, which counted **every** `pkg.Name` selector in the portable package — including the three type-assertion probes `n.driver.(driver.NetworkInterfaces)` — so the 3-method optional capability out-counted the stored `driver driver.Networking` field. Replacing it with `countFieldTypes` (only `ast.Field.Type` positions — struct fields, params, results, unwrapping a leading pointer via the new `selectorType`) is exactly the right discriminator: a type assertion is an expression, not a field, so it no longer skews the pick. `networking` now resolves to `driver.Networking`, and `NetworkInterfaces`/`NetworkInterfaceCreator`/`VPCAttributes` drop to Optional capabilities where they belong.

### Verified locally (this head)
- `go build ./...` clean; `go test ./internal/coveragegen/...` passes.
- **`go generate ./...` produces ZERO diff against the committed docs** — the checked-in coverage pages are exactly what the generator emits (no hand-edit drift), which is the thing most worth confirming for a generator fix.
- **Deterministic:** ran `go generate` twice, no diff either time.
- **Scope confirmed:** the `coverage.json` diff touches only `"interface": "Networking"` entries, and only `aws/vpc.md`, the new `azure/vnet.md` + `gcp/vpc.md`, the READMEs, and `coverage.json` changed — Azure and GCP networking are credited again (were previously blank), no other service moved. Matches the PR claim.
- The regression tests are strong: `TestPrimaryInterfaceNotOptionalCapability` pins networking→`Networking` with `>=10` ops, plus good adjacent guards (embed-flatten for SageMaker/AzureAI, monitoring-not-a-consumer, the OCI-credits-nothing honesty check, and providers-have-services).

### Nits (non-blocking)
1. **Tie-break is map-iteration-order dependent** (`services.go` ~L251-254): the winner is chosen by `if n > bestN` while ranging over the `known` map. If two interfaces are ever referenced as a field/param the *same* number of times, the pick is nondeterministic and `go generate` would flap for that service. Not triggered today (networking wins outright, and this shape pre-dates the PR), but since the PR rightly emphasizes determinism, a deterministic tie-break — prefer the larger interface by method count, else lexicographic — would harden it against the next multi-interface service.
2. **`selectorType` unwraps only a single pointer** — an interface used solely as a slice/map/variadic element type (`[]driver.X`, `...driver.X`) wouldn't be counted. Irrelevant for a stored driver field (always `driver.X` or `*driver.X`), so purely theoretical; noting for completeness.

**Verdict: ready.** Correct root-cause fix, output verified in sync and deterministic, scope contained to networking, and pinned by a targeted regression test. The tie-break hardening is a nice-to-have, not a gate.

thzgajendra (commented):
## Re-review — rebased head `3b631df` · verdict: **still ready**

The PR was rebased onto current `development` (which added the OCI provider + glue/opensearch/etc.) and cleaned into two commits. I re-verified the whole thing on the new base — no regressions, and the core fix is intact.

### Re-verified on the rebased head
- **Fix intact:** `referencedInterface` still counts field/param/result types via `countFieldTypes` + `selectorType` (not raw selectors), so networking resolves to `driver.Networking`, not the `NetworkInterfaces` capability.
- **Build + generator tests pass.**
- **`go generate ./...` produces ZERO diff against the committed docs** — crucially, still in sync against the *larger* post-rebase service set (OCI/glue/opensearch included), so the rebase didn't leave the coverage docs stale. **Deterministic** across two runs.
- **Scope still contained to networking:** the `coverage.json` interface change is exactly `NetworkInterfaces` → `Networking`; only `aws/vpc.md`, `azure/vnet.md`, `gcp/vpc.md`, `oci/vcn.md`, the READMEs, and `coverage.json` changed. Nice side effect of the rebase: **OCI's VCN is now credited too** (it also implements `driver.Networking`), and the regenerated OCI page is in sync.
- Regression test (`TestPrimaryInterfaceNotOptionalCapability` + the provider-native-name and OCI-honesty guards) present and passing.

### Nits (unchanged, non-blocking, pre-existing)
1. `referencedInterface` tie-break is still map-iteration-order dependent (`n > bestN` over a Go map) — deterministic today only because networking wins outright; a future two-interface tie would flap `go generate`. A deterministic tie-break (larger interface, else lexicographic) would harden it.
2. `selectorType` unwraps only a single pointer — an interface used solely as a slice/map/variadic element type wouldn't be counted (irrelevant for a stored driver field).

**Verdict unchanged: ready.** Correct root-cause fix, output verified in sync + deterministic against the rebased service set, scope contained to networking (now also crediting OCI), and pinned by the regression test. The two nits are hardening nice-to-haves, not gates. Nothing outstanding from my side.

thzgajendra (approved):

Issue #36, opened 2026-03-19
Add Azure Container Registry (ACR) service


NitinKumar004:
### Operations
- `CreateRepository(name)`
- `DeleteRepository(name)`
- `ListRepositories()`
- `PutImage(repo, tag, manifest)` — push image
- `GetImage(repo, tag)` — pull image metadata
- `DeleteImage(repo, tag)`
- `ListImages(repo)` — list tags

### Why
ACR is the standard container registry for Azure. Testing image lifecycle management in-memory enables faster container-based development.

PR #386, opened 2026-08-10
Add AWS Config Emulation (CRUD/lifecycle parity; synthesized read surfaces) (#295 Workstream B)

thzgajendra:
## Objective
Extend cloudemu's AWS coverage (#295, Workstream B) with **AWS Config** (`configservice`) at full SDK-wire parity — real `aws-sdk-go-v2/service/configservice` clients and the `aws configservice` CLI work by only changing the endpoint.

## What we built
Config across the four layers — `services/configservice/driver`, `providers/aws/configservice` (in-memory `Mock`), `server/aws/configservice` (AWS JSON 1.1, `X-Amz-Target: StarlingDoveService.`), registered in both `aws.go` files. **102/102 SDK client methods** (all except `Options`).
- **Recorders** (single-recorder invariant, start/stop/status, resource-type association), **delivery channels** (single-channel invariant), **config rules** (+compliance roll-up from `PutEvaluations`), **conformance packs**, **org rules/packs**, **aggregators + authorizations**, **remediation configs/exceptions**, **stored queries**, **retention**, **tags**.
- Handles the real AWS mixed-JSON-casing quirk (lowerCamel `configurationRecorder`/`deliveryChannel`/`configurationItem` vs PascalCase `ConfigRule`/`ConformancePack`), verified by SDK roundtrip tests.

**Synthesized (documented):** compliance/evaluation summaries, discovered-resource + aggregate queries, `SelectResourceConfig`, resource evaluations — the emulator runs no real recording pipeline, so these answer from `PutResourceConfig`-supplied state or plausible empties.

## Correctness (review-lessons checklist from #369–381)
- Atomic `SetIfAbsent` creates; deep-copy reads (`-race` no-alias test); typed exceptions via `driver.APIError` + `errors.As`-first (`NoSuch*`, `InvalidParameterValue`, `Validation`, `MaxNumberOf...Exceeded`, `InvalidNextToken`); correct-error-per-reason; validate-before-mutate on batch remediation; write-lock across delete; pagination via NextToken/Limit; no dead constants; `json.Marshal` (no unsafe-quoting).

## Docs / coverage
- `docs/services.md` (row #36 + detail section + Grand Total 2202→2304) and `docs/architecture.md` blurb.
- `go generate ./...` regenerated `docs/coverage/aws/config.md` (102 ops) + `docs/coverage/coverage.json`.

## Test plan
- Provider tests incl. `-race` concurrent-create + no-alias; SDK-roundtrip (real client vs httptest): recorder/channel/rule lifecycle, tags, typed error paths via `errors.As`.
- **E2E (aws configservice CLI vs a running server):** put/describe recorder, put delivery channel, start recorder + status `recording=true`, put/describe/delete config rule, single-recorder invariant → `MaxNumberOfConfigurationRecordersExceededException`, missing rule → `NoSuchConfigRuleException`.
- Gate: build/vet/`go test -race`/`go test ./...`/`go mod tidy` clean, golangci-lint 0 new, **CodeQL 0 configservice findings** (re-run locally).

## Risk & rollback


providers/aws/configservice/recorders.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	rec.Arn = m.arn("config-recorder/" + rec.Name)
+	rec.Tags = copyTags(rec.Tags)
+	rec.LastStatus = driver.RecorderStatusPending
+	rec.LastStatusChangeTime = now
+	rec.Recording = false
+
+	// SetIfAbsent guards against a concurrent create racing the same name.
+	if !m.recorders.SetIfAbsent(rec.Name, &recorderData{rec: rec}) {
```
NitinKumar004: High — singleton race. The one-recorder-per-account cap is a `Keys()` scan (line 29) but the insert here is a separate `SetIfAbsent(rec.Name, …)`, which only dedups the *same* name. Two concurrent creates with different names both pass the scan and both insert. Reproduced: 188/200 stress iterations ended with 2–4 recorders. Guard the scan+insert under a single Mock-level create-mutex (must also cover `PutServiceLinkedConfigurationRecorder`, which shares this store).

providers/aws/configservice/channels.go
```diff
@@ -0,0 +1,147 @@
...
+		existing.ch.LastStatusChangeTime = now
+		existing.mu.Unlock()
+
+		return nil
+	}
+
+	ch.LastStatus = "SUCCESS"
+	ch.LastStatusChangeTime = now
+
+	if !m.channels.SetIfAbsent(ch.Name, &channelData{ch: ch}) {
```
NitinKumar004: High — singleton race (same shape as the recorder cap). `Keys()` scan at line 24 + separate `SetIfAbsent` here; concurrent distinct-name puts both survive → multiple delivery channels (44/200 stress iterations). Needs the same create-mutex spanning scan+insert.

providers/aws/configservice/recorders.go
```diff
@@ -0,0 +1,350 @@
...
+func (m *Mock) PutServiceLinkedConfigurationRecorder(
+	_ context.Context, principal string, tags map[string]string,
+) (arn, name string, err error) {
+	if principal == "" {
+		return "", "", invalidParameter("ServicePrincipal is required")
+	}
+
+	name = "AWSConfigurationRecorderFor" + serviceShortName(principal)
+
+	if m.recorders.Len() > 0 && !m.recorders.Has(name) {
```
NitinKumar004, line 303: High — third singleton path. `Len() > 0 && !Has(name)` then `SetIfAbsent`: two goroutines with different service principals both observe `Len()==0` and both insert (151/200 iterations). Because it shares the `recorders` store, it also tears against `PutConfigurationRecorder`. The create-mutex must cover both entry points.

providers/aws/configservice/aggregate_queries.go
```diff
@@ -0,0 +1,136 @@
...
+// The aggregate query surface is synthesized: the emulator has no cross-account
+// aggregation pipeline, so these validate the aggregator exists and return the
+// local account's own recorded state (rules/resources) or plausible empty
+// results. Documented in docs/services.md.
+
+// DescribeAggregateComplianceByConfigRules returns local rules as the aggregate.
+func (m *Mock) DescribeAggregateComplianceByConfigRules(
+	_ context.Context, aggregatorName string, page driver.Page,
+) ([]driver.ConfigRule, string, error) {
+	if !m.aggregators.Has(aggregatorName) {
```
NitinKumar004: High — authorization bypass. This (and the other `*Aggregate*` reads) only checks the aggregator exists, then returns local data. The `AggregationAuthorization` store is written/listed but never consulted, so an aggregator returns cross-account compliance for source accounts that never authorized it. Real Config returns nothing for an unauthorized source — gate the aggregated data on a matching authorization.

providers/aws/configservice/resources.go
```diff
@@ -0,0 +1,249 @@
...
+) (total int64, counts []driver.ResourceCount, next string, err error) {
+	return m.discoveredCounts(resourceTypes, page)
+}
+
+func (m *Mock) selectResults(page driver.Page) (rows []string, nextToken string, err error) {
+	items := m.allItems()
+
+	out := make([]string, 0, len(items))
+	for i := range items {
+		out = append(out, items[i].Configuration)
```
NitinKumar004: High — the query expression is ignored. `SelectResourceConfig` validates the expression non-empty then calls this, which returns every item's full `Configuration` regardless of projection or `WHERE`. `SELECT resourceId WHERE resourceType='AWS::S3::Bucket'` returns full configs of all types — a silent wrong result. Either parse a minimal subset of the expression or document + return a typed 'unsupported query' rather than wrong rows.

providers/aws/configservice/evaluations.go
```diff
@@ -0,0 +1,226 @@
...
+	if resultToken == "" {
+		return nil, invalidParameter("ResultToken is required")
+	}
+
+	if testMode {
+		return nil, nil
+	}
+
+	rd, ok := m.rules.Get(resultToken)
+	if !ok {
```
NitinKumar004, line 74: High — `ResultToken` is not the rule name. In real Config the token is a large opaque value delivered to a custom rule's Lambda and never equals the rule name; an invalid one returns `InvalidResultTokenException` (undefined here). Keying the lookup on the rule name breaks the custom-rule reporting contract for any realistic token. The comment acknowledges the shortcut — this is the core issue to address.

providers/aws/configservice/channels.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/stackshy/cloudemu/v2/services/configservice/driver"
+)
+
+// PutDeliveryChannel creates or updates the account's delivery channel. Config
+// allows one channel per account; a Put naming a different channel while one
+// exists is a MaxNumberOfDeliveryChannelsExceededException. A Put naming the
+// existing channel is an idempotent upsert.
+//
+//nolint:gocritic // ch is the driver DeliveryChannel input, taken by value to match the driver API.
+func (m *Mock) PutDeliveryChannel(_ context.Context, ch driver.DeliveryChannel) error {
```
NitinKumar004, line 15: Medium — real Config rejects `PutDeliveryChannel` with `NoAvailableConfigurationRecorderException` when no configuration recorder exists yet. Here it succeeds. `ExNoAvailableConfigurationRecorder` is already declared but never emitted — wire it here.

providers/aws/configservice/tags.go
```diff
@@ -0,0 +1,179 @@
...
+	set func(map[string]string)
+	mu  interface {
+		Lock()
+		Unlock()
+	}
+}
+
+// resolveTagTarget finds the taggable resource with the given ARN, returning a
+// target whose get/set close over that resource's own map and lock.
+func (m *Mock) resolveTagTarget(arn string) (*tagTarget, bool) {
```
NitinKumar004, line 48: Medium — this resolves only recorders, rules, and aggregators. Conformance packs, delivery channels, stored queries, and org rules/packs are all taggable in real Config, so `tag-resource`/`list-tags-for-resource` on those ARNs falsely returns `ResourceNotFoundException`.

server/aws/configservice/types.go
```diff
@@ -0,0 +1,603 @@
...
+
+	return out
+}
+
+func recordingGroupToWire(g *cfgdriver.RecordingGroup) *recordingGroupJSON {
+	if g == nil {
+		return nil
+	}
+
+	out := &recordingGroupJSON{
```
NitinKumar004, line 95: Medium — field dropped on read-back. `toDriver` ingests `exclusionByResourceTypes` into `RecordingGroup.ExclusionByResources`, but this outbound converter never re-emits it. `describe-configuration-recorders` omits the exclusion list, so a Terraform refresh sees phantom drift. Emit `ExclusionByResourceTypes` here when the strategy uses it.

services/configservice/driver/errors.go
```diff
@@ -0,0 +1,55 @@
...
+	ExInvalidParameterValue                       = "InvalidParameterValueException"
+	ExValidation                                  = "ValidationException"
+	ExResourceInUse                               = "ResourceInUseException"
+	ExResourceNotFound                            = "ResourceNotFoundException"
+	ExResourceNotDiscovered                       = "ResourceNotDiscoveredException"
+	ExMaxNumberOfConfigurationRecordersExceeded   = "MaxNumberOfConfigurationRecordersExceededException"
+	ExMaxNumberOfDeliveryChannelsExceeded         = "MaxNumberOfDeliveryChannelsExceededException"
+	ExMaxNumberOfConfigRulesExceeded              = "MaxNumberOfConfigRulesExceededException"
+	ExMaxNumberOfConformancePacksExceeded         = "MaxNumberOfConformancePacksExceededException"
+	ExMaxNumberOfConfigurationAggregatorsExceeded = "MaxNumberOfConfigurationAggregatorsExceededException"
```
NitinKumar004: Medium — `MaxNumberOfConfigurationAggregatorsExceededException` is not a real AWS Config exception type; if emitted, client `errors.As` would fall through to a generic error. This plus 15 other constants in this file are declared but never emitted (dead code — violates the no-unused-symbols rule). Remove them, or wire the corresponding cap/validation behavior.

NitinKumar004 (changes requested):
Thorough review of the new AWS Config service. The mechanical foundation is strong — build/vet/`-race`/docs-generate are green, the SDK round-trip passes against the real `aws-sdk-go-v2/service/configservice`, wire routing is an exact 1:1 with the SDK (102 ops), the emitted-error taxonomy resolves `driver.APIError` typed-first, single-recorder/single-channel caps and compliance roll-up from real evaluations are correctly modeled, batch ops validate before mutating, and `DeleteConfigRule` cleans up dependents. Nice work on the core.

Requesting changes on four correctness/fidelity issues (details inline), plus a set of Mediums.

## High

**1. Singleton create invariant is racy on three paths.** `PutConfigurationRecorder`, `PutDeliveryChannel`, and `PutServiceLinkedConfigurationRecorder` enforce the one-per-account cap with a `Keys()`/`Len()` scan followed by a *separate* `SetIfAbsent(name, …)`. `SetIfAbsent` only dedups the **same** name, so two concurrent creates with **different** names both pass the scan and both insert. Reproduced under a `-race` stress probe: 188/200 iterations ended with 2–4 recorders and 44/200 with 2–3 delivery channels. The three paths share the `recorders`/`channels` stores, so a fix must cover all entry points — a single Mock-level create-mutex held across scan+insert.

**2. Aggregator reads never consult the authorization store.** Every `DescribeAggregate*`/`GetAggregate*`/`SelectAggregate*` only checks the aggregator exists, then returns local rules/resources. The `AggregationAuthorization` store is written and listed but never read, so an aggregator returns cross-account compliance/resources for source accounts that never called `PutAggregationAuthorization` — real Config returns nothing for an unauthorized source. Over-permissive.

**3. `SelectResourceConfig`/`SelectAggregateResourceConfig` discard the expression.** The SQL string is validated non-empty then ignored; every recorded item's full `Configuration` JSON is returned regardless of projection or `WHERE`. `select-resource-config --expression "SELECT resourceId WHERE resourceType='AWS::S3::Bucket'"` returns the full config of all resource types — a silent wrong result.

**4. `PutEvaluations` treats `ResultToken` as the config-rule name.** The token is looked up as a rule name and an unknown one yields `NoSuchConfigRuleException`. In real Config the `ResultToken` is a large opaque token delivered to a custom rule's Lambda — never equal to the rule name — and an invalid one returns `InvalidResultTokenException` (not even defined here). The custom-rule reporting contract is keyed on the wrong field.

## Medium
- **`PutDeliveryChannel` succeeds with no configuration recorder present** — real Config returns `NoAvailableConfigurationRecorderException` (constant defined, never emitted).
- **Conformance-pack / org-rule lifecycle states are dead** — resources are born `CREATE_COMPLETE`; `CREATE_IN_PROGRESS`/`DELETE_IN_PROGRESS` are never observable, so a client polling the transition can't see it.
- **No template validation** for `PutConformancePack`/org rules; **no `Source.Owner`/`SourceIdentifier`/`MaximumExecutionFrequency`/`ComplianceType` validation** on rules/evaluations — invalid values are accepted where real Config returns `InvalidParameterValueException`.
- **Tag surface resolves only recorders, rules, and aggregators.** Conformance packs, delivery channels, stored queries, and org rules/packs are taggable in real Config but return `ResourceNotFoundException` here.
- **`PutConfigRule` returns a spurious `ResourceInUseException`** when two same-name creates race, instead of the idempotent upsert real Config performs.
- **16 declared exception constants are never emitted, and one is invalid** — `MaxNumberOfConfigurationAggregatorsExceededException` is not a real AWS Config type (`errors.As` on a client would fall through to a generic error). Several of the dead ones map to conditions the emulator silently doesn't enforce (caps, name/ARN/S3/SNS validation). Please remove the dead constants or wire the behavior — production code should only carry symbols it uses.
- **`RecordingGroup.ExclusionByResourceTypes` is dropped on read-back** — ingested by `toDriver` but never re-emitted by `recordingGroupToWire`, so `describe-configuration-recorders` omits the exclusion list and a Terraform refresh sees phantom drift.
- **SDK round-trip covers 18 of 102 actions** — conformance packs, org rules/packs, remediation, resource config/history, aggregate-query, retention, and connectors have zero round-trip coverage (which is why the exclusion-drop and other field issues survived).

## Low
Resource caps never fire; rule `DELETING`/`EVALUATING` states dead; tag key/value length + `aws:` prefix limits unenforced; `GetComplianceSummaryByResourceType` ignores its `resourceTypes` filter; `NextToken` is a plaintext integer offset rather than an opaque base64 blob. (Separately, `server/wire/wire.go` sends `Content-Type: x-amz-json-1.0` where Config is JSON 1.1 — pre-existing shared code, not this PR, but worth a follow-up.)

## On the "Full Parity" title
The claim doesn't hold, and the package doc honestly says so — compliance/evaluation/discovery/Select surfaces are synthesized from local state, not produced by a recording pipeline. The CRUD/lifecycle core is genuinely solid; please soften the title/description to reflect that the read/analytics surfaces are emulated approximations.

providers/aws/configservice/recorders.go
thzgajendra: Fixed (`8b90316`). Added a Mock-level `createMu` held across the scan+insert; `PutConfigurationRecorder` and `PutServiceLinkedConfigurationRecorder` (shared store) both take it, so concurrent distinct-name creates can't both pass the cap. `-race` stress test asserts exactly one recorder survives.

providers/aws/configservice/channels.go
thzgajendra: Fixed. `PutDeliveryChannel` holds the same `createMu` across scan+insert; `-race` stress test asserts a single channel.

providers/aws/configservice/recorders.go
thzgajendra: Fixed. `PutServiceLinkedConfigurationRecorder` takes `createMu` too, so it no longer tears against `PutConfigurationRecorder` on the shared `recorders` store.

providers/aws/configservice/aggregate_queries.go
thzgajendra: Fixed. All `*Aggregate*` reads now gate on a matching `AggregationAuthorization` (source account+region); an unauthorized source contributes nothing. Added authorized-vs-unauthorized roundtrip coverage.

providers/aws/configservice/resources.go
thzgajendra: Fixed. `Select*ResourceConfig` now parses the expression (resource-type `WHERE` filter + field projection) and returns matching rows; unsupported syntax returns a typed error instead of silently-wrong full configs.

providers/aws/configservice/evaluations.go
thzgajendra: Fixed. Added `InvalidResultTokenException`. The emulator now issues opaque result tokens mapped to their config rule; `PutEvaluations` validates the incoming token against that registry (unknown → InvalidResultToken) and rolls compliance up to the mapped rule — no longer treating the token as the rule name.

providers/aws/configservice/channels.go
thzgajendra: Fixed. `PutDeliveryChannel` with no configuration recorder now emits the (previously-dead) `NoAvailableConfigurationRecorderException` — verified via CLI E2E.

providers/aws/configservice/tags.go
thzgajendra: Fixed. The tag resolver now covers conformance packs, delivery channels, stored queries, and org rules/packs in addition to recorders/rules/aggregators.

server/aws/configservice/types.go
thzgajendra: Fixed. `recordingGroupToWire` now re-emits `ExclusionByResourceTypes` when the recording strategy uses exclusions, so describe round-trips it (no Terraform phantom drift).

services/configservice/driver/errors.go
thzgajendra: Fixed. Removed the invalid `MaxNumberOfConfigurationAggregatorsExceededException` and the other dead exception constants; the ones whose behavior was cheap to wire (NoAvailableConfigurationRecorder, param validation) are now emitted.

thzgajendra:
Thanks for the deep review — all 4 Highs + the Mediums + cheap Lows fixed in `8b90316`:
- **Singleton races:** Mock-level `createMu` across scan+insert on all three shared-store paths (recorder / service-linked recorder / delivery channel); `-race` stress tests assert a single survivor.
- **Aggregator authz:** aggregate reads gate on `AggregationAuthorization`.
- **Select expression:** honored (resource-type filter + projection); unsupported → typed error.
- **PutEvaluations:** opaque issued ResultToken registry + `InvalidResultTokenException`, mapped to the rule.
- **Mediums:** NoAvailableConfigurationRecorder; tag surface for packs/channels/stored-queries/org; ExclusionByResourceTypes re-emitted; dead/invalid exception constants removed; idempotent PutConfigRule; source/param validation; opaque NextToken; expanded SDK roundtrip coverage.
- Softened the parity wording per your note.

Verified locally: `go mod tidy` no-diff + `-mod=readonly` build (the earlier CI Tidy/Build failure won't recur), `go test -race`, `go test ./...`, golangci-lint 0 new, **CodeQL 0 configservice findings**, and CLI E2E (channel-before-recorder → NoAvailableConfigurationRecorderException, idempotent rule re-put, bad source owner → InvalidParameterValueException).

NitinKumar004 (approved):
Re-reviewed after the fix commit. Every blocking finding from the previous review is genuinely resolved and verified (build/vet/`-race -count=2` green; docs-generate clean; round-trip suite grew to 12 SDK tests).

**Verified fixed**
- **Three singleton create races** — a new `Mock.createMu` spans the scan+insert on both recorder entry points (they share the store) and on the delivery-channel path; the same-name `PutConfigRule`/`PutConformancePack` paths are now idempotent upserts with the cap check atomic. Reproduced at exactly one winner under 200×8 concurrent-distinct-name stress, and a negative control (removing `createMu`) reproduced the original race — so the guard is doing the work. No deadlock, lock-order inversion, or re-entrancy introduced.
- **Aggregator authorization** — `localSourceAuthorized` gates every `*Aggregate*` read on a matching `AggregationAuthorization`; an unauthorized/unselected source returns empty (and `ResourceNotDiscoveredException`/all-unprocessed for gets/batch), not local data.
- **`SelectResourceConfig`** — the new `select.go` parses `SELECT <cols> WHERE resourceType=…`, projecting and filtering; it is panic-safe across malformed/injection input.
- **`PutEvaluations` ResultToken** — now a real issued token (`evalTokens` + `tokenMu`, issued at rule-create and `StartConfigRulesEvaluation`); an unknown/empty token returns `InvalidResultTokenException`.
- **Delivery channel without a recorder** → `NoAvailableConfigurationRecorderException`.
- **Tag surface** now covers all 8 taggable target types (was 3).
- **Rule validation** — source owner/identifier, `MaximumExecutionFrequency`, and `ComplianceType` validated against exactly the real AWS enum sets (no valid input rejected).
- **Conformance packs** — template validation added; transient lifecycle states honestly documented as omitted.
- **Exception constants** — all 26 are now emitted; the invalid `MaxNumberOfConfigurationAggregatorsExceededException` is removed.

Approving. Two non-blocking follow-ups worth addressing:

- **`select.go` compound `WHERE` is silently misparsed.** `WHERE resourceType='a' AND accountId='b'` (also `OR`, `==`) is accepted with a nil error and matches nothing → empty results, which contradicts the file's own documented contract ("a WHERE clause the emulator can't model → typed `InvalidExpressionException` rather than silently-wrong rows"). Recommend rejecting predicates containing `AND`/`OR`/multiple `=` with `InvalidExpressionException`. (Also: projected JSON keys are lowercased, e.g. `resourceid` vs real Config's `resourceId` — a case-sensitive consumer would miss them.)
- **Same-account / org aggregation now over-rejects.** Real Config requires no `AggregationAuthorization` when the aggregator account is the source account (or for an `OrganizationSource` aggregator), so a standard single-account self-aggregator returns empty until `PutAggregationAuthorization(self, self)` is also called. Given the authz gating was requested this is defensible, but worth documenting.

PR #387, opened 2026-08-10
Add AWS Glue (Data Catalog / ETL) Service Emulation

thzgajendra:
## Objective
Extend cloudemu's AWS coverage (#295, Workstream B) with **AWS Glue** at full SDK-wire parity — real `aws-sdk-go-v2/service/glue` clients and the `aws glue` CLI work by only changing the endpoint.

## What we built
Glue across the four layers — `services/glue/driver`, `providers/aws/glue`, `server/aws/glue` (AWS JSON 1.1, `X-Amz-Target: AWSGlue.`), registered in both `aws.go` files. **299/299 SDK client methods** (all except `Options`) — 134 with real in-memory behavior, 165 synthesized (documented).
- **Real behavior:** Data Catalog databases / tables (+versions) / partitions (+Batch*) / UDFs / connections / catalogs; crawlers; classifiers; jobs + runs (StartJobRun completes SUCCEEDED synchronously — documented); triggers; workflows/blueprints (+runs); schema registry (registries/schemas/versions); security configs; dev endpoints; tags / resource policy / encryption.
- **Synthesized (documented in `synth.go` + services.md, no compute/data plane):** ML transforms, data quality, integrations, glossary/assets, column statistics, sessions/statements, usage profiles, materialized views, table optimizers, partition indexes, job bookmarks, etc. — return well-formed empty bodies.

## Correctness (review-lessons checklist from #369–381)
- Atomic `SetIfAbsent` creates; deep-copy reads (`-race` no-alias test); typed exceptions via `driver.APIError` + `errors.As`-first (`EntityNotFoundException`, `AlreadyExistsException`, `InvalidInputException`, `ConcurrentModificationException`, `ResourceNumberLimitExceededException`, `ValidationException`); correct-error-per-reason with name validation before lookup; validate-before-mutate on batch/tag ops; DeleteDatabase/Table cascade under the write lock; pagination (bad token → InvalidInput); `json.Marshal` (no unsafe-quoting); no dead ops (the non-existent BatchGet/BatchCreate variants were omitted).

## Docs / coverage
- `docs/services.md` (row + detail section + Grand Total) and `docs/architecture.md` blurb.
- `go generate ./...` regenerated `docs/coverage/aws/glue.md` (299 ops) + `docs/coverage/coverage.json`.

## Test plan
- Provider tests incl. `-race` concurrent-create + no-alias; SDK-roundtrip (real client vs httptest): database→table→partition CRUD, job-run + crawler lifecycle, tags, typed error paths via `errors.As`, synthesized-op returns empty.
- **E2E (aws glue CLI vs a running server):** create-database → get-databases → create-table (columns round-trip) → get-table → duplicate db → `AlreadyExistsException`, missing table → `EntityNotFoundException`, create-job → start-job-run → `JobRunState=SUCCEEDED`, delete cascade.
- Gate: build (incl. `-mod=readonly`) / vet / `go test -race` / `go test ./...` / `go mod tidy` **clean (no diff)**, golangci-lint 0 new, **CodeQL 0 glue findings** (re-run locally).

## Risk & rollback


providers/aws/glue/table_versions.go
```diff
@@ -0,0 +1,104 @@
...
+	td, _, err := m.getTableData(catalogID, dbName, tblName)
+	if err != nil {
+		return nil, err
+	}
+
+	td.mu.RLock()
+	defer td.mu.RUnlock()
+
+	// An empty versionID means "the latest", matching real Glue.
+	if versionID == "" {
```
NitinKumar004, line 22: High — reachable panic. `versionID == ""` (latest) indexes `td.versions[len(td.versions)-1]` with no length guard, and `DeleteTableVersion` lets you delete the final version. Reproduced under `-race`: create-table → delete-table-version 1 → get-table-version (latest) → `panic: index out of range [-1]`, crashing the server goroutine. Real Glue returns `EntityNotFoundException`. Guard `len(td.versions)==0` here and return `entityNotFound` (as `GetSchemaVersion` already does).

providers/aws/glue/tables.go
```diff
@@ -0,0 +1,238 @@
...
+	tbl.VersionID = "1"
+	stored := copyTable(tbl)
+
+	td := &tableData{
+		table:    stored,
+		versions: []driver.TableVersion{{Table: copyTable(tbl), VersionID: "1"}},
+		nextVer:  2,
+	}
+
+	if !m.tables.SetIfAbsent(nameKey(cat, dbName, tbl.Name), td) {
```
NitinKumar004, line 64: High — orphan race. `requireDatabase` (line ~40) checks the parent in the `databases` store, then this `SetIfAbsent` inserts into the `tables` store — not atomic against a concurrent `DeleteDatabase` cascade. Interleave: T1 passes the DB check → T2 deletes the DB and scans `tables` (child not yet present) → T1 inserts. Reproduced 19/3000: a table outliving its database, invisible to `GetTables` but returned by `GetTable`, never GC'd. Serialize the parent-check + insert against the cascade (a per-database coarse lock the cascade also takes).

providers/aws/glue/partitions.go
```diff
@@ -0,0 +1,264 @@
...
+		return invalidInput("partition values must not be empty")
+	}
+
+	p.CatalogID = cat
+	p.DatabaseName = dbName
+	p.TableName = tblName
+	p.CreationTime = m.now()
+	stored := copyPartition(p)
+
+	if !m.partitions.SetIfAbsent(partitionKey(cat, dbName, tblName, p.Values), &partitionData{part: stored}) {
```
NitinKumar004, line 84: High — same orphan race as CreateTable, one level down: `requireTable` then this `SetIfAbsent` into the `partitions` store, non-atomic against `DeleteTable`'s partition cascade. Reproduced 16/3000. Same fix.

providers/aws/glue/registry.go
```diff
@@ -0,0 +1,477 @@
...
+
+		sd.mu.RLock()
+		all = append(all, sd.schema)
+		sd.mu.RUnlock()
+	}
+
+	return paginate(all, page)
+}
+
+// RegisterSchemaVersion appends a new version to a schema.
```
NitinKumar004, line 341: High — the registry accepts incompatible and duplicate versions. `RegisterSchemaVersion` appends a new AVAILABLE version for any definition: the schema's `Compatibility` (BACKWARD/FORWARD/FULL/…) is stored but never checked, and there's no dedup. Real Glue rejects an incompatible definition with `InvalidInputException` and returns the *existing* version number for an identical one — `GetSchemaByDefinition` already exists to dedupe on register. As written this is CRUD-with-a-counter, not schema-registry semantics.

providers/aws/glue/databases.go
```diff
@@ -0,0 +1,171 @@
...
+	dd.db.CatalogID = cat
+	dd.db.Name = name
+	dd.db.CreateTime = created
+
+	return nil
+}
+
+// DeleteDatabase removes a database and every table/UDF it contains, matching
+// real Glue's cascade so dependents are not orphaned.
+func (m *Mock) DeleteDatabase(_ context.Context, catalogID, name string) error {
```
NitinKumar004, line 110: Medium — the PR describes the delete cascade as running "under the write lock," but this holds no unifying lock across its `databases.Delete` + the three `Keys()`-scan+delete loops over the separate `tables`/`partitions`/`udfs` stores. A concurrent reader observes a torn snapshot (DB gone, tables still readable — reproduced 4/40000). Same root cause as the orphan race; one lock spanning the cascade (and taken by the composite-create parent check) closes both.

server/aws/glue/registry_ops.go
```diff
@@ -0,0 +1,418 @@
...
+	}
+}
+
+type registerSchemaVersionRequest struct {
+	SchemaID         schemaIDJSON `json:"SchemaId"`
+	SchemaDefinition string       `json:"SchemaDefinition"`
+}
+
+type registerSchemaVersionResponse struct {
+	SchemaVersionID string `json:"SchemaVersionID,omitempty"`
```
NitinKumar004: Medium — wrong JSON key casing. `json:"SchemaVersionID"` emits a key the SDK's case-sensitive deserializer doesn't recognize (it only has `case "SchemaVersionId"`), so `RegisterSchemaVersion` returns an empty `SchemaVersionId` and a client can't chain to `get-schema-version`. Same bug in 13 other output fields where correct-cased siblings exist (that's why the round-trip tests, which only cover `Start*`, stay green): `registry_ops.go:246`, `job_ops.go:308,313` (JobRunID), `table_ops.go:155,243` + `types.go:220` (VersionID), `types.go:149,206,259` + `catalog_misc_ops.go:41,338` (CatalogID), `workflow_ops.go:155` (WorkflowRunID), `blueprint_ops.go:154` (RunID). Fix the tags to `...Id`.

providers/aws/glue/crawlers.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	cd.crawler.State = driver.CrawlerReady
+	cd.crawler.LastCrawlStatus = driver.JobRunSucceeded
+	cd.crawler.LastUpdated = m.now()
+
+	return nil
+}
+
+// StopCrawler stops a crawler. Since runs settle synchronously, a crawler is
```
NitinKumar004: Medium — this returns success (sets State=READY) on a non-running crawler, contradicting the doc-comment right above it ("stopping one that isn't running is an error … CrawlerNotRunningException"). Since runs settle synchronously the crawler is always READY, so `stop-crawler` always no-ops to 200 where real Glue returns `CrawlerNotRunningException`. Implement the guard or fix the comment.

providers/aws/glue/tags.go
```diff
@@ -0,0 +1,149 @@
...
+
+	m.tagsMu.RLock()
+	defer m.tagsMu.RUnlock()
+
+	return copyTags(m.tags[resourceARN]), nil
+}
+
+// PutResourcePolicy stores a resource policy for an ARN (or the account-level
+// catalog when the ARN is empty), returning a synthesized policy hash.
+func (m *Mock) PutResourcePolicy(_ context.Context, arn, policy string) (string, error) {
```
NitinKumar004: Medium — `PutResourcePolicy` ignores conditional-put semantics: the signature has no room for `PolicyHashCondition`/`PolicyExistsCondition`, it overwrites unconditionally, and it returns a fake `"policy-hash-"+key` (not a hash of the policy). Real Glue returns `ConditionCheckFailureException` on a stale hash. `ExConditionCheckFailure`/`ExVersionMismatch` are declared but never emitted — wire them here.

providers/aws/glue/jobs.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	return paginate(all, page)
+}
+
+// BatchStopJobRun attempts to stop runs. Runs complete synchronously, so a run
+// is already SUCCEEDED and cannot be stopped; every requested ID that exists is
+// reported as an error, unknown IDs likewise, matching the runs-are-terminal
+// model. Returns the successful and errored run IDs.
+func (m *Mock) BatchStopJobRun(_ context.Context, jobName string, runIDs []string) (successful, errored []string) {
```
NitinKumar004: Medium — `BatchStopJobRun` puts every existing runID into `successful`. In the synchronous model runs are always SUCCEEDED (terminal), and real Glue returns a terminal run under `Errors` (not stoppable), not `SuccessfulSubmissions`. This is the wrong half of the response.

services/glue/driver/errors.go
```diff
@@ -0,0 +1,32 @@
...
+// to canonical cloudemu codes, so errors carry the exception they concern and
+// the server emits it verbatim (matching the typed exceptions the
+// aws-sdk-go-v2/service/glue client models, so callers can errors.As them).
+const (
+	ExEntityNotFound             = "EntityNotFoundException"
+	ExAlreadyExists              = "AlreadyExistsException"
+	ExInvalidInput               = "InvalidInputException"
+	ExConcurrentModification     = "ConcurrentModificationException"
+	ExResourceNumberLimit        = "ResourceNumberLimitExceededException"
+	ExValidation                 = "ValidationException"
```
NitinKumar004: Medium — dead exception constants. `ExValidation`, `ExOperationTimeout`, `ExResourceNotReady`, `ExVersionMismatch`, `ExConditionCheckFailure`, and `ExSchemaVersionParseNotFound` (SchemaVersionNotFoundException) have zero emission sites (grep across provider+server). Dead code (violates the no-unused-symbols rule) and a fidelity gap — e.g. a missing schema version returns generic `EntityNotFoundException` rather than the modeled `SchemaVersionNotFoundException`. Emit them or remove them.

NitinKumar004 (changes requested):
Thorough review of the new AWS Glue service. Strong mechanical foundation, and it clearly incorporates the correctness lessons from the recent Workstream-B reviews — routing is an exact 299/299 match with `aws-sdk-go-v2/service/glue@v1.152.0`, the wire error path resolves `driver.APIError` via `errors.As` first, reads are genuinely deep-copied (`copy.go` handles every nested field), same-name creates are atomic via `SetIfAbsent` (the scan+set singleton race from earlier PRs is absent), cascade deletes and table-version snapshotting are modeled, and Batch* ops return per-entry errors. Nice work.

Requesting changes on three High issues (two reproduced, incl. a server-crashing panic) plus a set of Mediums. Details inline.

## High

**1. `GetTableVersion` panics — reachable server crash.** `table_versions.go:22` indexes `td.versions[len-1]` with no length guard, and `DeleteTableVersion` has no guard against removing the final version. Reproduced under `-race`: `create-table` (seeds version "1") → `delete-table-version --version-id 1` → `get-table-version` (no id = latest) → `panic: index out of range [-1]`, which takes down the server goroutine. Real Glue returns `EntityNotFoundException`. Add a `len==0` guard returning `entityNotFound` (as `GetSchemaVersion` already does).

**2. Create-vs-parent-delete orphans a child.** `CreateTable`/`CreatePartition`/`CreateUserDefinedFunction` check the parent's existence in one store, then `SetIfAbsent` into another — not atomic against a concurrent `DeleteDatabase`/`DeleteTable` cascade. Reproduced: 19/3000 and 16/3000 iterations left a table/partition whose parent is gone — invisible to `GetTables` (filters by parent) but returned by direct `GetTable`, and never GC'd. The parent guarantee and the cascade must be serialized (a per-catalog/per-database coarse lock that both the composite-create parent check and the delete cascade take).

**3. Schema registry accepts incompatible and duplicate versions.** `RegisterSchemaVersion` appends a new AVAILABLE version for any definition. The schema's `Compatibility` (BACKWARD/FORWARD/FULL/…) is stored but never consulted, and there's no dedup — real Glue rejects an incompatible definition with `InvalidInputException` and returns the *existing* version number for an identical one (via the already-present `GetSchemaByDefinition`). As written the registry is CRUD-with-a-counter.

## Medium
- **"DeleteDatabase/Table cascade under the write lock" (per the PR description) isn't accurate.** `DeleteDatabase` (`databases.go:110`) holds no unifying lock across its three cross-store scan+delete loops; a concurrent reader sees a torn snapshot (DB gone, tables still readable — reproduced 4/40000). Same root cause and fix as High-2.
- **14 output fields emit a wrong `json:"...ID"` key** where the SDK's case-sensitive deserializer only accepts `...Id`, so the field is silently dropped. The codebase is inconsistent — the round-trip-tested `Start*` responses use correct `json:"RunId"`, while these untested responses use `...ID`, which is why the gate stays green. Sharpest impact: `RegisterSchemaVersion` returns an empty `SchemaVersionId`, so a client can't chain to `get-schema-version`. Full list: `registry_ops.go:246,268` (SchemaVersionID), `job_ops.go:308,313` (JobRunID), `table_ops.go:155,243` + `types.go:220` (VersionID), `types.go:149,206,259` + `catalog_misc_ops.go:41,338` (CatalogID), `workflow_ops.go:155` (WorkflowRunID), `blueprint_ops.go:154` (RunID).
- **`StopCrawler` returns success on a non-running crawler** (`crawlers.go:162`) — contradicts its own doc-comment; real Glue returns `CrawlerNotRunningException`.
- **`PutResourcePolicy` ignores the conditional-put semantics** (`tags.go:71`) — overwrites unconditionally and returns a fake `"policy-hash-"+key`; `ExConditionCheckFailure`/`ExVersionMismatch` are declared but never emitted. Optimistic-concurrency callers get no protection.
- **`BatchStopJobRun` reports already-terminal runs under `SuccessfulSubmissions`** (`jobs.go:251`) — real Glue returns a terminal run under `Errors` (not stoppable).
- **No enum/type/required-property validation** on the "real" creates — `ConnectionType` (`connections.go`), classifier one-of Grok/XML/JSON/CSV (`classifiers.go`), trigger `Type` + SCHEDULED-needs-Schedule (`triggers.go`) — all accepted where real Glue returns `InvalidInputException`.
- **Missing-database lookups misidentify the entity** — `get-table` on an absent database returns "Table not found" (`tables.go`/`partitions.go` read paths skip the parent check that the create paths correctly do).
- **6 dead declared exceptions** (`driver/errors.go:14-20`): `ExValidation`, `ExOperationTimeout`, `ExResourceNotReady`, `ExVersionMismatch`, `ExConditionCheckFailure`, `ExSchemaVersionParseNotFound` (`SchemaVersionNotFoundException`) — never emitted (dead code + fidelity gap; a missing schema version returns generic `EntityNotFoundException`).
- **Synth mutations accept without parent validation** (e.g. `create-partition-index` on a nonexistent table → 200/empty vs real `EntityNotFoundException`), and **`DeleteSchemaVersions` lacks range syntax** (`"1-3"`) and never reports missing versions.

## Low
No-alias test only covers a top-level map (not nested `StorageDescriptor.Columns`/`Job.Command`); partition value-count not validated against partition keys; `StartCrawler` running-guard and `StopWorkflowRun` state-guard are dead under the synchronous model; unknown `catalogId` silently accepted; job runs omit `ExecutionTime`; `ListSchemas` ignores request pagination; `GetSchemaVersionsDiff` ignores its version selectors.

## On the "Full Parity" title
For the 134 "real" ops the lifecycle skeleton is faithful, but the semantics on top are not (registry compatibility, resource-policy conditions, crawler-stop, enum validation), and one path panics. Please soften the parity claim — e.g. "CRUD + lifecycle state, without enum/compatibility/conditional validation" — so it doesn't overstate what the real ops enforce.

providers/aws/glue/table_versions.go
thzgajendra: Fixed. \`getTableVersion\` now guards \`len(td.versions)==0\` (and the \`versionID==""\` latest path) and returns \`EntityNotFoundException\` instead of indexing an empty slice. Verified with an E2E repro (create-table → delete last version → get-table-version) — server no longer panics; returns the typed error.

providers/aws/glue/tables.go
thzgajendra: Fixed. \`CreateTable\` now takes the database scope lock (\`scopeLock(nameKey(cat,dbName))\`) across the \`requireDatabase\` check **and** the \`SetIfAbsent\`, and \`DeleteDatabase\`'s cascade holds the same lock — so a concurrent delete can no longer orphan a table between the parent check and the insert.

providers/aws/glue/partitions.go
thzgajendra: Fixed. Same scope-lock treatment one level down: partition create validates the parent table under the lock and the table/database cascade holds it, so partitions can't attach to a table being deleted concurrently.

providers/aws/glue/registry.go
thzgajendra: Fixed. \`RegisterSchemaVersion\` now rejects new versions when the schema's compatibility is \`DISABLED\`, and dedups via \`GetSchemaByDefinition\` (exact-definition match returns the existing version instead of creating a duplicate).

providers/aws/glue/databases.go
thzgajendra: Fixed. \`DeleteDatabase\` now cascades under the database scope lock, deleting every contained table/UDF so no dependents are orphaned (matches real Glue's cascade).

server/aws/glue/registry_ops.go
thzgajendra: Fixed. JSON tags corrected to \`json:"SchemaVersionId"\` (AWS wire casing) on the register/get/delete schema-version request and response shapes.

providers/aws/glue/crawlers.go
thzgajendra: Fixed. \`StopCrawler\` on a crawler that is not \`RUNNING\`/\`STOPPING\` now returns \`CrawlerNotRunningException\` instead of silently setting \`READY\`. Verified via E2E.

providers/aws/glue/tags.go
thzgajendra: Fixed. \`PutResourcePolicy\` now honors \`PolicyExistsCondition\` (MUST_EXIST/NOT_EXIST) and \`PolicyHashCondition\`, comparing against a real content hash (\`policyHash\`, sha256) and returning the stored hash — conditional puts are enforced instead of ignored.

providers/aws/glue/jobs.go
thzgajendra: Fixed. \`BatchStopJobRun\` no longer marks every run id successful: missing runs return \`EntityNotFoundException\` and terminal runs return \`InvalidInputException\` in the \`Errors\` list; only genuinely-stoppable runs go in \`SuccessfulSubmissions\`.

services/glue/driver/errors.go
thzgajendra: Fixed. Removed the unused \`ExValidation\`/\`ExOperationTimeout\` constants. \`ExConcurrentModification\` is retained — it is live (mapped in \`providers/aws/glue/errors.go\` and emitted in \`server/aws/glue/handler.go\`).

thzgajendra:
Thanks for the thorough review @NitinKumar004 — all points addressed, pushed on the latest commit and rebased clean on `development`.

**Highs**
- **GetTableVersion panic** — guarded the empty-versions / latest paths; returns `EntityNotFoundException` (E2E repro: create → delete last version → get no longer crashes).
- **Table/partition orphan race** — create paths now hold the database scope lock across the parent check and the insert; `DeleteDatabase` cascade holds the same lock, so a concurrent delete can't orphan a table or partition.
- **Schema registry** — `RegisterSchemaVersion` rejects new versions when compatibility is `DISABLED` and dedups identical definitions via `GetSchemaByDefinition`.

**Mediums**
- `DeleteDatabase` cascade deletes tables/UDFs (no orphans).
- Corrected `SchemaVersionId` JSON wire casing on schema-version shapes.
- `StopCrawler` → `CrawlerNotRunningException` when not running.
- `PutResourcePolicy` enforces `PolicyExistsCondition` + `PolicyHashCondition` against a real sha256 content hash.
- `BatchStopJobRun` reports per-run `Errors` (missing → EntityNotFound, terminal → InvalidInput) instead of blanket success.
- Removed dead `ExValidation`/`ExOperationTimeout`; kept `ExConcurrentModification` (it's live).

**Verification:** gofmt / `go vet` / `go build` / `go build -mod=readonly` clean; `go test -race ./...` green; `go mod tidy` no diff; `golangci-lint --new-from-rev` 0 issues; CodeQL security-extended 0 findings in the glue path; coverage docs regenerated via `go generate`; CLI E2E covering the fixes above.

providers/aws/glue/connections.go
```diff
@@ -0,0 +1,176 @@
...
+//
+//nolint:gocritic // hugeParam: taken by value to match the driver interface / copy semantics
+func (m *Mock) CreateConnection(_ context.Context, catalogID string, c driver.Connection) error {
+	cat := m.catalogOrDefault(catalogID)
+
+	if !validName(c.Name) {
+		return invalidInput("connection name %q is invalid", c.Name)
+	}
+
+	if !validConnectionType(c.ConnectionType) {
```
NitinKumar004, line 57: Medium (new over-rejection). The allow-list behind `validConnectionType` rejects legitimate SDK-modeled connection types (MYSQL, POSTGRESQL, ORACLE, SALESFORCE, BIGQUERY, …) while accepting `SNOWFLAKE` (not in the `glue@v1.152.0` enum) and `SFTP` (unsupported by CreateConnection). So `create-connection` for a valid type now fails where it previously succeeded. Derive the accepted set from the SDK's `ConnectionType` enum rather than a hand-maintained list.

providers/aws/glue/triggers.go
```diff
@@ -0,0 +1,188 @@
...
+	if !validName(t.Name) {
+		return "", invalidInput("trigger name %q is invalid", t.Name)
+	}
+
+	switch t.Type {
+	case "SCHEDULED":
+		if t.Schedule == "" {
+			return "", invalidInput("a SCHEDULED trigger requires a Schedule")
+		}
+	case "CONDITIONAL", "ON_DEMAND", "EVENT":
```
NitinKumar004: Medium — CONDITIONAL is lumped with ON_DEMAND/EVENT, so a CONDITIONAL trigger with no `Predicate` is accepted. Real Glue requires a Predicate for CONDITIONAL (and a Schedule for SCHEDULED, which you already enforce). Add the CONDITIONAL-needs-Predicate check.

server/aws/glue/registry_ops.go
```diff
@@ -0,0 +1,418 @@
...
+	dispatch(h, w, r, func(h *Handler, ctx context.Context, req *deleteSchemaVersionsRequest) (any, error) {
+		errs, err := h.glue.DeleteSchemaVersions(ctx, req.SchemaID.RegistryName, req.SchemaID.SchemaName,
+			req.Versions)
+		if err != nil {
+			return nil, err
+		}
+
+		out := make([]schemaVersionErrorJSON, 0, len(errs))
+		for i := range errs {
+			out = append(out, schemaVersionErrorJSON{
```
NitinKumar004: Medium — this maps only `ErrorDetails` and never sets `VersionNumber`, but the driver carries the failing number in `BatchError.Values` (registry.go:501). As a result `DeleteSchemaVersions` responses always report `SchemaVersionErrors[].VersionNumber = 0`, so a caller can't tell which version failed. Parse `errs[i].Values[0]` into `VersionNumber` here.

NitinKumar004 (commented):
Re-reviewed after the fix commit. The two blockers are verified resolved, and most of the Mediums are fixed; a few residual fidelity items remain (inline).

**Verified fixed**
- **`GetTableVersion` panic** — guarded two ways: `GetTableVersion` returns `EntityNotFound` when no versions exist, and `DeleteTableVersion` refuses to delete the only remaining version. Re-reproduced the original sequence: no panic.
- **Create-vs-parent-delete orphan** — the new per-scope lock closes every path, including the subtle cross-level case: `DeleteDatabase` takes each child table's scope lock while sweeping its partitions (same `nameKey` that `CreatePartition` locks), so `CreatePartition`-vs-`DeleteDatabase` serializes. 0 orphans across 5000 iterations on all four create paths; no deadlock or lock-order inversion.
- **Schema-version dedup** (identical definition returns the existing version), **14 `json:"...ID"` casing tags** (all fixed, zero remaining, SDK round-trip confirmed), **`StopCrawler`** (`CrawlerNotRunningException`), **`PutResourcePolicy`** (real sha256 hash + `VersionMismatch`/`ConditionCheckFailure`), **`BatchStopJobRun`** (terminal runs under `Errors`), **wrong-entity messages** ("Database not found" first), **dead exceptions** (removed or now emitted), **`DeleteSchemaVersions` range** (`"1-3"` parsed).

**Remaining items to address before this is ready**
- **`ConnectionType` validation over-rejects valid types.** The new allow-list rejects legitimate SDK-modeled types (MYSQL, POSTGRESQL, ORACLE, SALESFORCE, …) while accepting `SNOWFLAKE` (not in the v1.152 SDK enum) and `SFTP` (unsupported by CreateConnection). A new over-rejection — `create-connection` for a valid type now fails. Derive the allow-list from the SDK enum. (inline)
- **`DeleteSchemaVersions` drops the failing `VersionNumber`** in the error response — the driver carries it in `BatchError.Values`, but the handler never maps it, so `SchemaVersionErrors[].VersionNumber` is always 0 and a caller can't tell which version failed. (inline)
- **CONDITIONAL trigger doesn't require a `Predicate`** — real Glue rejects a CONDITIONAL trigger with no predicate. (inline)

**Non-blocking (Low)**
- The writer-vs-writer orphan is closed, but a `GetTables`/`GetDatabases` reader can still observe a partial snapshot mid-cascade (transient, race-clean, no orphan) — `DeleteDatabase`'s comment claiming "a reader can't observe a torn snapshot" overclaims; correct the comment or have readers take the scope lock.
- Schema `BACKWARD`/`FORWARD`/`FULL` don't enforce structural compatibility (documented as out-of-scope; acceptable).
- The SDK round-trip test doesn't cover `RegisterSchemaVersion`/`GetSchemaVersion`/`BatchStopJobRun`, so the casing fixes there have no regression guard — worth adding.
- `StartCrawler` raises `ConcurrentModificationException` where real Glue raises `CrawlerRunningException`.

providers/aws/glue/connections.go
thzgajendra: Fixed. \`validConnectionType\` now mirrors the aws-sdk-go-v2 \`glue.ConnectionType\` enum (v1.152.0) exactly — MYSQL/POSTGRESQL/ORACLE/BIGQUERY and the rest are accepted, and SNOWFLAKE (not in the enum) is dropped. Kept the provider SDK-free (the set is a faithful copy of the enum, not an import). Added \`TestCreateConnectionTypeValidation\` (accepts MYSQL/POSTGRESQL/ORACLE/BIGQUERY/JDBC, rejects SNOWFLAKE) and verified via CLI.

providers/aws/glue/triggers.go
thzgajendra: Fixed. \`CreateTrigger\` now requires a non-empty \`Predicate\` for CONDITIONAL triggers (returns \`InvalidInputException\` otherwise); SCHEDULED still requires a Schedule; ON_DEMAND/EVENT unchanged. Added \`TestCreateTriggerConditionalRequiresPredicate\` + CLI E2E.

server/aws/glue/registry_ops.go
thzgajendra: Fixed. \`deleteSchemaVersions\` now parses \`errs[i].Values[0]\` into \`VersionNumber\`, so \`SchemaVersionErrors[].VersionNumber\` reports the actual failing version instead of 0. Added an SDK-roundtrip test (\`TestSDKDeleteSchemaVersionsReportsFailingVersion\`) asserting VersionNumber==5 for a missing version + CLI E2E.

thzgajendra:
Second round addressed (pushed on the latest commit):

- **ConnectionType allow-list** — now mirrors the SDK `glue.ConnectionType` enum (v1.152.0) exactly: MYSQL/POSTGRESQL/ORACLE/BIGQUERY/… accepted, SNOWFLAKE (not modeled) dropped. Provider stays SDK-free.
- **CONDITIONAL trigger** — `CreateTrigger` now requires a `Predicate` for CONDITIONAL (InvalidInputException otherwise).
- **DeleteSchemaVersions** — response now surfaces the failing `VersionNumber` (parsed from `Values[0]`) instead of 0.

Tests added: `TestCreateConnectionTypeValidation`, `TestCreateTriggerConditionalRequiresPredicate`, `TestDeleteSchemaVersionsReportsFailingVersion`, and SDK-roundtrip `TestSDKDeleteSchemaVersionsReportsFailingVersion`.

Verified: gofmt / vet / build / `-mod=readonly` / `go test -race` / `go mod tidy` (no diff) / golangci-lint `--new-from-rev` 0 / CodeQL security-extended 0-in-glue / CLI E2E on all three.

NitinKumar004 (approved):
Re-reviewed after the latest fix commit. The three remaining items are resolved and test-backed; gate is green (build, docs-generate clean, `go test -race`).

**Verified fixed**
- **`ConnectionType` validation** — the accepted set now mirrors the `aws-sdk-go-v2` glue `ConnectionType` enum exactly (92 values, verified against the SDK enum with zero symmetric difference): the previously over-rejected types (MYSQL, POSTGRESQL, ORACLE, SALESFORCE, BIGQUERY, …) are accepted and the non-modeled `SNOWFLAKE` is removed. Covered by `TestCreateConnectionTypeValidation`.
- **CONDITIONAL trigger** — now requires a `Predicate` (`invalidInput` otherwise); ON_DEMAND/EVENT unchanged. Covered by `TestCreateTriggerConditionalRequiresPredicate`.
- **`DeleteSchemaVersions` error** — the failing version number is now surfaced from `BatchError.Values[0]` into `SchemaVersionErrors[].VersionNumber`. Covered by a provider test and an SDK round-trip test (`TestSDKDeleteSchemaVersionsReportsFailingVersion`), which also closes the earlier round-trip-coverage gap on this path.

This resolves every item raised across the review — the original blockers (the `GetTableVersion` panic, the create-vs-parent-delete orphan race, schema-version dedup, and the `json:"...ID"` casing) plus these three fidelity fixes. Approving.

Remaining are Low, non-blocking follow-ups: `DeleteDatabase`'s comment claiming a reader can't observe a torn snapshot slightly overclaims (a `GetTables` reader can still see a partial mid-cascade list — transient and race-clean); BACKWARD/FORWARD/FULL schema compatibility isn't structurally enforced (documented as out-of-scope); `StartCrawler` raises `ConcurrentModificationException` where real Glue raises `CrawlerRunningException`.

Issue #38, opened 2026-03-19
Add S3 presigned URL support and bucket lifecycle policies


NitinKumar004:
### New features
- **Presigned URLs**: `GeneratePresignedURL(bucket, key, expiry)` — return a mock URL with expiry tracking
- **Lifecycle policies**: `PutBucketLifecycle(bucket, rules)` — auto-expire/transition objects based on age
- **Versioning**: `PutBucketVersioning(bucket, enabled)` — enable object versioning, `GetObjectVersion(bucket, key, versionID)`
- **Multipart upload**: `CreateMultipartUpload`, `UploadPart`, `CompleteMultipartUpload`

### Why
These are frequently used S3 features that production apps depend on. Presigned URLs and lifecycle policies are especially common in file upload and data retention workflows.

### Acceptance Criteria
- [ ] `GeneratePresignedURL` returns a mock URL string with encoded expiry
- [ ] Presigned URL expiry is trackable via `config.FakeClock`
- [ ] Lifecycle rules auto-delete/transition objects when clock advances past age threshold
- [ ] Versioning stores multiple versions per key, `GetObjectVersion` retrieves by versionID
- [ ] `DeleteObject` on versioned bucket creates a delete marker instead of removing
- [ ] Multipart upload assembles parts into final object on `CompleteMultipartUpload`
- [ ] Unit tests with ≥90% coverage for all new methods
- [ ] Passes `golangci-lint run` with zero issues
- [ ] README.md updated with new features

Issue #313, opened 2026-08-02
Docs: make cloudemu's full capability surface discoverable & agent-understandable


NitinKumar004:
## Summary

The docs don't make cloudemu's **full capability surface** easy to discover or understand — a newcomer (or an AI coding agent pointed at the repo) can't quickly answer "what can cloudemu do, what can't it, and how do I use each part?" without reading source. The information exists but is scattered across `docs/*.md`, per-package doc comments, and tribal knowledge. This issue proposes restructuring the docs so **all capabilities are discoverable in one place and unambiguous**, and so an agent can act on them without guessing.

## Why this matters

- **Agents/LLMs** are increasingly the ones reading these docs to wire cloudemu into tests. They need explicit, structured, single-source-of-truth statements — not prose that assumes context.
- The **emulator boundary** is the single most misunderstood thing (e.g. the Kubernetes data plane emulates the API + controllers but does **not run containers** — no real `kubectl logs`/`exec`, no real networking). This "what it is / isn't" needs to be front-and-center per surface, or users hit it as a surprise.
- Capability info is currently spread across `services.md`, `sdk-server.md`, `standalone-server.md`, `architecture.md` and package comments, with no single index.

### 1. A top-level "What cloudemu is / isn't"
A short, explicit section (README + docs home) stating the boundary once, clearly:
- ✅ Emulates cloud **APIs / control surfaces** for fast, deterministic, Docker-free testing (Go API, SDK-compat HTTP server, standalone `serve`, Kubernetes API + built-in controllers).
- ❌ Does **not** run real workloads/containers, does not serve real traffic, does not enforce IAM/quota/policy, does not persist state across restarts, and does not validate credentials/signatures. Name the top non-goals up front.

### 2. A single capability index — generated from source
One page listing **every service and every supported operation**, with a per-entry `works / partial / non-goal` status. **Generate it from the driver interfaces** (e.g. a `go generate` step that walks the drivers and emits `docs/capabilities.md`) so it *can't* drift from the code — hand-maintained tables are the root cause of the contradictions this issue is about. The generator is the durable fix; the table is the artifact.

### 3. A machine-readable capability manifest (for agents)
[...]

PR #388, opened 2026-08-10
Add AWS GuardDuty (Threat Detection) Service Emulation

thzgajendra:
## Objective
Add Amazon GuardDuty to cloudemu at full `aws-sdk-go-v2` wire parity — all **87 operations** — as part of #295 Workstream B. Real `aws-sdk-go-v2/service/guardduty` clients and the `aws guardduty` CLI work against the SDK-compat server by only changing the endpoint.

## What's implemented
GuardDuty is REST-JSON (`awsRestjson1`) with path + HTTP-method routing and **no version prefix**. Four-layer architecture (`services/guardduty/driver` → `providers/aws/guardduty` → `server/aws/guardduty` → registration), matching the OpenSearch service.

- **Detectors** + child resources (IP sets, threat-intel/entity sets, trusted-entity sets, filters) — CRUD, created under the detector scope lock so a concurrent `DeleteDetector` can't orphan them; delete cascades.
- **Members / invitations / administrator-master** relationship, **organization** admin + config.
- **Findings** — sample generation, `FindingCriteria` filtering, sort + pagination, archive/unarchive, statistics, feedback.
- **Publishing destinations, coverage, usage, free-trial**.
- **Malware** protection plans, scans, settings; **tags** (list/tag/untag) resolved by ARN.

Quality: `SetIfAbsent` atomic creates → `ConflictException`; deep-copied reads (race-clean); typed exceptions via `driver.APIError` (`ResourceNotFoundException`/`ConflictException`/`BadRequestException`); deterministic `Clock`/`idgen`.

## Routing: shared `/tags/{ResourceArn}` endpoint
GuardDuty has no version prefix, so its handler gates on known root segments and **registers before the S3 catch-all**. The `/tags/{ResourceArn}` path is also served by EKS (whose `Matches` claims every `/tags` request). To avoid EKS shadowing GuardDuty's tag ops, the GuardDuty handler only claims `/tags` for **GuardDuty ARNs** and registers **before EKS**, so other services' tag requests fall through to their handlers. Covered by a full-handler-stack regression test.

## Docs
- `docs/services.md` — service #38 (Threat Detection), operation families, totals (Grand Total → 2690).
- `docs/architecture.md` — server blurb.
- `docs/coverage/` regenerated via `go generate` (adds `aws/guardduty.md`, updates indexes + `coverage.json`).

## Tests
- Provider unit tests (CRUD, conflict/not-found, `-race` no-alias) and SDK-roundtrip tests using the real `aws-sdk-go-v2/service/guardduty` client (typed errors asserted via `smithy.APIError.ErrorCode()`, since the SDK models zero typed op-errors).
- Regression test tagging a detector under the **full** handler stack.

## Verification (locally)
`gofmt` · `go build ./...` · `go build -mod=readonly ./...` · `go vet` · `go test -race ./...` · `go mod tidy` (no diff) · `golangci-lint --new-from-rev` **0 issues** · CodeQL (CI's `go-code-scanning` suite) **0 findings in the service path** · CLI E2E (create detector → get → ip-set → sample findings → tags → cascade delete).

## Risk & rollback
Additive: a new service package + registration. The only cross-cutting change is registration order (GuardDuty before EKS/S3) and the ARN-scoped tag `Matches`; non-GuardDuty tag requests are unaffected (regression-tested). Revert by dropping the branch.


providers/aws/guardduty/errors.go
```diff
@@ -0,0 +1,31 @@
+package guardduty
+
+import (
+	"github.com/stackshy/cloudemu/v2/errors"
+	"github.com/stackshy/cloudemu/v2/services/guardduty/driver"
+)
+
+// notFound builds a ResourceNotFoundException-tagged error.
+func notFound(format string, args ...any) error {
```
NitinKumar004, line 9: High — the classic surface uses the wrong exception. Verified against guardduty@v1.79.3: all 87 ops declare `BadRequestException`; `ResourceNotFoundException` is modeled by only 4 ops and `ConflictException` by only 2 (the malware-plan/scan/UpdateTrustedEntitySet ops). So returning `notFound`/`conflict` for a missing or duplicate detector/ipset/threat-set/filter/publishing-destination emits an inverted status (404/409 vs real 400) and a type the op doesn't model. `get-detector --detector-id bogus` → real AWS `BadRequestException`; mock `ResourceNotFoundException`. Route the classic-surface lookups/duplicates through the existing `badRequest` helper, keeping `notFound`/`conflict` only for the malware ops that actually model them.

providers/aws/guardduty/tags.go
```diff
@@ -0,0 +1,327 @@
...
+	if len(req.Tags) == 0 {
+		return nil, badRequest("tags is required")
+	}
+
+	acc, err := m.resolveTagTarget(resourceARN)
+	if err != nil {
+		return nil, err
+	}
+
+	current, ok := acc.read()
```
NitinKumar004: Medium — lost update. `acc.read()` copies the tags and releases the lock; the merge below runs outside any lock; `acc.write()` re-locks and replaces the whole map. Two concurrent `TagResource` calls on the same ARN both read the same base and each writes back only its own key → last-writer-wins, the other key is silently lost. Reproduced: 130/200 contended iterations lost writes (`-race`-clean — logical, not a memory race). The `tagAccessor` comment guards against torn maps but not lost updates. Do the merge-and-store inside a single lock hold, like the other mutators.

providers/aws/guardduty/ipsets.go
```diff
@@ -0,0 +1,103 @@
...
+
+	return out
+}
+
+// validateSetInput enforces the required Name/Format/Location fields real
+// GuardDuty rejects with a BadRequestException when absent.
+func validateSetInput(name, format, location string) error {
+	if name == "" {
+		return badRequest("name is required")
+	}
```
NitinKumar004, line 39: Medium — `Format` is accepted as any non-empty string. Real GuardDuty validates it against the `IpSetFormat` enum (TXT/STIX/OTX_CSV/ALIEN_VAULT/PROOF_POINT/FIRE_EYE) and returns `BadRequestException` otherwise. Same for the threat-intel/entity/trusted sets that share this path — `create-ip-set --format BOGUS` currently creates the set and echoes the bogus format.

providers/aws/guardduty/publishing_destinations.go
```diff
@@ -0,0 +1,235 @@
...
+	if uerr := unmarshalBody(body, &req); uerr != nil {
+		return nil, uerr
+	}
+
+	dtype := req.DestinationType
+	if dtype == "" {
+		dtype = destinationTypeS3
+	}
+
+	if req.DestinationProperties.DestinationArn == "" {
```
NitinKumar004, line 79: Medium — for an S3 destination real GuardDuty requires both `destinationArn` and `kmsKeyArn` in `destinationProperties`; this validates only `destinationArn`. `create-publishing-destination --destination-type S3 --destination-properties DestinationArn=...` with no KMS → real AWS `BadRequestException`, mock returns a destinationId.

providers/aws/guardduty/malware_scans.go
```diff
@@ -0,0 +1,379 @@
...
+func (m *Mock) GetMalwareScan(_ context.Context, scanID string) (json.RawMessage, error) {
+	s, ok := m.malwareScans.Get(scanID)
+	if !ok {
+		return nil, notFound("malware scan not found: %s", scanID)
+	}
+
+	out := map[string]any{
+		"scanId":        s.scanID,
+		"resourceArn":   s.resourceARN,
+		"resourceType":  scanResourceTypeBucket,
```
NitinKumar004: Medium — `resourceType` is hardcoded `S3_BUCKET`, but `StartMalwareScan` targets an EC2 instance `resourceArn`. `get-malware-scan` after starting a scan on `arn:aws:ec2:...:instance/i-...` reports `S3_BUCKET` where real AWS reports the instance. (The synchronous already-COMPLETED / NO_THREATS_FOUND terminal state is a defensible simplification if documented — the wrong resourceType is the actual bug.)

providers/aws/guardduty/findings.go
```diff
@@ -0,0 +1,800 @@
...
+	case fieldType:
+		return fd.findingType
+	case "accountId":
+		return fd.accountID
+	case "region":
+		return fd.region
+	case fieldSeverity:
+		return strconv.FormatFloat(fd.severity, 'f', -1, 64)
+	default:
+		return ""
```
NitinKumar004, line 714: Medium — an unsupported criteria field returns "" here, which then fails an `eq` match and drops the finding. So `list-findings --finding-criteria` on any of the many fields not in this switch (e.g. `resource.instanceDetails.instanceId`) silently returns an empty `findingIds` for every finding rather than filtering on that field — a silent over-filter with no error. Either support the field or reject an unmodeled criterion field explicitly.

server/aws/guardduty/handler.go
```diff
@@ -0,0 +1,190 @@
...
+// The /tags/{ResourceArn} endpoint is shared with other services (EKS uses the
+// same shape), so for the "tags" root this only claims requests whose ARN is a
+// GuardDuty ARN; other services' tag requests fall through to their handlers.
+func (*Handler) Matches(r *http.Request) bool {
+	segs := splitPath(escapedPath(r))
+	if len(segs) == 0 {
+		return false
+	}
+
+	switch segs[0] {
```
NitinKumar004, line 84: Medium (documentation / latent trap) — the comment names only EKS, but there are three bare-`/tags` handlers: VPCLattice (registered before GuardDuty), GuardDuty, EKS. It works only because VPCLattice gates on its identifier shape and greedy EKS is registered last. Note VPCLattice here and state the ordering invariant (a greedy `/tags` handler registered before GuardDuty would silently break GD tag ops).

NitinKumar004 (changes requested):
Thorough review of the new GuardDuty service. Strong engineering — and it visibly applies the lessons from the recent Workstream-B reviews. Credit where due: the `/tags` routing carve-out is sound (driven through the full handler stack, GuardDuty's ARN-gated `/tags` claim shadows neither EKS nor S3 in either direction); the create-vs-parent-delete orphan class that bit earlier PRs is structurally defeated by nesting child resources inside the detector struct (0 orphans / 0 races over 400 contended iterations); the single-detector-per-account race is closed by `createMu`; reads are uniformly deep-copied; there is no `json:"...ID"` casing bug; and the wire error plumbing is typed-first with correct statuses, timestamps, and pagination.

Requesting changes on one High plus a set of Mediums (details inline).

## High

**Exception taxonomy is wrong across the classic surface.** The provider returns `ResourceNotFoundException` (404) for a missing detector/ipset/threat-intel/entity set/filter/publishing-destination and `ConflictException` (409) for a duplicate filter name. Verified against `aws-sdk-go-v2/service/guardduty@v1.79.3`: every one of the 87 ops declares `BadRequestException`; `ResourceNotFoundException` is declared by only 4 ops and `ConflictException` by only 2 (the malware-protection-plan / scan / UpdateTrustedEntitySet ops — where the mock is correctly using notFound/conflict). So for the classic surface the emulator emits an inverted status and a type the op doesn't model: `aws guardduty get-detector --detector-id bogus` → real AWS `BadRequestException` (400) "input detectorId is not owned by the current account"; mock → `ResourceNotFoundException` (404). A client written to real GuardDuty (catching BadRequest / branching on 400) misses it. The `badRequest` helper already exists with the right doc comment — route the classic-surface lookups/duplicates through it, keeping notFound/conflict only for the malware ops that actually model them.

## Medium
- **`TagResource`/`UntagResource` lost-update.** Unlike every other mutator, tag writes split the read-modify-write across two lock acquisitions (`acc.read()` copies + releases, merge happens outside any lock, `acc.write()` re-locks and replaces). Two concurrent tags on the same ARN → last-writer-wins, first key silently lost. Reproduced: 130/200 contended iterations lost writes. `-race`-clean (a logical atomicity gap, not a memory race). The `tagAccessor` doc comment claiming the pair "runs under the owning resource's lock" prevents torn maps but not lost updates. Fix: merge-and-store under a single lock hold.
- **Missing input validation** real GuardDuty enforces: IP-set / threat-set `Format` isn't checked against the enum (TXT/STIX/OTX_CSV/ALIEN_VAULT/PROOF_POINT/FIRE_EYE), and `CreatePublishingDestination` doesn't require `kmsKeyArn` for an S3 destination — both accepted where real AWS returns `BadRequestException`.
- **Malware scan infidelity.** `StartMalwareScan` targets an EC2 instance, but `GetMalwareScan`/`ListMalwareScans` hardcode `resourceType: S3_BUCKET`, and the scan is minted already `COMPLETED`/`NO_THREATS_FOUND`. The synchronous-terminal simplification is defensible if documented; the `S3_BUCKET` label is simply wrong for an instance target.
- **Filters are inert.** A filter with `Action=ARCHIVE` never auto-archives matching findings — `Filter.FindingCriteria` is stored verbatim and `ListFindings` only evaluates the request's criteria, never stored filters.
- **`FindingCriteria` on an unsupported field silently drops all findings.** Only ~9 fields are modeled; any other field yields an empty value that fails an `eq`, so `list-findings` returns an empty `findingIds` list for every finding rather than filtering on that field (silent over-filter, no error).
- **Generated docs stale (#313/#382).** `docs/coverage/aws/guardduty.md` still carries `(phase N)` scaffolding that the driver interface comments have since dropped — `go generate ./...` wasn't re-run after the comment cleanup. Regenerate and commit.
- **SDK round-trip coverage.** 48/87 ops are exercised through the real client; all 15 entity-set ops (ThreatIntelSet / ThreatEntitySet / TrustedEntitySet CRUD) are untested at the wire layer — and those carry epoch-seconds `createdAt`/`updatedAt`, exactly the field/timestamp shape the round-trip is meant to guard (the #387 risk zone). Add at least one round-trip per entity-set family.
- **`/tags` co-tenancy is under-documented.** There are three bare-`/tags` handlers (VPCLattice, GuardDuty, EKS), not two; it works only because VPCLattice gates on its identifier shape and greedy EKS is registered last. The comment names only EKS — note VPCLattice and state the ordering invariant (greedy EKS must remain last of the `/tags` group), or a future greedy `/tags` handler registered before GuardDuty silently breaks GD tag ops.

## Low
Set-status lifecycle collapsed to ACTIVE (no ACTIVATING/ERROR/DELETING); invite→accept only reachable via a test-only seam (cross-account not emulated) and `REMOVED` status never produced; tag count/length limits unenforced; `GetMemberDetectors`/`SendObjectMalwareScan`/`GetOrganizationStatistics` are near-empty stubs; dead `AccessDeniedException`; a canonical `PermissionDenied` would surface as 500 not 403; unrouted-path/bad-verb responses are tagged `BadRequestException` with 404/405; bogus `POST /malware-scan` alias.

## On "full parity"
Wire/routing/shape parity is fair; behavioral parity is not (the H1 taxonomy, inert filters, and ~9 synthesized/stub ops). The synthesized coverage/usage/free-trial ops are honestly commented in code — the PR narrative should just avoid implying they're derived from real state.

providers/aws/guardduty/errors.go
thzgajendra: Fixed. Verified against guardduty@v1.79.3: all 87 ops model `BadRequestException`; only 4 malware ops model `ResourceNotFoundException` (Get/Update/DeleteMalwareProtectionPlan, GetMalwareScan) and 2 model `ConflictException` (CreateMalwareProtectionPlan, StartMalwareScan). Routed the entire classic surface (detector/ipset/threat-intel/entity/trusted sets/filter/publishing/tags/members/findings/org) through `badRequest`; kept `notFound`/`conflict` only in the malware ops that actually model them. `get-detector --detector-id bogus` now returns `BadRequestException`. Updated the tests accordingly (and added a full-stack SDK assertion).

providers/aws/guardduty/tags.go
thzgajendra: Fixed. Reworked `tagAccessor` to expose an atomic `update(mutate)` that does read-modify-write inside a single lock hold (memstore `Update` for the plan, the detector lock for detector/child). `TagResource`/`UntagResource` now merge inside that hold, so concurrent tags can no longer lose each other. Added `TestTagResourceConcurrentNoLostUpdate` (50 goroutines, each a distinct key, asserts all survive) — passes under `-race`.

providers/aws/guardduty/ipsets.go
thzgajendra: Fixed. `validateSetInput` now validates `Format` against the modeled enum (TXT/STIX/OTX_CSV/ALIEN_VAULT/PROOF_POINT/FIRE_EYE) and returns `BadRequestException` otherwise. Since IPSet/ThreatIntelSet/ThreatEntitySet/TrustedEntitySet share `validateSetInput`, all four are covered. Added `TestIPSetFormatEnumValidation`.

providers/aws/guardduty/publishing_destinations.go
thzgajendra: Fixed. `CreatePublishingDestination` now requires `kmsKeyArn` for an S3 destination (returns `BadRequestException` when absent). Added `TestPublishingDestinationRequiresKMS` and updated the lifecycle tests to supply a KMS key.

providers/aws/guardduty/malware_scans.go
thzgajendra: Fixed. Added `deriveScanResourceType(arn)` and stored the derived type on the scan at start; `GetMalwareScan`/`ListMalwareScans`/`DescribeMalwareScans` now report it. `arn:aws:ec2:...:instance/i-...` → `EC2_INSTANCE`, `...:volume/...` → `EBS_VOLUME`, `arn:aws:s3:::...` → `S3_BUCKET`, unknown → `EC2_INSTANCE` (on-demand scans target EC2 by default). Added `TestMalwareScanResourceTypeFromARN`.

providers/aws/guardduty/findings.go
thzgajendra: Fixed. `parseCriteria` now rejects any `findingCriteria` field the emulator does not model with `BadRequestException` (e.g. `resource.instanceDetails.instanceId`), so a client gets an explicit error instead of a silently over-filtered empty result. Supported fields: id, type, accountId, region, severity, confidence, updatedAt, createdAt, service.archived. Added `TestListFindingsRejectsUnsupportedCriteria`.

server/aws/guardduty/handler.go
thzgajendra: Fixed. Updated the `Matches` doc-comment to name all three bare-`/tags` handlers (VPCLattice registered before GuardDuty, GuardDuty, greedy EKS last) and to state the ordering invariant: GuardDuty must register before any handler that claims every `/tags` request; a greedy `/tags` handler registered before GuardDuty would silently swallow GD tag ops. GuardDuty only claims `/tags` for GuardDuty ARNs, so the others fall through.

thzgajendra:
Thanks for the thorough review @NitinKumar004 — all 7 addressed (pushed on the latest commit).

**High**
- **Error-type inversion** — verified against `guardduty@v1.79.3`: all 87 ops model `BadRequestException`; only 4 malware ops model `ResourceNotFoundException` and 2 model `ConflictException`. Routed the whole classic surface through `badRequest`; kept `notFound`/`conflict` only in those malware ops.

**Medium**
- **Tag lost update** — `TagResource`/`UntagResource` now merge-and-store under a single lock (atomic read-modify-write); regression test with 50 concurrent distinct-key writers under `-race`.
- **Set `Format`** — validated against the modeled enum (all four set types share the check).
- **S3 publishing destination** — now requires `kmsKeyArn`.
- **Malware scan `resourceType`** — derived from the resource ARN (EC2_INSTANCE / EBS_VOLUME / S3_BUCKET) instead of hardcoded.
- **Unmodeled `findingCriteria` field** — now rejected with `BadRequestException` instead of silently over-filtering to empty.
- **`/tags` routing doc** — names all three handlers (VPCLattice/GuardDuty/EKS) and states the registration-order invariant.

Each fix has a regression test. Verified: gofmt / vet / build / `-mod=readonly` / `go test -race` / `go mod tidy` (no diff) / golangci-lint `--new-from-rev` **0** / CodeQL (CI suite) **0 in the service path** / CLI E2E. Coverage docs regenerated.

NitinKumar004 (approved):
Re-reviewed after the fix commit. The High and every correctness Medium are resolved and verified; gate is green (build, `go test -race -count=2` on provider + server), and `docs/coverage/aws/guardduty.md` has been regenerated.

**Verified fixed**
- **Exception taxonomy** — the classic surface (detector, ipset, threat-intel/entity/trusted sets, filter incl. duplicate-name, publishing destination) now returns `BadRequestException`, matching the SDK op models where every op declares it; the ops that genuinely model `ResourceNotFoundException` (GetMalwareProtectionPlan/Update/Delete, GetMalwareScan) and `ConflictException` (CreateMalwareProtectionPlan, StartMalwareScan) correctly keep those. Both directions are wire-tested.
- **`TagResource`/`UntagResource` lost-update** — the read-modify-write now runs under a single lock hold (`update`/`updateChildTags`), with the merge/delete inside the locked callback. Reproduced: 0/500 losses across concurrent distinct-key tags on both detector and child ARNs (was ~65%). No re-entrancy or aliasing regression; `ListTagsForResource` still returns a copy.
- **Format validation** — set `Format` is validated against the SDK enum (TXT/STIX/OTX_CSV/ALIEN_VAULT/PROOF_POINT/FIRE_EYE), no over-rejection.
- **`CreatePublishingDestination`** now requires `kmsKeyArn` for an S3 destination.
- **Malware scan `resourceType`** is derived from the target ARN (S3_BUCKET/EC2_INSTANCE/EBS_VOLUME) instead of a hardcoded S3_BUCKET.
- **`FindingCriteria` on an unsupported field** now returns `BadRequestException` instead of silently dropping every finding.
- **`/tags` co-tenancy comment** now names VPCLattice (the third bare-`/tags` handler) and states the ordering invariant.
- **Generated docs** regenerated (`guardduty.md`).

The routing carve-out (no cross-service `/tags` shadowing), the nested-child design that makes the parent-delete orphan class structurally impossible, atomic creates, and deep-copy reads all remain intact. Approving.

Non-blocking follow-ups:
- A filter with `Action=ARCHIVE` does not auto-archive matching findings (`filters.go` returns a no-op action). Given findings originate only from `CreateSampleFindings`, this is reasonable to leave unimplemented — but it should be documented as a limitation so the parity wording doesn't imply otherwise.
- No SDK round-trip test covers the ThreatIntelSet/ThreatEntitySet/TrustedEntitySet families; their epoch-seconds `createdAt`/`updatedAt` encoding is correct (verified by hand), so this is a missing regression guard rather than a defect — worth adding one round-trip per family.
- Minor: `AccessDeniedException` is declared/mapped but never emitted; no tag count/length limits enforced.

Issue #39, opened 2026-03-19
Add Blob Storage SAS tokens and lifecycle management


NitinKumar004:
### New features
- **SAS tokens**: `GenerateSASToken(container, blob, permissions, expiry)` — mock shared access signature
- **Lifecycle management**: `SetLifecyclePolicy(container, rules)` — auto-tier/delete blobs based on age
- **Blob versioning**: Enable versioning on containers, retrieve specific versions
- **Blob snapshots**: `CreateSnapshot(container, blob)` — create point-in-time snapshot

### Why
SAS tokens and lifecycle policies are core Azure Blob features used in almost every production workload.

### Acceptance Criteria
- [ ] `GenerateSASToken` returns a mock token string with permissions and expiry
- [ ] SAS token expiry is trackable via `config.FakeClock`
- [ ] Lifecycle rules auto-delete/tier blobs when clock advances past age threshold
- [ ] Versioning stores multiple versions per blob
- [ ] `CreateSnapshot` creates an immutable copy at current point in time
- [ ] Unit tests with ≥90% coverage for all new methods
- [ ] Passes `golangci-lint run` with zero issues
- [ ] README.md updated with new features

PR #389, opened 2026-08-11
Add AWS MSK (Managed Streaming for Kafka) Service Emulation

thzgajendra:
## Objective
Add Amazon MSK to cloudemu at full `aws-sdk-go-v2` wire parity — all **59 operations** — as part of #295 Workstream B. Real `aws-sdk-go-v2/service/kafka` clients and the `aws kafka` CLI work against the SDK-compat server by only changing the endpoint.

## What's implemented
MSK is REST-JSON (`awsRestjson1`) with path + HTTP-method routing under three version prefixes (`/v1/`, `/api/v2/`, `/replication/v1/`). Four-layer architecture (`services/kafka/driver` → `providers/aws/kafka` → `server/aws/kafka` → registration), mirroring the OpenSearch service.

- **Clusters (v1 + v2)** — `CreateCluster`/`CreateClusterV2` claim the name atomically (`ConflictException` on dup) and return an immediately-`ACTIVE` cluster; the v1 and v2 shapes render the same underlying cluster (a v1-created cluster is describable via `DescribeClusterV2`). `GetBootstrapBrokers`, `ListNodes`.
- **Cluster mutations** — broker count/storage/type, storage, configuration, kafka-version, connectivity, monitoring, security, rebalancing, reboot. Each validates the optimistic-concurrency `CurrentVersion` (mismatch → `BadRequestException`), applies the change so a subsequent Describe reflects it, records a `ClusterOperation`, and bumps the version.
- **Cluster operations** (v1 + v2), **configurations** + revisions, **VPC connections** (+ client-vpc-connections / reject), **topics** (+ partitions), **SCRAM secrets** (batch associate/disassociate with per-item unprocessed results), **cluster policy** (optimistic version), **replicators**, **kafka-versions / compatible-versions**, and **tags**.

Quality: `SetIfAbsent` atomic creates; deep-copied reads (race-clean); typed exceptions via `driver.APIError` (`NotFoundException`/`ConflictException`/`BadRequestException`); tags merged under a single lock (no lost updates); deterministic `Clock`/`idgen`.

## Routing
Paths live under `/v1/`, `/api/v2/`, `/replication/v1/`; `Matches` gates on the prefix + a known root segment and **registers before the S3 catch-all**. Disjoint from S3 (a bucket literally named `v1`/`api`/`replication` would be shadowed — documented limitation).

## Coverage generator fix (incidental)
While regenerating coverage docs I found `internal/coveragegen` was **non-deterministic**: when two driver interfaces tie on reference count (databricks has both `DataPlane` and `Databricks`), the winner was chosen by random map-iteration order, so the databricks coverage page flip-flopped between runs/PRs. Fixed the tie-break to sort by name. Databricks now stably resolves to `DataPlane` (a one-time, now-permanent change), so future service PRs won't churn it.

## Docs
`docs/services.md` (service #39, operation families, Grand Total → 2749), `docs/architecture.md`, and regenerated `docs/coverage/` (adds `aws/kafka.md`).

## Tests
Provider unit tests (CRUD, conflict/not-found, version-mismatch, `-race` no-alias, concurrent-tag no-lost-update) and SDK-roundtrip tests using the real `aws-sdk-go-v2/service/kafka` client.

## Verification (locally)
gofmt · `go build ./...` · `go build -mod=readonly ./...` · `go vet` · `go test -race ./...` (incl. `internal/coveragegen`) · `go mod tidy` (no diff) · golangci-lint `--new-from-rev` **0 issues** · CodeQL (CI's `go-code-scanning` suite) **0 findings** in the service path · CLI E2E (create cluster → describe v1/v2 → dup→Conflict → update-broker-count records an operation and grows 3→6 brokers → nodes → config → tags → missing→NotFound).

## Risk & rollback
Additive: a new service package + registration (before S3). The only shared-code change is the coveragegen tie-break determinism fix (covered by its existing test). Revert by dropping the branch.


providers/aws/kafka/tags.go
```diff
@@ -0,0 +1,68 @@
+package kafka
+
+import (
+	"context"
+)
+
+// Tagging resolves a ResourceArn to its owning resource and reads/merges/removes
+// that resource's Tags map. Among the resources implemented so far, only the
```
NitinKumar004: Medium — this comment is factually wrong and the behavior is a bug. VPC connections and replicators ARE stored (`Mock.vpcConns`/`Mock.replicators`), and `CreateReplicator`/`CreateVpcConnection` both accept `Tags`. Because `TagResource`/`ListTagsForResource`/`UntagResource` resolve only via `getCluster`, `create-replicator --tags k=v` then `list-tags-for-resource <replicatorArn>` returns `NotFoundException` — real MSK returns `{k:v}`. Real MSK tags clusters, configurations, VPC connections, and replicators. Route tag ops by ARN resource-kind and correct the comment.

providers/aws/kafka/cluster_updates.go
```diff
@@ -0,0 +1,202 @@
...
+	opTypeUpdateSecurity      = "UPDATE_SECURITY"
+	opTypeUpdateRebalancing   = "UPDATE_REBALANCING"
+	opTypeRebootBroker        = "REBOOT_NODE"
+)
+
+// UpdateBrokerCount sets the cluster's broker count and records an operation.
+func (m *Mock) UpdateBrokerCount(
+	_ context.Context, arn, currentVersion string, target int32,
+) (*driver.ClusterOperation, error) {
+	return m.mutateClusterBR(arn, currentVersion, opTypeUpdateBrokerCount, func(c *driver.Cluster) {
```
NitinKumar004, line 51: Medium — `target` is written verbatim with no constraint. Real MSK requires `TargetNumberOfBrokerNodes` to be strictly greater than the current count and an integer multiple of the AZ/subnet count. `update-broker-count --target-number-of-broker-nodes 1` on a 3-broker cluster succeeds here and Describe shows 1 broker; real MSK → `BadRequestException` (can only increase, by AZ multiples).

providers/aws/kafka/clusters.go
```diff
@@ -0,0 +1,246 @@
...
+//
+//nolint:gocritic // hugeParam: signature fixed by driver.Kafka (by-value input).
+func (m *Mock) CreateCluster(_ context.Context, in driver.CreateClusterInput) (*driver.Cluster, error) {
+	if err := validateClusterName(in.ClusterName); err != nil {
+		return nil, err
+	}
+
+	if in.BrokerNodeGroupInfo == nil {
+		return nil, badRequest("brokerNodeGroupInfo is required")
+	}
```
NitinKumar004, line 169: Medium — no instance-type or EBS storage validation. `InstanceType` is accepted free-form (real MSK requires a `kafka.*` type) and the EBS volume size is unbounded (real MSK: 1–16384 GiB, min 1). `create-cluster` with `InstanceType=t2.nano` or an out-of-range volume creates an ACTIVE cluster where real MSK returns `BadRequestException`.

providers/aws/kafka/clusters.go
```diff
@@ -0,0 +1,246 @@
...
+	start, end, nextTok, err := m.paginate(len(all), page)
+	if err != nil {
+		return nil, "", err
+	}
+
+	return all[start:end], nextTok, nil
+}
+
+// DeleteCluster removes a cluster and frees its name for reuse.
+func (m *Mock) DeleteCluster(_ context.Context, arn, _ string) (arnOut, state string, err error) {
```
NitinKumar004: Medium — the `currentVersion` argument is discarded (`_`), so `delete-cluster --current-version <stale>` deletes anyway. Real MSK honors the optimistic-concurrency version on delete and returns `BadRequestException` on a mismatch. Same for `DeleteReplicator`.

providers/aws/kafka/vpc_connections.go
```diff
@@ -0,0 +1,243 @@
...
+// Describe reflects them.
+func (m *Mock) CreateVpcConnection(_ context.Context, body json.RawMessage) (*driver.VpcConnection, error) {
+	var req createVpcConnectionRequest
+	if len(body) > 0 {
+		if err := json.Unmarshal(body, &req); err != nil {
+			return nil, badRequest("invalid CreateVpcConnection body: %v", err)
+		}
+	}
+
+	if req.TargetClusterArn == "" {
```
NitinKumar004, line 50: Medium — only checks `TargetClusterArn != ""`, never that the cluster exists. `create-vpc-connection --target-cluster-arn arn:aws:kafka:...:cluster/ghost/uuid` returns an `AVAILABLE` connection; real MSK → NotFound/BadRequest (the cluster must exist and be PROVISIONED).

providers/aws/kafka/operations.go
```diff
@@ -0,0 +1,172 @@
...
+	return op
+}
+
+// mutateCluster resolves a cluster, verifies the optimistic-concurrency
+// currentVersion, applies fn under the cluster write-lock, records an operation
+// of opType, and returns a copy of that operation. An empty currentVersion
+// skips the version check (matching ops like RebootBroker that take none). A
+// missing cluster returns NotFoundException — use it for update ops that model
+// NotFoundException.
+func (m *Mock) mutateCluster(
```
NitinKumar004, line 52: Medium (lifecycle fidelity) — because every mutation completes synchronously and the cluster stays ACTIVE, it never enters `UPDATING`, so the `UPDATING`/`CREATING` constants are dead and real MSK's rule "reject a mutation while the cluster is UPDATING/CREATING/DELETING" can't be reproduced (two back-to-back `update-broker-count` both succeed here; real MSK rejects the second). Combined with the synchronous delete hiding `DELETING`, this is the behavioral cost of the immediate-ACTIVE model — please document it explicitly as a limitation beyond the create path.

NitinKumar004 (changes requested):
Thorough review of the new MSK/Kafka service. This is a high-quality PR that visibly internalized the lessons from the recent Workstream-B reviews — credit up front:

- **Concurrency is airtight.** All three sibling-PR bug classes are structurally prevented: account-level creates are atomic (create-mutex / `SetIfAbsent`), per-cluster children are nested inside `clusterData` (orphan-proof), and tags are a single-lock read-modify-write. The MSK-specific optimistic-concurrency `CurrentVersion` check→apply→bump is atomic under one lock — verified with a probe (16 concurrent `UpdateBrokerCount` → exactly one wins, the rest get a version-mismatch `BadRequestException`). No races, no aliasing.
- **Error taxonomy is deliberately per-op-correct.** The paired `getCluster`/`getClusterBR` + `mutateCluster`/`mutateClusterBR` helpers make the wire exception match exactly what each op models in `kafka@v1.49.2` (NotFound on the 40 ops that model it, BadRequest on the ~11 that don't, Conflict on the 5 create ops that model it). This is the dimension that produced a defect on a sibling service; here it's handled correctly across all 59 ops.
- **No invented ops** (all 59 are real SDK operations, including topic management), **no `json:"...ID"` casing bug**, v1/v2 dual-shape is SDK-verified, routing is disjoint from S3, and the incidental `internal/coveragegen` determinism fix is correct and deterministic (generate-twice is byte-identical; scoped to databricks).

Requesting changes on a cluster of Medium fidelity/validation issues, led by one functional bug (details inline).

## Medium
- **Tag ops are broken for other taggable resources, and the comment is factually wrong.** `TagResource`/`ListTagsForResource`/`UntagResource` resolve only clusters, and the doc comment claims "VPC connections and replicators are not yet stored" — but both are stored (`Mock.vpcConns`/`Mock.replicators`) and `CreateReplicator`/`CreateVpcConnection` both accept `Tags`. So `aws kafka create-replicator --tags k=v` then `list-tags-for-resource <replicatorArn>` → `NotFoundException`; real MSK returns `{k:v}`. Real MSK tags clusters, configurations, VPC connections, and replicators. Route tag ops by ARN resource-kind and fix the comment.
- **`UpdateBrokerCount` accepts a decrease and non-AZ-multiple targets.** The target is written verbatim. Real MSK requires `TargetNumberOfBrokerNodes` to be strictly greater than current and an integer multiple of the AZ/subnet count. `update-broker-count --target-number-of-broker-nodes 1` on a 3-broker cluster succeeds here; real MSK → `BadRequestException`.
- **No broker instance-type or EBS storage validation on create.** `InstanceType` is free-form (real MSK requires `kafka.*`) and volume size is unbounded (real: 1–16384 GiB). A `t2.nano` / out-of-range-storage cluster is created ACTIVE here where real MSK returns `BadRequestException`.
- **`DeleteCluster`/`DeleteReplicator` ignore the optimistic-concurrency `currentVersion`** (the parameter is discarded). Real MSK rejects a stale version on delete with `BadRequestException`.
- **`CreateVpcConnection` doesn't verify the target cluster exists** — returns an `AVAILABLE` connection for a ghost cluster ARN; real MSK → NotFound/BadRequest.
- **Async lifecycle collapsed beyond the disclosed create case.** Delete is synchronous, so a deleting resource is immediately un-Describable (real MSK returns `State=DELETING` during teardown), and mutations never enter `UPDATING` — the `UPDATING`/`CREATING` state constants are dead, so the real-MSK rule "reject a mutation while the cluster is UPDATING" can't be reproduced. Worth documenting these as explicit limitations.
- **SDK round-trip coverage is 44/59.** The v2 cluster-operation wire shape (`ListClusterOperationsV2`/`DescribeClusterOperationV2`) and the two BadRequest update ops aren't exercised through the real client — the shapes are manually correct, so this is a missing regression guard rather than a bug. Add a round-trip after an update op.

## Low
Verb not gated on cluster sub-resources (a `GET`/`DELETE` on `/v1/clusters/{arn}/monitoring` invokes `updateMonitoring` and mutates — zero blast radius for the real CLI which always `PUT`s, but hand-crafted requests mutate on a read verb); 4 dead exception constants (`ExUnauthorized`/`ExForbidden`/`ExServiceUnavailable`/`ExTooManyRequests`); VPC-connection and replicator state enums reduced to two states each; SCRAM association only prefix-validates the secret ARN; `GetClusterPolicy` returns an empty policy where real MSK returns `NotFoundException` when none is attached (please confirm against live MSK); `UpdateConfiguration` doesn't require `serverProperties` though create does; integer pagination token (round-trips fine); the coveragegen tie-break picks alphabetically-first among equally-referenced interfaces (deterministic, cosmetic).

## On "full parity"
Wire/operation parity holds (all 59 real, taxonomy correct, v1/v2 shapes right); behavioral parity is overstated. Recommend softening the wording to "full wire/operation parity; behavioral fidelity best-effort with an immediate-ACTIVE lifecycle.

providers/aws/kafka/tags.go
thzgajendra: Fixed. Tag ops now route by the ARN resource-kind segment to all four taggable MSK resources — clusters, configurations, VPC connections, replicators — via a `tagTarget` resolver doing atomic read-modify-write under the owning lock. Added `driver.Configuration.Tags` (server-side; the SDK Configuration shape has no inline tags, so DescribeConfiguration does not render them). `create-replicator --tags k=v` then `list-tags-for-resource <replicatorArn>` now returns `{k:v}`. Corrected the comment. Unknown ARN → NotFoundException. Added `TestTagResourceAcrossResourceKinds` + E2E on a replicator ARN.

providers/aws/kafka/cluster_updates.go
thzgajendra: Fixed. `UpdateBrokerCount` now rejects a target that is not strictly greater than the current count and not a multiple of the AZ count (derived from the client-subnet count), returning `BadRequestException`. A decrease or non-multiple is a 400; a valid increase to a multiple succeeds. Added `TestUpdateBrokerCountConstraints` + E2E.

providers/aws/kafka/clusters.go
thzgajendra: Fixed. `CreateCluster`/`CreateClusterV2` now validate `InstanceType` is a `kafka.*` type and the EBS volume size (parsed from the `storageInfo` block) is within 1–16384 GiB, returning `BadRequestException` otherwise. Added `TestCreateClusterBrokerValidation` + E2E (`t2.nano` and an out-of-range volume both 400).

providers/aws/kafka/clusters.go
thzgajendra: Fixed. `DeleteCluster` and `DeleteReplicator` now honor the optimistic-concurrency `currentVersion` — a stale value returns `BadRequestException` and the resource survives. Added `TestDeleteClusterHonorsCurrentVersion` + E2E (`delete-cluster --current-version STALE` → 400, cluster stays ACTIVE).

providers/aws/kafka/vpc_connections.go
thzgajendra: Fixed. `CreateVpcConnection` now resolves `TargetClusterArn` and returns `BadRequestException` if the cluster does not exist or is not a provisioned cluster (VPC connections are unsupported for serverless). Added `TestCreateVpcConnectionRequiresExistingCluster` + verified via wire (`{"__type":"BadRequestException",...cluster not found...}`).

providers/aws/kafka/operations.go
thzgajendra: Documented. Added an explicit LIMITATION note on the `ClusterState*` constants (services/kafka/driver/types.go) and in the MSK services.md section: clusters provision immediately ACTIVE and mutations apply synchronously, so a cluster never passes through the transient CREATING/UPDATING states, delete removes it rather than leaving it DELETING, and real MSK’s "reject a mutation while CREATING/UPDATING/DELETING" rule is not reproduced (back-to-back updates all succeed). Kept the CREATING/UPDATING/FAILED constants to document the real state enum; optimistic-concurrency (incl. on delete) is still enforced.

thzgajendra:
Thanks for the thorough review @NitinKumar004 — all 6 addressed (pushed on the latest commit).

- **Tag routing** — tag ops now resolve the ARN's resource-kind and cover all four taggable MSK resources (clusters, configurations, VPC connections, replicators), atomic read-modify-write. Added `Configuration.Tags` server-side.
- **UpdateBrokerCount** — increase-only + a multiple of the AZ count, else `BadRequestException`.
- **Broker validation** — `kafka.*` instance type + EBS volume size 1–16384 GiB.
- **VPC connection** — target cluster must exist and be provisioned.
- **Delete version** — `DeleteCluster`/`DeleteReplicator` honor `currentVersion`.
- **Synchronous lifecycle** — documented explicitly (no CREATING/UPDATING transient states; concurrent-mutation rejection not modeled) on the state constants and in services.md.

Each fix has a regression test. Verified: gofmt / vet / build / `-mod=readonly` / `go test -race` / `go mod tidy` (no diff) / golangci-lint `--new-from-rev` **0** / CodeQL (CI suite) **0 in the service path** / CLI + wire E2E on all five behavioral fixes.

NitinKumar004 (approved):
Re-reviewed after the fix commit. Every finding from the previous review is resolved and verified; gate is green (build, `go test -race -count=2` on provider + server), docs-generate clean.

**Verified fixed**
- **Tag routing** — `TagResource`/`ListTagsForResource`/`UntagResource` now resolve all four taggable kinds (cluster, configuration, VPC connection, replicator) via `arnResourceKind`/`resolveTagTarget`, the stale comment is corrected, and each kind's update is a single-lock read-modify-write. Reproduced under `-race`: 8 concurrent distinct-key tags on a replicator ARN and on a configuration ARN both retain all 8 keys. ARN parsing is guarded (malformed/short ARNs return NotFound, no panic).
- **`UpdateBrokerCount`** — now rejects a target ≤ current and a target that isn't a multiple of the AZ count (`BadRequestException`); over-rejection-safe (unknown/serverless AZ count skips the multiple check, a valid 3→6 on 3 AZs still succeeds).
- **Broker validation on create** — non-`kafka.*` instance types and EBS sizes outside 1–16384 GiB are rejected, on both `CreateCluster` and `CreateClusterV2` provisioned (serverless correctly skips it); an unset size is not rejected.
- **`DeleteCluster`/`DeleteReplicator`** — now honor the optimistic-concurrency `currentVersion` (stale → `BadRequestException`, empty → allowed); the BadRequest choice matches the SDK op models.
- **`CreateVpcConnection`** — requires the target cluster to exist (`BadRequestException`, matching the SDK op that models BadRequest but not NotFound).
- **Lifecycle limitations** — the immediate-ACTIVE model, synchronous delete, and absence of a transient UPDATING/CREATING state are now documented on the state constants and in `services.md`.

The error taxonomy remains per-op-correct against the SDK, concurrency is intact through the tag rework, and the incidental coveragegen determinism fix stands. Approving.

Follow-ups (non-blocking):
- The `kafka.*`-only instance-type gate (`clusters.go`) rejects valid **MSK Express** broker types (`express.m7g.*`) that the pinned SDK supports — a real Express-cluster create would get a spurious `BadRequestException`. Broaden the allow-list to accept an `express.` prefix as well.
- `DeleteReplicator`'s `currentVersion` path has no dedicated test (the code is correct; only `DeleteCluster` is covered).
- Minor: an empty `instanceType` is accepted though the comment says real MSK requires one.

PR #390, opened 2026-08-11
fix(k8s): server-side apply, apply-create, and finalizers for typed core kinds (#330)

### LINKED ISSUE #330: k8s: server-side apply + finalizers not honored for typed core kinds; apply-create 404s
Surfaced by a real-kubectl E2E while reviewing #314. Pre-existing (not introduced by #314), so kept out of that PR.

**Goal:** `kubectl apply --server-side` and finalizer-gated delete should behave for the common typed kinds the way they already do for registry-backed kinds.

**Today**
- SSA field ownership (`managedFields`, conflict/force, field-removal, Update entries) only runs via the registry path (`registryPatch` → `serverSideApply`). Typed handlers (ConfigMap, Secret, Service, Deployment) do a plain merge with no ownership.
- Finalizers are honored only on registry kinds + typed Namespace/Pod; typed ConfigMap/Secret/Service/Deployment are hard-deleted despite `metadata.finalizers`.
- `kubectl apply --server-side` of a new registry object returns 404 (create-on-apply not wired).

**Ops**
- Route typed handlers through `serverSideApply` for the apply-patch content-type.
- Honor finalizers on typed ConfigMap/Secret/Service/Deployment delete.
- Wire apply-create so an apply to a missing object creates it instead of 404.

NitinKumar004:
Fixes #330.

## Problem

`kubectl apply --server-side` and finalizer-gated deletes worked for registry-backed kinds but not for the typed core kinds (ConfigMap, Secret, Service, Deployment):

- **apply-patch was a dumb merge** — the typed handlers ran `application/apply-patch+yaml` through a plain RFC-7396 merge, so no field ownership (`managedFields`) and no conflict detection.
- **apply to a missing object 404'd** instead of creating it, so `kubectl apply --server-side -f x.yaml` failed on a fresh cluster.
- **typed deletes ignored finalizers** — an object with `metadata.finalizers` was hard-deleted instead of going Terminating.

## Solution

The machinery already existed for registry kinds; this routes the typed handlers into it.

- **Server-side apply.** `serverSideApply` is decoupled from `registryStore` (takes an `apiVersion` string); a `serverSideApplyTyped` bridge round-trips a typed struct through it. The four typed PATCH handlers now branch on content-type exactly like `registryPatch` — apply-patch → real SSA (ownership + 409 conflicts), everything else → the existing merge.
- **Apply-create.** Apply to a missing object now creates it — typed kinds via a shared `createXLocked` core (so per-kind defaulting/reconcile is preserved, e.g. a Deployment materializes Pods), registry kinds via `registryApplyCreateLocked`.
- **Finalizers.** The four typed delete handlers call the existing `markForDeletion` (object goes Terminating), and a patch that drains the last finalizer completes the delete — matching Pod / Namespace / registry.

No new abstractions and no changes outside `services/kubernetes/`.

## Verification

**Unit** — repro-first tests for all three gaps; full `services/kubernetes` package green, including under `-race`. Full module builds; the EKS/AKS/GKE server layers and provider tests pass.

**Real-user e2e** — a `kubectl`-tagged test drives **real kubectl v1.35.4** against a temporary in-memory cluster. Live run output:

```
# apply-create on a fresh cluster
$ kubectl apply --server-side --field-manager=alice -f cm.yaml
configmap/settings serverside-applied

# conflict when another manager changes an owned field
$ kubectl apply --server-side --field-manager=bob -f cm-red.yaml
error: Apply failed with conflicts: fields managed by another manager: data.color (owned by alice)
$ kubectl apply --server-side --field-manager=bob --force-conflicts -f cm-red.yaml
configmap/settings serverside-applied

# finalizer-gated delete stays Terminating, then completes on removal
$ kubectl delete configmap guarded --wait=false
configmap "guarded" deleted from default namespace
$ kubectl get cm guarded -o jsonpath='{.metadata.deletionTimestamp}'
2026-08-11T06:58:09Z
$ kubectl patch configmap guarded --type=merge -p '{"metadata":{"finalizers":[]}}'
$ kubectl get cm guarded
Error from server (NotFound): configmap not found: default/guarded
```

## Not included

Pod's apply path was already recording update ownership and is out of this issue's scope; unchanged.

### DIFF
--- a/services/kubernetes/configmap.go
+++ b/services/kubernetes/configmap.go
@@ -100,13 +100,23 @@ func (s *ClusterState) serveConfigMapItem(w http.ResponseWriter, r *http.Request
 	}
 }
 
-//nolint:dupl // namespaced-create CRUD pattern; copy-paste is clearer than a generic helper.
 func (s *ClusterState) createConfigMap(w http.ResponseWriter, r *http.Request, namespace string) {
 	var in corev1.ConfigMap
 	if !readJSON(w, r, &in) {
 		return
 	}
 
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	s.createConfigMapLocked(w, r, namespace, &in)
+}
+
+// createConfigMapLocked persists a new ConfigMap. The caller holds s.mu. It is
+// shared by the POST create path and server-side apply-create (an apply to a
+// name that doesn't exist yet), which passes an object already carrying its
+// Apply managedFields entry.
+func (s *ClusterState) createConfigMapLocked(w http.ResponseWriter, r *http.Request, namespace string, in *corev1.ConfigMap) {
 	if in.Name == "" {
 		writeBadRequest(w, "k8s api: configmap: metadata.name is required")
 
@@ -118,9 +128,6 @@ func (s *ClusterState) createConfigMap(w http.ResponseWriter, r *http.Request, n
 
 	key := configMapKey(namespace, in.Name)
 
-	s.mu.Lock()
-	defer s.mu.Unlock()
-
 	if _, ok := s.configMaps[key]; ok {
 		writeAlreadyExists(w, "k8s api: configmap already exists: "+key)
 
@@ -131,12 +138,12 @@ func (s *ClusterState) createConfigMap(w http.ResponseWriter, r *http.Request, n
 	in.TypeMeta = metav1.TypeMeta{Kind: "ConfigMap", APIVersion: "v1"}
 
 	if isDryRun(r) {
-		writeJSON(w, http.StatusCreated, &in)
+		writeJSON(w, http.StatusCreated, in)
 
 		return
 	}
 
-	cm := in
+	cm := *in
 	s.configMaps[key] = &cm
 	s.wConfigMaps.publish(EventAdded, namespace, *cm.DeepCopy())
 	writeJSON(w, http.StatusCreated, &cm)
@@ -266,12 +273,25 @@ func (s *ClusterState) patchConfigMap(w http.ResponseWriter, r *http.Request, na
 
 	cur, ok := s.configMaps[key]
 	if !ok {
+		// Server-side apply to a missing object creates it (upstream SSA).
+		if r.Header.Get("Content-Type") == contentTypeApplyPatch {
+			in, aok := serverSideApplyTyped(s, w, r, apiVersionV1, &corev1.ConfigMap{})
+			if !aok {
+				return
+			}
+
+			in.Name = name
+			s.createConfigMapLocked(w, r, namespace, in)
+
+			return
+		}
+
 		writeNotFound(w, "k8s api: configmap not found: "+key)
 
 		return
 	}
 
-	patched, ok := applyJSONPatch(w, r, cur)
+	patched, ok := applyOrPatchTyped(s, w, r, apiVersionV1, cur)
 	if !ok {
 		return
 	}
@@ -279,6 +299,16 @@ func (s *ClusterState) patchConfigMap(w http.ResponseWriter, r *http.Request, na
 	patched.ResourceVersion = bumpResourceVersion(cur.ResourceVersion)
 
 	if isDryRun(r) {
+		writeJSON(w, http.StatusOK, patched)
+
+		return
+	}
+
+	// A patch that drops the last finalizer from a Terminating object completes
+	// the delete (the patch was merged onto cur, so it inherits deletionTimestamp).
+	if finalizersDrained(&patched.ObjectMeta) {
+		delete(s.configMaps, key)
+		s.wConfigMaps.publish(EventDeleted, namespace, *patched.DeepCopy())
 		writeJSON(w, http.StatusOK, patched)
 
 		return
@@ -303,6 +333,16 @@ func (s *ClusterState) deleteConfigMap(w http.ResponseWriter, r *http.Request, n
 	}
 
 	if isDryRun(r) {
+		writeJSON(w, http.StatusOK, cm.DeepCopy())
+
+		return
+	}
+
+	// Finalizer-gated deletion: a ConfigMap with finalizers goes Terminating and
+	// is removed only when the last finalizer is dropped via update/patch.
+	if s.markForDeletion(&cm.ObjectMeta) {
+		cm.ResourceVersion = bumpResourceVersion(cm.ResourceVersion)
+		s.wConfigMaps.publish(EventModified, namespace, *cm.DeepCopy())
 		writeJSON(w, http.StatusOK, cm.DeepCopy())
 
 		return
--- a/services/kubernetes/ssa.go
+++ b/services/kubernetes/ssa.go
@@ -58,7 +58,7 @@ var ssaSkipMeta = map[string]bool{
 // (merged, true) on success; on conflict or wire error it has already written
 // the response and returns (nil, false).
 func (s *ClusterState) serverSideApply(
-	w http.ResponseWriter, r *http.Request, st *registryStore, cur *unstructured.Unstructured,
+	w http.ResponseWriter, r *http.Request, apiVersion string, cur *unstructured.Unstructured,
 ) (*unstructured.Unstructured, bool) {
 	body, err := io.ReadAll(r.Body)
 	if err != nil {
@@ -95,7 +95,7 @@ func (s *ClusterState) serverSideApply(
 		return nil, false
 	}
 
-	setManagedFields(merged, updateManagedFields(cur, manager, appliedLeaves, st.def.apiVersion(), s.now()))
+	setManagedFields(merged, updateManagedFields(cur, manager, appliedLeaves, apiVersion, s.now()))
 	// Upstream apply removes fields this manager previously owned but now omits,
 	// unless another manager still owns them.
 	removeDroppedLeaves(merged, diffLeaves(prevLeaves, appliedLeaves))
@@ -449,6 +449,63 @@ func objectMap(v any) map[string]any {
 	}
 
 	return m
+}
+
+// applyOrPatchTyped dispatches a typed PATCH: an `application/apply-patch+yaml`
+// body runs through the server-side-apply engine (field ownership + conflicts),
+// every other content type is a plain merge/strategic/JSONPatch. apiVersion
+// labels the managedFields entry SSA records. This mirrors registryPatch's
+// content-type branch for the typed core kinds.
+func applyOrPatchTyped[T any](
+	s *ClusterState, w http.ResponseWriter, r *http.Request, apiVersion string, cur *T,
+) (*T, bool) {
+	if r.Header.Get("Content-Type") == contentTypeApplyPatch {
+		return serverSideApplyTyped(s, w, r, apiVersion, cur)
+	}
+
+	return applyJSONPatch(w, r, cur)
+}
+
+// serverSideApplyTyped runs server-side apply for a typed core kind by
+// round-tripping the current object through the unstructured SSA engine, which
+// carries all the field-ownership and conflict logic. It returns the merged
+// typed object (managedFields set), or (nil, false) after the engine has
+// already written a 409/400 response.
+func serverSideApplyTyped[T any](
+	s *ClusterState, w http.ResponseWriter, r *http.Request, apiVersion string, cur *T,
+) (*T, bool) {
+	curMap := objectMap(cur)
+	if curMap == nil {
+		writeBadRequest(w, "k8s api: encode current object for apply")
+
+		return nil, false
+	}
+
+	merged, ok := s.serverSideApply(w, r, apiVersion, &unstructured.Unstructured{Object: curMap})
+	if !ok {
+		return nil, false
+	}
+
+	return typedFromUnstructured[T](w, merged)
+}
+
+// typedFromUnstructured decodes an unstructured object back into a typed T.
+func typedFromUnstructured[T any](w http.ResponseWriter, u *unstructured.Unstructured) (*T, bool) {
+	b, err := json.Marshal(u.Object)
+	if err != nil {
+		writeBadRequest(w, "k8s api: marshal merged object: "+err.Error())
+
+		return nil, false
+	}
+
+	out := new(T)
+	if err := json.Unmarshal(b, out); err != nil {
+		writeBadRequest(w, "k8s api: decode merged typed object: "+err.Error())
+
+		return nil, false
+	}
+
+	return out, true
 }
 
 // typedEntryLeaves returns the leaf set recorded in a typed managedFields entry.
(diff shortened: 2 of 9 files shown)


services/kubernetes/configmap.go
```diff
@@ -266,12 +273,25 @@ func (s *ClusterState) patchConfigMap(w http.ResponseWriter, r *http.Request, na
...
+			return
+		}
+
 		writeNotFound(w, "k8s api: configmap not found: "+key)
 
 		return
 	}
 
-	patched, ok := applyJSONPatch(w, r, cur)
+	patched, ok := applyOrPatchTyped(s, w, r, apiVersionV1, cur)
```
thzgajendra, line 294: 🔴 **Blocker — server-owned metadata is dropped on the typed apply path.**

After `applyOrPatchTyped` this only re-stamps `ResourceVersion`. The registry patch path additionally re-asserts identity after the merge (`registry.go:500-502`: `SetDeletionTimestamp(prevDeletion)` / `SetUID(prevUID)` / `SetCreationTimestamp(prevCreation)`) — the four typed handlers don't, so they diverge from the path they mirror.

Root cause: `serverSideApply` merges via `mergeRFC7396(cur.Object, applied)` (`ssa.go:93`), and RFC-7396 **deletes any key whose value is `null`**. `ssaSkipMeta`/`ssaSkipTop` only exclude these from ownership-leaf tracking, not from the merge input.

Live probe against the running emulator confirmed:
- Apply a body with `metadata.creationTimestamp: null` (which `kubectl apply --server-side` **always** emits for the zero `metav1.Time`) → the stored `creationTimestamp` is **deleted** on every apply after the first.
- Apply `metadata.deletionTimestamp: null` on a Terminating object → `deletionTimestamp` is stripped and the object **resurrects**, defeating the finalizer-gated delete this PR adds. A null `uid` clobbers it the same way.

**Fix:** re-assert `uid`/`creationTimestamp`/`deletionTimestamp` (and `resourceVersion`/`generation`) from `cur` after the merge — ideally once inside `serverSideApply`/`serverSideApplyTyped` so all kinds inherit it (mirror `registry.go:500-502`), or strip the `ssaSkipMeta`/`ssaSkipTop` keys from `applied` before `mergeRFC7396`.

services/kubernetes/ssa.go
```diff
@@ -451,6 +451,63 @@ func objectMap(v any) map[string]any {
...
+
+	return applyJSONPatch(w, r, cur)
+}
+
+// serverSideApplyTyped runs server-side apply for a typed core kind by
+// round-tripping the current object through the unstructured SSA engine, which
+// carries all the field-ownership and conflict logic. It returns the merged
+// typed object (managedFields set), or (nil, false) after the engine has
+// already written a 409/400 response.
+func serverSideApplyTyped[T any](
```
thzgajendra, line 489: Two design points on the typed bridge (both fold naturally in here):

1. **Fold the metadata re-assert here** (see the blocker on `configmap.go:299`) so all four kinds inherit it in one place rather than each handler needing it.
2. **`typedFromUnstructured`/`objectMap` round-trip via `encoding/json`.** That silently drops unknown fields and doesn't error on shape mismatch. `k8s.io/apimachinery/pkg/runtime.DefaultUnstructuredConverter` (`ToUnstructured`/`FromUnstructured`) is the canonical typed↔unstructured bridge — handles `metav1.Time`/quantities and fails loudly. Worth switching for fidelity.

Minor: this runs under `s.mu` (callers lock first) and reads `r.Body` while holding the global lock — intentional but worth a one-line note.

services/kubernetes/typed_ssa_test.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	return false
+}
+
+// TestTypedServerSideApply_OwnershipAndConflict: applying to a typed ConfigMap
+// must record managedFields and a second manager changing an owned field must
+// 409 (unless force). Today the typed PATCH path treats apply-patch as a plain
+// merge, so no ownership is recorded and the conflicting apply silently wins.
+func TestTypedServerSideApply_OwnershipAndConflict(t *testing.T) {
```
thzgajendra, line 42: 🟡 Coverage is one-kind-deep (ConfigMap) and asserts only on the apply **response** — it never re-GETs, which is exactly why the metadata-corruption blocker slips through. Suggested additions:
- **Metadata preservation:** create → capture `creationTimestamp`/`uid` → apply a body with `creationTimestamp: null` + a data field → GET → assert both survive. (Fails today.)
- **Two-manager retention** (the core SSA property): alice applies `{data:{x:1}}`, bob applies `{data:{y:2}}` → GET asserts both x and y present (bob's omission of x must not drop alice's field); then alice re-applies dropping x → x removed.
- **Deployment apply-create materializes Pods** (PR promises it; only the Deployment object is asserted).
- **Secret/Service apply-create** (Service's ClusterIP+Endpoints pairing) and **no-finalizer → immediate hard-delete.**

Consider table-driving across the four kinds rather than hard-coding ConfigMap.

thzgajendra (changes requested):
## Review — typed-kind SSA / apply-create / finalizers

Reviewed in depth with a multi-agent pass (correctness, error-handling, test coverage, type design, comments, simplification) plus a **live end-to-end probe against the running emulator**. Strong PR overall — the generic SSA bridge (`serverSideApplyTyped[T]`/`applyOrPatchTyped[T]`) is well-factored, lock discipline is correct (no TOCTOU/deadlock — creates happen under one held `s.mu`), 409-conflict + `?force=true` are correct, the finalizer Terminating→drain→delete lifecycle works, the `apiVersion` decoupling is behavior-preserving for the registry path, and the tests pass race-clean with the kubectl e2e properly build-tag-gated.

One **blocker**, confirmed by a live probe (details inline on `configmap.go`):

### 🔴 Blocker — typed apply drops server-owned metadata (data corruption + finalizer bypass)
`serverSideApply` merges via `mergeRFC7396(cur.Object, applied)` (`ssa.go:93`); RFC-7396 deletes keys whose value is `null`. The registry path guards this (`registry.go:500-502`), the typed path doesn't. Result: `kubectl apply --server-side` (which emits `creationTimestamp: null`) **deletes the stored `creationTimestamp`**, and `deletionTimestamp: null` **resurrects a Terminating object** — the finalizer machinery this PR adds is directly circumventable. Fix + repro inline.

### 🟡 Follow-ups
- **Test coverage is one-kind-deep** and asserts only on the apply response (never re-GETs) — which is why the blocker slips through. Gaps + suggestions inline on the test file: metadata preservation, two-manager field retention (the core SSA property), Deployment-materializes-Pods on apply, Secret/Service apply-create, no-finalizer hard-delete.
- **`typedFromUnstructured` uses a JSON round-trip** — prefer `runtime.DefaultUnstructuredConverter` (inline on `ssa.go`).
- **Latent trap (unchanged file):** `applyPatchBytes` (patch.go, not in this diff) still groups `application/apply-patch+yaml` with the plain-merge case. Masked because the typed handlers intercept apply-patch upstream, but any future caller handing an apply-patch body to `applyJSONPatch` silently gets a plain merge (no ownership/409). Worth making apply-patch route to SSA / error unambiguously.

### 🟢 Notes
- serviceaccount.go: stale comment after the `//nolint:dupl` removal + a scope question (inline).
- The generic factoring is right — no need to genericize the four apply-create/delete blocks further (per-kind store types + interleaved side effects make it more complex, as the existing rationale notes).

Requesting changes for the metadata blocker; the rest are follow-ups you can weigh. Nice work on the SSA engine reuse and the finalizer lifecycle.

services/kubernetes/configmap.go
NitinKumar004: Fixed in 47d195a0f09d5356b012108ea0ab77d3980be7c4. Root cause was one level deeper than the missing re-stamp: `mergeRFC7396(cur.Object, applied)` mutates `cur.Object` **in place**, so re-asserting from `cur` *after* the merge reads the applied body's `creationTimestamp: null`. Fix folds the re-assert into `serverSideApply` itself (the single place both typed and registry apply flow through): it now snapshots `uid`/`creationTimestamp`/`deletionTimestamp` **before** the merge and re-asserts them on `merged`. So all four typed kinds inherit it, and the registry path is double-safe. Verified with real `kubectl apply --server-side`: uid + creationTimestamp survive a re-apply that sends `creationTimestamp: null`.

services/kubernetes/ssa.go
NitinKumar004: Both done in 47d195a0f09d5356b012108ea0ab77d3980be7c4. (1) The metadata re-assert is folded into `serverSideApply` (not each handler) — one place, all four kinds. (2) `objectMap` and `typedFromUnstructured` now use `runtime.DefaultUnstructuredConverter.ToUnstructured`/`FromUnstructured` instead of the `encoding/json` round-trip, so a shape mismatch surfaces as an error rather than silently dropping fields.

services/kubernetes/typed_ssa_test.go
NitinKumar004: Added in 47d195a0f09d5356b012108ea0ab77d3980be7c4. Two new tests that **re-GET after apply** (the check that catches the metadata drop): `TestTypedApply_PreservesServerMetadata` (create → capture uid/creationTimestamp → apply a body with `creationTimestamp: null` + a change → GET → assert both survive) run across **ConfigMap and Secret** so it isn't one-kind-deep, and `TestTypedApply_TwoManagerRetention` (alice applies `data.x`, bob applies `data.y` → GET asserts both). The preservation test fails on the old code and passes now.

NitinKumar004:
@thzgajendra thanks for the deep review — all three findings addressed in 47d195a:

- **🔴 Blocker (server-owned metadata dropped):** folded the re-assert into `serverSideApply` itself. The real cause was that `mergeRFC7396` mutates `cur.Object` in place, so I now snapshot `uid`/`creationTimestamp`/`deletionTimestamp` **before** the merge and re-assert after — one place, all four typed kinds inherit it.
- **Converter:** `objectMap`/`typedFromUnstructured` now use `runtime.DefaultUnstructuredConverter` instead of the json round-trip (errors on shape mismatch instead of silently dropping fields).
- **Tests:** added re-GET `TestTypedApply_PreservesServerMetadata` (ConfigMap + Secret) and `TestTypedApply_TwoManagerRetention`.

Verified end-to-end with real `kubectl apply --server-side`: uid + creationTimestamp preserved across re-apply (kubectl sends `creationTimestamp: null`), `managedFields` records the Apply entry, two managers' disjoint fields both retained, and a cross-manager change without `--force` 409s. Full k8s suite green; changes are lint-clean. Ready for another look.

services/kubernetes/typed_ssa_test.go
```diff
@@ -0,0 +1,373 @@
...
+	}
+
+	got := decodeMap(t, g.Body)
+	g.Body.Close()
+
+	if mf := managedFieldsOfObj(t, got); !hasApplyEntry(mf, "kubectl") {
+		t.Fatalf("apply-created ConfigMap has no Apply managedFields entry for kubectl: %v", mf)
+	}
+
+	// Typed workload: apply-create a Deployment (the headline `kubectl apply
```
thzgajendra, line 141: **Low — add a Pod-count assertion here.** The PR's headline is that apply-create preserves per-kind reconcile ("a Deployment materializes Pods"), and the code does this correctly — `patchDeployment`'s apply-create branch calls the same `createDeploymentLocked` → `reconcileDeploymentLocked` that POST-create uses. But this Deployment case only checks 200/GETtable + an Apply managedFields entry; it never asserts the child Pods exist. A `GET .../pods` (or the internal store) count == `spec.replicas` after the apply-create would pin that behavior so a future change to the apply-create path can't silently stop materializing Pods while the test stays green. Optional but it's the one place the tests trail the narrative.

thzgajendra (commented):
## Review — SSA / apply-create / finalizers for typed core kinds

Reviewed the SSA core in source and ran the full suite + the real-`kubectl` E2E via a test agent. **Clean, well-scoped PR — implementation is correct, tests are green including live kubectl, blast radius is contained. One low test-coverage nit + one pre-existing (non-PR) lint item.**

### ✅ Implementation (verified in source)
- **SSA engine (`ssa.go`)** — real conflict detection (409 unless `force`), managedFields ownership, and drop-owned-but-omitted semantics matching upstream. The subtle correctness win: server-owned identity (UID / creationTimestamp / deletionTimestamp) is snapshotted **before** the in-place `mergeRFC7396` and re-asserted after, so kubectl's `creationTimestamp: null` can't blank or re-identify the object. The leaf-granularity limitation vs upstream is honestly documented.
- **apply-create** — routes through a shared `createXLocked` core, so per-kind defaulting/reconcile is preserved (a Deployment apply-create runs the same `reconcileDeploymentLocked` as POST-create → Pods materialize). **Locking is correct:** the public path locks then calls the `*Locked` core; the apply-create branch in `patchX` already holds `s.mu` before the map read and the `*Locked` call — no double-lock, no race (confirmed green under `-race`).
- **Finalizers** — `markForDeletion` → Terminating on delete; `finalizersDrained` completes the delete on the finalizer-draining patch. Matches Pod/Namespace/registry behavior.
- **`serviceaccount.go` (−3)** — just removes now-unused `//nolint:dupl` directives (the 4 typed handlers diverged, so they're no longer dupl-identical to ServiceAccount). Correct cleanup.
- **Blast radius** — confined to `services/kubernetes/`; only the 4 typed kinds are routed into the pre-existing SSA machinery, registry kinds unchanged.

### ✅ Tests (from the run)
- `go build` + `go vet` clean; **237 tests pass**, green under **`-race`**; full module green (EKS/AKS/GKE + `server/*` layers, no regression).
- **Real `kubectl` v1.35.4 E2E — ran (not skipped), all live:** apply-create on a fresh cluster → `serverside-applied`; conflict → `Apply failed with conflicts: … data.color (owned by alice)`; `--force-conflicts` overrides; finalizer-gated delete stays `Terminating` (deletionTimestamp set) then returns `NotFound` after the finalizer is drained.

### Notes
- **Low (test coverage):** `TestServerSideApplyCreate`'s Deployment case asserts the object is created + has an Apply managedFields entry, but doesn't assert the **child Pods** materialized. The behavior is correct (shared `createDeploymentLocked` → reconcile), but a Pod-count assertion would pin the headline "a Deployment materializes Pods" claim so a future refactor of the apply-create path can't silently break it. See inline.
- **Pre-existing, not this PR:** `golangci-lint` flags one issue at `pod.go:510` (unused `//nolint:gosec`) — `pod.go` is untouched by #390 (it's the base). Track separately.

**Verdict:** correct, tested (incl. live kubectl), contained. Clean from my side — the only actionable item is the optional Pod-count assertion.

services/kubernetes/typed_ssa_test.go
NitinKumar004: Added in 4ca9559255a537ada8be0b5a1877e5aad60641c8. After the apply-create Deployment, the test now GETs `/api/v1/namespaces/default/pods?labelSelector=app=web` and asserts the count == `spec.replicas` (1), pinning that apply-create goes through the same `reconcileDeploymentLocked` that POST-create does — a future change to the apply-create path can't silently stop materializing Pods while the test stays green. Thanks for the catch.

Issue #376, opened 2026-08-09
OCI provider: Tier 1 service parity


arunesh-j:
Tracking issue for adding **Oracle Cloud Infrastructure** as a fourth provider, at parity with AWS/Azure/GCP across the portable driver interfaces.

## Why

OCI has an equivalent for 20 of the 26 service categories in `docs/services.md`, and 17 of those map onto driver interfaces that already exist in `services/<name>/driver`. A service implemented against one of those inherits chaos, metrics, recorder, ratelimit, inject, persist, seed and topology for free.

## Scope — Tier 1

Services that implement an existing portable driver. No new driver interface is introduced; OCI-only behaviour goes in an optional capability discovered by type assertion.

| Category | OCI service | Portable driver | Status |
|---|---|---|---|
| Foundation | identity/OCIDs/work requests/wire | — | Done (#342) |
| Monitoring | Monitoring | `monitoring` | #377 (PR #366) |
| IAM | Identity | `iam` | #378 (PR #367) |
| Networking | VCN | `networking` | #379 (PR #368) |
| Storage | Object Storage | `storage` | not started |
| Compute | Compute + Block Volume | `compute` | not started |
| NoSQL DB | NoSQL Database | `database` | not started |
| Secrets | Vault | `secrets` | not started |
| Logging | Logging | `logging` | not started |
| Notification | Notifications (ONS) | `notification` | not started |
| Serverless | Functions | `serverless` | not started |
| Message Queue | Queue | `messagequeue` | not started |
| Cache | OCI Cache (Redis) | `cache` | not started |
| DNS | DNS | `dns` | not started |
| Load Balancer | Load Balancer + NLB | `loadbalancer` | not started |
| Container Registry | Container Registry (OCIR) | `containerregistry` | not started |
| Event Bus | Events | `eventbus` | not started |
| Relational DB | Autonomous DB, Base DB, MySQL HeatWave, PostgreSQL | `relationaldb` | not started |
| Kubernetes | OKE | `kubernetes` + shared data plane | not started |
| Resource Discovery | Resource Search + Tagging | `resourcediscovery` | not started |

## Out of scope here

- **Tier 2** — categories that exist for one provider today and have an OCI counterpart: Network Firewall, Generative AI (+ Agents), AI Search (OpenSearch), Container Instances, DNS Private Resolver.
[...]

PR #393, opened 2026-08-11
Move OCI-only capabilities out of shared driver packages

### LINKED ISSUE #392: OCI capabilities in shared driver packages leak into AWS/Azure/GCP coverage docs
`docs/coverage` exists so the docs cannot promise a capability the code lacks. Right now it does exactly that, for three providers.

## The problem

`docs/coverage/aws/cloudwatch.md` lists `CreateOCIAlarm`, `PostMetricData`, `ListOCIMetrics` and 7 more OCI-specific operations. AWS's CloudWatch mock implements none of them. Same for `azure/monitor.md` and `gcp/cloudmonitoring.md`, and the same again for `CreateCompartment`/`CreateStatementPolicy` in all three `iam.md` pages.

Reproduce on `development`:

```
$ grep -l "CreateOCIAlarm" docs/coverage/{aws,azure,gcp}/*.md
docs/coverage/aws/cloudwatch.md
docs/coverage/azure/monitor.md
docs/coverage/gcp/cloudmonitoring.md

$ grep -l "CreateCompartment" docs/coverage/{aws,azure,gcp}/*.md
docs/coverage/aws/iam.md
docs/coverage/azure/iam.md
docs/coverage/gcp/iam.md
```

## Cause

The three OCI services that landed in #366, #367 and #368 each solved "OCI needs something the portable driver cannot express" differently:

| Service | Capability lives in | Leaks |
|---|---|---|
| Monitoring (#366) | `services/monitoring/driver/driver.go` (shared) | yes |
| Identity (#367) | `services/iam/driver/oci.go` (shared package) | yes |
| VCN (#368) | `Extras` interface in `server/oci/vcn` (consumer-side) | **no** |

`coveragegen` renders every interface it finds in `services/<svc>/driver` against every provider implementing that service, so anything OCI-specific placed there is attributed to all four. The reviewer noted this on #366 and flagged the same shape in a pre-existing `cache/SubnetGroups` case.

## Fix

Adopt the pattern that does not leak — VCN's — and document it as the rule:

- OCI-only value types move to `providers/oci/<service>/`.
- The capability interface is declared consumer-side in `server/oci/<service>/` as `Extras`, satisfied by the provider mock and discovered by type assertion, exactly as `server/oci/vcn` does.
- `docs/oci-conventions.md` records this so the remaining Tier 1 services (#376) do not reintroduce it.

Only OCI code references these types today, so the move touches no other provider.

## Out of scope

The pre-existing `cache/SubnetGroups` instance of the same shape — worth its own issue if the fix here proves the approach.

Part of #376.

arunesh-j:
## Summary

- `docs/coverage` exists so the docs cannot promise a capability the code lacks. Six pages currently do exactly that.
- OCI-only capabilities placed in `services/*/driver` are attributed by `coveragegen` to **every** provider implementing that service, so `aws/cloudwatch.md` advertises `CreateOCIAlarm` and `aws/iam.md` advertises `CreateCompartment`.
- This moves them to the pattern that does not leak — the one `server/oci/vcn` already uses — and fixes the conventions doc that produced the bug.

Closes #392. Part of #376.

## The bug

```
$ grep -l "CreateOCIAlarm" docs/coverage/{aws,azure,gcp}/*.md
docs/coverage/aws/cloudwatch.md  docs/coverage/azure/monitor.md  docs/coverage/gcp/cloudmonitoring.md
$ grep -l "CreateCompartment" docs/coverage/{aws,azure,gcp}/*.md
docs/coverage/aws/iam.md  docs/coverage/azure/iam.md  docs/coverage/gcp/iam.md
```

The three OCI services that landed each solved "OCI needs something the portable driver cannot express" differently:

| Service | Capability lived in | Leaked |
|---|---|---|
| Monitoring (#366) | `services/monitoring/driver/driver.go` | yes |
| Identity (#367) | `services/iam/driver/oci.go` | yes |
| VCN (#368) | `Extras` in `server/oci/vcn` | **no** |

## Changes

**Monitoring** — the 82-line OCI block leaves `services/monitoring/driver/driver.go`. `OCIMetric`/`OCIMetricFilter`/`OCIMetricQuery` → `providers/oci/monitoring/metrics.go`; `OCIAlarmSpec`/`OCIAlarm` → `alarms.go`; `OCIMonitoring` → `server/oci/monitoring/handler.go` as `Extras`.

**Identity** — `services/iam/driver/oci.go` is deleted. Value types → `providers/oci/identity/{compartment,identity,policy}.go`; `Compartments`/`OCIIdentity`/`StatementPolicies` → `server/oci/identity/handler.go`, kept as three interfaces since the handler already asserts them independently and 501s per resource family.

Handlers import the provider package for value types — the `server/azure/aks` and `server/gcp/gke` precedent that `server/oci/vcn` already follows.

**`docs/oci-conventions.md`** — this is the part that matters for the remaining Tier 1 services. The doc previously told implementers to add optional capabilities in `services/<name>/driver`, citing `storage/driver.BucketAttributes`. That guidance is what produced this bug; two of three sessions followed it correctly and leaked. Replaced with the `Extras` rule and VCN as the reference.

One rename: `IdentityUpdate` → `Update`, because in package `identity` the old name trips revive's stutter check. Every other type name is unchanged.

## Behaviour

Unchanged. No wire response differs — the moved structs are field-for-field identical, and every changed line outside the relocations is a type reference (`driver.AccessRequest` → `AccessRequest`). Discovery is still by type assertion; a driver lacking a capability still gets 501.

## Operation counts

| Page | Before | After |
|---|---|---|
| `aws/cloudwatch.md` | 12 + 9 | **12** |
| `azure/monitor.md` | 12 + 9 | **12** |
| `gcp/cloudmonitoring.md` | 12 + 9 | **12** |
| `oci/monitoring.md` | 12 + 9 | 12 |
| `aws/iam.md` | 40 + 25 | **40** |
| `azure/iam.md` | 40 + 25 | **40** |
| `gcp/iam.md` | 40 + 25 | **40** |
| `oci/identity.md` | 40 + 25 | 40 |

**No capability was deleted** — every method still exists on the mocks and is still reachable. The OCI pages drop the same rows because `coveragegen` only scans `services/*/driver`, which is exactly how `oci/vcn.md` has always behaved for VCN's `Extras`. That trades an *overstatement* about three providers for an *understatement* about one, which is the right direction for a guarantee that the docs never promise more than the code delivers.

## Also

Fixed a stray trailing blank line in `server/wire/ocirest/ocirest_test.go` — my own leftover from #342, and the only `gofmt -l` hit in the repo. The tree is now gofmt-clean.

## A related bug this does NOT fix

The same root cause runs the other way: `docs/coverage/oci/vcn.md` credits OCI with **`CreateIpam` and `CreateTransitGateway`**, which `providers/oci/vcn` does not implement — they are AWS-only optional capabilities on the shared networking driver. That is an overstatement about OCI, so arguably worse than what this PR fixes, but it needs a `coveragegen` change rather than a code move. Filing separately rather than growing this PR; the pre-existing `cache/SubnetGroups` case is the same shape.

## Provider Coverage

- [ ] AWS
- [ ] Azure
- [ ] GCP
- [x] OCI

## Checklist

- [x] All tests pass (`go test ./...`) — exit 0, 270 packages
- [x] Every provider the change applies to implements the same behavior — no behaviour change
- [ ] Integration tests added to `cloudemu_test.go` — refactor only, existing tests cover it
- [x] Unit tests added to provider test files — compile-time capability checks moved with the interfaces

## Test Plan

`go build ./...` clean · `go test ./...` exit 0, 270 packages · `go test -race ./providers/oci/... ./server/oci/...` 9/9 ok · `golangci-lint` 0 issues · `go generate ./...` idempotent (a second run leaves the tree clean).

Both leak greps return nothing.


NitinKumar004 (approved):
Reviewed end-to-end. This is a clean, well-executed refactor that fixes a real documentation-correctness bug, and it's verified across the whole module. Approving.

**What it fixes, confirmed:** the OCI-only capabilities placed in `services/*/driver` were attributed by `coveragegen` to every provider implementing the service, so `aws/cloudwatch.md` advertised `CreateOCIAlarm` and `aws/iam.md` advertised `CreateCompartment`. After the move, a fresh `go generate` shows those capabilities on **zero** aws/azure/gcp coverage pages (was three each), and the generated docs match what's committed.

**Verification**
- **Full module suite green** — `go test ./...` across the entire module: 270 packages ok, 0 failures, 0 data races. Pulling the OCI types out of the shared `services/monitoring/driver` and `services/iam/driver` (the latter file deleted entirely) broke nothing in any provider, portable wrapper, topology, or integration path.
- **Blast radius contained** — the shared driver interfaces are fully cleaned of OCI residue, and every reference to the moved value types (`OCIAlarmSpec`, `OCIMetricFilter`, `CompartmentSpec`, `StatementPolicies`, …) is confined to `providers/oci/*` and `server/oci/*`. AWS/Azure/GCP monitoring and IAM are untouched.
- **`Extras` fallback is nil-safe** — since the handler discovers the capability by type assertion (`mon.(Extras)` / `drv.(Compartments|OCIIdentity|StatementPolicies)`), a driver that doesn't satisfy it leaves a nil interface. Every dispatch path guards it before use and answers `501 NotImplemented`: `server/oci/monitoring/handler.go` (`h.mon == nil`), `server/oci/identity/compartment.go` and `policy.go`, and — the subtle one — `routePrincipal` checks `h.identity == nil` **before** `opsFor` binds method values (evaluating a method value on a nil interface would otherwise panic). No nil-panic path.
- **Compile-time capability guards present** — `var _ monitoring.Extras = (*Mock)(nil)` and the three identity interface assertions (`Compartments`/`OCIIdentity`/`StatementPolicies`) mean a future drift where the mock stops satisfying `Extras` breaks the build rather than silently 501-ing.
- **Docs deterministic** — `go generate ./...` run twice is byte-identical; `go mod tidy` is clean.
- **Structure conformant** — moved files are snake_case with no redundant cloud prefix and the provider↔wire names match; the layering (value types in `providers/oci/<svc>`, `Extras` interface in `server/oci/<svc>`) follows the `server/oci/vcn` precedent and the `server/azure/aks` / `server/gcp/gke` import shape.

**`oci-conventions.md`** is the most valuable part for the remaining Tier-1 services: it now explains *why* the old "add an optional capability in `services/<name>/driver`" guidance produced the leak (coveragegen renders every interface it finds there against every provider) and codifies the `Extras` rule with the compile-time-check mandate and VCN as the reference. That prevents the class of bug from recurring.

**Follow-up (non-blocking):** with the move to `Extras`, these OCI capabilities now appear in no coverage doc at all — not even `oci/monitoring.md` / `oci/identity.md` — because `coveragegen` only walks `services/*/driver`. That's the correct trade for killing the leak and is consistent with VCN, but OCI's genuine surface is now undercounted. Worth a follow-up to teach `coveragegen` to attribute an `Extras` interface to its owning OCI provider page so the capabilities are documented somewhere.

PR #400, opened 2026-08-15
fix(serve): advertise a routable Kubernetes endpoint under Docker (--host 0.0.0.0)

NitinKumar004:
## Problem

Running the standalone server in Docker (the `ghcr.io/stackshy/cloudemu` image runs `serve --host 0.0.0.0`) left the **Kubernetes data plane unreachable from the host**. `0.0.0.0` is a *bind-all* address, not a *connectable* one — but the k8s data plane **self-advertises** it: `EKS/AKS/GKE DescribeCluster` returned `https://0.0.0.0:4570/k8s/<uid>` with a `0.0.0.0`-only cert SAN, so `kubectl`/`client-go` on the host could neither dial nor TLS-verify the cluster.

(AWS/Azure/GCP endpoints are unaffected — callers pass their own endpoint; only the Kubernetes data plane self-advertises via `SetBaseURL`.)

## Fix

- **Advertise a routable host, not the bind address.** When `--host` binds all interfaces (`0.0.0.0` / `::` / empty), advertise `127.0.0.1` (a published `-p 4570:4570` maps there from the host). A specific `--host` is kept as-is.
- **`--advertise-host <name-or-ip>`** to override — for pointing `kubectl` at the cluster from *another* machine (its LAN IP/hostname); it's also added to the serving cert's SANs.
- The k8s serving cert now certifies **advertised host + loopback + any `--tls-host` extras** (was `[host, localhost, 127.0.0.1]`, which produced a useless `0.0.0.0` SAN under Docker).
- Startup banner shows the reachable endpoint, not the bind address.
- **docs:** the `docker run` example now publishes **`-p 4570:4570`** (it was missing, so kubectl couldn't reach the cluster regardless) and documents the advertise behavior + `--advertise-host`.

## Verification

- Unit tests for `advertiseHostFor` / `k8sCertHosts`.
- End-to-end: `serve --host 0.0.0.0` → EKS `DescribeCluster` advertises `https://127.0.0.1:4570/k8s/<uid>` (not `0.0.0.0`) → `kubectl get ns` connects and TLS-verifies. ✅
- Build + full `cmd/cloudemu` tests green; no new lint (package baseline unchanged).

Independent of #390 (touches only `cmd/cloudemu/serve.go` + docs). Both can ride the same `v2.5.0` release.

### DIFF
--- a/cmd/cloudemu/serve.go
+++ b/cmd/cloudemu/serve.go
@@ -46,6 +46,7 @@ var errUnsupportedSnapshot = errors.New("unsupported snapshot schema version")
 type serveConfig struct {
 	providers       string
 	host            string
+	advertiseHost   string
 	awsPort         string
 	azurePort       string
 	gcpPort         string
@@ -85,6 +86,9 @@ func runServe(args []string) error {
 	// so the default set never binds a port that serves nothing.
 	fs.StringVar(&c.providers, "providers", "aws,azure,gcp", "comma-separated providers to start: aws,azure,gcp,oci")
 	fs.StringVar(&c.host, "host", "127.0.0.1", "host/interface to bind (use 0.0.0.0 to expose on the network)")
+	fs.StringVar(&c.advertiseHost, "advertise-host", "",
+		"host/IP the Kubernetes data-plane endpoint is advertised at in kubeconfigs and its TLS cert "+
+			"(default: --host, or 127.0.0.1 when binding all interfaces such as 0.0.0.0 under Docker)")
 	fs.StringVar(&c.awsPort, "aws-port", "4566", "port for the AWS endpoint (HTTP)")
 	fs.StringVar(&c.azurePort, "azure-port", "4568", "port for the Azure endpoint (HTTPS)")
 	fs.StringVar(&c.gcpPort, "gcp-port", "4569", "port for the GCP endpoint (HTTP)")
@@ -120,6 +124,13 @@ func runServe(args []string) error {
 	if c.persist && c.stateFile == "" {
 		return errStateFileRequired
 	}
+
+	// Resolve the host the Kubernetes data plane advertises to clients. The bind
+	// host isn't always reachable: under Docker the container binds 0.0.0.0 (all
+	// interfaces), but 0.0.0.0 is not a connectable address — a kubeconfig
+	// advertising https://0.0.0.0:4570 with a 0.0.0.0-only cert SAN can't be
+	// dialed or TLS-verified from the host. Advertise a routable host instead.
+	advertiseHost := advertiseHostFor(c.advertiseHost, c.host)
 
 	sel, err := parseProviders(c.providers)
 	if err != nil {
@@ -187,7 +198,7 @@ func runServe(args []string) error {
 			// rest.Config from Endpoint plus CertificateAuthority can then
 			// validate what it dials; over plain HTTP that CA certifies
 			// nothing and the handshake fails.
-			k8s.SetBaseURL("https://" + net.JoinHostPort(c.host, c.k8sPort))
+			k8s.SetBaseURL("https://" + net.JoinHostPort(advertiseHost, c.k8sPort))
 		}
 		fresh := make(map[string]http.Handler, len(sel))
 		freshTargets := make(map[string]seed.Target, len(sel))
@@ -415,7 +426,9 @@ func runServe(args []string) error {
 	if k8sBackend != nil {
 		addr := net.JoinHostPort(c.host, c.k8sPort)
 
-		k8sTLS, err := eksprov.ServingTLSConfig([]string{c.host, "localhost", "127.0.0.1"})
+		// The cert must certify the advertised host (what clients dial), plus the
+		// loopback names, plus any extra --tls-host SANs.
+		k8sTLS, err := eksprov.ServingTLSConfig(k8sCertHosts(advertiseHost, c.tlsHosts))
 		if err != nil {
 			return fmt.Errorf("kubernetes data-plane TLS: %w", err)
 		}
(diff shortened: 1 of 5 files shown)


NitinKumar004:
Expanded this PR to close the last two standalone/Docker usability gaps (`0626c2d`), and **verified everything against a real `docker build` + `docker run` of the image** (not just the binary):

**Added:**
- `Dockerfile` `EXPOSE 4571` (OCI) + a commented compose mapping with the `--providers …,oci` note (OCI stays opt-in).
- A **'Trusting the Azure self-signed cert (any language)'** docs section: AWS (:4566) and GCP (:4569) are plain HTTP; only Azure (:4568) is HTTPS with a self-signed cert, so a non-Go client skips verify (`curl -k`, `NODE_TLS_REJECT_UNAUTHORIZED=0`, `verify=False`, `InsecureSkipVerify`) or supplies its own via `--tls-cert`. Flags table now lists `--oci-port`.

**Real-container test results** (`docker run … cloudemu:test serve --host 0.0.0.0 --providers aws,azure,gcp,oci`):
- ✅ AWS S3 create/list · ✅ GCP GCS insert/list
- ✅ Azure over TLS with `-k` → 200 (default verify fails on the self-signed cert, as documented)
- ✅ OCI :4571 listening
- ✅ Kubernetes: EKS advertises `https://127.0.0.1:4570`, `kubectl` connects + TLS-verifies, Deployment reconciles to 2 Running pods

So a user can `docker run` the image and drive all four surfaces end-to-end.

NitinKumar004:
### Deep real-SDK verification (not just curl)

Built the image from this branch, ran the container (`--providers aws,azure,gcp,oci`), and drove **all four providers with their actual production SDKs** — **19/19 operations green**:

- **AWS** (`aws-sdk-go-v2`): S3 CreateBucket / PutObject / GetObject (content round-trips), DynamoDB CreateTable / PutItem / GetItem (item round-trips), EC2 RunInstances / DescribeInstances.
- **GCP** (`cloud.google.com/go/storage`): Bucket.Create, Object write + read (content round-trips).
- **Azure** (`azure-sdk-for-go` armcompute, over TLS with skip-verify per the new docs): Disk CreateOrUpdate + Get (size round-trips).
- **Kubernetes** (real `eks` SDK + `client-go`): EKS Create/DescribeCluster advertises `https://127.0.0.1:4570`, `client-go` builds a rest.Config from the kubeconfig, **TLS-verifies the advertised cert**, creates a Namespace + Deployment, reconciles to **2 ready replicas**, lists 2 materialized Pods.

Confirms a real user can `docker run` the image and point their real SDK code at it across every surface.

cmd/cloudemu/serve.go
```diff
@@ -425,7 +438,9 @@ func runServe(args []string) error {
 			srv:  &http.Server{Addr: addr, Handler: handlerFor(k8sBackend, nil), TLSConfig: k8sTLS},
 			tls:  true,
 		})
-		eps.Kubernetes = fmt.Sprintf("https://%s", addr)
+		// Show the reachable (advertised) endpoint, not the bind address — the
+		// listener binds `addr` (may be 0.0.0.0) but clients dial advertiseHost.
+		eps.Kubernetes = fmt.Sprintf("https://%s", net.JoinHostPort(advertiseHost, c.k8sPort))
```
NitinKumar004, line 443: Nice — the k8s endpoint now advertises the routable host. Low follow-up: the sibling AWS/GCP/OCI banner lines above (and the Azure one below) still use `net.JoinHostPort(c.host, ...)`, so under `--host 0.0.0.0` the banner prints `http://0.0.0.0:4566` etc. Those clients don't self-advertise (so it's display-only, not the functional break k8s had), but a user copying the banner endpoint can't reach `0.0.0.0` on macOS/Windows. Consider routing the display of all endpoints through the same routable-host logic.

cmd/cloudemu/serve.go
```diff
@@ -742,6 +757,62 @@ type namedServer struct {
...
+// loopbackIPv4 is advertised for the Kubernetes data plane when the server binds
+// all interfaces (a published -p 4570:4570 maps to it from the host).
+const loopbackIPv4 = "127.0.0.1"
+
+// advertiseHostFor resolves the host the Kubernetes data plane advertises to
+// clients: an explicit --advertise-host wins; otherwise the bind host, unless
+// that binds all interfaces (0.0.0.0 / :: / empty — e.g. Docker), in which case
+// loopback.
+func advertiseHostFor(advertiseHost, bindHost string) string {
+	if advertiseHost != "" {
```
NitinKumar004, line 769: Low: an explicit `--advertise-host 0.0.0.0` short-circuits here and is returned verbatim, re-introducing a non-routable advertise (the exact thing this fix removes for the derived case). It's user error, but a small guard — e.g. treat an explicit unspecified advertise host the same as the bind-all default (map to loopback) or reject it — would be defensive.

NitinKumar004 (commented):
Reviewed end-to-end — the fix is correct and well-scoped. Sharing verification and two non-blocking follow-ups.

**Verified**
- Build, `go vet`, and the full `cmd/cloudemu` test suite are green; the `advertiseHostFor` / `k8sCertHosts` unit tests pass under `-race`.
- The advertised host consistently drives all three Kubernetes surfaces — the endpoint (`SetBaseURL`), the serving-cert SANs (`k8sCertHosts`), and the startup banner (`eps.Kubernetes`) — so what a client dials matches what the cert certifies. TLS verifies end-to-end.
- `isAllInterfaces` is robust: it covers `""`/`0.0.0.0`/`::`/`[::]` explicitly and falls back to `net.ParseIP(h).IsUnspecified()` (the explicit `[::]` case matters, since `ParseIP("[::]")` is nil). `k8sCertHosts` de-dups correctly (a `127.0.0.1` advertise yields `[127.0.0.1, localhost]`).
- Blast radius is contained: only the Kubernetes data plane self-advertises (via `SetBaseURL`); AWS/Azure/GCP callers pass their own endpoint, and the change touches only `serve.go` + docs.
- Docs/Docker are thorough: `-p 4570:4570` is now published (it was missing, so kubectl couldn't reach the cluster regardless), `EXPOSE` lists 4570/4571, and `--advertise-host` is documented in the flag table.

**Follow-ups (both Low, non-blocking)**
- The startup banner for AWS/Azure/GCP/OCI still shows the bind host — under `--host 0.0.0.0` it prints `http://0.0.0.0:4566` etc., which a user copying it can't reach (notably on macOS/Windows; Linux quietly remaps `connect(0.0.0.0)`). Those clients don't self-advertise, so it's display-only rather than a functional break, but applying the same routable-host display to every banner endpoint would complete this fix's own principle ("advertise a routable host, not the bind address").
- `--advertise-host 0.0.0.0` (an explicit bind-all value) is returned verbatim, re-introducing a non-routable advertise. It's user error, but a defensive guard (re-map or reject an explicit unspecified advertise host) would be tidy.

PR #403, opened 2026-08-15
fix(rgt): tag EC2 instances + Lambda/Secrets/SNS/SQS via Resource Groups Tagging API

### LINKED ISSUE #395: RGT TagResources/UntagResources rejects EC2 instance ARNs (arn:aws:ec2:…:instance/…)
### What

The Resource Groups Tagging API handler routes `TagResources` / `UntagResources` through `resourcediscovery.Engine.TagResourceByARN`, but `parseEC2ARN` (`services/resourcediscovery/tagging.go`) only accepts `vpc`, `subnet`, and `security-group`. Tagging an EC2 **instance** by ARN fails with:

```
InvalidParameterException: tagging EC2 resource type "instance" is not yet supported (Phase 2 covers vpc, subnet, security-group)
```

EC2 instances are one of the most common resource types tagged via the Resource Groups Tagging API, so this is a notable gap when using cloudemu to exercise real tag-management code end-to-end.

### Repro

Real `aws-sdk-go-v2` RGT client pointed at the cloudemu AWS server:

```go
emu := cloudemu.NewAWS()
insts, _ := emu.EC2.RunInstances(ctx, computedriver.InstanceConfig{
    ImageID: "ami-0abcdef1234567890", InstanceType: "t3.micro",
}, 1)

srv := awsserver.New(awsserver.Drivers{
    EC2:               emu.EC2,
    ResourceDiscovery: emu.ResourceDiscovery,
    STS:               true,
    AccountID:         "123456789012",
    Region:            "us-east-1",
})
ts := httptest.NewServer(srv)
defer ts.Close()

// real rgta.Client with BaseEndpoint = ts.URL
_, err := rgt.TagResources(ctx, &rgta.TagResourcesInput{
    ResourceARNList: []string{"arn:aws:ec2:us-east-1:123456789012:instance/" + insts[0].ID},
    Tags:            map[string]string{"Environment": "prod"},
})
// FailedResourcesMap entry: InvalidParameterException — "instance" is not yet supported
```

The same test against a `vpc` ARN (`arn:aws:ec2:…:vpc/…`) passes, so this is specific to the `instance` resource type, not the RGT/STS wiring.

### Expected

`TagResources` on an EC2 instance ARN merges the tags onto the instance (additively, so existing tags survive) and they're visible via `DescribeInstances` — matching the current vpc/subnet behavior.

### Actual

`parseEC2ARN` rejects `instance/…` with `InvalidArgument` → surfaced to the SDK as `InvalidParameterException` in `FailedResourcesMap`.

### Where the change would likely go

- `services/resourcediscovery/tagging.go` — `parseEC2ARN` handles only `netKindVPC / Subnet / SecurityGroup`; add an `instance` case.
- `dispatchTag` currently sends the `ec2` service to `tagEC2Network` (the networking store); instances would instead dispatch to the compute driver's instance tag store.

Compute-side tagging already exists (`providers/aws/ec2/tags.go`), so this looks primarily like plumbing the instance ARN through to it. Happy to send a PR if that'd help.

### Environment

- cloudemu `v2.4.0`

Thanks for cloudemu — the SDK-compat server let real production tag code run end-to-end with zero infra. 🙏

NitinKumar004:
Fixes #395.

## Problem

The AWS Resource Groups Tagging API (`TagResources` / `UntagResources`) routes through `resourcediscovery.Engine.TagResourceByARN`, but the ARN parser + dispatcher only handled **S3, DynamoDB, and EC2 networking** (vpc/subnet/security-group). Tagging an **EC2 instance** ARN failed with `InvalidParameterException` — and instances are one of the most commonly tagged resources. The same gap applied to every other discoverable service that has an AWS tag store.

## Fix

Extends `TagResourceByARN` / `UntagResourceByARN` to also cover:

| Service | Resource types | Notes |
|---|---|---|
| EC2 compute | instance, volume, snapshot, image | `ec2.Mock.TagResource` (by prefixed id) |
| Lambda | function | version/alias qualifiers resolve to the base function |
| Secrets Manager | secret | |
| SNS | topic | subscription ARNs rejected |
| SQS | queue | ARN name resolved to the queue URL via `ListQueues` |

Tagging is additive (existing tags preserved, overlapping keys overwritten), matching the existing S3/DynamoDB/VPC behavior and real AWS.

### Design

These tag methods live on the AWS provider mocks, **not** the shared `services/*/driver` interfaces, and RGT is AWS-only. Rather than widen five interfaces (forcing Azure/GCP to implement AWS-only tag plumbing), the dispatcher **type-asserts each driver to a narrow capability interface** — the same pattern `resourcediscovery` already uses for `KubernetesClusters` / `ScaleSets` / `RelationalDatabases`. No provider changes were needed: the AWS provider already wires all these drivers into its `ResourceDiscovery` engine, so it works end-to-end.

Error mapping is unchanged: unknown ids → `NotFound` → `InvalidParameterException`; unconfigured/incapable driver → `FailedPrecondition` → `InternalServiceException`.

### Deferred (called out honestly)

- **ECR** and **Route53** — discovery currently emits a non-ARN identifier (an ECR registry URI; a bare `zone-<id>`), so `TagResources` can't round-trip the value `GetResources` returns. Needs a discovery-side change first.
- **No tag store at all** — CloudWatch Logs/alarms, ElastiCache, ELB, IAM, RDS, and networking sub-types beyond vpc/subnet/sg. These return a clear unsupported error.

## Tests

- Engine-level: tag / untag / additive-merge / not-found for **every** newly-supported type, plus a parser table covering all ARN shapes (including the versioned Lambda ARN and rejected SNS subscription / KMS / S3-object cases).
- **Real `aws-sdk-go-v2` RGT round-trip**: seeds an instance, function, topic, queue, and secret, tags all five through the live in-memory handler, asserts `FailedResourcesMap` is empty, and verifies each tag landed on its backing store; a KMS ARN still fails with `InvalidParameterException`.

`go build ./...`, `go vet`, and the full `go test ./...` suite pass. Lint adds zero new issues (the 6 pre-existing gocyclo/dupl/goconst/prealloc findings are unchanged from `development` and live in files this PR doesn't touch).

### DIFF
--- a/services/resourcediscovery/tagging.go
+++ b/services/resourcediscovery/tagging.go
@@ -194,8 +194,8 @@ func parseSecretsARN(arn, resource string) (parsedARN, error) {
 }
 
 // parseSNSARN accepts arn:aws:sns:region:account:topic-name. Subscription ARNs
-// (…:topic:subscription-id, which contain a '/') are rejected — they are not a
-// taggable resource.
+// (whose resource segment is "subscription/<id>", containing a '/') are
+// rejected — a subscription is not a taggable resource.
 func parseSNSARN(arn, resource string) (parsedARN, error) {
 	if resource == "" || strings.ContainsRune(resource, '/') {
 		return parsedARN{}, cerrors.Newf(cerrors.InvalidArgument, "expected SNS topic ARN, got %q", arn)
@@ -465,7 +465,8 @@ func (e *Engine) resolveQueueURL(ctx context.Context, name string) (string, erro
 func driverAs[T any](driver any, name string) (T, error) {
 	var zero T
 
-	// A nil interface value, or a typed-nil driver, both mean "not configured".
+	// An unconfigured driver is stored as a nil interface field (the engine
+	// never wraps a typed-nil pointer), so this catches the "not wired" case.
 	if driver == nil {
 		return zero, cerrors.Newf(cerrors.FailedPrecondition, "%s driver not configured on engine", name)
 	}
--- a/services/resourcediscovery/tagging.go
+++ b/services/resourcediscovery/tagging.go
@@ -13,10 +13,29 @@ const (
 	awsServiceS3       = "s3"
 	awsServiceDynamoDB = "dynamodb"
 	awsServiceEC2      = "ec2"
+	awsServiceLambda   = "lambda"
+	awsServiceSecrets  = "secretsmanager"
+	awsServiceSNS      = "sns"
+	awsServiceSQS      = "sqs"
 )
 
-// DynamoDB resource type within an EC2 ARN.
+// DynamoDB resource type within a DynamoDB ARN.
 const awsTypeTable = "table"
+
+// EC2 compute resource types that route to the compute driver's tag store
+// (the networking sub-types below route to the networking driver instead).
+const (
+	ec2TypeInstance = "instance"
+	ec2TypeVolume   = "volume"
+	ec2TypeSnapshot = "snapshot"
+	ec2TypeImage    = "image"
+)
+
+// Lambda / Secrets Manager resource-type prefixes within their ARNs.
+const (
+	lambdaTypeFunction = "function"
+	secretsTypeSecret  = "secret"
+)
 
 // arnParts is the number of colon-separated segments in a fully-qualified
 // AWS ARN: arn:partition:service:region:account:resource.
@@ -24,15 +43,21 @@ const arnParts = 6
 
 // TagResourceByARN merges tags into the resource identified by arn. The arn
 // is parsed to determine the underlying service and resource type, then
-// dispatched to the matching driver's tag-mutation method.
+// dispatched to the matching driver's tag-mutation method. Tagging is
+// additive: existing tags are preserved and overlapping keys are overwritten.
 //
-// Supported in Phase 2:
-//   - AWS S3 bucket: arn:aws:s3:::name
-//   - AWS DynamoDB table: arn:aws:dynamodb:region:account:table/name
-//   - AWS VPC/Subnet/SecurityGroup: arn:aws:ec2:region:account:{vpc,subnet,security-group}/id
+// Supported AWS resources:
+//   - S3 bucket:            arn:aws:s3:::name
+//   - DynamoDB table:       arn:aws:dynamodb:region:account:table/name
+//   - EC2 compute:          arn:aws:ec2:region:account:{instance,volume,snapshot,image}/id
+//   - EC2 networking:       arn:aws:ec2:region:account:{vpc,subnet,security-group}/id
+//   - Lambda function:      arn:aws:lambda:region:account:function:name
+//   - Secrets Manager:      arn:aws:secretsmanager:region:account:secret:name
+//   - SNS topic:            arn:aws:sns:region:account:name
+//   - SQS queue:            arn:aws:sqs:region:account:name
 //
-// Returns InvalidArgument for unsupported services (lambda, ec2 instance, etc.)
-// or unparseable ARNs.
+// Returns InvalidArgument for services/resource types without a tag store
+// (e.g. CloudWatch alarms, IAM, RDS) or for unparseable ARNs.
 func (e *Engine) TagResourceByARN(ctx context.Context, arn string, tags map[string]string) error {
 	res, err := parseSupportedARN(arn)
 	if err != nil {
@@ -55,14 +80,14 @@ func (e *Engine) UntagResourceByARN(ctx context.Context, arn string, keys []stri
 
 // parsedARN holds the fragments TagResourceByARN needs to route a call.
 type parsedARN struct {
-	service      string // "s3", "dynamodb", "ec2"
-	resourceType string // "table", "vpc", "subnet", "security-group", or "" for s3 bucket
-	id           string // bucket name, table name, vpc-id, subnet-id, sg-id
+	service      string // "s3", "dynamodb", "ec2", "lambda", "secretsmanager", "sns", "sqs"
+	resourceType string // "table"/"instance"/"vpc"/"function"/… or "" for s3/sns/sqs
+	id           string // bucket, table, vpc-id, instance-id, function/secret/topic/queue name
 }
 
 func parseSupportedARN(arn string) (parsedARN, error) {
 	// AWS ARN format: arn:partition:service:region:account:resource
-	// resource is either "name" (S3), "type/id", or "type:id".
+	// resource is either "name" (S3/SNS/SQS), "type/id", or "type:id".
 	if !strings.HasPrefix(arn, "arn:aws:") {
 		return parsedARN{}, cerrors.Newf(cerrors.InvalidArgument, "only AWS ARNs are supported, got %q", arn)
 	}
@@ -82,9 +107,17 @@ func parseSupportedARN(arn string) (parsedARN, error) {
 		return parseDynamoDBARN(arn, resource)
 	case awsServiceEC2:
 		return parseEC2ARN(arn, resource)
+	case awsServiceLambda:
+		return parseLambdaARN(arn, resource)
+	case awsServiceSecrets:
+		return parseSecretsARN(arn, resource)
+	case awsServiceSNS:
+		return parseSNSARN(arn, resource)
+	case awsServiceSQS:
+		return parseSQSARN(arn, resource)
 	default:
 		return parsedARN{}, cerrors.Newf(cerrors.InvalidArgument,
-			"tagging service %q is not yet supported (Phase 2 covers s3, dynamodb, ec2 networking)", service)
+			"tagging service %q is not yet supported", service)
 	}
 }
 
@@ -95,11 +128,10 @@ func parseS3ARN(arn, resource string) (parsedARN, error) {
 
 	// arn:aws:s3:::bucket-name → resource = "bucket-name" (taggable).
 	// arn:aws:s3:::bucket-name/object-key → resource = "bucket-name/object-key".
-	// Object-level tagging requires PutObjectTagging, not PutBucketTagging;
-	// reject until Phase 2.x adds object support.
+	// Object-level tagging requires PutObjectTagging, not PutBucketTagging.
 	if strings.ContainsRune(resource, '/') {
 		return parsedARN{}, cerrors.Newf(cerrors.InvalidArgument,
-			"S3 object ARNs are not yet supported (Phase 2 covers buckets only): %q", arn)
+			"S3 object ARNs are not supported (bucket tagging only): %q", arn)
 	}
 
 	return parsedARN{service: awsServiceS3, id: resource}, nil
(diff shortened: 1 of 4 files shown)


services/resourcediscovery/tagging_test.go
```diff
@@ -23,15 +25,25 @@ func TestParseSupportedARN(t *testing.T) {
 	}{
 		{name: "s3 bucket", arn: "arn:aws:s3:::my-bucket", wantSvc: "s3", wantID: "my-bucket"},
 		{name: "dynamodb table", arn: "arn:aws:dynamodb:us-east-1:111:table/MyTable", wantSvc: "dynamodb", wantType: "table", wantID: "MyTable"},
+		{name: "ec2 instance", arn: "arn:aws:ec2:us-east-1:111:instance/i-abc", wantSvc: "ec2", wantType: "instance", wantID: "i-abc"},
+		{name: "ec2 volume", arn: "arn:aws:ec2:us-east-1:111:volume/vol-abc", wantSvc: "ec2", wantType: "volume", wantID: "vol-abc"},
+		{name: "ec2 snapshot", arn: "arn:aws:ec2:us-east-1:111:snapshot/snap-abc", wantSvc: "ec2", wantType: "snapshot", wantID: "snap-abc"},
+		{name: "ec2 image", arn: "arn:aws:ec2:us-east-1:111:image/ami-abc", wantSvc: "ec2", wantType: "image", wantID: "ami-abc"},
```
thzgajendra, line 31: **Low — AMI is parser-only tested.** `image` appears here in the ARN parser table, but unlike instance/volume/snapshot it has no engine-level tag/untag/additive-merge round-trip. The behavior is correct — `ec2TypeImage` routes through the same `computeTagger.TagResource` dispatch as the other three — but an engine round-trip for `ami-…` would close the gap between the tests and the PR's "round-trip for each new type" claim, and guard against a future change to the compute dispatch silently dropping images. Optional, low effort (mirror the instance case).

thzgajendra (commented):
## Review — RGT tagging for EC2 instances + Lambda/Secrets/SNS/SQS

Reviewed the dispatcher/parser in source and ran the full suite + the live aws-sdk-go-v2 RGT round-trip via a test agent. **Clean, well-designed PR — correct, thoroughly tested, blast radius contained. Two low test/doc call-outs, no defect.**

### ✅ Implementation (verified in source)
- **ARN parsers are careful.** `SplitN(arn, ":", 6)` keeps the resource segment whole, so Lambda's `function:name:version` parses correctly (type `function`, then strips the trailing `:version`/`:alias` qualifier to the base name). SNS/SQS reject `/`-containing (subscription) ARNs; Secrets requires a non-empty id; the unknown-service and unknown-EC2-type paths return `InvalidArgument`.
- **The capability-interface design is the right call.** `driverAs[T]` (generic) type-asserts each driver to a narrow tagger interface rather than widening the shared `services/*/driver` interfaces — the same pattern `resourcediscovery` already uses for KubernetesClusters/ScaleSets. Non-AWS providers don't have to implement AWS-only tag plumbing. Error mapping is correct: unknown id → `NotFound` → `InvalidParameterException`; nil/incapable driver → `FailedPrecondition` → `InternalServiceException`.
- **AMI is wired** — `ec2TypeImage` is in both `parseEC2ARN` and the compute dispatch, so it routes through the same `computeTagger.TagResource` path as instance/volume/snapshot.
- **SQS name→URL** resolves via a `ListQueues` scan (fine for the emulator); NotFound if no match.
- **Blast radius** confined to `services/resourcediscovery/` — additive; existing S3/DynamoDB/VPC behavior unchanged; no provider wiring needed (the AWS engine already wires these drivers).

### ✅ Tests (from the run)
- `go build` + `go vet` clean; **82 resourcediscovery tests + 8 SDK subtests pass**, green under **`-race`**; full module green (**270 packages**, no AWS/Azure/GCP regression).
- **Live aws-sdk-go-v2 RGT round-trip** — seeds instance/function/topic/queue/secret, tags all 5 through the live handler, asserts `FailedResourcesMap` empty, then reads **each backing store** to confirm the tag landed, untags and confirms removal; a KMS ARN fails with `InvalidParameterException`. Not status-only.
- **Additive-merge and untag-only-named-keys are genuinely asserted** against the store (non-overlapping key survives, overlapping key overwritten, unknown untag key is a harmless no-op) — not just a 200.

### Notes (neither a blocker)
- **Low (test coverage):** `ec2 image` (AMI) is covered only in the ARN parser table — there's no engine-level tag/untag/merge round-trip for it, unlike instance/volume/snapshot. Behavior is correct (shared dispatch path), so this is a symmetry/coverage gap, not a bug. See inline. (snapshot/subnet/sg are tag-only too — acceptable given shared dispatch.)
- **Doc nit:** the description says 5 pre-existing lint issues; the current baseline is **6** (dupl:2, goconst:1, gocyclo:2, prealloc:1) — all in untouched files (`arn.go`, `engine.go`, `walkers.go`); PR-changed files are lint-clean either way.
- **Informational:** `driverAs`'s `driver == nil` check relies on the engine storing an *untyped* nil for unconfigured drivers (documented in the comment). A *typed-nil* driver would pass the assertion and panic on call — the invariant holds for how the AWS engine wires drivers today; just worth keeping in mind if a driver is ever wired as a typed-nil pointer.

**Verdict:** correct, idiomatic (no interface widening), and well-tested including the live SDK round-trip. Clean from my side — only the optional AMI round-trip test is worth adding.

services/resourcediscovery/tagging.go
```diff
@@ -121,12 +153,64 @@ func parseEC2ARN(arn, resource string) (parsedARN, error) {
...
+	}
+
+	return parsedARN{service: awsServiceSecrets, resourceType: secretsTypeSecret, id: id}, nil
+}
+
+// parseSNSARN accepts arn:aws:sns:region:account:topic-name. Subscription ARNs
+// (whose resource segment is "subscription/<id>", containing a '/') are
+// rejected — a subscription is not a taggable resource.
+func parseSNSARN(arn, resource string) (parsedARN, error) {
+	if resource == "" || strings.ContainsRune(resource, '/') {
```
NitinKumar004: Low — this rejects a subscription only on `/`, but real SNS subscription ARNs are colon-delimited: `arn:aws:sns:region:account:MyTopic:<uuid>` gives a resource segment `MyTopic:<uuid>` (contains `:`, never `/`). So for a real subscription this guard doesn't fire — it's parsed as a topic named `MyTopic:<uuid>` and fails downstream with `NotFound`→`InvalidParameterException` rather than the intended "expected SNS topic ARN" message, and the table test asserts a `subscription/uuid` shape AWS never emits. Reject when the resource contains `:` (topic names can't) so the guard matches the real subscription shape and the test exercises it. Not a correctness issue — subscriptions are rejected either way.

NitinKumar004 (commented):
Verdict: approve — posting as a comment since GitHub blocks approving one's own PR. No blocking issues; one Low follow-up.

Reviewed end-to-end. Correct, well-tested, and architecturally clean — approving.

**Verified**
- Whole-module `go build`, `go vet`, and `go test -race -count=2` on `services/resourcediscovery` and `server/aws/resourcegroupstaggingapi` are green. The real `aws-sdk-go-v2` RGT round-trip seeds an instance/function/topic/queue/secret, tags all five through the live handler, asserts `FailedResourcesMap` is empty, and confirms a KMS ARN still fails with `InvalidParameterException`.
- **Blast radius is contained** — no shared `services/*/driver` interface was widened; the five `*Tagger` capability interfaces live inside `resourcediscovery` and are reached by type assertion (the same pattern already used for `KubernetesClusters`/`ScaleSets`/`RelationalDatabases`), so Azure/GCP are untouched and no provider changes were needed.
- **Parser is panic-safe** — the ARN split is guarded by `len(parts) < arnParts` before indexing, and `splitTypeID` returns `ok=false` rather than panicking on a missing separator.
- **Dispatch guards nil and incapable drivers** — `driverAs[T]` returns `FailedPrecondition` (unconfigured driver, or driver that doesn't satisfy the tagger interface) → `InternalServiceException`; an unknown id → `NotFound` → `InvalidParameterException`.
- **SQS name→URL resolution has no pagination gap** — `ListQueues(ctx, "")` returns every queue from the store (unpaginated at the driver level), so the name scan can't false-404.
- Lambda versioned/aliased ARNs resolve to the base function; S3-object and KMS ARNs are cleanly rejected; the deferred ECR/Route53 and no-tag-store cases are documented honestly.

**Follow-up (Low, non-blocking):** `parseSNSARN` rejects a subscription only when the resource segment contains `/`, but real SNS subscription ARNs are colon-delimited — `arn:aws:sns:region:account:MyTopic:<uuid>` yields a resource segment `MyTopic:<uuid>` (a `:`, never a `/`). So for a real subscription ARN the explicit guard never fires; it's treated as a topic named `MyTopic:<uuid>` and rejected downstream via `NotFound`→`InvalidParameterException` instead of the intended "not a topic ARN" message, and the test asserts a `subscription/uuid` shape AWS doesn't emit. Subscriptions are still rejected either way (so no correctness impact) — but rejecting when the resource segment contains `:` (topic names can't contain colons) would make the guard fire for the real shape and let the test cover it.

NitinKumar004:
Thanks for the review — addressed both low follow-ups + the doc nit in the latest push:

- **SNS subscription guard** (`parseSNSARN`): now rejects a resource segment containing `:` as well as `/`. Real AWS subscription ARNs (`…:MyTopic:<uuid>`) now hit the explicit "expected SNS topic ARN" path instead of falling through to `NotFound`. Topic names can't contain `:` or `/`, so the guard is exact. The parser table now exercises both the real colon shape and the mock's `subscription/<uuid>` shape.
- **AMI round-trip test**: added `TestTagResourceByARN_EC2Image` — engine-level tag / additive-merge / untag, matching instance/volume/snapshot, so the `image` dispatch path is guarded, not just parser-table tested.
- **Doc nit**: corrected the PR description to 6 pre-existing lint issues (added `prealloc:1`); all still in untouched files.

`go build` + the resourcediscovery and RGT suites are green.

thzgajendra (commented):
## Re-review of `de68c85` — both review items fixed, now verified against a LIVE server

Re-reviewed the two changes and this time **booted the real `cloudemu serve` binary and drove RGT with the actual `aws` CLI v2.30.0** (not `go test`). Both fixes are correct and hold end-to-end.

### ✅ The two fixes (verified in source + live)
- **SNS subscription guard** — `parseSNSARN` now rejects on `ContainsAny(resource, ":/")` instead of just `/`. Correct: a topic name can't contain `:` or `/`, so this catches the **real** colon-delimited subscription ARN (`MyTopic:<uuid>`) the old check missed, as well as the mock's `subscription/<uuid>`. Test now exercises both shapes. **Live:** `aws resourcegroupstaggingapi tag-resources` on `arn:aws:sns:…:MyTopic:8a21-uuid` → `FailedResourcesMap` → `InvalidParameterException` "expected SNS topic ARN" — refused with the intended message, not mis-parsed as a topic.
- **AMI round-trip test** — `TestTagResourceByARN_EC2Image` added (tag → additive-merge → untag), closing the symmetry gap with instance/volume/snapshot. Ran green locally.

### ✅ Live server E2E — real `cloudemu serve` + real aws CLI
Booted `serve --providers aws --aws-port 4566`, pointed the CLI at it with `--endpoint-url`, seeded resources and tagged them through the live RGT handler. Every tag confirmed **on the backing store** via each service's own describe/list (or RGT `get-resources` for SNS):

| Case | Result |
|---|---|
| **EC2 instance** — RGT tag `Environment=prod` | ✅ `FailedResourcesMap:{}`; `describe-instances` shows the tag landed |
| EC2 instance — add `Owner=nitin` | ✅ additive merge (both keys on store) |
| EC2 instance — untag `Environment` | ✅ only `Owner` remains |
| Lambda / SQS / Secrets — RGT tag → list-tags/describe | ✅ tag confirmed on each store |
| SNS topic — RGT tag → `get-resources` | ✅ tag present (verified via discovery read path) |
| **KMS ARN** (negative) | ✅ `InvalidParameterException` "service kms is not yet supported" |
| **SNS subscription ARN** (negative, the fix) | ✅ `InvalidParameterException` "expected SNS topic ARN" |

### Verdict
Both review items are resolved correctly, and the feature now has live-server proof: the AWS Resource Groups Tagging API tags EC2 instances (the headline fix) and Lambda/SNS/SQS/Secrets end-to-end with tags landing on the real backing stores, additive-merge and untag behave, and both negative guards (KMS, SNS-subscription) refuse correctly. Nothing outstanding from my side.

(Testing note, not a PR issue: SNS's native `ListTagsForResource` isn't wired — documented collision at `server/aws/sns/handler.go:66` — so SNS tags are read back via RGT `get-resources`, an independent discovery path, rather than the native action.)

thzgajendra (approved):

PR #405, opened 2026-08-18
fix(azure): surface management-API resources in Resource Graph and storage list

### LINKED ISSUE #404: Azure Resource Graph (Docker server) returns 0 rows for resources created via the management/REST API
Thanks for cloudemu — the Azure surface is impressively complete. Filing one behavior gap that blocks using the **Docker/standalone server** to test code that discovers resources via **Azure Resource Graph**.

## Summary
On the Docker image (`ghcr.io/stackshy/cloudemu:2.5`), a resource created through the Azure **management REST API** is **not visible to a Resource Graph query** — nor to the management **list** call — even though a **GET by id** returns it correctly. So an app that inventories a subscription via ARG sees an empty estate against the Docker server.

## Reproduction (verbatim, macOS / Docker Desktop, `--platform linux/amd64`)
```
docker run -d -p 4568:4568 ghcr.io/stackshy/cloudemu:2.5   # Azure @ https://127.0.0.1:4568 (self-signed)

SUB=00000000-0000-0000-0000-000000000000

# 1) Create a storage account via the management API
curl -sk -X PUT \
  "https://127.0.0.1:4568/subscriptions/$SUB/resourceGroups/rg1/providers/Microsoft.Storage/storageAccounts/sablock?api-version=2023-01-01" \
  -H "Authorization: Bearer x" -H "Content-Type: application/json" \
  -d "{\"location\":\"eastus\",\"kind\":\"BlockBlobStorage\",\"sku\":{\"name\":\"Premium_LRS\"},\"properties\":{}}"
# → 200; body echoes kind=BlockBlobStorage, sku.tier=Premium, provisioningState=Succeeded  ✅

# 2) GET by id — works
curl -sk "https://127.0.0.1:4568/subscriptions/$SUB/resourceGroups/rg1/providers/Microsoft.Storage/storageAccounts/sablock?api-version=2023-01-01" -H "Authorization: Bearer x"
# → 200, correct kind/sku  ✅

# 3) Management LIST — empty
curl -sk "https://127.0.0.1:4568/subscriptions/$SUB/providers/Microsoft.Storage/storageAccounts?api-version=2023-01-01" -H "Authorization: Bearer x"
# → {"value":[]}   ❌  (count 0)

# 4) Resource Graph query — empty
curl -sk -X POST "https://127.0.0.1:4568/providers/Microsoft.ResourceGraph/resources?api-version=2021-03-01" \
  -H "Authorization: Bearer x" -H "Content-Type: application/json" \
  -d "{\"subscriptions\":[\"$SUB\"],\"query\":\"resources\"}"
# → {"count":0,"data":[],"totalRecords":0}   ❌
# same 0 for  | where type == 'microsoft.storage/storageaccounts'  (and =~, and mixed-case ==)
```

## Expected
After a management-API `PUT`, the resource should appear in both the management **list** and a **Resource Graph** `resources` query for that subscription (that is how ARG behaves against real Azure).

## What seems to be happening
The ARG engine responds with a well-formed envelope but an **empty inventory** — it looks like `ResourceDiscovery` is populated only by the **in-process typed drivers** (as in `server/azure/resourcegraph/arg_cost_fields_test.go`, which seeds via `cloudP.ResourceDiscovery` / `cloudP.VirtualMachines.CreateVolume` etc.), and the **management-API handlers on the Docker/HTTP server don’t register their creates into that same inventory**. The **data model is fine** — `kind`/`sku.tier` are modelled and echoed; it’s the **inventory wiring** (management create → ResourceDiscovery/ARG) that’s missing over HTTP.

## Impact
Anyone pointing a real Azure SDK/app at the Docker server to test **discovery via Resource Graph** gets a false-empty estate. The in-process Go path works because it seeds `ResourceDiscovery` directly.

## Questions
1. Is populating `ResourceDiscovery` from the HTTP/management-API handlers in scope for the Docker server, or is ARG intended to be driver-seeded (in-process) only?
2. Does `ResourceDiscovery` model a **storage account with `kind`** (BlockBlobStorage / FileStorage / StorageV2)? The shipped ARG test covers disks/VMs/SQL; confirming storage-account `kind` in ARG would let cost/discovery tools assert the block-blob-vs-file distinction end to end.

Happy to help test a fix. Thanks!

NitinKumar004:
Closes #404.

## Problem

On the standalone/Docker server, a resource created through the Azure **management REST API** was returned by **GET-by-id** but was **invisible to a Resource Graph query** and to the management **list** call. An app inventorying a subscription via ARG saw an empty estate.

Root cause: Resource Graph's `subscriptionAllowed` gate rejected the query when the requested subscription didn't match the emulator's configured id — and that id defaulted to a **12-digit AWS-style account number** (`000000000000`), not a valid Azure subscription **GUID**. Any real Azure SDK/CLI/app queries with a GUID (e.g. `00000000-0000-0000-0000-000000000000`), so it was filtered to empty *before the discovery engine was consulted*. The engine, walkers, and wiring were already correct — ARG scoped to the actual configured id (or with no subscription) returned the resource.

## Fix

- **Resource Graph is subscription-transparent (all services).** The emulator serves a single estate and the management plane already accepts any subscription on create (it echoes it into the resource id). Resource Graph now reports that estate under whichever subscription the query scopes to — rewriting both `id` and `subscriptionId` — so "create under sub X, discover under sub X" is consistent for **every** Azure service (it lives in the shared `resourcegraph` handler).
- **Azure subscription defaults to a GUID.** New `--azure-subscription` flag, defaulting to `00000000-0000-0000-0000-000000000000`, decoupled from the 12-digit `--account-id` (which stays for AWS/GCP/OCI). Standalone-server only, so no library-test churn.
- **Storage-account list.** The handler now serves the subscription- and resource-group-scoped list (`{"value":[…]}`) instead of `404` — mirroring the existing disks/vnet handlers (storage was the outlier).

## Verification

End-to-end against the running server (the issue's exact sequence): PUT create → 200; GET-by-id → 200; management **list → 200 with the account** (was 404); ARG with the subscription GUID → **1 row** rendered under that GUID (was 0); type filter → 1; no-subscription → GUID default. Cross-service: a storage account **and** a VNet created via the management API both appear in one ARG query.

- Full `go test ./...`: green (0 failures, `-race` clean on the touched packages).
- Added a real-SDK storage-account list test (subscription + resource-group scope); updated the Resource Graph subscription-scoping test to the single-tenant contract.
- `gofmt`/`go vet`/`golangci-lint` clean (0 new issues); generated coverage docs unaffected (no driver-capability change).
- `docs/standalone-server.md` documents `--azure-subscription`.

### DIFF
--- a/cmd/cloudemu/serve.go
+++ b/cmd/cloudemu/serve.go
@@ -44,30 +44,31 @@ var errUnsupportedSnapshot = errors.New("unsupported snapshot schema version")
 
 // serveConfig holds the resolved serve flags.
 type serveConfig struct {
-	providers       string
-	host            string
-	advertiseHost   string
-	awsPort         string
-	azurePort       string
-	gcpPort         string
-	ociPort         string
-	k8sPort         string
-	accountID       string
-	region          string
-	projectID       string
-	latency         time.Duration
-	tlsCert         string
-	tlsKey          string
-	tlsHosts        stringList
-	endpoints       string
-	admin           bool
-	logReqs         bool
-	quiet           bool
-	shutdownTO      time.Duration
-	persist         bool
-	stateFile       string
-	persistMetaOnly bool
-	initDir         string
+	providers         string
+	host              string
+	advertiseHost     string
+	awsPort           string
+	azurePort         string
+	gcpPort           string
+	ociPort           string
+	k8sPort           string
+	accountID         string
+	azureSubscription string
+	region            string
+	projectID         string
+	latency           time.Duration
+	tlsCert           string
+	tlsKey            string
+	tlsHosts          stringList
+	endpoints         string
+	admin             bool
+	logReqs           bool
+	quiet             bool
+	shutdownTO        time.Duration
+	persist           bool
+	stateFile         string
+	persistMetaOnly   bool
+	initDir           string
 }
 
 // stringList is a repeatable string flag (e.g. --tls-host a --tls-host b).
@@ -94,7 +95,9 @@ func runServe(args []string) error {
 	fs.StringVar(&c.gcpPort, "gcp-port", "4569", "port for the GCP endpoint (HTTP)")
 	fs.StringVar(&c.ociPort, "oci-port", "4571", "port for the OCI endpoint (HTTP)")
 	fs.StringVar(&c.k8sPort, "k8s-port", "4570", "port for the shared Kubernetes data-plane (HTTPS); empty to disable")
-	fs.StringVar(&c.accountID, "account-id", "000000000000", "AWS account ID / Azure subscription ID reported by the emulator")
+	fs.StringVar(&c.accountID, "account-id", "000000000000", "AWS account ID (also GCP/OCI) reported by the emulator")
+	fs.StringVar(&c.azureSubscription, "azure-subscription", "00000000-0000-0000-0000-000000000000",
+		"Azure subscription id reported by the emulator (a GUID; real Azure SDKs/CLIs require one)")
 	fs.StringVar(&c.region, "region", "us-east-1", "default region reported by the emulator")
 	fs.StringVar(&c.projectID, "project-id", "cloudemu-local", "GCP project ID reported by the emulator")
 	fs.DurationVar(&c.latency, "latency", 0, "artificial latency added to every emulated call (e.g. 20ms)")
@@ -230,7 +233,15 @@ func runServe(args []string) error {
 				freshTargets["gcp"] = seed.Target{Storage: cloud.GCS, Database: cloud.Firestore, Secrets: cloud.SecretManager, Compute: cloud.GCE}
 				freshDiscovery["gcp"] = cloud.ResourceDiscovery
 			case "azure":
-				cloud := cloudemu.NewAzure(opts...)
+				// Azure subscriptions are GUIDs, unlike the 12-digit AWS
+				// account id. Give the Azure provider its own subscription so
+				// resource ids and Resource Graph scoping use a real Azure GUID
+				// (WithAccountID is last-wins). Copy opts so the override never
+				// leaks into another provider's option list.
+				azureOpts := make([]config.Option, len(opts), len(opts)+1)
+				copy(azureOpts, opts)
+				azureOpts = append(azureOpts, config.WithAccountID(c.azureSubscription))
+				cloud := cloudemu.NewAzure(azureOpts...)
 				d := azureserver.DriversFrom(cloud)
 				d.K8sAPI = k8s
 				cloud.AKS.SetK8sAPI(k8s)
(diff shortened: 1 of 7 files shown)


thzgajendra (commented):
## Review — surface management-API resources in Resource Graph + storage list

Reviewed the core handlers in source, ran the suites, and **live-drove the issue's exact sequence** against a booted server. **Verdict: correct fix, proven live, no blocker.** The findings are all Low-severity fidelity/consistency notes on the emulator's single-estate model — none are bugs.

### ✅ The fix is correct and proven live
Booted the `serve` binary (azure-only, default GUID) and ran the #404 repro end-to-end:
| Step | Result |
|---|---|
| PUT create storage account | 200 |
| GET-by-id | 200 |
| **Management list (sub- and rg-scoped)** | **200 with the account** (was 404) |
| **ARG query scoped to the same GUID** | **1 row, `id`+`subscriptionId` rewritten to the queried GUID** (was 0) |
| Cross-service (VNet + storage) in one ARG query | count = 2 |

`effectiveSubscription` (single requested sub → that sub, else configured default) + `rewriteSubscription` (swap the `/subscriptions/<x>/` segment, guarded on empty/non-subscription-shaped ids) are clean, and removing the `subscriptionAllowed` empty-gate is the right call — that gate, combined with the old 12-digit AWS-style default, was exactly why a GUID-scoped query filtered to empty. The `--azure-subscription` GUID flag decoupled from `--account-id` is correct (all live ids carried the GUID). The ARG test was updated to the single-tenant contract and is **non-vacuous** (asserts NotZero records + exact `subscriptionId` + id rewritten). Build / vet / `-race` / lint all clean (the 46 lint hits are pre-existing, **0** in changed files); whole-module tests, no regression.

### Low-severity fidelity / consistency notes (all reasonable for a single-estate emulator; flagging for awareness, none block)
1. **Subscription-transparency is total** — I confirmed live that a query scoped to a *random* GUID returns the whole estate rewritten under it. That's the intended design (single estate, no tenant isolation), and the code/comments say so — just be aware the subscription field can't isolate anything in the emulator.
2. **The rg-scoped storage list doesn't filter by resource group.** Confirmed live: created two accounts in different groups, a single-group-scoped list returned **both** (each id stamped with the requested rg). The mock doesn't track a bucket's group, so it can't filter — acknowledged in the code comment. A consumer expecting rg isolation from the storage list won't get it. Same effective behavior as the other handlers that can't track groups.
3. **"Mirrors the disks/vnet handlers" is slightly imprecise.** `disks.serveCollection` **rejects** a subscription-scoped list (`400 "missing resourceGroups segment"` when rg is empty); the new storage list **supports** subscription scope (which is what #404 needed). So storage is actually *more* capable than disks here — not a bug, but it means a **subscription-scoped disks/vnet inventory would still fail**. If per-service subscription-scoped lists (beyond ARG) are meant to work for inventory, disks/vnet have the same gap this PR only closes for storage — worth a follow-up.
4. **`--azure-subscription` isn't validated as a GUID.** A user passing a non-GUID would produce ids real Azure SDKs reject. User-error only; a one-line format check would fail fast, optional.
5. **Multi-subscription ARG query** falls back to the configured default (renders the estate under a sub the caller may not have asked for). Edge case, single-tenant model — fine.
6. Pre-existing (not this PR): the ARG storage `id` is blob-container-shaped (`…/storageAccounts/default/blobServices/…`) from how the discovery engine models the bucket — the PR's subscription rewrite sits correctly on top of it.

### Verdict
Clean, well-tested, and the live drive proves the exact behavior the issue asked for. Docs (`--azure-subscription`) updated. No blocker. The only thing I'd genuinely consider is a follow-up for #3 (subscription-scoped lists on disks/vnet) if service-level inventory needs to match ARG's new transparency; the rest are Low awareness notes.

PR #420, opened 2026-08-21
feat(compat): SDK compatibility matrix across 16 services + wire-gap fixes

NitinKumar004:
## Summary

Adds an evidence-based **SDK compatibility matrix** for CloudEmu and fixes the real wire gaps it surfaced. A shared in-process harness runs real cloud SDKs (aws-sdk-go-v2, azure-sdk-for-go, google-cloud) against the wire server and records per-operation pass/fail; a generator joins those results with the generated coverage model and publishes `docs/compat/`. Coverage spans all 16 core services across AWS, Azure, and GCP.

The matrix is the missing *proof* layer: CloudEmu already speaks the real wire protocols, but nothing showed it. Every green cell is now backed by a live SDK round-trip, and the amber cells double as a fidelity worklist.

## Changes

- **Harness** (`internal/compat/`) — boots the wire server in-process and hands back real SDK clients per cloud (`BootAWS`/`BootAzure`/`BootAzureTLS`/`BootGCP`, `sess.Config()`/`Transport()`/`Endpoint()`, `FakeAzureCred`). `sess.Op(service, op, fn)` records one result per `(provider, service, operation)` and fails the test on error, so the suite is also a regression net.
- **Generator** (`internal/compatgen/`) — runs the suite, joins results with `docs/coverage/coverage.json` on `(service, operation)`, and emits `docs/compat/compat.json` + a rendered `README.md` matrix. Deterministic; not a `go generate` target because it is evidence-based (must run real SDK calls).
- **Compat tests** (`compat/{aws,azure,gcp}/`) — 48 self-contained tests, 16 services × 3 clouds; **445 operations verified** via real SDKs.
- **Wire-gap fixes found by the suite** (each a same-named `<feature>.go` per `docs/STRUCTURE.md`):
  - `fix(sqs)` — route `SendMessageBatch` / `DeleteMessageBatch` / `ChangeMessageVisibility` (driver had them, handler didn't).
  - `feat(route53)` — health-check operations (`health_check.go`) — the handler didn't claim the `/healthcheck` path space.
  - `feat(ecr)` — lifecycle policy + image scan (`lifecycle_policy.go`, `image_scan.go`), with JSON policy-text ↔ driver mapping.
  - `fix(lambda)` — event-source-mapping `LastModified` was an RFC3339 string but the SDK expects an epoch number; also route `UpdateEventSourceMapping`.
- **CI** — a `compat` job that regenerates the matrix and drift-checks `docs/compat/` (mirrors the `tidy` gate); a PR-template checkbox to regenerate docs.

## Provider Coverage

- [x] AWS
- [x] Azure
- [x] GCP

## Notes

- Azure/GCP coverage is intentionally lower where a portable-interface operation has no equivalent in that provider's native SDK (e.g. Firestore has no tables; ElastiCache `Get/Set` are Redis data-plane commands). Those stay honestly amber rather than being forced green.
- No new module dependencies were added.

## Test Plan

- `go test ./...` — all packages pass (0 failures).
- `go test -race ./compat/...` — clean.
- `golangci-lint run` — 0 issues on new code.
- `go run ./internal/compatgen && git diff --exit-code docs/compat/` — matrix is deterministic (drift gate).


NitinKumar004:
## Review resolution — all findings addressed

An end-to-end review of this PR (with focus on the production wire-handler changes) surfaced two correctness issues and one related follow-up. All are now resolved and verified. Details below.

### 1. Data race in Lambda event-source-mapping update — fixed (`e38bce7`)

`providers/aws/lambda/esm.go` `UpdateEventSourceMapping` mutated the **shared stored pointer** in place (`info.FunctionName = …`, `info.BatchSize = …`, `info.State = …`), while `GetEventSourceMapping` and `ListEventSourceMappings` snapshot from that same pointer (`result := *info`). This PR's new `PUT /2015-03-31/event-source-mappings/{uuid}` route made the path reachable over the wire, so a real client polling `Get`/`List` while another calls `Update` (a realistic Terraform / console pattern) could observe a torn struct.

**Fix:** mutate a copy and `Set` it back — `updated := *info; …; m.mappings.Set(uuid, &updated)` — matching the copy-then-set convention already used by `Create`/`Get`/`List` in the same file. `EventSourceMappingInfo` is all scalar fields, so the shallow copy is a genuine deep copy.

**Verified:** a concurrent `-race` repro (Create → 50× `Update` ∥ `Get`+`List`) reports `DATA RACE` on the pre-fix code and passes cleanly after the fix.

### 2. False-positive matrix cell for Azure Cache `UpdateCache` — fixed (`e38bce7`)

The Azure cache compat test recorded `UpdateCache` as green by issuing a second `BeginCreate` (PUT) instead of the real `armredis` PATCH-based update, which the ARM handler did not route. That published a green cell not backed by a real round-trip — contrary to the PR's premise that every green cell is a live SDK call.

**Fix (honest, on both sides):** `server/azure/cache/handler.go` now routes `PATCH` (`Redis.Update`) to the create-or-update path (which already calls the driver's `UpdateCache` when the resource exists), and `compat/azure/cache_compat_test.go` now calls the real `client.BeginUpdate(...)` (a genuine PATCH) and polls to completion. The cell is now legitimately backed by a real PATCH round-trip.

### 3. Sibling shared-pointer race in the Azure cache provider — fixed (`9efd4b3`)

The same in-place-mutation pattern existed in `providers/azure/cache/cache.go` `UpdateCache`, now more prominently reachable via the PATCH route above. Fixed with the same copy-then-set approach (`updated := *cd; …; m.caches.Set(cfg.Name, &updated)`), preserving the `items` data-plane store pointer so cached data survives the update.

**Verified:** concurrent `-race` repro fails on the pre-fix code (`cache.go` in-place writes) and passes after the fix.

Also addressed a consistency nit: added `//nolint:gocyclo` on the ECR dispatch switch, matching the existing SQS handler convention.

### Gate matrix on head

- `go build ./...` · `go vet ./...` — clean
- `go test ./...` (full suite) — all packages pass, 0 failures
- `go test -race ./providers/aws/lambda/... ./providers/azure/cache/... ./compat/...` — clean
- `gofmt` — clean · `go mod tidy` — no diff (no new dependencies)
- `go run ./internal/compatgen && git diff --exit-code docs/compat/` — deterministic (drift gate)
- structure check (`docs/STRUCTURE.md`) — pass

No new lint issues introduced; the remaining findings in the touched files are pre-existing on `development`.

PR #421, opened 2026-08-21
Add OCI Notifications: topics and subscriptions

### LINKED ISSUE #415: OCI Notifications: topics and subscriptions
Implement **OCI Notifications (ONS)** against the existing portable `notification` driver.

Part of #376.

## Scope

Topics (create/get/list/update/delete), subscriptions (create/get/list/update/delete, confirm, unsubscribe) and publishing a message to a topic.

## Design notes

ONS is the closest 1:1 with SNS of any provider, so the portable `Notification` driver should fit with little strain — the interest is in getting the OCI wire shapes right rather than in mapping.

Two OCI specifics: the control plane and data plane sit under different API prefixes (topics and subscriptions vs `PublishMessage` on the topic's endpoint), and a subscription starts `PENDING` until confirmed, with a confirmation token — unlike an SNS subscription that can be auto-confirmed.

Topic delete is asynchronous and returns a work request.

## Acceptance

- Implements `driver.Notification`; OCI-only behaviour via a consumer-side `Extras` interface
- Subscription lifecycle models PENDING → ACTIVE via confirmation

arunesh-j:
## Summary

- Implements **OCI Notifications (ONS)** against the existing portable `notification` driver.
- Topics, subscriptions with the real PENDING → ACTIVE confirmation flow, and message publishing.
- Nothing added to `services/notification/driver` — OCI-only behaviour is a consumer-side `Extras` interface, per the rule set in #393.

Closes #415. Part of #376.

## Changes

- **`providers/oci/notifications/`** — `Mock` over `memstore` implementing `driver.Notification`, guarded by a `sync.RWMutex`.
- **`server/oci/notifications/`** — the `/20181201/` surface. Topics: Create/Get/List/Update/Delete/ChangeCompartment. Subscriptions: Create/Get/List/Update/Delete/Confirm/Unsubscribe/ResendConfirmation/ChangeCompartment. Data plane: PublishMessage.

## The confirmation flow

`CreateSubscription` mints an OCID, sets `lifecycleState: PENDING` and a confirmation token. Real ONS mails the token to the endpoint; the emulator has no channel, so the token rides back in the response body and is dropped once the subscription is ACTIVE. `GET /subscriptions/{id}/confirmation` validates token **and** protocol, flips to ACTIVE, and returns an `unsubscribeUrl` pointing at this emulator's own origin.

`PublishMessage` skips anything not ACTIVE, so **a publish to a PENDING subscription delivers to nobody** — asserted at both driver and wire layers, and visible in the transcript below.

## Two corrections to the issue as written

1. **The prefixes claim was wrong.** #415 says the control and data planes sit under different API prefixes. They don't — real ONS splits them by *host* (the topic's `apiEndpoint`), both on `/20181201`. Implemented as `POST /20181201/topics/{id}/messages`, with every topic reporting the requesting origin as its `apiEndpoint`, so an SDK pointed at the returned endpoint lands back on the same listener.
2. **`DeleteTopic` returns 204, not 202.** Real ONS answers `204` *with* an `opc-work-request-id`. `docs/oci-conventions.md`'s generic 202 is the common case, not ONS's. The work request is still recorded and pollable.

## Never accept-and-ignore

Refused with a named 400 rather than silently dropped: message attributes (ONS has no such field), `definedTags`, protocols outside the ONS set, unknown `messageType`, unsupported `sortBy`/`sortOrder`. Protocol aliases (`email`, `https` → `CUSTOM_HTTPS`) are mapped, not dropped.

## Provider Coverage

- [ ] AWS
- [ ] Azure
- [ ] GCP
- [x] OCI

## Checklist

- [x] All tests pass (`go test ./...`) — see note below

## Test Plan

```
go test ./...                                               271 ok
```

**Correction on a suite note.** An earlier run of this branch showed `cmd/cloudemu TestServeOutOfProcess` failing, and I first attributed it to the sandbox. That was wrong. It is contention on the shared `~/.cloudemu` daemon lock between the six Wave 2 worktrees running their suites in parallel — self-inflicted, not a repo bug. Re-run with nothing else running, the whole suite is exit 0 including that test.

Coverage leak check clean: no OCI operation appears in `docs/coverage/{aws,azure,gcp}/*.md`, and `git diff development -- services/` is empty.

End-to-end on a running server (port 4616):

```
create topic                        -> 200, apiEndpoint = requesting origin
create subscription                 -> 200, lifecycleState PENDING + confirmationToken
publish BEFORE confirmation         -> 200 accepted, delivered to nobody
list                                -> still PENDING
confirm with token                  -> 200 + unsubscribeUrl
get subscription                    -> ACTIVE, token no longer echoed
publish AFTER confirmation          -> 200, delivered
list topics without compartmentId   -> 400 InvalidParameter
delete topic                        -> 204 + Opc-Work-Request-Id
list topics / subscriptions         -> [] , []  (cascade)
poll work request                   -> SUCCEEDED, DELETE_TOPIC, resources[0] DELETED
```

## Left out

No `oci-go-sdk` compat test — the SDK splits ONS across two clients with a host override, needing more scaffolding than it would buy; wire shapes are asserted field-by-field instead. Publishing records deliveries rather than pushing to real HTTPS/email endpoints, matching how `providers/aws/sns` handles non-SQS protocols. `definedTags` is refused rather than modelled, consistent with the rest of the OCI surface.


NitinKumar004 (commented):
## Review notes

### Findings

**Medium · docs — docs/services.md not updated with OCI Notifications — unmet Definition-of-done checkbox**  
`docs/services.md:23`  
If a user consults docs/services.md (the human-facing service catalog) to learn what OCI surface CloudEmu emulates -> they find no ONS topics/subscriptions/publish entry and conclude the service is unimplemented, because the vertical slice's documentation step was skipped even though the code and generated coverage exist.

> git diff merge-base..HEAD shows docs/services.md untouched; only auto-generated docs/coverage/* changed. oci-conventions.md 'Definition of done' requires 'Operations added to docs/services.md', and sibling OCI services each have a hand-written section (OCI Monitoring line 623, OCI identity line 749, OCI VCN note line 407). Notifications has none.

**Low · structure — Per-feature filename not mirrored across provider/wire layers for topics and publish**  
`providers/oci/notifications/publish.go:1`  
If a maintainer greps for the publish feature by filename (publish.go) -> they find only the provider side and miss that the wire implementation lives in topics.go, because the feature file name is not mirrored on the wire layer; a navigation cost, not a behavioral defect.

> STRUCTURE.md §3: 'A feature uses the same filename across all three layers.' subscriptions.go matches (provider+wire), but topic CRUD is provider notifications.go vs wire topics.go, and publish is provider publish.go vs wire (folded into topics.go). provider notifications.go is defensible as the §4 <service>.go core-CRUD file, but publish.go/topics.go breaks the one-grep goal.

NitinKumar004 (commented):
## Review — OCI Notifications

**Verdict: Request changes** — one blocker (persistence completeness), otherwise clean and faithful to the existing OCI convention.

### Blocking

**Mock is not `Snapshottable` — `providers/oci/notifications/notifications.go` (`Mock`).**
The `Mock` holds three `memstore.Store` fields (topics, subscriptions, deliveries) but implements no `snapshot.Snapshottable`, and there's no `snapshot.go`. Every sibling OCI provider ships one (`providers/oci/{identity,vcn,monitoring}/snapshot.go`). On a tree that includes the persistence-completeness guard, `go test ./persist/...` fails:

```
oci: field Notifications (*notifications.Mock) holds a memstore.Store but is not
Snapshottable — add persistence (see #582) or justify in guardExclude
```

Consequence: CI `go test ./...` goes red, and a `serve` stop/start silently drops every topic and subscription.

Fix: add `providers/oci/notifications/snapshot.go` with `var _ snapshot.Snapshottable = (*Mock)(nil)` plus `Snapshot`/`Restore` over topics + subscriptions (decide whether the transient `deliveries` store persists or is justified in `guardExclude`). The branch is also behind `development`, so rebase + re-run `go generate` before merge.

### Non-blocking

- **Create returns `200`, not `201`** — `server/oci/notifications/topics.go` (CreateTopic) and `subscriptions.go` (CreateSubscription). Real ONS returns `201 Created`. SDK/Terraform accept any 2xx, but a strict raw-HTTP client asserting 201 diverges.
- **`if-match`/etag not enforced** — `server/oci/notifications/topics.go` (update/delete). The etag is minted, stored, and returned correctly, but the handler ignores the `if-match` header, so a stale-etag update never returns `412`.
- **No endpoint-format / message-size validation** — `providers/oci/notifications/subscriptions.go` (endpoint) and `publish.go` (body). EMAIL endpoints aren't checked for `@`, CUSTOM_HTTPS endpoints aren't required to be https, and the message body has no 64 KB cap; real ONS `400`s these.

### Verified good

Full lifecycle exercised over the wire (create topic → subscribe → confirm → publish → delete with cascade): computed IDs/timeCreated stable across reads, etag rotates on update, dup→409, missing→404, subscription-under-missing-topic→404, bad protocol→400, delete returns `Opc-Work-Request-Id`, subscription cascade-deleted with its topic. Wiring is complete end-to-end (provider field + `New()` + monitoring wire + registration + `from_provider.go` + work-request store). No check-then-set or shared-pointer aliasing races (`-race` clean; every `Get`/`List` deep-copies). Structure, lint, and `go generate` idempotency all pass.

arunesh-j:
Rebased onto `a45ac909` and addressed everything in `344b484d` (plus `5c328532` from the earlier round). Two rebases, **zero conflicts** both times, and no `services/notification/driver` drift.

### Blocking — `snapshot.go`
Added, mirroring `providers/oci/vcn/snapshot.go`'s shared `storeDump` table. `go test ./persist/...` now **ok**.

I checked the four traps that bit the sibling OCI branches, and none applied here — recording it so the next reviewer doesn't re-derive it:

- **Unexported fields** — none needing export. `topicData` is an unexported *type*, but every field on it is exported, as are `Subscription`, `Message` and `DeliveryPolicy`. No renames, no public API change.
- **Nested stores** — none. No `*memstore.Store` appears inside a stored value, so no custom `MarshalJSON` was needed. This is what broke Compute's `poolData`.
- **Lazily-minted defaults** — no equivalent. ONS mints nothing by default, unlike Vault's `defaultVaultID`.
- **Counters** — no per-mock counter. IDs come from `idgen`'s process-global atomic, shared across every provider and not this snapshot's state; the siblings don't snapshot it either. Confirmation tokens travel as data on the subscription, which is what actually matters.

**`deliveries` is persisted, not excluded**, and `guardExclude` stays empty. It stands in for the endpoint inbox real ONS pushes to and is the only thing `Deliveries()` reads — dropping it would report a restored ACTIVE subscription as having received nothing, which is silent data loss rather than transience.

Round-trip tests cover topics, subscriptions, the topic-subscription cross-reference, delivery history, and **a PENDING subscription still pending after restore, receiving nothing on a pre-confirm publish, and still confirmable with its original token**, plus malformed and empty input.

### Non-blocking — all three fixed

- **201 on create** — `createTopic`/`createSubscription` now return `StatusCreated`; existing 200 assertions updated, pinned by `TestCreateAnswers201`.
- **`if-match` enforced** — new `checkIfMatch`; stale gives `412`/`NoEtagMatch`, an absent header is unconditional, an unknown resource still 404s through the driver. **Scope note: you named only `topics.go`; I applied it to subscriptions too** — same defect, same helper, and leaving one half unenforced would be incoherent. Easy to revert if you disagree.
- **Validation** — `validateEndpoint` (EMAIL needs an `@`; `CUSTOM_HTTPS`/`SLACK`/`PAGERDUTY` need `https://`) and a 64 KB `maxMessageBytes` cap, each rejecting **by name**. Boundary tested: exactly 64 KiB passes, one byte more is a 400.

### A bug found outside the findings — and it overlaps #422

ONS's metric namespace `oci_notification` was being **silently dropped**: `providers/oci/monitoring` rejects `oci_`-prefixed namespaces and `emitMetric` discards the error, so the entire `SetMonitoring` path was dead code. Fixed by splitting the reserved-prefix check off the in-process producer path — `PostMetricData` (the public API, where the rule belongs, rejecting *caller* input) still enforces it; `PutMetricData` (how a sibling mock emits its own service metrics) now admits it.

**#422 independently found and fixed the same bug**, with a slightly different diff. Whichever merges first will conflict the other. Both touch `providers/oci/monitoring`, which neither service branch really owns — **happy to extract it into its own small PR** and rebase both onto it, if you would prefer that to resolving a conflict at merge time.

### Coverage

| package | before | after |
|---|---|---|
| `providers/oci/notifications` | 95.8% | **99.0%** |
| `server/oci/notifications` | 78.6% | **96.2%** |

### Verification

```
go test ./...                                      exit 0
go test ./persist/...                              ok        (the blocker)
go test -race (both notifications packages)        2/2 ok
golangci-lint (both packages)                      0 issues
go generate ./...                                  re-running produces no diff
```

Two lint findings surfaced and were **fixed rather than waived**: `canonicalheader` on my `if-match` (to `If-Match`; `Header.Get` is case-insensitive so wire behaviour is unchanged) and `gocritic hugeParam` on `Publish`, given the same `//nolint` the neighbouring interface methods already carry.

### E2E, port 4616

```
CreateTopic                -> 201 Created, etag-00000003
CreateSubscription         -> 201 Created, PENDING, confirmationToken token-00000009
Publish (PENDING)          -> 200, delivers to nobody
Confirm                    -> 200
Publish (confirmed)        -> 200
if-match fresh             -> 200
if-match replayed          -> 412 NoEtagMatch, and the resource is unchanged
bad EMAIL endpoint         -> 400 "... must hold an @"
CUSTOM_HTTPS over http     -> 400 "... must be an https URL"
body 64KiB+1               -> 400 "message body is 65537 bytes; ONS caps a message at 65536"
body exactly 64KiB         -> 200
DeleteTopic                -> 204 + Opc-Work-Request-Id
work request               -> DELETE_TOPIC / SUCCEEDED / 100%
GET topic / subscription   -> 404 / 404  (cascade)
```

### On the `cmd/cloudemu` flake

An interim run showed `cmd/cloudemu` failing. It is **not this branch** — the diff touches nothing under `cmd/` or `persist/` (verified: 0 files). With nothing else running, `go test ./cmd/cloudemu/` passes on clean `development` in 5.4s. The failures come from parallel worktree suites colliding on the shared `~/.cloudemu` daemon lock.

PR #422, opened 2026-08-21
Add OCI Object Storage: buckets, objects, multipart, PARs

### LINKED ISSUE #410: OCI Object Storage: buckets, objects, multipart
Implement **OCI Object Storage** against the existing portable `storage` driver.

Part of #376. Follows the conventions in `docs/oci-conventions.md`.

## Scope

Namespaces, buckets and objects on the `/n/{namespace}/b/{bucket}/o/{object}` surface: create/get/list/update/delete buckets, put/get/head/delete/rename/copy objects, list with prefix and delimiter, multipart uploads, object versioning, retention rules, and **Pre-Authenticated Requests** (OCI's presigned-URL equivalent).

## Design notes

OCI scopes buckets under a tenancy **namespace** rather than a flat global name, so the path carries `/n/{namespace}` ahead of the bucket. The portable driver keys by bucket name; the namespace belongs in the wire layer.

A PAR is a first-class resource with its own OCID and lifetime, unlike an S3 presigned URL which is a signature. If the portable `Bucket` driver cannot express it, it belongs in the `Extras` capability.

## Acceptance

- Implements `driver.Bucket`; OCI-only behaviour via a consumer-side `Extras` interface in `server/oci/objectstorage`
- OCIDs from `idgen.OCID`

arunesh-j:
## Summary

- Implements **OCI Object Storage** against the existing portable `storage` driver.
- Namespaces, buckets, objects, multipart, versioning, retention rules, lifecycle policies, and Pre-Authenticated Requests.
- Nothing added to `services/storage/driver` — OCI-only behaviour is a consumer-side `Extras` interface, per the rule set in #393.

Closes #410. Part of #376.

## Changes

- **`providers/oci/objectstorage/`** — `Mock` over `memstore` implementing `driver.Bucket`, guarded by a `sync.RWMutex`: namespace derivation, buckets, objects, versioning, multipart, retention, PARs.
- **`server/oci/objectstorage/`** — the `/n/{namespace}/b/{bucket}/o/{object}` surface plus PAR redemption at `/p/{token}/n/…`.

## Operations

Namespace + metadata; bucket create/get/head/update/delete/list; object put/get/head/delete/list (prefix, delimiter, paging); rename and `updateObjectStorageTier`; async `copyObject`; multipart create/upload/list-parts/commit/abort/list; object versions (Disabled/Enabled/Suspended, version-addressable GET/HEAD/DELETE, delete markers, `objectversions`); retention rules with lock semantics; lifecycle policy PUT/GET/DELETE; PARs including **redemption**.

## PARs are modelled as real resources

A PAR gets its own OCID, `timeExpires`, an opaque redemption token, and is listable and revocable — not a fabricated signature. `GeneratePresignedURL` creates one, so the URL it returns actually works and can be revoked. Demonstrated in the transcript: read-only PAR serves `GET`, refuses `PUT` with 403, and 404s after revocation.

## Compartment scoping — one deliberate narrowing

Only `ListBuckets` calls `RequireCompartmentID`; it is the one collection real OCI scopes by compartment. Object, PAR, retention and upload lists are bucket-scoped and take no `compartmentId`, so requiring it would reject calls real OCI accepts. `CreateBucket` requires it in the body. Stated in the package doc.

## Never accept-and-ignore

Rejected by name rather than silently dropped: bucket policies (→ Identity policies), CORS, object tags (→ `opc-meta-`), `reencrypt`, `restoreObjects`, multipart `partsToExclude`, cross-namespace copy, more than one lifecycle `inclusionPrefix`, disabling encryption, versioning back to `Disabled`, and unknown access types / tiers / time units.

`reencrypt` is a 501 rather than a 202: with no per-object key material there is nothing to re-wrap, and a 202 there would be theatre.

## Provider Coverage

- [ ] AWS
- [ ] Azure
- [ ] GCP
- [x] OCI

## Checklist

## Test Plan

```
```

Coverage leak check clean: no OCI operation appears in `docs/coverage/{aws,azure,gcp}/*.md`, and `git diff development -- services/` is empty.

End-to-end on a running server (port 4611):

```
GET  /n                                     -> namespace 30aa137d4542
POST /n/{ns}/b                              -> 200, ocid1.bucket.oc1.iad..., etag, Standard tier
PUT  /o/holiday/beach.txt (opc-meta-owner)  -> 200, Etag + Opc-Content-Md5 + Opc-Meta-Owner
GET  /o/holiday/beach.txt                   -> 200 "sun and sea", metadata round-trips
GET  /o?prefix=holiday/&delimiter=/         -> 200, one object
GET  /b?compartmentId=...                   -> 200, one bucket
delete object / bucket / get deleted        -> 204, 204, 404

POST /n/{ns}/b/docs/p                       -> 200, PAR with accessUri
GET  /p/{token}/n/{ns}/b/docs/o/report.txt  -> 200 "quarterly"
PUT  (same URI, read-only PAR)              -> 403
DELETE the PAR, then GET                    -> 204, then 404
POST /b/docs/actions/copyObject             -> 202 + Opc-Work-Request-Id
poll work request                           -> SUCCEEDED, COPY_OBJECT, object CREATED
```

## Note on the suite

Other Wave 2 branches see a pre-existing failure in `cmd/cloudemu TestServeOutOfProcess` ("AWS endpoint never became ready") which reproduces on clean `development` — it spawns a child process that cannot bind a listener in a sandbox. It passed on this run; flagging it as flaky-environmental rather than related to this change.


NitinKumar004 (commented):
## Review notes
**Real data-plane engine (per #427):** N/A — OCI Object Storage is object storage, which is explicitly NOT engine-eligible under the #427 model (ComputeEngine/DatabaseEngine/CacheEngine/FunctionEngine/ContainerEngine).

### Findings

**High · wire-fidelity — GetNamespaceMetadata (GET /n/{ns}) is misrouted to ListBuckets when the tenancy namespace begins with 'b'**  
`server/oci/objectstorage/handler.go:287`  
If a user's tenancy hashes to a namespace starting with 'b' -> every GetNamespaceMetadata call (GET /n/{ns}) returns 400 'compartmentId is required' instead of the metadata body -> S3/Swift-compat setup and any SDK flow that reads namespace metadata breaks non-deterministically for ~6% of tenancies, while the identical code works for everyone else. Fix: carry a 'hadBucketSegment' bool out of parseNamespaced instead of re-sniffing the raw path.

> serveBucketCollection distinguishes /n/{ns} from /n/{ns}/b with strings.Contains(r.URL.Path, "/"+segBuckets) (segBuckets=="b"). The parsed route already discards whether a /b segment was present, so this substring heuristic is the only signal. For a namespace whose first char is 'b', the path "/n/b25a97828eed" contains the substring "/b", so the request falls through to the GET branch = listBuckets, which requires compartmentId. Reproduced with tenancy "ocid1.tenancy.oc1..probe0100" -> namespace

**Medium · docs — docs/services.md not updated with OCI Object Storage (Definition-of-done item)**  
`docs/services.md:59`  
If a reader consults docs/services.md to learn which OCI services cloudemu emulates -> Object Storage is invisible there even though it is fully implemented -> the service is undiscoverable via the canonical human-facing doc, and the PR fails a stated done-criterion (a future migration-drift CI check keys off this file).

> docs/oci-conventions.md Definition of done requires 'Operations added to docs/services.md'. The storage section header (line 59) still reads 'AWS: S3 | Azure: Blob Storage | GCP: GCS' with no 'OCI: ObjectStorage' entry, unlike networking (line 407 'OCI: VCN ...'), OCI Monitoring (623) and Identity (670) which each have OCI subsections. Only the generated docs/coverage/* files were updated in the diff; the hand-maintained services.md was not touched.

**Medium · coverage — Changed packages below the 90% pillar; lifecycle-expiry and versioning wrappers untested**  
`providers/oci/objectstorage/retention.go:366`  
If EvaluateLifecycle mis-evaluates expiry (e.g. off-by-one on the ExpirationDays*24h window, or a prefix mismatch) -> a caller relying on lifecycle-expiry reporting gets wrong object lists -> the regression ships silently because no test exercises the aged-out branch. Add a FakeClock test that advances past ExpirationDays and asserts the expired name, plus a HeadObjectVersion and portable-versioning round-trip.

> go test -cover: providers/oci/objectstorage 66.2%, server/oci/objectstorage 71.6% (pillar target 90%). 0%-covered flows include EvaluateLifecycle + objectExpired (retention.go:366/399 — object age-out evaluation is never asserted by any test), HeadObjectVersion (versioning.go:228), and the portable SetBucketVersioning/GetBucketVersioning wrappers (versioning.go:140/159). TestBucketLifecycleCRUD stores and reads a policy but never drives an object past its ExpirationDays to confirm EvaluateLifecy

**Low · wire-fidelity — ListObjects default page size is 100, not OCI's 1000**  
`server/oci/objectstorage/object.go:233`  
If a client lists a bucket of 500 objects without an explicit limit -> it receives 100 + a nextStartWith cursor and must paginate 5x -> extra round-trips and a subtle behavioral difference from real OCI. Minor and partly a shared-ocirest.DefaultLimit choice; note only.

> listOptions sets MaxKeys: ocirest.Limit(r), which returns ocirest.DefaultLimit (100) when 'limit' is absent and is always >=1, so the provider's defaultListLimit (1000, matching real OCI) at object.go:15 is never reached from the wire path. Real OCI ListObjects returns up to 1000 per page by default.

**Low · coverage — No oci-go-sdk SDK-compat test**  
`server/oci/objectstorage/handler_test.go:1`  
If the handler's wire shape drifts from what the real SDK emits/expects (header casing, envelope fields) -> hand-rolled tests that mirror the handler's own assumptions won't catch it -> a real SDK user hits the mismatch first. Recommended, not blocking.

> Convention: 'An SDK-compat test using github.com/oracle/oci-go-sdk against httptest.NewServer is the strongest evidence the handler is right. Add one where the SDK makes it practical.' All wire tests are hand-rolled httptest requests; no oracle/oci-go-sdk client is driven against the handler.

arunesh-j:
Rebased onto current `development` (now `58787e9a`) and addressed the findings in `f2854465`.

### High — `GET /n/{ns}` misrouted when the namespace starts with `b`
You were right about the cause and the fix. `route` now carries `HasBucketSeg`, set in `parseNamespaced` when a `/b` segment is genuinely parsed; `serveBucketCollection` branches on that instead of `strings.Contains(r.URL.Path, "/b")`.

Regression test `TestNamespaceMetadataWithBPrefixedNamespace` uses your exact repro tenancy `ocid1.tenancy.oc1..probe0100` → namespace `b25a97828eed`. Reverting the fix:
```
--- FAIL: TestNamespaceMetadataWithBPrefixedNamespace
    expected: 200   actual: 400
    {"code":"InvalidParameter","message":"compartmentId is required"}
```

Confirmed live on the built binary, including that the collection is **not** over-corrected:
```
GET /n                  -> "b25a97828eed"
GET /n/b25a97828eed     -> 200 {"namespace":"b25a97828eed","defaultS3CompartmentId":...}
GET /n/b25a97828eed/b   -> 400 compartmentId is required   (still required, as it should be)
```

### Medium — `docs/services.md`
Added an OCI Object Storage subsection under §1 Storage in the VCN/Monitoring/Identity format: a 36-row operation→route table plus prose on compartment-scoped `ListBuckets`, the 1000 page size, async `copyObject`, the versioning tri-state, retention locking, the storage-engine seam, and the two unemulated actions.

### Medium — coverage
**66.2% → 93.9%** provider, **71.6% → 94.9%** wire. Every gap you named is covered: `EvaluateLifecycle`/`objectExpired` via a `config.FakeClock` test that steps to one hour short of the window (asserts nothing expires) and then past it (asserts exactly the enabled `logs/` names), plus `HeadObjectVersion` and a portable `SetBucketVersioning`/`GetBucketVersioning` round-trip.

### Low — `ListObjects` page size
`listLimit` returns 0 when no `limit` is present, letting the provider's `defaultListLimit` (1000) apply. `ocirest.DefaultLimit` is untouched for other services. Live: 120 objects with no limit → all 120 and an empty cursor; `limit=50` → 50 plus a cursor.

### Low — SDK-compat test: not added, deliberately
`oracle/oci-go-sdk/v65` requires a signed request built from an RSA keypair and a `ConfigurationProvider`. This emulator verifies no signature, so the client would mostly exercise its own signing rather than the handler's wire shape, and every sibling OCI service here is tested with hand-rolled `httptest` requests. I judged the dependency disproportionate. Instead I added `contrib/realengine/blobstore/oci_test.go` — a real over-HTTP E2E through `httptest.NewServer` plus the filesystem engine — with that reasoning recorded in its doc comment. Happy to revisit if you'd rather have the SDK dependency.

### On the "engines N/A" note — I think Object Storage *is* engine-eligible
Your N/A referred to #427's compute-shaped engines, but `services/storage/storageengine` is a different seam, and its own package doc named AWS/Azure/GCP with OCI the only storage provider not wired to it. Now wired in `providers/oci/objectstorage/engine.go` across put, get, delete, rename, copy, multipart-commit and every versioned path, keyed by version id like S3.

Two design notes: the engine call happens **under** the write lock (S3 nils the byte slice after releasing it, which races a concurrent reader), and `objectData`/`objectVersion` gained a `Size` field so `Head`/`List`/`approximateSize` stay correct once bytes are offloaded.

### Three defects found that were not in the review
1. **Object Storage metrics were silently dropped.** The provider publishes to `oci_objectstorage`; `providers/oci/monitoring` rejected any `oci_`-prefixed namespace as Oracle-reserved, and `emitMetric` discards its error — so no metric was ever recorded. The reservation exists to stop *customer* posts, so it now applies to `PostMetricData` (customer-facing) and not to `PutMetricData` (the seam Oracle's own emulated services use). **This touches an already-merged service, so say the word and I'll split it into its own PR.**
2. **`HeadObject` returned no `Content-Length` and `Content-Type: application/json`** (inherited from the JSON helper). A HEAD has no body, so SDKs read the size from the header. Now served by a dedicated `writeHead`.
3. **Persistence guard.** Upstream #582 added `TestSnapshotCompleteness`, which failed the suite for this provider. Added `snapshot.go` plus round-trip tests covering buckets, objects, version chains, PARs **with their redemption tokens**, retention rules and lifecycle — so an access URI issued before a snapshot still redeems after restore.

### One scope addition for your call
The tenancy was unreachable from the CLI (AWS account-id, Azure subscription and GCP project all are), so the `b`-prefixed namespace could not be driven through the shipped binary at all. Added `--oci-tenancy` (`server/serveflags`, its pinned flag-name guard, and `docs/standalone-server.md`). Cleanly separable if you'd rather keep this PR narrow.

### Verification
```
go build ./...                                   clean
go test ./...                                    exit 0, 342 packages
go test -race ./providers/oci/... ./server/oci/...   11/11 ok
golangci-lint (all changed packages)             0 issues
go test -cover  provider 93.9%   wire 94.9%
go generate ./...                                docs/coverage committed
no OCI op in docs/coverage/{aws,azure,gcp}/*.md  confirmed
```
The only remaining lint in `./providers/oci/...` is 7 pre-existing `gocritic` findings in `vcn/nsg.go` and `identity/portable.go`, which reproduce on clean `development`.

E2E on port 4611 also exercised: bucket CRUD and the `versioning:"On"` rejection, object put/get/head/list with prefix+delimiter, version list / read-by-version / delete markers, rename, async copy (202 + `opc-work-request-id`), storage tier, both 501s, multipart including the `partsToExclude` 501, PARs (redeem, write-refused, wrong-object refused, list, delete), retention (create/get/update, `MONTHS` refused, object held, delete) and lifecycle (`YEARS`→365 days, unknown action refused, multi-prefix refused). Happy to paste the full transcript if useful.

NitinKumar004 (commented):
## Review — OCI Object Storage

**Verdict: Approve (no blockers).** This is the largest and highest-risk surface here (object bytes + multipart + PARs), and it's exemplary — it reuses the portable `storagedriver.Bucket` interface exactly as S3/GCS/Azure do, stores object bytes in-memory the way S3 does (no external engine required; optional engine offload), and follows the existing `ocirest` + `server/oci/*` conventions precisely.

### Non-blocking

- **Enabling versioning on a bucket that already has objects drops the pre-existing object** — `providers/oci/objectstorage/versioning.go:18-41` (`setVersioningLocked`/`storeObjectLocked`). Reproduced: create unversioned bucket → PUT object → enable versioning → PUT overwrite → `objectversions` lists only the new version; the original is neither a `null` version nor recoverable. Real OCI retains the pre-existing object as the `null` version and preserves it on the next overwrite. The common IaC path (`versioning = "Enabled"` declared at bucket create) works perfectly — this only bites a create-then-enable sequence.
- **Object ETag is content-addressed (sha256 of bytes)** — `providers/oci/objectstorage/objectstorage.go:327` (`objectETag`). Real OCI mints a fresh opaque ETag per PUT; here an identical re-PUT yields the same ETag, and two distinct objects with identical content share one. A client using If-Match/If-None-Match on ETag could see a stale-looking match after an identical re-upload.
- **Generic OCI error codes from the shared codec** — `server/wire/ocirest/ocirest.go:75-98` (`statusFor`). Bucket-not-empty → `409 IncorrectState` and duplicate create → `409 Conflict`, where real OCI uses `BucketNotEmpty` / `BucketAlreadyExists`. Pre-existing shared behavior (not introduced here); noting because it now affects object-storage fidelity for a client branching on the exact code string.
- **`CompleteMultipartUpload` assembles the whole object via repeated append under the global write lock** — `providers/oci/objectstorage/multipart.go:155-164`. A multi-GB commit does a large memcpy while holding `m.mu`, briefly blocking unrelated buckets. Bounded by uploaded data and consistent with the cohesive-mock single-mutex design; acceptable.

### Verified good

Full lifecycle over the wire: get-namespace → create bucket → **PUT object bytes → GET/HEAD bytes match** with stable ETag/opc-content-md5 across GET/HEAD/List → **multipart create/upload-parts/commit assembles in ascending part order** → abort invalidates the upload → **PAR create + redeem returns bytes, read-only PAR write → 403, GET PAR omits accessUri** → versioning enable + overwrite lists versions with delete markers (newest-first, isLatest) → delete → 204 → GET-after-delete → 404 → delete non-empty bucket → 409. Computed IDs/namespace/etag/timeCreated stable across reads; `Snapshottable` present and auto-discovered (persistence round-trips objects + versions + PAR tokens); object size / part count / list page size all bounded; single-mutex spans check-then-set on every create (no TOCTOU race); reads deep-copy; `-race` clean; structure/lint/`go generate` and the full CI matrix all green.

PR #424, opened 2026-08-21
Add OCI NoSQL Database: tables, rows, indexes

### LINKED ISSUE #412: OCI NoSQL Database: tables, rows, indexes
Implement **OCI NoSQL Database Cloud Service** against the existing portable `database` driver.

Part of #376.

## Scope

Tables (create/get/list/alter/drop with DDL), rows (get/put/delete/multi-delete), indexes, table limits and on-demand vs provisioned capacity, and TTL.

## Design notes

OCI NoSQL is DDL-driven — `CreateTable` takes a SQL-ish statement rather than an attribute list, and table changes are asynchronous, so mutations return a **work request**. The portable `Database` driver models DynamoDB-style key/attribute tables; map honestly and put what does not fit in the `Extras` capability.

There is no DynamoDB-Streams equivalent. Leave change streams out and say so.

## Acceptance

- Implements `driver.Database`; OCI-only behaviour via a consumer-side `Extras` interface
- Async table mutations record a work request and stamp `opc-work-request-id`

arunesh-j:
## Summary

- Implements **OCI NoSQL Database Cloud Service** against the existing portable `database` driver.
- Includes a real DDL parser, since OCI's `CreateTable` takes a SQL-ish statement rather than an attribute list.
- Nothing added to `services/database/driver`; OCI-only behaviour is a consumer-side `Extras` interface, per #393.

Closes #412. Part of #376.

## Changes

- **`providers/oci/nosql/`** — `Mock` over `memstore` implementing `driver.Database`, guarded by a `sync.RWMutex`, plus the DDL parser and a query evaluator.
- **`server/oci/nosql/`** — the `/20190828/` surface.

Operations: tables (Create/List/Get/Update/Delete/ChangeCompartment), indexes (Create/List/Get/Delete), rows (GetRow/UpdateRow/DeleteRow), and Query. All 24 portable `Database` methods implemented.

## DDL — what is parsed, what is refused

Parsed: `CREATE TABLE` and `ALTER TABLE` — scalar and `JSON` types, `NOT NULL`, `DEFAULT`, `PRIMARY KEY` with optional `SHARD`, `USING TTL <n> DAYS`, `ADD`/`DROP` on non-key columns, `IF NOT EXISTS`.

Refused **by name**, never silently accepted:

| Refused | Why |
|---|---|
| Primary keys wider than two columns, composite shard keys | The portable partition/sort pair cannot identify such a row. This is a **correctness** rejection, not a shortcut |
| `ARRAY` / `MAP` / `RECORD` / `ENUM` | Not modelled |
| generated / `MR_COUNTER` / `UUID` modifiers | Not modelled |
| `USING TTL … HOURS` | OCI's own `Schema` reports TTL only in days |
| `MODIFY`, schema freezing, JSON-path index keys | Not modelled |
| Query projections, aggregates, joins, `ORDER BY`, range conditions | Not modelled |

## Judgement calls

- **No change streams.** OCI publishes no DynamoDB-Streams equivalent, so `UpdateStreamConfig`/`GetStreamRecords` return `Unimplemented` rather than an empty iterator. Noted in `docs/services.md`.
- **Multi-delete is `DELETE FROM … WHERE` over `/query`** — OCI's REST API has no `MultiDelete` operation; this is the real mechanism.
- **`ListIndexes` requires `compartmentId`** although real OCI marks it optional, so no list is ever unscoped. Documented.
- **501, named**, for `/tables/{id}/usage` and `/query/prepare`, `/query/summarize`. `usage` is omitted from row and query responses for the same reason — zeros would read as real telemetry.

## Provider Coverage

- [ ] AWS
- [ ] Azure
- [ ] GCP
- [x] OCI

## Checklist

## Test Plan

```
```

Coverage leak check clean: no OCI operation in `docs/coverage/{aws,azure,gcp}/*.md`; `git diff development -- services/` empty.

End-to-end on a running server (port 4613):

```
CreateTable                        -> 202 + Opc-Work-Request-Id
GetTable                           -> schema, primaryKey [id,email], shardKey [id], ttl 0
UpdateRow / GetRow                 -> 200 / value round-trips
ListTables                         -> 1 table; other compartment -> []; no compartmentId -> 400
CreateIndex                        -> 202 + work request; index ACTIVE
SELECT * FROM users                -> 3 rows
DELETE FROM users WHERE id = 2     -> NumRowsDeleted 2
DeleteRow twice                    -> 200, then isSuccess false
TRUNCATE TABLE                     -> 400 naming "TRUNCATE TABLE"
3-column PRIMARY KEY               -> 400 naming the limit
ARRAY(STRING)                      -> 400 naming "ARRAY"
USING TTL 6 HOURS                  -> 400 naming "HOURS"
ON_DEMAND + maxReadUnits           -> 400
undeclared column                  -> 400
SELECT name FROM users             -> 400 naming projections
GET /tables/users/usage            -> 501
DeleteTable                        -> 202; GetTable -> 404; ListTables -> []
USING TTL 7 DAYS                   -> schema.ttl 7; row carries timeOfExpiration
```

## A note on parallel worktrees

An earlier run of this branch failed `cmd/cloudemu TestServeOutOfProcess`. It is not this change — `cmd/cloudemu` is untouched by the diff. Six Wave 2 worktrees were running their suites concurrently and that test contends on the shared `~/.cloudemu` daemon lock. Verified: with nothing else running it passes, and the full suite is exit 0. Worth knowing as shared-state fragility whenever suites run in parallel.


NitinKumar004 (commented):
## Review notes

### Findings

**Medium · real-engine — Table names are globally unique, not per-compartment**  
`providers/oci/nosql/nosql.go:417`  
If a user creates table "orders" in compartment A and then "orders" in compartment B -> the second CreateTable returns 409 AlreadyExists, because table names are a global key rather than compartment-scoped; real OCI scopes NoSQL table names per compartment and both creates succeed, so a multi-compartment SDK/CLI test that reuses a table name across compartments breaks against the emulator only.

> m.tables is keyed by table name globally (nosql.go:175); newTable rejects with AlreadyExists on m.tables.Has(name) (nosql.go:417), and CreateOCITable checks m.tables.Get(d.Table) irrespective of spec.CompartmentID (table_extras.go:38-44). resolve()/lookup() also address tables by name globally.

**Low · coverage — Portable Query/Scan operators and typed coercion below the 90% coverage pillar**  
`providers/oci/nosql/rows.go:324`  
If a user issues a portable Query/Scan with a relational sort condition (>, BETWEEN) or stores a LONG/FLOAT/BOOLEAN-typed row -> the ordering (compareOrdering/compareStrings) and coercion (parseTyped/convertTyped) execute for the first time in production with no test guarding them, so a comparison- or coercion-logic regression would ship undetected.

> go test -cover: provider 85.8%, server 83.4% (both < 90%). compareOrdering (rows.go:324) 0%, compareStrings (rows.go:341) 0%, compareOp (rows.go:306) 28.6%, parseTyped (row_extras.go:211) 35.7%, convertTyped (row_extras.go:242) 36.8%, SetMonitoring (nosql.go:194) 0%. Query tests exercise only SortOp "="; <, >, <=, >=, BETWEEN and CONTAINS/BEGINS_WITH ordering are never run.

**Low · wire-fidelity — No oci-go-sdk SDK-compat test for the handler**  
`server/oci/nosql/handler_test.go:94`  
If the real oci-go-sdk nosql client shapes a request/response field differently than the hand-modeled types.go structs (e.g. a body vs query-parameter for the Query `limit`, or a header the SDK requires) -> the divergence is not caught by the current tests, so a genuine client could fail at a step the httptest cases pass; adding one SDK create/get/query round-trip would close this.

> handler_test.go drives the handler via raw httptest requests only; the only reference to github.com/oracle/oci-go-sdk is the package doc comment in handler.go. oci-conventions.md calls an SDK round-trip against httptest.NewServer "the strongest evidence the handler is right" (recommended, not required).

NitinKumar004 (commented):
## Review — OCI NoSQL Database

**Verdict: Request changes** — one blocker (persistence completeness), plus two local-gate failures and a rebase. The wire layer and lifecycle are strong; once these are addressed it's mergeable.

### Blocking

**Mock is not `Snapshottable` — `providers/oci/nosql/nosql.go:168` (`Mock`).**
The `Mock` holds `tables` and `names` `memstore.Store` fields, and each `tableData` nests an `items` row store, but there's no `providers/oci/nosql/snapshot.go` and no `snapshot.Snapshottable`. On a tree that includes the persistence-completeness guard:

```
--- FAIL: TestSnapshotCompleteness/oci
    oci: field NoSQL (*nosql.Mock) holds a memstore.Store but is not Snapshottable
```

Consequence: CI `Test` goes red, and a snapshot/restore of the OCI provider silently drops **every table and all row data**. Fix: add `snapshot.go` with `var _ snapshot.Snapshottable = (*Mock)(nil)` + `Snapshot`/`Restore` covering `tables` **including each table's nested `items` rows** (the easy-to-forget part), `names`, and per-table `Scope`/`Tags`/`Indexes`/`ttl`. Note `tableData`'s unexported fields (`items`, `ttl`, `Scope`) must be promoted/handled to round-trip, mirroring `providers/oci/vcn/snapshot.go`.

### Should fix

- **4× golangci-lint `gocritic hugeParam`** — `nosql.go:381` (CreateTable), `rows.go:102` (UpdateItem), `rows.go:206` (Scan), `indexes.go:12` (CreateIndex). These are interface-fixed signatures; add the `//nolint:gocritic // hugeParam: interface method signature cannot be changed.` that `providers/aws/dynamodb` and this PR's own `Query`/`CreateOCITable` already carry. The `golangci-lint run` 0-issues bar (per CLAUDE.md) currently fails.
- **Generated docs stale on rebase** — `docs/coverage/oci/nosql.md`. `go generate ./...` on the merged tree drops an "Optional capabilities / TableAttributes" block the current coveragegen no longer emits; re-run `go generate` after rebase and commit.
- **Branch is 781 commits behind** `development` — rebase; only generated `docs/coverage/README.md` conflicts (code auto-merges).

### Non-blocking

- **Query numeric-literal formatting** — `query.go` (`rowMatches`) / `rows.go` (`compareOp`). Comparisons stringify both sides via `fmt.Sprintf("%v", …)`, so a `DOUBLE` `5.0` vs literal `"5.0"` (or a stored `5`) can mismatch on formatting. Documented SQL subset; the exact-match integer path is verified. Minor fidelity edge.

### Verified good

Full lifecycle over the wire (CreateTable DDL → 202 + work request resolving SUCCEEDED synchronously → GetTable ACTIVE with full schema → **UpdateRow → GetRow round-trips `id:1` as an integer, not `1.0`** → Query → CreateIndex → delete → 404): computed OCID/timeCreated byte-stable; dup→409, missing→404, row-under-missing-table→404, bad DDL type→400, ON_DEMAND-with-units→400. Async work-request wiring correct and complete; error taxonomy faithful; rollback on CreateTable name/DDL mismatch; create paths span check-then-set under one write lock (no TOCTOU race); Get/List/Query/Index deep-copy (no aliasing; `-race` clean); honest `501` stubs for `usage`/`query/prepare`/`summarize`/change-streams; structure conformant.

arunesh-j:
Rebased onto `a45ac909` (781 commits) and addressed everything across `5da59199` and `95228653`. Only `docs/coverage/README.md` conflicted, exactly as you predicted; code auto-merged and there was no `services/database/driver` drift.

### Blocking — `snapshot.go`

Added, mirroring `providers/oci/vcn/snapshot.go`'s shared `storeDump` table. `go test ./persist/...` now **ok**.

You were right that the nested `items` store is the easy-to-forget part — it needed custom `MarshalJSON`/`UnmarshalJSON` on `tableData`, carrying rows as a map and rebuilding the store on the way back. Renaming those two methods away makes `TestSnapshotRestoreRoundTrip` panic with `invalid memory address or nil pointer dereference`, which is exactly how the same shape failed on the Compute branch.

`items` and `ttl` travel through those methods; `tableData` is an unexported type, so there is no public API change. No lazily-minted defaults or counters to carry — `Mock` holds only `tables`, `names`, `opts` and `monitoring`, and OCIDs are minted at create and stored on the table.

**One trap that was not on anyone's list:** JSON decodes every number to `float64`, so an `INTEGER` key came back as `float64(1)` after restore. Added `retypeRow`, re-fitting rows to their declared column types. Reverting it fails with `expected: int64(1) / actual: float64(1)`.

Round-trip tests cover tables including their rows, `names`, per-table `Scope`/`Tags`/`Indexes`/`ttl`, a TTL'd row still carrying its expiry, and malformed and empty input.

### Per-compartment table names

The store is now keyed by `compartment\x00name`, following `providers/aws/sagemaker`'s `scopedKey`. The OCID still resolves tenancy-wide via `names`, and every Extras method takes `compartmentID` alongside `tableNameOrId` — which mirrors OCI's own rule that a compartment is required when the path parameter is a table name and optional when it is an OCID. `compartmentId` was added to `UpdateTableDetails`, and `fromCompartmentId` to `ChangeTableCompartmentDetails`.

**How the portable driver addresses a table**, since #423 and #425 have the same finding and the answers should agree: it has no compartment in its shape, so it reads the compartment new resources default to, then the *sole* compartment holding that name — and leaves a name held by two others unaddressable rather than picking between them. `ListTables` returns exactly what that resolves. Cross-surface use keeps working, and ambiguity is never guessed.

Reverting `tableKey` to a bare name fails three tests with your exact `AlreadyExists: table "users" already exists`.

Also removed `OCITableScope`, which was declared on `Extras` and never called.

### Engines (#427) — N/A confirmed, not assumed

`config` defines six engine hooks: Database, Cache, Function, Compute, Container, Storage. `DatabaseEngine` is relational only — it provisions a host and port for a client to connect with master credentials, is typed against `services/relationaldb/driver`, and `dbengine` dispatches on `postgres`/`mysql`/`aurora-*`/`redshift`. OCI NoSQL sits on `services/database/driver` (key-value/document); there is no NoSQL engine hook, and no sibling NoSQL provider (DynamoDB, Cosmos, Firestore) wires one.

### Gate failures you named

- **4x `gocritic hugeParam`** — fixed with the same targeted directive `providers/aws/dynamodb` and this PR's own `Query`/`CreateOCITable` already carry. `golangci-lint` is now **0 issues**.
- **Stale generated docs** — `go generate ./...` re-run after the rebase and committed; running it a second time produces no diff.

### Coverage

Provider **85.8% -> 91.0%**, server **83.4% -> 95.3%**. Every function you named is covered: all sort and filter operators on numeric and lexical keys, LONG/FLOAT/DOUBLE/NUMBER/BOOLEAN/BINARY/TIMESTAMP/JSON coercion including failures and DDL defaults, `SetMonitoring` via a capture stub, and the handler's work-request, malformed-body and error branches.

### Numeric literals — fixed, not documented away

`rowMatches` is now type-aware: a condition on a numeric column compares by value, so `DOUBLE` 5 matches literal `5.0`, while other columns stay textual and `STRING "007"` still does not equal `7`. The portable `compareOp` cannot do this because it has no column types, so its equality limit is documented explicitly and points at `rowMatches`.

I chose fixing over documenting because a silent wrong answer is the accept-and-ignore shape this project has flagged repeatedly.

### SDK-compat test — declined, with your questions answered

`oci-go-sdk` is a ~500-package monolith and vendoring it for one round-trip is disproportionate to a recommended-not-required item. I answered your two specific questions from the OCI spec instead: Query's `limit` is a **body** field on `QueryDetails` (matching `queryRequest`), and the handler sets `opc-request-id`, `opc-work-request-id` and `opc-next-page` — the SDK requires no header we do not set. Worth a follow-up issue rather than this PR.

### Verification

```
go build ./...                                    clean
go test ./...                                     exit 0
go test ./persist/...                             ok        (the blocker)
go test -race (both nosql packages)               2/2 ok
golangci-lint (both nosql packages)               0 issues
go test -cover   provider 91.0%   server 95.3%
go generate ./...                                 committed and idempotent
git diff origin/development -- services/database/ empty (shared driver untouched)
no OCI op in docs/coverage/{aws,azure,gcp}/*.md   confirmed
```

### E2E, port 4613

```
CreateTable                  -> 202 + Opc-Work-Request-Id
GetTable                     -> ACTIVE, primaryKey [id,sku], shardKey [id]
GetRow                       -> {"id":1,"qty":3,"sku":"widget"}   integer, not 1.0
Query                        -> 1 item
CreateIndex / GetIndex       -> 202, then ACTIVE
PRIMARY KEY (a,b,c)          -> 400 naming the 2-column limit
SHARD(a,b)                   -> 400 same
RECORD(x INTEGER)            -> 400 naming "RECORD"
GENERATED ALWAYS AS IDENTITY -> 400 naming the modifier
USING TTL 6 HOURS            -> 400 naming "HOURS"
ALTER ... MODIFY qty LONG    -> 400 naming the action
ALTER ... DROP id            -> 400, id is part of the primary key
same name in two compartments-> both created, distinct OCIDs, distinct capacity modes
row written to A             -> invisible from B (404)
delete B, then A             -> 202 each; A survives B's delete; both 404 after
```

### On the `cmd/cloudemu` flake

Interim runs showed `cmd/cloudemu` persist/crash tests failing with shared-daemon collision markers while a sibling session ran its own suite. It is **not this branch** — the diff touches nothing under `cmd/` or `persist/` (verified: 0 files), and with nothing else running `go test ./cmd/cloudemu/` passes on clean `development` in 5.4s. The cause is parallel worktree suites colliding on the shared `~/.cloudemu` daemon lock.

PR #425, opened 2026-08-21
Add OCI Vault: vaults, keys, secrets

### LINKED ISSUE #413: OCI Vault: vaults, keys, secrets
Implement **OCI Vault** against the existing portable `secrets` driver.

Part of #376.

## Scope

Vaults, master encryption keys, secrets and secret bundles/versions. Create/get/list/update/schedule-deletion/cancel-deletion for each, plus reading a secret bundle by version or stage.

## Design notes

OCI Vault is a **superset** of the portable `secrets` driver — it carries KMS key management alongside secret storage, and the portable interface has only 7 operations. Keys, vaults and rotation belong in the `Extras` capability.

OCI deletes are **scheduled**, not immediate: `ScheduleSecretDeletion` sets a deletion time and `CancelSecretDeletion` reverses it, unlike an AWS immediate delete. Model that rather than mapping it to a hard delete.

Secret versions carry stages (`CURRENT`, `PENDING`, `PREVIOUS`, `DEPRECATED`), which the portable version model does not have.

## Acceptance

- Implements `driver.Secrets`; OCI-only behaviour via a consumer-side `Extras` interface
- Vault and key mutations are async in real OCI — record a work request where they are

arunesh-j:
## Summary

- Implements **OCI Vault** against the existing portable `secrets` driver.
- OCI Vault is a superset of that driver — it puts KMS key management behind the secrets service — so most of the surface lives in a consumer-side `Extras` interface. Nothing added to `services/secrets/driver`, per #393.

Closes #413. Part of #376.

## Changes

- **`providers/oci/vault/`** — `Mock` over `memstore` implementing `driver.Secrets`, one `sync.RWMutex` guarding all four stores: vaults, master encryption keys and key versions, OCI-shaped secrets, secret versions and stages.
- **`server/oci/vault/`** — `/20180608` management and `/20190301` retrieval, claimed separately in `Matches`.

## How `DeleteSecret` maps onto scheduled deletion

The interesting design question in this service. `DeleteSecret(name)` takes the same path as `ScheduleSecretDeletion`, with `timeOfDeletion` at OCI's **soonest permitted value** — one day out — rather than the 30-day default an unspecified wire call receives.

The secret moves to `PENDING_DELETION`, keeping its OCID and every version. All seven portable operations then treat it as gone (`NotFound` from Get/Put/GetValue/ListVersions/Delete; absent from `ListSecrets`), while the OCI surface still lists it as `PENDING_DELETION` and `CancelSecretDeletion` restores it — the same soft delete the AWS Secrets Manager mock exposes. Nothing reaps; a scheduled resource stays pending until cancelled.

One deliberate consequence: **a pending secret releases its name**, so a portable delete→recreate works. `CancelSecretDeletion` then fails with `AlreadyExists` if the name was retaken, rather than producing two live secrets with the same name.

Deletion windows are enforced — 7–30 days for vaults and keys, 1–30 for secrets; out-of-window or unparseable `timeOfDeletion` is a 400.

## Judgement calls

- **Portable creates need a vault.** OCI has no secret outside one, so the first portable `CreateSecret` mints `cloudemu-default-vault` and an AES-256 key. Tested that a second create reuses them, and that 16 concurrent creates mint exactly one vault.
- **Crypto endpoints are claimed, not faked.** `encrypt`/`decrypt`/`sign`/`verify`/`generateDataEncryptionKey`/`exportKey` answer 501 naming the gap, because CloudEmu stores no key material. A fake ciphertext would be worse than an honest refusal.
- **Compartments**: `RequireCompartmentID` on `/vaults`, `/keys`, `/secrets`; deliberately not on `keyVersions`, `secrets/{id}/versions`, `getByName` or any `/20190301` path — the real API takes none there and the parent resource already names the compartment.
- **Status codes follow the SDK response shapes**: 200 + body for vault/key mutations (`ScheduleVaultDeletionResponse` carries a `Vault`), 204 headers-only for secret and version deletions (those responses carry no body), 202 for `changeCompartment`.
- **Never accept-and-ignore** — 400 naming the field for `definedTags`, `restoreFromFile`, `restoreFromObjectStore`, `externalKeyManagerMetadata`, `autoKeyRotationDetails`, `externalKeyReference`, `desiredState`, `secretRules`, `rotationConfig`, `secretGenerationContext`, a non-BASE64 `contentType`, and reshaping a key after create.

## Provider Coverage

- [ ] AWS
- [ ] Azure
- [ ] GCP
- [x] OCI

## Checklist

## Test Plan

```
```

Coverage leak check clean: no OCI operation in `docs/coverage/{aws,azure,gcp}/*.md`; `git diff development -- services/` empty. `docs/coverage/oci/vault.md` lists exactly the 7 portable operations, as expected with capabilities consumer-side.

**Lock sweep** over all 63 `*Mock` methods: every one of the 42 exported methods takes `m.mu` exactly once, no locking method calls another locking method, no unexported helper touches `m.mu`, and there is no `store.Update` anywhere — so no closure can re-enter its store.

End-to-end on a running server (port 4614):

```
create vault -> create key -> create secret        -> opc-work-request-id stamped
write version 2                                    -> 200
read CURRENT bundle                                -> "hunter3"
read PREVIOUS stage                                -> "hunter2"
version stages                                     -> [PREVIOUS] / [CURRENT,LATEST]
schedule secret deletion                           -> 204, PENDING_DELETION, timeOfDeletion 2026-09-20
cancel deletion                                    -> 204, ACTIVE
list without compartmentId                         -> 400
poll work request                                  -> CREATE_VAULT, SUCCEEDED
rotate key (new key version)                       -> 200
encrypt                                            -> 501, names the gap
```

OCIDs came out as `ocid1.{vault,key,keyversion,vaultsecret}.oc1.iad.…`.

## A note on parallel worktrees

The session first reported `cmd/cloudemu TestServeOutOfProcess` as a pre-existing failure. **That was wrong** and I have re-checked it: with nothing else running, the test passes on clean `development` in under 6s. The failure is contention on the shared `~/.cloudemu` daemon lock between the six Wave 2 worktrees running suites in parallel. This branch's full suite is exit 0.

## Left out

KMS crypto operations (disclosed as 501), OCI secret rules and rotation targets, tag namespaces, and vault backup/restore — each rejected by name rather than silently dropped.


NitinKumar004 (commented):
## Review notes
**Real data-plane engine (per #427):** N/A — OCI Vault is a secrets/KMS service.

CI: the **Build** failure is a flaky runner I/O error (`fork/exec .../link: input/output error`), not a code defect — a re-run should clear it. Findings below are the real items.

### Findings

**Medium · real-engine — Secret name uniqueness enforced globally across all vaults instead of per-vault**  
`providers/oci/vault/secret.go:117`  
If a user creates a secret named db-password in vault A, then calls CreateSecret (CreateOCISecret) for db-password in vault B -> the handler returns 409 AlreadyExists, but real OCI creates it, because validateSecretSpecLocked (secret.go:117) calls secretByNameLocked (secret.go:398) which scans every secret in every vault and never compares spec.VaultID (OCI scopes secret-name uniqueness to the vault). CancelOCISecretDeletion (secret.go:339) reuses the same global check and can likewise wrongly refuse a restore when an unrelated vault reused the name.

**Medium · docs — docs/services.md not updated for OCI Vault (Definition-of-done miss)**  
`docs/services.md:1389`  
If a user or contributor consults docs/services.md (the human service reference) for OCI secrets -> the ~30-op Vault/KMS surface is invisible: the Secrets header (line 1389) still lists only AWS/Azure/GCP and there is no OCI Vault section, unlike every sibling OCI service (VCN line 407, Monitoring 623, Identity 749). oci-conventions.md Definition-of-done explicitly requires 'Operations added to docs/services.md'; only the generated coverage pages (7 portable ops) were updated, and those cannot describe the OCI-only Extras surface.

**Low · docs — Garbled doc comment on VaultManagement interface**  
`server/oci/vault/extras.go:7`  
If a maintainer reads the VaultManagement doc comment (extras.go:7-11) -> they get an out-of-order, incoherent sentence (a dangling 'container above a secret at all...' fragment and a //nolint:revive directive that runs on into prose), because the comment lines were reordered incorrectly in a tidy edit. The nolint still functions; cosmetic only.

**Low · coverage — server/oci/vault wire handler coverage 79.7% (below 90% pillar)**  
`server/oci/vault/handler.go:246`  
If a client sends an unsupported verb (e.g. DELETE /vaults/{id}) or walks a multi-page list -> those paths ship untested (notFound 0%, listKeys 37.5%, listSecretVersions 50%, getSecretVersion 54.5%, paginate next-page 70%), so a regression in the negative-verb or pagination branch would pass the suite silently.

**Low · coverage — No oci-go-sdk SDK-compat test**  
`server/oci/vault/handler_test.go:31`  
If the wire envelope subtly diverges from what the real oci-go-sdk KMS/Vaults/Secrets client expects (field casing, work-request waiter behavior) -> the hand-rolled map[string]any assertions may still pass while the actual SDK client would fail, because no test round-trips through the real SDK marshaller/waiter (convention calls this the strongest evidence, framed as 'ideally').

NitinKumar004 (commented):
## Review — OCI Vault (KMS + Secrets)

**Verdict: Request changes** — one persistence blocker, one real state-name fidelity bug, and a rebase. State guards and scope are otherwise correct.

### On the red CI

The failing **"Build & Vet"** job is an **infrastructure flake, not a code defect** — the runner hit a disk I/O error corrupting its build cache:

```
go: error obtaining buildID for go tool link: fork/exec .../tool/linux_amd64/link: input/output error
```

which cascaded into spurious `no such file or directory` import errors across unrelated packages (azureai, bedrock, cache, …). `go build ./...` and `go vet ./...` pass clean locally on the pushed SHA. **Re-run the job.** (Separately, the PR reports `mergeable: CONFLICTING` — a real rebase requirement, below.)

### Blocking

**Mock is not `Snapshottable` — `providers/oci/vault/` (no `snapshot.go`).**
`vault.Mock` holds four `memstore.Store` fields (`vaults`, `keys`, `keyVersions`, `secrets`) plus default IDs, but ships no `snapshot.go` and implements no `snapshot.Snapshottable` — the only `providers/oci/*` package missing it. On the merged tree:

```
oci: field Vault (*vault.Mock) holds a memstore.Store but is not Snapshottable — add persistence (see #582)
```

Consequence: CI `Test` goes red on rebase, and Vault state is silently dropped by every snapshot/restore. Fix: add `providers/oci/vault/snapshot.go` (+ test) implementing `Snapshottable` for all four stores (deep-copying key versions, secret versions, and tag maps), mirroring `providers/oci/vcn/snapshot.go`.

### Should fix

**KMS keys and key versions report `lifecycleState: "ACTIVE"` — real OCI reports `ENABLED` — `providers/oci/vault/vault.go:31` (`StateActive`), `key.go:142`, `key.go:445`.**
Per the OCI SDK `KeyLifecycleStateEnum` (`CREATING, ENABLING, ENABLED, DISABLING, DISABLED, DELETING, DELETED, PENDING_DELETION, …`), KMS keys/key-versions **never report `ACTIVE`** — that value belongs to Vaults and Secrets (both correctly `ACTIVE` here). Live-confirmed: `POST /20180608/keys` returned `"lifecycleState":"ACTIVE"`. Cascade: `terraform-provider-oci`'s `oci_kms_key` waits for `ENABLED` after create/rotate, so an `ACTIVE` key hangs or errors the apply; SDK consumers comparing against `KeyLifecycleStateEnabled` misclassify every key. Fix: use `ENABLED`/`DISABLED` for the key + key-version surface (keep `ACTIVE` for vault + secret).

### Non-blocking

- **No `enableKey`/`disableKey` and no `DISABLED` state** — `providers/oci/vault/key.go`. The key lifecycle is `ACTIVE ⇄ PENDING_DELETION` only; OCI's enable/disable management ops and `DISABLING/DISABLED/ENABLING` states are absent. Fits Track-1 scope (metadata only, no crypto) but is a genuine feature gap — recommend a follow-up ticket, provided the `ENABLED` naming above is fixed now.
- **`scheduleSecretDeletion` returns `204`; sibling key/vault deletions return `200` + body** — `server/oci/vault/secret.go:184`. Worth confirming the status code + body against the OCI spec for consistency with the key/vault paths.
- **Branch is behind `development` / CONFLICTING** — rebase; only generated `docs/coverage/README.md` + `oci/README.md` conflict, resolved by re-running `go generate`.

### Verified good

Full lifecycle over the wire (create vault → key ACTIVE → **rotate a PENDING_DELETION key → 409 IncorrectState** → schedule→PENDING_DELETION with timeOfDeletion → cancel→ACTIVE → create secret → get bundle round-trips content → schedule secret deletion): state guards present and correct (`CreateKeyVersion`/`UpdateOCISecret`/`ScheduleSecretVersionDeletion` all guard on live state; current secret version can't be scheduled). Scope correct — no real crypto: `encrypt/decrypt/sign/verify/generateDataEncryptionKey/exportKey` return `501` naming the gap; secrets store base64 content in-memory (matches the AWS KMS metadata-only precedent). Wiring complete; `opc-request-id`/`opc-work-request-id` headers + OCID formats correct; name check-then-set atomic under one mutex (`-race` clean); reads deep-copy.

arunesh-j:
Rebased onto `adb83d85` (789 commits) and addressed everything in `73c67190`.

### On the red CI — your diagnosis was right
Only the two generated coverage files conflicted, exactly as you predicted; took the upstream side and re-ran `go generate`. `services/secrets/driver` gained no methods, so there was no upstream drift to absorb. `go build ./...` and the whole suite pass on the rebased tree, consistent with the flaky-runner reading. The rebase push should give the job a clean run.

### Blocking — `snapshot.go`
Added `providers/oci/vault/snapshot.go`, mirroring `vcn/snapshot.go` with the shared `storeDump` table so Snapshot and Restore cannot drift. All four stores; secret versions travel with their secret.

**The non-obvious part:** `defaultVaultID`/`defaultKeyID` had to go into the snapshot too. Without them a restored mock mints a *second* default vault on the next portable create and orphans every secret it just restored.

`snapshot_test.go` covers all four stores, vault→key→secret cross-references, version stages, scheduled deletion surviving restore **and still being cancellable afterwards**, portable-default continuity, deep-copy, and empty/malformed input. `go test ./persist/...` now passes.

### Should fix — KMS `ENABLED`
Four sites in `key.go` (142, 269, 321, 445) now use a new `StateEnabled`. Worth noting: **two existing tests asserted `ACTIVE` for keys** — they encoded the bug — and were corrected. `TestKeyStatesAreEnabledAndResourceStatesAreActive` pins both vocabularies so the split cannot regress.

I added only `ENABLED`, not `DISABLED`: without `enableKey`/`disableKey` (which you scoped out) `DISABLED` is unreachable and would be dead vocabulary. Follow-up ticket wants `enableKey`/`disableKey` plus `ENABLING`/`DISABLING`/`DISABLED` together.

### The 204 question — it is correct, and now pinned
Secrets management and KMS are **separate OCI services** that merely share the `/20180608` prefix. The SDK's `ScheduleSecretDeletionResponse` carries only `opc-request-id`/`opc-work-request-id` — no entity, hence 204 — while `ScheduleKeyDeletionResponse` carries a `Key`, hence 200 + body. The asymmetry is real OCI's, not ours.

Documented in `writeSecretAction` and pinned by `TestDeletionActionStatusCodesMatchTheService` so it is not "tidied" into consistency later. Step 13 of the transcript below shows both live.

### Per-vault secret names
Uniqueness is now scoped to the vault; `CancelOCISecretDeletion` no longer consults unrelated vaults. The portable lookup resolves across vaults but **rejects an ambiguous name** with `InvalidArgument` naming both vaults, rather than silently picking one — the same answer #423 settled on for log groups, so #423/#424/#425 stay consistent.

One addition worth flagging: portable `CreateSecret` now also refuses a name another vault already holds, otherwise it would happily create a secret its own reads would immediately call ambiguous.

Load-bearing — reverting the `s.VaultID == vaultID` clause fails four tests across both packages (`TestSecretNamesAreUniquePerVault`, `TestCancelDeletionAllowsTheSameNameInAnotherVault`, `TestPortableRejectsAnAmbiguousName`, `TestSameSecretNameInTwoVaults`).

### Also
`docs/services.md` OCI Vault section (32 ops across both API versions, plus namespace semantics, the ENABLED/ACTIVE split, scheduled deletion, staging, rejected inputs, the 501 crypto surface). Garbled `VaultManagement` comment repaired.

### Verification
```
go test ./...                                      exit 0, 451 packages
go test ./persist/...                              ok      (the blocking guard)
go test -race ./providers/oci/... ./server/oci/... 11/11 ok
golangci-lint (both vault packages)                0 issues
go test -cover   provider 96.9%   wire 94.1%  (wire was 79.7%)
go generate ./...                                  docs/coverage committed
```

One lint note: `hugeParam` on `CreateSecret`'s `cfg` appeared after the rebase because upstream grew `driver.SecretConfig`. The signature is fixed by the interface, so it carries a targeted `//nolint:gocritic`. **`azure/keyvault` and `gcp/secretmanager` trip the identical issue and are currently unfixed** — worth a separate sweep rather than a drive-by here.

### E2E, port 4614
```
create key in vault A     -> "lifecycleState":"ENABLED"   (vault itself: ACTIVE)
bundles                   -> CURRENT from-vault-a-v2 ; PREVIOUS from-vault-a-v1
db-password in vault B    -> 200
getByName                 -> A ...0008 (vault ...0001) ; B ...0017 (vault ...0010)
duplicate within vault A  -> 409 "secret \"db-password\" already exists in vault ...0001"
schedule secret deletion  -> 204 + Opc-Work-Request-Id  -> PENDING_DELETION 2026-10-07
cancel                    -> 204 -> ACTIVE
rotate key                -> keyversion ...0023 ENABLED ; both versions ENABLED
encrypt                   -> 501 "stores no key material"
KEY schedule/cancel       -> 200 + body (PENDING_DELETION, then ENABLED)
list                      -> two db-password secrets, distinct OCIDs, distinct vaults
```

PR #427, opened 2026-08-22
Real data-plane engines across AWS, Azure, and GCP

NitinKumar004:
## Real data-plane engines across AWS, Azure, and GCP

Give CloudEmu's most-used resources an **opt-in real data plane** so unmodified apps, SDKs, CLIs, and IaC run *real* workloads — real SQL, Redis, function code, VMs, and containers — against the emulator, not just control-plane mocks. The in-memory, no-Docker default is unchanged: every engine is additive and active only when a caller wires one in.

### How it fits — the pluggable engine seam

```mermaid
flowchart LR
  client["Real SDK / CLI / Terraform"] -->|"cloud wire protocol"| wire["server/&lt;cloud&gt;/&lt;svc&gt;\nwire handler"]
  wire --> provider["providers/&lt;cloud&gt;/&lt;svc&gt;\nin-memory driver"]
  provider -->|"CreateInstance / Invoke / RunTask …"| helper["shared hook\n(dbengine · cacheengine · funcengine\n· computeengine · containerengine)"]
  helper -->|"nil engine = no-op\n(synthetic endpoint / stub)"| synthetic["in-memory default\n(unchanged)"]
  helper -->|"engine wired"| iface["config.&lt;X&gt;Engine\n(interface in core)"]
  iface -.-> realengine["contrib/realengine\n(no-Docker: embedded-postgres,\nminiredis, python/node subprocess)"]
  iface -.-> dockerengine["contrib/dockerengine\n(Docker: mysql, VM compute,\ncontainers, azure-functions host)"]
  realengine --> backing[("real Postgres / Redis /\nfunction process")]
  dockerengine --> containers[("real Docker containers")]
```

**Five pluggable interfaces** live in `config/engine.go` (`DatabaseEngine`, `CacheEngine`, `FunctionEngine`, `ComputeEngine`, `ContainerEngine`). Each has **one shared provider-side helper** (`services/*/…engine`) that every cloud routes through, so AWS/Azure/GCP cannot diverge. Engine **backings live outside core** (core `go.mod` stays dependency-free): `contrib/realengine` (no-Docker) and `contrib/dockerengine` (Docker).

### What now has a real data plane

| Capability | AWS | Azure | GCP | Backing |
|---|---|---|---|---|
| **Postgres** | RDS, Aurora | PostgreSQL Flexible | Cloud SQL, AlloyDB | embedded-postgres (no-Docker) |
| **Redis** | ElastiCache | Azure Cache | Memorystore | miniredis (no-Docker) |
| **MySQL** | RDS | MySQL Flexible | Cloud SQL | Docker `mysql:8.0` |
| **Functions** | Lambda | **Functions** | Cloud Functions | subprocess + Docker host image |
| **VMs** | EC2 | — | — | Docker (boot-script → console) |
| **Containers** | ECS | **Container Instances** *(new svc)* | **Cloud Run Jobs** *(new svc)* | Docker |

Azure Container Instances and GCP Cloud Run Jobs were built as **new services** (driver + provider + wire + `docs/coverage`).

### End-to-end example (real MySQL, real SQL)

```go
cloud := cloudemu.NewAWS(config.WithDatabaseEngine(mysql.New(0)))   // contrib/dockerengine/mysql
// aws rds create-db-instance --engine mysql …  ->  a real mysql:8.0 container
// aws rds describe-db-instances                ->  the SDK's Endpoint.Address:Port
db, _ := sql.Open("mysql", dsnFromSdkResponse)  // connect a real client
db.Exec("CREATE TABLE …"); db.Query("SELECT …") // real SQL round-trips
```

### Real-operation parity (not just create)

The operations users actually perform against these resources behave like the real cloud:

- **RDS / Cloud SQL:** restore-from-snapshot, point-in-time restore, and Cloud SQL **clone** each provision their *own* real database (keyed by the new id — a clone never aliases or drops the source); `ModifyInstance`/`ModifyServer`/`ModifyDBCluster` **rotate the master password** through the engine so the new credential authenticates. `CreateInstance` releases the provider lock across the cold-start provision.
- **ElastiCache:** `CreateReplicationGroup` (the path most IaC uses), not just `CreateCacheCluster`, resolves to real Redis.
- **Lambda / Cloud Functions:** code from **S3** / **`gs://`** is fetched in-process and run as real code; published **layers** are overlaid into the package (Node `node_modules` resolves, Python lands on `sys.path`). A missing backend/object now **fails loudly** instead of echoing.
- **ECS / ACI / Cloud Run:** essential-container exit stops a task; a non-exiting sidecar no longer hangs `RunTask`; EC2 batch `RunInstances` rolls back on partial failure; ACI honors `Never`/`OnFailure`/`Always`; Cloud Run runs `TaskCount` real executions with per-task index env.
- **Lifecycle:** `Provider.Close()` cascades teardown to any wired engine.

### Validation

Every backing ships a **real-user e2e** (real SDK create → connect a real client / invoke → real workload → delete). The Docker e2es run **real containers** and were executed green in-session (MySQL round-trips, EC2 boot-script → `GetConsoleOutput`, ECS/ACI/Cloud Run tasks → exit codes + logs, and a real Python function doubling its input inside the official `azure-functions` host). No-Docker e2es use embedded-postgres / miniredis / host `python3`+`node`. CI gates: build · vet · test · `-race` · gofmt · `go mod tidy` · `docs/coverage` generate · Lint · gosec · govulncheck · CodeQL — now including a **`contrib-dockerengine`** module job.

**Wire-format fidelity (always-on, not gated by an engine):** Cloud SQL `IPAddresses[].type` reports `PRIMARY` (matching real GCP), and `DescribeTasks` surfaces `exitCode`/`reason`/`runtimeId` for stopped ECS tasks.

**Security:** MySQL DDL identifiers flow through a single allowlist-validating `quoteIdent` barrier; zip extraction (both function engines) is guarded against path-traversal and zip-bombs. `go/sql-injection` stays active repo-wide — a scoped CodeQL config exempts only the opt-in MySQL Docker helper, whose first-party allowlist barrier CodeQL cannot model (MySQL has no `pq`-style library quoter).

### Module layout (provider-wise, one subpackage per engine)

```
contrib/dockerengine/  internal/{dockerx,dtest}  mysql/ compute/ container/ azurefunctions/
contrib/realengine/    internal/rtest            postgres/ redis/ functions/
```
Constructors read like `providers/<svc>`: `mysql.New(0)`, `compute.New()`, `container.New()`, `azurefunctions.New()`, `postgres.New(0)`, `redis.New()`, `functions.New()`.

### Additive core changes (no behavior change with no engine wired)

- `dbengine.NewMultiEngine` + `IsMySQLFamily` — one `DatabaseEngine` can back both Postgres and MySQL, routed by engine family.
- `rdsdriver.Instance.ConnectionName` (GCP-only), `services/ecs/driver.Container.{ExitCode,Reason,RuntimeID}`, optional `ConsoleReader` compute capability.
- Op-surface fills: Lambda reserved-concurrency + Layers, Cloud Functions get/set IAM policy, GCE machine-types.

### Deliberately out of scope (follow-up PRs)

EC2 `RegisterImage`/`aws_ami`, Azure Service Bus Topics/Subscriptions, Cloud Run *Services* (Jobs shipped), S3 bucket sub-resources, DynamoDB Streams, GCP MIG, Azure VMSS.


NitinKumar004:
## End-to-end review — real-operation parity, edge cases, CI, security

Full verification is green: `go build/vet`, `go test ./...` (282 pkgs), `go test -race` (131 pkgs, 0 races), `go generate` (0 docs drift, 51 services), core `go.mod` stays dependency-free, both contrib modules build/vet/lint/test clean, and `docker ps -a` shows zero leaked containers across the happy paths. The findings below are about real-cloud *parity* — operations that report success but don't behave like the real service.

### Blocking — real-operation parity

1. **GCP Cloud SQL `CloneInstance` aliases the source database.** `providers/gcp/cloudsql/subresources.go` sets `cloneCfg.DBName = src.DBName`; the postgres engine resolves the same `dbName` on the shared server, so writes through the clone mutate the source and deleting the clone runs `DROP DATABASE` on the still-live source. Give the clone its own dbName (and either real-copy data or document clone as schema-only).

2. **Restore-from-snapshot never engine-backs the restored instance.** `RestoreInstanceFromSnapshot`/`RestoreClusterFromSnapshot` (AWS RDS) and Cloud SQL restore return a 200 + endpoint but never call `dbengine.Provision` — the endpoint resolves to nothing. Route restore through the same provision path as `CreateInstance`.

3. **`ModifyInstance`/`ModifyServer` silently drops password rotation.** The wire layer parses `MasterUserPassword` but the providers (AWS RDS, Azure postgres/mysql flex) never read it, so `modify --master-user-password` succeeds while the role password never changes. Re-run the role/user upsert on a non-empty password.

4. **ECS essential-container rollup ignores `Essential`.** `providers/aws/ecs/engine.go` stops the task only when *every* container exits; a task whose essential container exits alongside a long-running sidecar never flips to STOPPED, so a client polling `DescribeTasks` hangs. Stop the task when an essential container exits.

5. **Multi-container `RunToCompletion` runs sequentially and can hang.** `contrib/dockerengine/container/container.go` blocks on each container in turn; a main + non-exiting sidecar blocks `RunTask`/`CreateContainerGroup` forever. Start+wait concurrently, or gate `RunToCompletion` to single-container until fixed.

6. **EC2 batch `RunInstances` leaks a container on partial failure.** A mid-batch `Provision` error returns zero instance IDs while earlier iterations already have live containers not yet recorded in `m.instances` — unreachable by `TerminateInstances`. Adopt ECS's non-aborting per-item pattern, or roll back on failure.

7. **No lifecycle teardown for engine-backed resources.** No `Provider.Close()` cascades to a wired engine, `cmd/cloudemu` never references the engine interfaces (real engines are library-only today), and `docker run` omits `--rm`, so an ungraceful host exit permanently leaks every container. Provide a `Provider.Close()` that cascades, and document engine lifecycle as the caller's responsibility.

### Blocking — CI

8. **`contrib/dockerengine` has zero CI coverage.** It's a separate module, so the root-scoped `build-vet`/`test`/`lint`/`gosec`/`govulncheck`/CodeQL jobs never touch the riskiest code in the PR. Add a `contrib-dockerengine` job mirroring `contrib-realengine` (ubuntu runners have Docker, so it can run the real e2es).

### Parity gaps (real users hit these paths)

- **AWS `CreateReplicationGroup` isn't engine-backed** — only `CreateCacheCluster` is, but most users provision ElastiCache via replication groups; they get a non-resolving endpoint.
- **Lambda `S3Bucket`/`S3Key` and GCF `sourceArchiveUrl` fall back to stub/echo** — the deployment paths Terraform/SAM/CDK actually use silently don't run real code.
- **ACI ignores `RestartPolicy`** (only `Never` is handled; default `Always` should restart).
- **Cloud Run Jobs `TaskCount > 1`** runs one real execution and synthesizes the rest.
- **Shared MySQL master-username** fully revokes the earlier instance's grants (DROP USER + recreate) vs. the Postgres engine's password-only reset — surface in user-facing docs.
- **Azure Functions needs a CloudEmu-specific `_CLOUDEMU_HANDLER` app setting** to invoke real code — document prominently (the one provider where stock tooling doesn't invoke real code out of the box).
- **Lambda layers are bookkeeping only** — never merged into the real exec env, so a function importing a layer's deps `ImportError`s at invoke despite CRUD success.
- **Provider mutex held across engine provision I/O** (~6–10s cold start) serializes unrelated reads like `DescribeDBInstances`; real cloud answers instantly.

### Wire-format callouts (correct + tested, but always-on — not gated by an engine)

- Cloud SQL `IPAddresses[0].Type` `"PRIVATE"` → `"PRIMARY"` (matches real GCP better).
- ECS `DescribeTasks` now emits `exitCode`/`reason`/`runtimeId` for stopped tasks (was dropped); a stopped no-engine container reports `exitCode: 0`.

### Security / CI status

- Zip-slip (both engines) and MySQL SQL-injection allowlist: **verified sound by hand-trace, no bypass.**
- CodeQL still flags the 2 MySQL DDL lines — its Go SQL-injection query doesn't recognize a regexp allowlist as a sanitizer, and DDL identifiers can't be parameterized. Genuine false-positive; the inline allowlist (`^[A-Za-z0-9_$-]{1,64}$`) is a strict superset-safe barrier.
- **Non-blocking:** `contrib/dockerengine/azurefunctions` extract lacks the zip-bomb caps (`maxUnzipBytes`/`maxUnzipTotal`/`maxZipEntries`) its sibling `contrib/realengine/functions` has — port them for symmetry.
- Top-level `README.md` has no mention/link to the real-engine capability; coveragegen (driver-only) can't see the two Wave-6 wire-only ops (GCF get/set IAM, GCE machineTypes) — tracking-issue items.

PR #433, opened 2026-08-23
feat(storage): opt-in real object-storage engine (S3 · Azure Blob · GCS)

NitinKumar004:
## Opt-in real object-storage engine (S3 · Azure Blob · GCS)

Extend the real-data-plane seam to **object storage**. Wire an engine and object bytes are persisted to a real backing (a real filesystem today) instead of living only in RAM — so uploads survive in the store for the process's lifetime, are inspectable with ordinary tools, and don't bloat memory. The in-memory default is **completely unchanged**; the engine turns on only when you ask for it.

### How it fits (mirrors the existing seam)

Unlike DatabaseEngine/CacheEngine (which hand back a `host:port` a client dials), objects are always served *through* cloudemu — so `StorageEngine` is a **byte transport** (like FunctionEngine):

```go
type StorageEngine interface {
    Put(ctx, StorageObject) error
    Get(ctx, StorageRef) (StorageObject, error)
    Delete(ctx, StorageRef) error
    Copy(ctx, dst, src StorageRef) error
}
```

- One shared helper `services/storage/storageengine` (nil-engine no-op, `cerrors`-wrapped) that **all three providers route through**, so S3/Blob/GCS can't diverge.
- The in-memory `Mock` keeps every object's **metadata** (ETag, versioning, tags, multipart, lifecycle); the engine holds only the **bytes** — the DB analogy (Mock keeps the record, the engine keeps the data).
- Added to `config.Options.EngineClosers()` so `Provider.Close()` cascades teardown.

### What's wired

`PutObject` / `GetObject` / `DeleteObject` / `CopyObject` / `CompleteMultipartUpload` route bytes through the helper when an engine is set (S3 also routes each **version's** bytes under its version id). Multipart parts stay assembled in memory; only the finished object goes to the engine.

**Key refactor:** `Size` is decoupled from `len(Data)` (a stored `Size` field), so `HeadObject`/`ListObjects`/version listings stay correct once bytes move out of `obj.Data`. The no-engine path keeps `obj.Data` and reports `Size == len(Data)` — byte-for-byte identical.

### Backing + validation

`contrib/realengine/blobstore` is a **real filesystem** store (bytes written to real files under a root dir; path-traversal-guarded; `Close()` removes a temp dir it created). Docker MinIO backing is a follow-up.

Each cloud ships a **real-user e2e** with the real SDK (`aws-sdk-go-v2/service/s3`, `azblob`, `cloud.google.com/go/storage`): create bucket → put → get (bytes match) → head/list (size + content-type) → copy → delete → not-found, **and reads the object straight off disk under the store root to prove real persistence**. Plus provider-level engine tests and a helper unit test.

No driver-interface change, no factory change, no docs regeneration (config-only engine). Core `go.mod` stays dependency-free (backing lives in `contrib/realengine`).


NitinKumar004:
## Review — VERDICT: APPROVE

Full pass over the seam + all three providers. The `Size`-decoupling refactor is complete and consistent (every size path reads the stored `Size`/`v.size`, set at every store site), the no-engine path is byte-for-byte unchanged, S3 version refs are consistent between write and read (delete markers never touch the engine), and the `blobstore` traversal guard rejects escaping keys. `build`/`vet`/`test`/`-race` on all packages pass; the three real-SDK e2es round-trip and read bytes off `eng.Root()` with no temp-dir leak; `go generate` yields no docs drift; 0 new lint. Follow-ups below, all being addressed.

### [GAP] S3 versioning × engine is untested
The highest-complexity paths ship without coverage: `GetObjectVersion` engine-load, `DeleteObjectVersion` engine-delete, `recomputeCurrentLocked` rebuilding a current object with `Data=nil` under the latest version id, and suspended-bucket `"null"`-version routing. The S3 e2e only covers a non-versioned bucket. Adding an S3 provider-level engine test + a versioned round-trip (enable versioning → 2 puts → `GetObjectVersion` each → delete newest → `GetObject` returns older).

### [NIT] Suspended-bucket `DeleteObject` orphans engine bytes
`providers/aws/s3/s3.go` — on a versioning-suspended bucket, `PutObject` stores bytes under version `"null"`, but `DeleteObject` deletes ref `{Bucket,Key,Version:""}`, so the `"null"` bytes are never removed. Metadata is dropped so the object correctly 404s (no functional impact); it's a byte leak in the backing. Delete under the object's actual `VersionID`.

### [NIT] Inconsistent engine-Delete error handling
S3 ignores the delete error (`_ = storageengine.Delete(...)`) while Blob/GCS `return err`. Harmless with the filesystem backing (not-exist is already swallowed), but inconsistent — aligning all three.

NitinKumar004:
Follow-ups resolved in `53a3aff`:
- **[GAP] S3 versioning × engine tests** — added S3 provider-level engine tests: put/get round-trip, head/list Size, copy, delete, multipart, a **versioned round-trip** (each version's bytes resolve under its version id; a top-level delete only stamps a marker, prior versions keep their bytes), and a **suspended-bucket purge** test.
- **[NIT] Suspended-bucket orphan** — `DeleteObject` now purges the removed object's bytes under its version id (`null` on a suspended bucket); Enabled still skips (delete-marker only).
- **[NIT] Delete-error consistency** — Blob/GCS `DeleteObject` now best-effort (`_ = storageengine.Delete`), matching S3.

Verified: targeted + `-race` on all 3 providers pass, full suite 0 fail, 0 new lint, gofmt clean.

PR #434, opened 2026-08-23
feat(dynamodb): faithful FilterExpression + ConditionExpression (in-memory)

NitinKumar004:
## DynamoDB expression fidelity (real query semantics, in-memory)

Make cloudemu's **in-memory** DynamoDB actually understand the query language — no real backing, no Docker. Real apps use expressions the emulator silently dropped or got wrong; now they behave like DynamoDB.

### Before → after

Previously the emulator only honored ` AND ` between simple `field op value` clauses and compared everything as strings (`25` matched the string `"25"`). Everything else — `OR`, `NOT`, parentheses, `IN`, `BETWEEN`, `attribute_exists`, `contains()`, `size()`, document paths — was **dropped or rejected**. Now:

```
FilterExpression: "attribute_exists(email) AND status IN (:a, :b) AND size(#tags) > :n"
                  "begins_with(sk, :p) OR NOT (n BETWEEN :lo AND :hi)"
ConditionExpression (Put/Update/Delete): "attribute_not_exists(pk) AND size(#n) < :m"
```
all evaluate faithfully, with **type-aware** comparison (`25` ≠ `"25"`).

### How

- New **`services/database/driver/expr`** — a real tokenizer → recursive-descent parser → AST → type-aware evaluator over the native `map[string]any` item model (93% covered). Precedence `NOT > AND > OR`; parentheses; `IN`/`BETWEEN`; functions `attribute_exists`/`attribute_not_exists`/`attribute_type`/`begins_with`/`contains`/`size`; `#name`/`:value` substitution; document paths (`a.b[0]`). A shared library — reusable by other NoSQL providers later.
- **Filters** (Query/Scan): the raw `FilterExpression` + name/value maps flow to the driver via new `QueryInput`/`ScanInput` struct fields (no interface-method change → **no docs regeneration**) and evaluate per item through `expr`, replacing the flat AND-only path. Paging/Limit/Count/ScannedCount/ordering unchanged.
- **Conditions** (Put/Update/Delete): evaluated with the full grammar against the pre-write item; **extended to UpdateItem and DeleteItem** (was PutItem-only), returning `ConditionalCheckFailedException` on failure.
- The **no-expression path is byte-for-byte behavior-identical** (legacy `Filters` retained for direct driver callers).

### Validation

Real `aws-sdk-go-v2/service/dynamodb` e2es against the wire server assert every new grammar form (functions, `IN`, `OR`/`NOT`, `BETWEEN`, `size`, `contains`, `begins_with`, typed `25` ≠ `"25"`, and Put/Update/Delete conditions). Full suite + `-race` green, no docs drift, 0 new lint.

### Scope
DynamoDB only. Cosmos SQL and Firestore structured queries are different grammars → separate follow-ups. KeyCondition `begins_with`/`BETWEEN`, UpdateExpression `ADD`/`DELETE`/arithmetic, and ProjectionExpression are follow-up slices.

### DIFF
--- a/providers/aws/dynamodb/dynamodb.go
+++ b/providers/aws/dynamodb/dynamodb.go
@@ -13,6 +13,7 @@ import (
 	cerrors "github.com/stackshy/cloudemu/v2/errors"
 	"github.com/stackshy/cloudemu/v2/internal/memstore"
 	"github.com/stackshy/cloudemu/v2/services/database/driver"
+	"github.com/stackshy/cloudemu/v2/services/database/driver/expr"
 	mondriver "github.com/stackshy/cloudemu/v2/services/monitoring/driver"
 )
 
@@ -277,7 +278,10 @@ func (m *Mock) Query(_ context.Context, input driver.QueryInput) (*driver.QueryR
 		return nil, err
 	}
 
-	matched := m.matchQueryItems(td, pkField, skField, &input)
+	matched, err := m.matchQueryItems(td, pkField, skField, &input)
+	if err != nil {
+		return nil, err
+	}
 
 	limit := input.Limit
 	if limit <= 0 {
@@ -323,7 +327,12 @@ func resolveKeyFields(td *tableData, indexName string) (pkField, skField string,
 
 func (m *Mock) matchQueryItems(
 	td *tableData, pkField, skField string, input *driver.QueryInput,
-) []map[string]any {
+) ([]map[string]any, error) {
+	node, err := compileFilter(input.FilterExpression, input.ExprNames, input.ExprValues)
+	if err != nil {
+		return nil, err
+	}
+
 	allItems := td.items.All()
 
 	var matched []map[string]any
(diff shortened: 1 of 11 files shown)


NitinKumar004:
## Review — VERDICT: APPROVE

Adversarial pass over the parser, evaluator, and all three integration points; every gate green.

- **Parser** — recursive-descent, correct precedence `OR→AND→NOT→primary`: `a=:x OR b=:y AND c=:z` parses as `Or{a, And{b,c}}`, `NOT a=:x AND b=:y` as `And{Not{a}, b}`; parentheses override; malformed input returns `InvalidArgument`, no panics/infinite loops (EOF-clamped, ~27 malformed cases asserted).
- **Evaluator** — type-aware: `25` vs `"25"` is non-equal *and* non-orderable; same-type numbers numeric / strings lexical / bytes bytewise; missing paths short-circuit to false (no crash); `attribute_exists` treats present-with-nil as existing (DynamoDB NULL); `IN`/`BETWEEN` typed, `BETWEEN` inclusive; `a.b[0]` document paths walk with bounds checks.
- **Filter integration** — parses once, evaluates per key-matched item before paging; `FilterExpression==""` falls through to the untouched legacy path; `PageOrdered`/Limit/ordering/Count/ScannedCount unchanged.
- **Condition integration** — `gateCondition` runs before every Put/Update/Delete mutation and returns on failure (nothing mutates past a failed gate); evaluated against the existing item (nil→empty map); unmet→`ConditionalCheckFailedException`, malformed→`ValidationException`; UpdateExpression actions untouched.
- **No interface/doc drift** — struct fields only; `go generate` produces empty `docs/` diff. Cosmos/Firestore build+test green (ignore the new fields).

### Out-of-scope NITs (pre-existing, not this PR)
- The DynamoDB wire codec (`server/aws/dynamodb/types.go` `fromAttributeValue`) decodes only S/N/BOOL/NULL/L/M — `B`/`SS`/`NS`/`BS` fall through, so over the wire, binary/set semantics differ from the native evaluator (which handles `[]byte`/lists correctly). Worth a follow-up in the codec.
- `attribute_type` with set codes (`SS`/`NS`/`BS`) can't match, since the native model collapses sets to lists (`L`) — documented in the code.

PR #435, opened 2026-08-23
feat(dynamodb): KeyCondition begins_with/BETWEEN + ProjectionExpression

NitinKumar004:
## Summary

Completes DynamoDB expression fidelity begun in #434 by handling the two remaining query-shaping expressions, reusing the shared \`services/database/driver/expr\` package.

### KeyConditionExpression — \`begins_with\` and \`BETWEEN\`
The wire parser previously accepted only \`pk = :v\` and \`pk = :v AND sk <op> :v2\`. It now parses the full official sort-key grammar:
- \`sk = | < | <= | > | >= :v\` (unchanged)
- \`sk BETWEEN :lo AND :hi\` (inclusive)
- \`begins_with(sk, :v)\`

The driver \`KeyCondition\` struct and provider \`applySortCondition\` already supported \`BEGINS_WITH\`/\`BETWEEN\`, so this is a wire-parser-only change. The top-level split fires on the **first** \` AND \` so BETWEEN's internal \`AND\` stays with the sort clause.

### ProjectionExpression
GetItem, Query and Scan now honor \`ProjectionExpression\` — returning only the requested paths. Added exported \`expr.ParseProjection\` + \`expr.Project\` reusing the existing lexer and path resolver: supports top-level names, nested \`a.b\` (structure rebuilt), list indexes \`a[n]\`, \`#alias\` resolution, and silently omits absent paths. Applied post-driver over native items; parsed once per page.

## Why
Real apps and IaC drive DynamoDB with these expressions constantly; without them queries returned wrong key ranges and full items where a projection was requested. All in-memory — no Docker, helps every user.

## Notes
- No \`Database\` interface method change → no docs/coverage regeneration.
- Set-based expressions are unaffected (out of scope; tracked for the UpdateExpression PR).

## Tests
- \`expr/project_test.go\`: parse (empty/paths/alias/errors) + project (subset, nested map, list index, absent omitted, passthrough).
- Real aws-sdk-go-v2 e2e: BETWEEN + begins_with in \`TestDDBQueryPartitionAndSort\`; projection over GetItem (subset, nested, absent), Query and Scan in \`TestDDBProjectionExpression\`.
- build, vet, gofmt, 0 lint, \`-race\`, full suite — all green.

### DIFF
--- a/server/aws/dynamodb/advanced.go
+++ b/server/aws/dynamodb/advanced.go
@@ -8,6 +8,7 @@ import (
 	cerrors "github.com/stackshy/cloudemu/v2/errors"
 	"github.com/stackshy/cloudemu/v2/server/wire"
 	dbdriver "github.com/stackshy/cloudemu/v2/services/database/driver"
+	"github.com/stackshy/cloudemu/v2/services/database/driver/expr"
 )
 
 // updateItem handles UpdateItem. Supports the common cases:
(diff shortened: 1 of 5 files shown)


NitinKumar004 (commented):
Reviewed for correctness, blast radius, and fidelity to real DynamoDB. Verdict: no blockers. Two fidelity gaps found and fixed on this branch:

1. **List-index projection padded gaps with `nil`.** Projecting `tags[1]` on `["a","b","c"]` returned `["null","b"]`; real DynamoDB compacts to just the projected elements in index order (`["b"]`). Reworked `expr.Project` to accumulate projected indexes sparsely and materialize a compacted list, and to deep-copy the result so it shares no structure with the stored item. Added unit + SDK-level coverage (single index, multiple out-of-order indexes, no-alias).

2. **GetItem validated the ProjectionExpression only when the item existed.** A malformed projection against a missing key returned an empty response instead of `ValidationException`. Parsing now happens before the lookup, matching Query/Scan.

Also confirmed: the first-`AND`-only key split correctly keeps BETWEEN's internal `AND` with the sort clause; the `expr`→`updateExpr` rename avoids shadowing the imported package; no blast radius to the Cosmos/Firestore consumers of `expr` (only additive exports); Count/LastEvaluatedKey are unaffected by projection.

Known follow-ups (rare, pre-existing edges): space-less sort operators (`sk>:v`) and `begins_with` on a numeric sort key are accepted loosely; overlapping projected paths aren't rejected.

PR #436, opened 2026-08-23
feat(dynamodb): UpdateExpression grammar (ADD/DELETE/arithmetic/if_not_exists/list_append) + real set types

NitinKumar004:
## Summary

Completes the DynamoDB expression surface begun in #434/#435, all through the shared \`services/database/driver/expr\` engine and with no \`Database\` interface change (so no docs regeneration).

### UpdateExpression — full grammar
A new parser + evaluator in \`expr\` handles the forms the wire handler previously dropped:
- **SET** operands: \`a = a + :n\`, \`a = a - :n\` (number-only), \`if_not_exists(path, :v)\`, \`list_append(a, b)\`, plus literals/paths and **nested / list-indexed** targets.
- **REMOVE** including \`path[i]\` (shifts later elements down).
- **ADD**: number (atomic increment, creates the attribute when absent) and **set union**.
- **DELETE**: **set difference**; an emptied set removes the attribute.

The raw expression flows via new \`UpdateItemInput\` fields (\`UpdateExpression\`/\`ExprNames\`/\`ExprValues\`); the legacy \`Actions\` form is retained for direct-API callers. A single shared \`driver.ApplyUpdate\` routes **all three providers** (DynamoDB, Cosmos, Firestore) through the same evaluator, replacing three identical SET/REMOVE loops. Path mutation is **copy-on-write**, so a shallow item copy never aliases the stored item.

### Real set types (SS / NS / BS)
Previously sets were indistinguishable from lists. The wire codec now decodes/encodes \`SS\`/\`NS\`/\`BS\` (and the \`B\` binary scalar), modeled as distinct native types (\`expr.StringSet\`/\`NumberSet\`/\`BinarySet\`) that the evaluator understands for \`attribute_type\`, \`contains\`, equality, and ADD/DELETE set algebra.

### Edge fixes (from the #435 review)
- **KeyConditionExpression** now parses through the shared lexer, so it is spacing-tolerant (e.g. \`sk>:v\`), and enforces the restricted key grammar.
- **ProjectionExpression** rejects overlapping document paths (\`a, a.b\`).
- Known limitation: \`begins_with\` on a numeric sort key isn't rejected — the emulator doesn't model attribute types, so schema-based validation isn't available (separate change).

## Why
Real apps and IaC drive counters, sets, and conditional list edits through UpdateExpression constantly; without the grammar and real set types these operations failed or silently misbehaved. All in-memory — no Docker.

## Tests
- \`expr\`: unit coverage for the update parser/evaluator (arithmetic, if_not_exists, list_append, ADD/DELETE number+set, nested paths, copy-on-write, error cases), set algebra, and the key-condition parser (spacing-tolerant, BETWEEN, begins_with, aliases, rejections).
- Real aws-sdk-go-v2 e2e: \`TestDDBUpdateExpressionGrammar\` (SET/ADD/DELETE incl. set union/difference) and \`TestDDBSetTypesRoundTrip\` (SS/NS/BS/B). Cosmos + Firestore provider tests exercise the shared raw-expression path.
- build, vet, gofmt, tidy, \`-race\`, full suite — all green.

### DIFF
--- a/providers/aws/dynamodb/dynamodb.go
+++ b/providers/aws/dynamodb/dynamodb.go
@@ -194,6 +194,8 @@ func (m *Mock) GetItem(_ context.Context, table string, key map[string]any) (map
 }
 
 // UpdateItem applies partial updates to an existing item.
+//
+//nolint:gocritic // hugeParam: interface method signature cannot be changed.
 func (m *Mock) UpdateItem(_ context.Context, input driver.UpdateItemInput) (map[string]any, error) {
 	m.mu.Lock()
 
@@ -212,18 +214,11 @@ func (m *Mock) UpdateItem(_ context.Context, input driver.UpdateItemInput) (map[
 	}
 
 	oldItem := copyItem(item)
-	updated := copyItem(item)
 
-	for _, action := range input.Actions {
-		switch action.Action {
-		case "SET":
-			updated[action.Field] = action.Value
-		case "REMOVE":
-			delete(updated, action.Field)
-		default:
-			m.mu.Unlock()
-			return nil, cerrors.Newf(cerrors.InvalidArgument, "unsupported action: %s", action.Action)
-		}
+	updated, err := driver.ApplyUpdate(copyItem(item), input)
+	if err != nil {
+		m.mu.Unlock()
+		return nil, err
 	}
 
 	td.items.Set(k, updated)
--- a/providers/azure/cosmosdb/cosmosdb.go
+++ b/providers/azure/cosmosdb/cosmosdb.go
@@ -252,6 +252,8 @@ func (m *Mock) GetItem(_ context.Context, table string, key map[string]any) (map
 }
 
 // UpdateItem applies partial updates to an existing document in a container.
+//
+//nolint:gocritic // hugeParam: interface method signature cannot be changed.
 func (m *Mock) UpdateItem(_ context.Context, input driver.UpdateItemInput) (map[string]any, error) {
 	m.mu.Lock()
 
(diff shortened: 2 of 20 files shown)


NitinKumar004 (commented):
Reviewed for copy-on-write safety, parser robustness, set algebra, and the KeyCondition rewrite against the parser it replaces. No blockers found. Two fidelity gaps fixed on this branch:

1. **Invalid document paths were silently coerced.** `REMOVE a[0]` where `a` is a map destroyed it (`{"a":[]}`), and `SET a.b` where `a` is a scalar or absent silently created a map. Both now return a ValidationException, matching real DynamoDB — assignPath/removePath reject a path that traverses the wrong container type.
2. **Operands now read the pre-update image.** UpdateExpression operands resolve against a pre-update snapshot, so `SET a = :v, b = a` reads the original `a` for `b` (DynamoDB evaluates all clauses against the item as it was before the update). The snapshot is a shallow copy — cheap, and safe because writes are copy-on-write.

Also verified and confirmed clean: nested SET/REMOVE and list_append/set-union/difference never mutate the shared stored item; the parser has no panic paths; the KeyCondition rewrite preserves every previously-valid form (now stricter on malformed input); ADD/DELETE type mismatches error rather than corrupt; legacy Actions ADD-rejection is preserved for Cosmos/Firestore.

Added coverage: invalid-document-path errors, pre-update image, list_append/nested-REMOVE no-mutate, and an SDK-level ADD type-mismatch rejection.

Known limitation (documented): NumberSet is float64, so it doesn't preserve DynamoDB's full decimal precision — consistent with how scalar N is already modeled.

PR #437, opened 2026-08-23
feat(firestore): faithful structured-query WHERE (OR / IN / array-contains / unary) + expr numeric coercion

NitinKumar004:
## Summary

Makes in-memory Firestore queries faithful by translating the StructuredQuery `where` clause into the shared `expr` AST and evaluating it in the handler, replacing the previous stringly flat-filter path (which only did AND of six relational ops with `fmt.Sprintf` comparisons).

### WHERE fidelity
- **fieldFilter**: all operators — `EQUAL`/`NOT_EQUAL`/`<`/`<=`/`>`/`>=`, plus `ARRAY_CONTAINS`, `IN`, `NOT_IN`, `ARRAY_CONTAINS_ANY`.
- **compositeFilter**: `OR` as well as `AND` (previously OR was rejected).
- **unaryFilter**: `IS_NULL` / `IS_NOT_NULL`.
- Type-aware throughout (a number never equals a string; missing fields don't match).

Evaluation moved handler-side: the full collection is fetched and each document is matched with `expr.Eval`, so nested composites and the array/list operators are honored exactly.

### expr numeric coercion (shared prerequisite)
Firestore decodes integers as `int64`; DynamoDB decodes numbers as `float64`. The evaluator now coerces `int64`/`int`→`float64` in equality and ordering, so cross-type numeric comparisons work (`2 == 2.0`). Guarded so DynamoDB behavior is byte-identical (its values are already float64).

## Scope / next
This is the WHERE slice. `orderBy` (direction-aware multi-key), `offset`, `startAt`/`endAt` cursors, and `select` projection are the next Firestore PR. `IS_NAN`/`IS_NOT_NAN` are documented as unmodeled (no expr primitive; rare).

## Why
Real apps and the Firestore SDK drive OR, `in`, and `array-contains` constantly; the emulator previously rejected OR and silently mishandled the array operators. All in-memory — no external Firestore.

## Tests
- `expr`: `TestNumericCoercionAcrossTypes` (int64/float64 equality + ordering, number≠string).
- Real `cloud.google.com/go/firestore` REST SDK: `TestDatabaseQueryOperators` drives `in`, `not-in`, `array-contains`, `array-contains-any`, numeric range over int fields, `!=`, `IS_NULL`, and a composite `OR`.
- build, vet, gofmt, tidy, zero docs drift, `-race`, full suite — all green.

### DIFF
--- a/server/gcp/firestore/firestore_lifecycle_test.go
+++ b/server/gcp/firestore/firestore_lifecycle_test.go
@@ -824,6 +824,7 @@ func TestDatabaseQueryOperators(t *testing.T) {
 		"p0": {"status": "active", "tags": []string{"red", "new"}, "score": 10},
 		"p1": {"status": "pending", "tags": []string{"blue"}, "score": 20},
 		"p2": {"status": "archived", "tags": []string{"red", "old"}, "score": 30, "owner": nil},
+		"p3": {"score": 5}, // no status/tags/owner — must be excluded from not-in / !=
 	}
 	for id, fields := range docs {
 		if _, err := coll.Doc(id).Set(ctx, fields); err != nil {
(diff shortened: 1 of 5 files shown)


NitinKumar004 (commented):
Reviewed for translation correctness, the coercion's blast radius, and the negation-with-missing-field trap. Two real correctness bugs found and fixed on this branch:

1. **not-in included documents missing the field (and null fields).** `Not(In)` returns true when the field is absent (In → false → Not → true), so a document without the field matched not-in. Real Firestore excludes absent and null fields. Fixed: not-in is now `And(field <> null, Not(In))`, which excludes both. Added a field-absent document to the test so not-in and `!=` assert the exclusion.
2. **array-contains substring-matched on string fields.** The shared Contains node does string-substring for DynamoDB's `contains()`, so `array-contains "arch"` matched a `"archived"` string field. Fixed in the translation (not the shared evaluator): array-contains / array-contains-any are gated on the field being an array. Added an array-contains-on-scalar assertion (expects zero matches).

Also fixed a minor: an OR with a match-all sub-filter now short-circuits to match-all.

Verified clean: the int64→float64 coercion is a genuine no-op for DynamoDB (float64-only) and Cosmos (JSON float64) — no previously-unequal values become equal; IS_NULL/IS_NOT_NULL correctly exclude missing fields; `!=`→`<>` is consistent; the removed flat-filter path has no other callers; limit now correctly applies after filtering.

PR #438, opened 2026-08-23
feat(firestore+cosmos): complete query fidelity — Firestore orderBy/cursors/select + Cosmos SQL

NitinKumar004:
## Summary

The final PR of the query-fidelity initiative, closing out both remaining tracks. All handler-side, reusing the shared `expr` evaluator; no `Database` interface change (no docs regeneration).

### Firestore — structured-query completion
Building on the WHERE fidelity from #437:
- **ORDER BY** — multi-key, direction-aware, with the implicit `__name__` tiebreaker Firestore appends.
- **offset**, **limit**, and **startAt / StartAfter / endAt / EndBefore** cursors (positional comparison over the order-by tuple).
- **select** — field projection (retaining the document name).

### Cosmos DB — real SQL queries
New `services/database/driver/cosmossql` package (lexer + recursive-descent parser) compiles a Cosmos SQL query into a shared `expr` WHERE AST plus a projection / order-by / paging plan; the handler evaluates it instead of discarding the SQL and returning everything.
- **WHERE**: `= != <> < <= > >=`, `AND`/`OR`/`NOT`, `IN`, `BETWEEN`, `STARTSWITH`, `CONTAINS`, `ARRAY_CONTAINS` (array-gated so it never substring-matches a string field), `IS_DEFINED`, `IS_NULL`, `@parameters`, dotted/indexed paths (FROM alias stripped).
- **SELECT**: `*`, field list with `AS` aliases, `SELECT VALUE`, `DISTINCT`, `TOP n`.
- **ORDER BY** (ASC/DESC), **OFFSET/LIMIT**, and **COUNT/SUM/AVG/MIN/MAX** aggregates.
- Offset-based `x-ms-continuation` paging over the logical result set; the response envelope widened to carry bare scalars (`SELECT VALUE` / aggregates).
- The partition-key header is advisory (cross-partition), matching the emulator's pk-only item identity.

## Why
Real apps and IaC drive `orderBy`/cursors on Firestore and `SELECT … WHERE … ORDER BY` on Cosmos constantly; the emulator previously dropped Firestore ordering/paging and ignored Cosmos SQL entirely (full scan). Both now behave faithfully, all in-memory — no external Firestore/Cosmos.

## Tests
- `cosmossql`: parser unit tests (WHERE eval, IN/BETWEEN/NOT, functions incl. array-gated ARRAY_CONTAINS, projection/VALUE/aggregate, ORDER BY/OFFSET/LIMIT, DISTINCT/TOP, error cases).
- Real `cloud.google.com/go/firestore` SDK: `TestDatabaseQueryOrderingAndPaging` (orderBy asc/desc, StartAfter+limit, offset, EndBefore, select).
- Real `azcosmos` SDK: `TestCosmosQueryFidelity` (parameters, ORDER BY, VALUE, aliased projection, COUNT, STARTSWITH/BETWEEN/ARRAY_CONTAINS/IN/OFFSET-LIMIT); the divergence test rewritten to assert faithful WHERE (15 of 30 docs).
- build, vet, gofmt, tidy, zero docs drift, `-race`, full suite — all green.

### DIFF
--- a/server/azure/cosmosdb/cosmos_lifecycle_test.go
+++ b/server/azure/cosmosdb/cosmos_lifecycle_test.go
@@ -5,7 +5,8 @@
 //
 // The Cosmos HTTP surface (server/azure/cosmos/handler.go) covers: account
 // probe, virtual databases, container create/list/read/delete, document
-// create/list/read/replace/delete, and query (SQL body ignored — full scan).
+// create/list/read/replace/delete, and query (SQL WHERE / ORDER BY /
+// projection / aggregates evaluated).
 // TTL and the change feed are driver-level only (no HTTP endpoint), so those
 // journeys mix SDK writes with direct driver calls on the same provider, per
 // the suite SDK-test-setup notes.
@@ -15,9 +16,10 @@
 //   - CreateItem with a duplicate id succeeds (blind upsert; real Cosmos: 409).
 //   - ReplaceItem ignores IfMatchEtag (real Cosmos: 412 on stale etag).
 //   - DeleteItem on a missing document returns 204 (real Cosmos: 404).
-//   - Queries ignore the SQL text and the partition key (full scan).
-//   - Item identity is the partition-key value only, so two docs with the
-//     same pk but different ids collide (real Cosmos keys on (pk, id)).
+//   - Queries evaluate the SQL WHERE/ORDER BY/projection but treat the
+//     partition key as advisory (cross-partition), since item identity is the
+//     partition-key value only, so two docs with the same pk but different ids
+//     collide (real Cosmos keys on (pk, id)).
 //   - PATCH (PatchItem) is not routed at all (405 MethodNotAllowed).
 package cosmosdb_test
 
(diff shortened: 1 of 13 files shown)


NitinKumar004 (commented):
Reviewed for parser panic-safety, WHERE-translation correctness, aggregate math, cursor semantics, and paging composition. One substantive bug found and fixed, plus two minors:

1. **NOT over an absent field wrongly matched** (the three-valued-logic trap). `NOT (c.status = 'deleted')` on a document missing `status` matched, where real Cosmos excludes undefined fields. Fixed: a NOT over a single-field predicate (comparison, STARTSWITH/CONTAINS, IN, BETWEEN, ARRAY_CONTAINS) now requires that field to exist. `NOT IS_DEFINED` stays unguarded (it must match absent fields). NOT over a composite AND/OR predicate keeps two-valued behavior — a genuine limitation of the two-valued evaluator, now documented in the divergence list and covered by a test.
2. **Empty aggregates** — `MIN`/`MAX`/`AVG`/`SUM` over zero values now return an empty result instead of `[null]`; `COUNT` still returns `[0]`.
3. Robustness: negative `TOP`/`OFFSET`/`LIMIT` and malformed numeric literals now error.

Verified clean by the review and confirmed: no parser panic paths on malformed SQL; Firestore cursor semantics (all four boundary cases + desc flip + partial cursors) correct; `expr.Project` deep-copies for select (no store aliasing); alias stripping only touches the head segment; `!=`→`<>` and array-gated ARRAY_CONTAINS correct; paging composition drops/dups nothing; TTL filtering preserved.

Added coverage: guarded-NOT (present matches / absent excluded, NOT IS_DEFINED still matches absent) and empty-aggregate shape.

PR #495, opened 2026-08-24
fix(iam): serve GetAccountAuthorizationDetails

NitinKumar004:
## Summary
Dispatches `iam.GetAccountAuthorizationDetails`, which aggregates users/groups/roles/policies with their attached policies in paginated form (Marker/MaxItems, Filter by entity type). Previously undispatched.

Wire aggregation lives in a new `server/aws/iam/account_authorization_details.go`. `UserPolicyList`/`GroupPolicyList` are emitted empty (no stored inline-policy state today). SDK round-trip test creates users/roles/policies then asserts they appear in the GAAD response.


NitinKumar004:
Review findings:

**minor — invalid Filter EntityType silently ignored.** An unrecognized `Filter.member` value is dropped instead of rejected. Real GAAD returns `ValidationError`. Fix: default case → ValidationError.

**nit — PermissionsBoundaryUsageCount.** `managedPolicyDetail` omits the field real GAAD emits. Adding it (0).

NitinKumar004:
Fixes applied (5c69f61):
- **invalid Filter → ValidationError (minor)** — an out-of-enum `Filter` value is now rejected with ValidationError instead of silently ignored. New test `TestSDKGetAccountAuthorizationDetailsInvalidFilter`.
- **PermissionsBoundaryUsageCount (nit)** — now emitted (0) on every managed-policy detail, matching real GAAD.

PR #496, opened 2026-08-24
fix(sqs): add MessageMoveTask DLQ-redrive family

NitinKumar004:
## Summary
Dispatches `StartMessageMoveTask` / `CancelMessageMoveTask` / `ListMessageMoveTasks` — DLQ redrive that moves messages from a source (DLQ) ARN back to the destination queue and tracks task status/progress. Previously undispatched.

Modeled via an AWS-only `messageMover` optional interface on the messagequeue driver. The move is synchronous (no RUNNING/CANCELLING intermediate — that's deferred to the async-state-machine tier). SDK round-trip test: start→list→cancel plus message movement verification.

### DIFF
--- a/providers/aws/sqs/movetask.go
+++ b/providers/aws/sqs/movetask.go
@@ -54,6 +54,12 @@ func (m *Mock) StartMessageMoveTask(_ context.Context, sourceARN, destARN string
 		return "", errors.Newf(errors.NotFound, "no queue found for source arn %q", sourceARN)
 	}
 
+	// Real SQS requires the source to be a dead-letter queue that some other queue
+	// redrives to. cloudemu deliberately relaxes this: it accepts any existing
+	// queue as a redrive source, so a user can seed a queue and redrive it without
+	// first configuring a maxReceiveCount RedrivePolicy. This is a permissive
+	// divergence (cloudemu accepts what AWS would reject), never the reverse.
+
 	var destURL string
 
 	if destARN != "" {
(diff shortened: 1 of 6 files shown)


NitinKumar004:
Review findings:

**major — drainSourceQueue non-atomic overwrite.** `drainSourceQueue` snapshots `src.messages`, releases `src.mu`, moves messages, then re-locks and does `src.messages = remaining` computed from the *stale* snapshot. Any message delivered to the DLQ during the move window is silently dropped by the overwrite. Fix: hold `src.mu` across the whole drain, or subtract only the specific moved messages from the current slice.

**minor — TaskHandle on terminal tasks.** ListMessageMoveTasks populates TaskHandle for terminal tasks; real SQS only sets it while RUNNING. Since our move is synchronous (immediately terminal), TaskHandle should be empty.

**minor — SourceArn not validated as a DLQ.** StartMessageMoveTask accepts any SourceArn; real SQS rejects a source with no queue redriving to it. Adding the validation.

NitinKumar004:
Fixes applied (34db3c5):
- **drain atomicity (major)** — `drainSourceQueue` now removes only the exact messages it moved (by identity) from the current source slice, instead of overwriting with a stale snapshot; messages delivered to the DLQ mid-move are preserved.
- **TaskHandle on terminal tasks (minor)** — ListMessageMoveTasks now omits TaskHandle for terminal tasks (only RUNNING carries it); synchronous moves are terminal, so it's empty.
- **SourceArn-must-be-a-DLQ (minor) — deliberate.** Keeping the permissive behavior: cloudemu accepts any existing queue as a redrive source (so a user can seed + redrive without pre-configuring a maxReceiveCount RedrivePolicy). This is a permissive divergence (accepts what AWS rejects), never the reverse — documented in the code.

PR #497, opened 2026-08-24
fix(dynamodb): add DynamoDB Streams data-plane wire handler

NitinKumar004:
## Summary
Adds the `DynamoDBStreams_20120810.*` data-plane surface (DescribeStream / GetShardIterator / GetRecords / ListStreams) as a new wire handler `server/aws/dynamodb/streams.go`, registered in `server/aws/aws.go`. The change-record buffer and `GetStreamRecords` already existed — this exposes them over the real Streams protocol. No driver change. SDK round-trip test: item writes → DescribeStream → GetShardIterator → GetRecords.


NitinKumar004:
Review findings:

**major — per-record StreamViewType inferred, not configured.** `viewTypeForRecord` derives StreamViewType from which images happen to be present, so on a NEW_AND_OLD_IMAGES table an INSERT (only NewImage) reports NEW_IMAGE. Fix: thread the table's configured `cfg.StreamViewType` (already fetched via DescribeTable) and emit it verbatim.

**minor — GetShardIterator doesn't validate ShardId.** A mismatched `ShardId` is ignored; real Streams returns ResourceNotFoundException. Fix: reject unknown ShardId.

**nit — GetRecords Limit unclamped.** Limit is neither defaulted nor capped at 1000. Clamping to [1,1000].

NitinKumar004:
Fixes applied (4fd1af8):
- **StreamViewType (major)** — every record now reports the stream's configured view type (threaded from DescribeTable) instead of one inferred from which images are present. Asserted across INSERT/MODIFY/REMOVE on a NEW_AND_OLD_IMAGES table.
- **GetShardIterator ShardId (minor)** — an unknown ShardId is now rejected with ResourceNotFoundException. New test `TestStreamsGetShardIteratorUnknownShard`.
- **GetRecords Limit (nit)** — clamped to [1,1000], defaulting an unset limit to 1000.

PR #499, opened 2026-08-24
feat(aws): clock-driven async state settling (pending/creating/RUNNING/PENDING_VALIDATION)

NitinKumar004:
## Summary
Adds realistic post-creation state transitions for AWS resources that real clouds settle asynchronously, via a shared **lazy, read-time** mechanism driven purely by `config.Clock` — no goroutines or timers. **Off by default** (`config.WithAsyncSettle()`), so it's a zero-behavior-change addition; every existing test stays green unedited.

### Mechanism — `internal/settle`
A `settle.Window{Intermediate, ReadyAt}` whose `Observe(clock.Now(), final)` returns the intermediate state until `readyAt = createdAt + settleDuration`, then the stored final state. It's a pure function (no mutation on read, composes with existing RLocks), and it's an *overlay* on top of the existing event-driven `statemachine/` — the logical/stored state stays terminal (what internal consumers and transition-legality checks expect); only the Describe surface reports the intermediate. `config.Options.AsyncSettle` gates it (`SettleDuration` returns 0 when off ⇒ inactive window ⇒ immediate final state). Durations are small (1–3s) so a real SDK waiter succeeds on its first poll while an immediate Describe still observes the intermediate.

### Wirings (6 findings)
| Service | Transition |
|---|---|
| ec2 | RunInstances `pending → running`; CreateVolume `creating → available` |
| rds | CreateDBInstance `creating → available`; CreateDBSnapshot `creating → available` (+ source `backing-up` dip, PercentProgress 0→100); RebootDBInstance `available → rebooting → available` |
| sfn | StartExecution `RUNNING → SUCCEEDED`; StopExecution during RUNNING → `ABORTED` (was a no-op) |
| acm | RequestCertificate `PENDING_VALIDATION → ISSUED`, exposing the DNS-validation CNAME while pending; ListCertificates filter follows the transition |

### Architecture
- No new methods on any shared driver interface → **no Azure/GCP/OCI mirrors, coverage docs unchanged.**
- Where a resource is stored as a shared cross-cloud driver struct (EC2 volumes, RDS instances/snapshots — used by Azure flex/SQL), settling lives in an AWS-provider-internal parallel store, not on the shared struct. Where there's an AWS-internal wrapper (ec2 instanceData, sfn execData, acm certData), it's a single field on that wrapper.
- `Observe` is reserved for the Describe surface; stored state stays logical.

Each transition has a FakeClock-driven test asserting intermediate → final. Full suite green (286 pkgs), compat suite green, coverage + terraform-tidy gates clean.

### DIFF
--- a/config/options.go
+++ b/config/options.go
@@ -62,6 +62,25 @@ type Options struct {
 	// StorageEngine optionally persists object-storage bytes to a real backing
 	// (opt-in). Nil (the default) keeps object bytes in-memory.
 	StorageEngine StorageEngine
+
+	// AsyncSettle, when true, makes resources report a realistic intermediate
+	// state (pending/creating/PENDING_VALIDATION/RUNNING) for a short settle
+	// window after creation before their final state. Default false, which keeps
+	// every resource reporting its terminal state immediately (the historical
+	// behavior). See internal/settle.
+	AsyncSettle bool
+}
+
+// SettleDuration returns d when asynchronous state settling is enabled and 0
+// when it is not, so a provider can write `settle.Pending(state, now,
+// o.SettleDuration(settle.DefaultInstanceSettle))` at one call site and have the
+// window be inactive (immediate final state) unless the caller opted in.
+func (o *Options) SettleDuration(d time.Duration) time.Duration {
+	if o.AsyncSettle {
+		return d
+	}
+
+	return 0
 }
 
 // OCIRegion returns the region OCI services should use, substituting an OCI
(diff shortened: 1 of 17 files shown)


NitinKumar004:
Review findings (all opt-in / AsyncSettle-on only, but real):

**major — EC2 lifecycle staleness.** Stop/Terminate/Start/Reboot within the settle window mutate stored State but leave the launch `pending` window, so DescribeInstances reports `pending` for an already-stopped/terminated instance. Fix: clear the settle window on every lifecycle transition.

**major — RDS lifecycle staleness.** Start/StopInstance don't clear `instSettle`, so a stopped just-created instance still reports `creating`. Fix: clear the window in the lifecycle transition.

**major — EC2 instance-state-name filter uses stored state.** Filtering `instance-state-name=pending` right after launch returns EMPTY (filter reads stored `running`) while unfiltered Describe returns the instance as `pending` — self-contradictory. Fix: evaluate the filter against the observed state.

**major — SFN StartSyncExecution breaks the sync contract.** It runs through the RUNNING overlay and returns RUNNING with empty output; a synchronous execution must return its terminal result. Fix: StartSyncExecution bypasses the overlay.

**minor — SFN GetExecutionHistory** returns a full ExecutionSucceeded history while the execution is observably RUNNING. Fix: truncate to in-flight events while unsettled.

**nit — ACM GetCertificate** returns issued PEM while the cert is PENDING_VALIDATION; real ACM returns RequestInProgressException. Fix: gate while pending.

Fixing all six + extending the real-user SDK wire tests to lock them.

PR #500, opened 2026-08-24
feat(ec2): ReplaceNetworkAclAssociation + subnet↔NACL association model

NitinKumar004:
## Summary
Implements `ec2.ReplaceNetworkAclAssociation` — the last undispatched op of its tracker line — by introducing the subnet↔network-ACL **association** concept that was missing at every layer.

### Official AWS contract (API reference)
`AssociationId` + `NetworkAclId` (both required) → `newAssociationId`; idempotent; errors `InvalidAssociationID.NotFound` / `InvalidNetworkAclID.NotFound`. "By default when you create a subnet, it's automatically associated with the default network ACL."

### Architecture fit
- **Driver** (`services/networking/driver`): new `NetworkACLAssociation` type + `Associations` field on `NetworkACL`; new **`NetworkACLAssociator` optional interface** (type-asserted, AWS-only → no Azure/GCP/OCI mirrors), mirroring the existing `VPCAttributes`/`SubnetAttributes` optional pattern.
- **AWS provider** (`providers/aws/vpc`): a default NACL is auto-created per VPC (mirroring default-SG / main-route-table); new subnets auto-associate with it; `ReplaceNetworkACLAssociation` moves a subnet with a fresh id; `DescribeNetworkACLs` returns associations; VPC/subnet teardown cleans them up; a subnet-associated NACL can't be deleted (DependencyViolation). The association store mirrors the existing `rtAssocs` route-table pattern.
- **Wire** (`server/aws/ec2`): dispatch + `<associationSet>` on `DescribeNetworkAcls` + `<newAssociationId>` response; the two not-found cases map to their distinct AWS codes.

### Blast radius
Introducing default NACLs (which the teardown code already anticipated via its `!acl.IsDefault` guards) touched one pagination test, updated to expect the default. Topology, compat, and all provider tests pass unchanged.

Tests: provider-level lifecycle + a real-user `aws-sdk-go-v2` round-trip through the wire (create VPC+subnet → default association → replace onto custom ACL → new id → describe reflects it → both error codes). Full suite green (286 pkgs), compat green, coverage regenerated.

### DIFF
--- a/docs/coverage/aws/vpc.md
+++ b/docs/coverage/aws/vpc.md
@@ -241,6 +241,14 @@ IPAMResources is an OPTIONAL AWS capability exposing IPAM's view of the
 | `GetIpamResourceCidrs` |  |
 | `ModifyIpamResourceCidr` |  |
 
+### NetworkACLAssociator
+
+NetworkACLAssociator is an OPTIONAL capability, discovered by type assertion.
+
+| Operation | Description |
+| --- | --- |
+| `ReplaceNetworkACLAssociation` |  |
+
 ### NetworkInsights
 
 NetworkInsights is an OPTIONAL AWS capability (type-asserted). It covers both
--- a/docs/coverage/azure/vnet.md
+++ b/docs/coverage/azure/vnet.md
@@ -241,6 +241,14 @@ IPAMResources is an OPTIONAL AWS capability exposing IPAM's view of the
 | `GetIpamResourceCidrs` |  |
 | `ModifyIpamResourceCidr` |  |
 
+### NetworkACLAssociator
+
+NetworkACLAssociator is an OPTIONAL capability, discovered by type assertion.
+
+| Operation | Description |
+| --- | --- |
+| `ReplaceNetworkACLAssociation` |  |
+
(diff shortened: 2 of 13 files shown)


NitinKumar004:
Review findings (no blockers/majors; opt to fix the two minors):

**minor — cross-VPC association allowed.** ReplaceNetworkACLAssociation only checks the target ACL exists, not that it's in the subnet's VPC. Real EC2 rejects a cross-VPC move with InvalidParameterValue ("network ACL and subnet must be in the same VPC"). Fix: compare the subnet's VPC and the ACL's VPC, return InvalidArgument→InvalidParameterValue.

**minor — fragile error-code classification.** `writeNetworkACLAssocErr` distinguishes InvalidAssociationID.NotFound vs InvalidNetworkAclID.NotFound via `strings.Contains(err, "association")`; a NetworkAclId literally containing "association" would misroute. Real acl-ids are hex so it never fires, but I'll make it robust (match the fixed message prefix / typed discrimination).

**nit — DryRun** not short-circuited (pre-existing across all EC2 handlers; deferred).

Fixing the two minors.

PR #501, opened 2026-08-24
fix(rds): DescribeDBParameters returns the engine-default parameter set

NitinKumar004:
## Summary
`DescribeDBParameters` (and its cluster sibling `DescribeDBClusterParameters`) returned only the user-modified parameters — 0 for a fresh group — where real RDS always returns the engine's full default parameter set overlaid with user changes.

### Approach
A curated, representative engine-default table (`engine_defaults.go`) keyed by parameter-group family: MySQL-family (mysql/mariadb/aurora-mysql) and PostgreSQL-family (postgres/aurora-postgresql) each get ~13–14 of the parameters IaC tooling and operators most commonly read/tune (max_connections, character_set_server, shared_buffers, work_mem, …) with realistic default values, `ApplyType`/`DataType`/`Description`/`Source=engine-default`; a small generic fallback keeps any family non-empty. Modeling all several-hundred real parameters isn't useful or maintainable — this is the faithful, pragmatic middle.

- `DescribeDBParameters` now returns `mergedParameters(family, userParams)`: engine defaults with user modifications overlaid (`Source=user`, keeping the default's metadata).
- Always-present `default.<family>` groups (never explicitly created) are synthesized from the family in their name, so `DescribeDBParameters("default.mysql8.0")` resolves.
- `ResetDBParameterGroup` now correctly reverts parameters to `engine-default`; the cluster variant mirrors all of this.
- Removed the now-unused `paramsToDriver` (no dead code).

No driver-interface change (provider-internal), so no cross-cloud impact; coverage unchanged. Provider + SDK-wire tests updated to the correct behavior, plus a new real-user test asserting the default group returns engine defaults. Full suite green (286 pkgs), compat green.

### DIFF
--- a/providers/aws/rds/engine_defaults.go
+++ b/providers/aws/rds/engine_defaults.go
@@ -1,0 +1,142 @@
+package rds
+
+import (
+	"sort"
+	"strings"
+
+	rdsdriver "github.com/stackshy/cloudemu/v2/services/relationaldb/driver"
+)
+
+// Real RDS parameter groups always report the engine's full default parameter
+// set (hundreds of entries) overlaid with any user modifications. Modeling every
+// parameter is neither useful nor maintainable, so we seed a curated,
+// representative set of the parameters IaC tooling and operators most commonly
+// read/tune per engine family. DescribeDBParameters returns these
+// engine-defaults merged with the group's user-modified values.
+
+const (
+	sourceEngineDefault = "engine-default"
+	sourceUser          = "user"
+	applyDynamic        = "dynamic"
+	applyStatic         = "static"
+	dataInteger         = "integer"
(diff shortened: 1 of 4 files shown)


NitinKumar004:
Review: APPROVED. Two minor modeling gaps noted, both deferred as out-of-scope:
- **Source filter unhonored** — DescribeDBParameters ignores a `Source=user` filter (now returns the merged set). Faithfully honoring it needs a `Source` param on the driver interface; tracking separately.
- **default-group metadata (nit)** — `DescribeDBParameters("default.mysql8.0")` resolves but `DescribeDBParameterGroups` still 404s that name. Params-only synthesis is the intended scope.

PR #502, opened 2026-08-24
feat(cloudtrail): LookupEvents records real API activity

NitinKumar004:
## Summary
`cloudtrail.LookupEvents` always returned 0 events and the event wire shape modeled only EventId/EventName. Now the wire server records a CloudTrail management event for every served API request, and LookupEvents queries them with the full attribute/time filters, pagination, and complete event fields.

### Architecture fit
- **Generic observer on the core dispatcher** (`server/server.go`): an optional `SetObserver(func(*http.Request))` post-dispatch hook — protocol-agnostic, nil by default (zero behavior change). The core stays service-agnostic; it does not know about CloudTrail.
- **AWS-only `EventRecorder` optional interface** (`services/cloudtrail/driver`, type-asserted → no cross-cloud mirrors). `server/aws.New` wires the observer to it when CloudTrail is present, deriving EventName (X-Amz-Target op or query Action), EventSource + AccessKeyId (SigV4 credential scope), and a ReadOnly heuristic.
- **Provider** (`providers/aws/cloudtrail`): `RecordEvent` stamps EventTime/EventID and builds the CloudTrailEvent JSON (clock + region stay in the provider); a bounded ring buffer (2000) keeps a long-lived process bounded. `LookupEvents` filters by EventName/EventSource/Username/ReadOnly/AccessKeyId/EventId + time window, newest-first, paginated (MaxResults capped at 50).
- **Wire** (`server/aws/cloudtrail`): parses LookupAttributes/StartTime/EndTime and serializes EventTime/EventSource/Username/ReadOnly/AccessKeyId/CloudTrailEvent (was EventId/EventName only).

Tests: provider (record/lookup/filters/pagination/ring-buffer) + a real-user `aws-sdk-go-v2` round-trip (EC2 RunInstances/DescribeInstances → LookupEvents returns them with EventSource/EventTime/CloudTrailEvent, filter narrows to one). Full suite green (286 pkgs), compat green, coverage regenerated.

### DIFF
--- a/docs/coverage/aws/cloudtrail.md
+++ b/docs/coverage/aws/cloudtrail.md
@@ -68,6 +68,18 @@ AWS's `cloudtrail` service · portable interface `driver.CloudTrail` · [AWS ind
 | `UpdateEventDataStore` |  |
 | `UpdateTrail` |  |
 
+## Optional capabilities
+
+Discovered by type assertion; only some providers implement these.
+
+### EventRecorder
+
+EventRecorder is an OPTIONAL capability, discovered by type assertion. The
+
+| Operation | Description |
+| --- | --- |
+| `RecordEvent` |  |
+
 ## Not in scope
 
 _Not documented yet. See the [emulator boundary](../../../README.md) for cloudemu-wide non-goals._
--- a/docs/coverage/coverage.json
+++ b/docs/coverage/coverage.json
@@ -1817,6 +1817,17 @@
         "name": "UpdateTrail"
       }
     ],
+    "optionalCapabilities": [
+      {
+        "name": "EventRecorder",
+        "doc": "EventRecorder is an OPTIONAL capability, discovered by type assertion. The",
+        "operations": [
+          {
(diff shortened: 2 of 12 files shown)


NitinKumar004:
Review findings (fixing all):
- **major — pagination duplicates.** Integer-offset NextToken shifts when new events prepend (newest-first), and the page-1 LookupEvents call records itself before page 2 → duplicate rows across pages. Fix: stable cursor keyed on the last event's id (resume strictly after it).
- **minor — self-recording pollution.** Recording CloudTrail's own read-only LookupEvents lets a polling client fill + evict the 2000-entry log with its own lookups. Fix: skip recording CloudTrail's own read-only ops.
- **minor — O(n) trim on the hot path.** Past the cap, every request re-copies the whole slice under the lock. Fix: amortized high-water trim.
- **nit — tie ordering.** Equal-timestamp events returned oldest-first. Fix: secondary sort by event id (total newest-first order).

Issue #582, opened 2026-08-25
Evolve persist/ to per-driver, identity-preserving snapshotting (full-surface)


NitinKumar004:
Today persist/ hand-writes a bespoke JSON struct + export/restore per resource kind, so it covers only 4 kinds (buckets, tables, secrets, instances) and restores instances with fresh IDs (breaking cross-references). Make persistence generic, full-surface, and identity-preserving.

Approach (fits the existing optional-interface pattern; mirrors LocalStack's per-service model but with portable driver-JSON instead of pickle):
- [ ] Add an optional Snapshottable interface (Snapshot()/Restore()) that persist/ type-asserts per service, like the existing AWS-only/Azure-only optional interfaces
- [ ] Implement it on providers by dumping/loading their memstore.Store maps via All()/Set() — keys=IDs preserved, so cross-references survive restart (this also defuses restore-ordering)
- [ ] Collapse persist/ into generic orchestration (iterate services -> Snapshottable -> map[service]json.RawMessage), retiring the per-kind structs
- [ ] Keep SchemaVersion + IncludeAssets; add version-compat guard on load
- [ ] Extend coverage beyond the 4 kinds to the full provider surface (networking, IAM, lambda, sqs, eventbridge, ...)
- [ ] (Follow-up) add SCHEDULED save strategy alongside the current on-shutdown + manual snapshot

Refs: LocalStack per-service persistence (pickle + assets, save strategies, known ordering/port gaps) — https://docs.localstack.cloud/aws/capabilities/state-management/persistence/

NitinKumar004:
## Implementation plan (grounded in docs/architecture.md)

persist stays a **cross-service engine** consuming Layer-2 drivers generically (like services/resourcediscovery, which already 'walks every driver a provider holds'), never concrete provider types. A mock opts into persistence via an **optional interface** (the established type-assertion pattern); persist type-asserts each driver to Snapshottable. State dumps/reloads through memstore.Store All()/Set() keyed by ID → **identity-preserving** (IDs + cross-references survive restore).

**Components**
1. `internal/snapshot` leaf pkg: `Snapshottable{ Snapshot(ctx, includeAssets) (json.RawMessage, error); Restore(ctx, json.RawMessage) error }` (no deps → any mock imports it, like memstore).
2. `memstore.Store[V].Snapshot()/LoadSnapshot()` — dump All() as map[id]V, reload via Set(id,v). Per-store dump becomes ~2 lines.
[...]

PR #616, opened 2026-08-25
feat(persist): per-driver identity-preserving snapshotting (Phase 1, refs #582)

NitinKumar004:
## What

Phase 1 of #582 — evolves `persist/` from its 4-kind bespoke design toward **per-driver, identity-preserving snapshotting**.

A mock that implements the new optional `internal/snapshot.Snapshottable` interface serializes its **entire** in-memory state and restores it under the **same identities** (resource ids, and id-string cross-references between resources). `persist` type-asserts each service driver to this interface and calls it; for any service whose mock does not implement it yet, it **falls back to the existing bespoke `exportX`/`restoreX` path**, so Azure/GCP and un-migrated services keep working unchanged.

### Delivered

- **`internal/snapshot`** — leaf interface (`Snapshot`/`Restore`, only `context` + `encoding/json`), so any mock can implement it without an import cycle.
- **`internal/memstore`** — generic `Store.Snapshot()` / `Store.LoadSnapshot()` helpers that dump/restore `map[string]V` **under the same keys** (identity-preserving), with a round-trip test.
- **`services/database/driver/expr`** — public `RetypeItem`/`RetypeValue` (item-number reconstruction lifted out of `persist`, now shared, no duplication).
- **AWS 4 mocks implement `Snapshottable`**: `s3` (storage), `dynamodb` (database), `secretsmanager` (secrets), `ec2` (compute). Each dumps **all** its state — every `memstore.Store` plus non-store state (atomic id/IP counters, client-token & subnet-IP maps, managed-visibility, ASG policies, S3 version history, bucket config sub-resources). Value types with **unexported data** are round-tripped via exported snapshot structs; `sync.Mutex`/`func`/transient settle fields are never serialized. `includeAssets=false` omits S3 object bytes (metadata-only), mirroring the current default. EC2 re-registers restored instances with the state machine so lifecycle transitions still validate.
- **`persist/`** — generic export/restore producing `map[serviceName]json.RawMessage` under `ProviderState.Services`, with bespoke fallback. `SchemaVersion` bumped to **2**; a v1 snapshot is rejected on load (snapshots are dev-only — clean break). The generic path enumerates services via the existing `seed.Target`, so **no provider packages are imported into `persist`** and **no call-site changes** were needed in `cmd/cloudemu/serve.go` or the admin snapshot endpoint.

## Why

The old compute path recreated instances via `RunInstances` on restore, minting **fresh** instance ids and dropping cross-references. The headline win here: **instance ids and references survive**. New test `TestIdentityPreservedAcrossRestore` creates an S3 bucket+object, a DynamoDB table+item, a secret, and an EC2 instance launched with a security group; exports to bytes; restores into a **fresh** `NewAWS()`; and asserts the EC2 instance has the **same** instance id, its SG reference resolves, the instance is still transitionable (state machine re-registered), and the object bytes / item / secret round-trip.

## Remaining #582 work (Phase 2)

Roll `Snapshottable` out to Azure/GCP and the remaining ~16 services, then retire the bespoke fallback path.

## Gates

`go build ./...`, `go vet`, scoped `go test` (incl. the identity test), `-race` on `persist`/`memstore` and the four AWS provider packages, `golangci-lint` (zero new issues), and `gofmt` — all green.

### DIFF
--- a/internal/memstore/store.go
+++ b/internal/memstore/store.go
@@ -2,6 +2,7 @@
 package memstore
 
 import (
+	"encoding/json"
 	"sort"
 	"sync"
 )
@@ -145,6 +146,32 @@ func (s *Store[V]) Filter(fn func(key string, value V) bool) map[string]V {
 	}
 
 	return result
+}
+
+// Snapshot serializes the whole store as a JSON object keyed by id
+// (map[string]V), so a restore can reinstate every value under its exact key.
+// Identity is preserved: the keys are the resource ids and round-trip unchanged.
+// V must be JSON-serializable for the dump to be lossless — a value type with
+// unexported data fields needs an exported snapshot form or custom JSON methods.
+func (s *Store[V]) Snapshot() ([]byte, error) {
+	return json.Marshal(s.All())
+}
+
+// LoadSnapshot restores a store from Snapshot's output, Set-ing each value under
+// the same key it was dumped with, so ids and any id-string cross-references
+// survive. It merges onto whatever the store already holds (a fresh store is
+// empty), leaving existing unrelated keys untouched.
+func (s *Store[V]) LoadSnapshot(data []byte) error {
+	var items map[string]V
+	if err := json.Unmarshal(data, &items); err != nil {
+		return err
+	}
+
+	for k, v := range items {
+		s.Set(k, v)
+	}
+
+	return nil
 }
 
 // SortedValues returns the values ordered by their keys. List endpoints
--- a/internal/memstore/store_snapshot_test.go
+++ b/internal/memstore/store_snapshot_test.go
@@ -1,0 +1,66 @@
+package memstore
+
+import (
+	"testing"
+)
+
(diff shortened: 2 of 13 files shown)


NitinKumar004:
Review — verified against the pushed commit (b1be7470).
- build / vet / gofmt clean.
- Identity preservation holds: after Export→Restore into a fresh provider, an EC2 instance keeps the same instance ID, its SG cross-reference resolves, and it stays FSM-transitionable — confirmed for stopped-state, multi-instance, and reservation cross-ref cases.
- No dropped unexported data fields across ec2/s3/dynamodb/secretsmanager snapshot structs.
- expr package extraction is additive-safe: all callers repointed, no dead code, byte-identical logic.
- SchemaVersion 1→2 guard rejects incompatible snapshots; bespoke fallback intact for un-migrated services.
- -race clean; persist imports no provider package.
Verdict: approved.

PR #617, opened 2026-08-25
feat(persist): identity-preserving snapshots for Azure + GCP core mocks (Phase 2a, refs #582)

NitinKumar004:
## What

Phase 2a of #582 (persist evolution). Implements identity-preserving `internal/snapshot.Snapshottable` (`Snapshot`/`Restore`) on the Azure and GCP mocks for the four service kinds persist already orchestrates — Storage, Database, Secrets, Compute — mirroring the four AWS mocks merged in Phase 1 (#616).

Eight new `snapshot.go` files:

- **Azure**: `blobstorage` (storage), `cosmosdb` (database), `keyvault` (secrets), `virtualmachines` (compute)
- **GCP**: `gcs` (storage), `firestore` (database), `secretmanager` (secrets), `compute`/GCE (compute)

Each dumps the mock's full state: every `memstore.Store` (via the `Store.Snapshot`/`LoadSnapshot` helpers) plus non-store state (atomic ID/IP/sequence counters, cross-ref maps, per-container/table sequence counters, Cosmos account attributes). Value types with unexported data fields are promoted to exported snapshot structs so `json.Marshal` doesn't silently drop them:

- **VM / GCE instanceData**: `engineBacked` (plus Azure's `PowerState`/`Generalized`/`Identity`/`NICRefs`, all already exported).
- **blobstorage blobObject**: the full Lease Blob state (`leaseState`/`leaseID`/`leaseDurationSec`/`leaseExpiresAt`/`leaseBreakAt`/`leaseModTimeAtAcquire`) and `appendBlocks`; container snapshots and `snapshotSeq` captured too.
- **keyvault**: multi-vault (secrets + keys per vault); `secretVersion` and `keyVersion` unexported fields promoted, with RSA/EC private-key material serialized as PKCS#1 / SEC1 DER and oct keys as raw bytes. The default vault is restored through `m.vault(defaultVault)` so it stays aliased to the shared Secrets-driver stores.

Sync primitives, transient multipart/staging state, and settle overlays are intentionally not serialized.

**Compute identity preservation** (the point of the phase): Azure VM and GCE previously minted fresh instance IDs on restore via the driver-replay `RunInstances` path. Now restored instances keep their stored ID and ID-string cross-references (security groups), and each is re-registered with the state machine at its **stored** state (copying `ec2/snapshot.go`), so a restored stopped instance stays stopped and remains transitionable.

## Why

Completes identity preservation across all three clouds' four core kinds. The existing generic persist orchestration type-asserts each driver to `Snapshottable`, so it picks these up automatically — `persist/persist.go` is unchanged. `cmd/cloudemu/serve.go` already builds the Azure/GCP `seed.Target`s from the raw `*Mock` types, so the assertion hits.

## Tests

`persist/identity_azure_test.go` and `persist/identity_gcp_test.go` mirror the AWS `identity_test.go`: create a resource of each of the 4 kinds + a compute instance with a security-group cross-ref, `ExportAll` → JSON → `RestoreAll` into a fresh `NewAzure()`/`NewGCP()`, then assert the compute instance keeps the SAME ID, its SG cross-ref resolves, it's transitionable (Stop succeeds), and the storage/database/secret payloads survive.

## Follow-up

Phase 2b = replace the `seed.Target` enumeration so persist reaches the remaining services beyond these four core kinds.

### DIFF
--- a/persist/identity_azure_test.go
+++ b/persist/identity_azure_test.go
@@ -1,0 +1,135 @@
+package persist_test
+
+import (
+	"context"
+	"encoding/json"
+	"testing"
+
+	cloudemu "github.com/stackshy/cloudemu/v2"
+	"github.com/stackshy/cloudemu/v2/persist"
+	"github.com/stackshy/cloudemu/v2/seed"
+	computedriver "github.com/stackshy/cloudemu/v2/services/compute/driver"
+	dbdriver "github.com/stackshy/cloudemu/v2/services/database/driver"
+	secretsdriver "github.com/stackshy/cloudemu/v2/services/secrets/driver"
+)
+
+// TestIdentityPreservedAcrossRestoreAzure is the Azure analogue of the AWS
+// identity guarantee: after a full Export→JSON→Restore into a FRESH Azure
+// provider, resource identifiers and id-string cross-references survive
+// unchanged. In particular a virtual machine keeps the SAME instance id and its
+// security-group reference — the thing the old driver-replay compute path could
+// not do (it minted a fresh id via RunInstances).
+func TestIdentityPreservedAcrossRestoreAzure(t *testing.T) {
+	ctx := context.Background()
+
+	src := cloudemu.NewAzure()
+
+	// Blob container + blob (with bytes).
+	if err := src.BlobStorage.CreateBucket(ctx, "app-data"); err != nil {
+		t.Fatalf("create container: %v", err)
+	}
+	if err := src.BlobStorage.PutObject(ctx, "app-data", "config.yaml", []byte("port: 8080"), "text/yaml", nil); err != nil {
+		t.Fatalf("put blob: %v", err)
+	}
+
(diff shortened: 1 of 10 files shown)


NitinKumar004:
Review — verified against the pushed commit.
- build / vet / gofmt clean.
- Identity preservation verified for Azure + GCP through persist.ExportAll→RestoreAll into a fresh provider: same instance ID, cross-ref resolves, FSM restored at the stored state (stopped stays stopped). Storage/DB/secret payloads round-trip.
- No dropped unexported fields across the 8 value types; Key Vault RSA/EC key DER round-trips to a usable key post-restore.
- persist.go unchanged — the generic type-assert path genuinely reaches the Azure/GCP raw *Mock fields in production (serve.go), not a silent fallback.
- -race clean; no cross-provider or SDK imports; leaf interface intact.
Verdict: approved.

PR #619, opened 2026-08-25
fix(blobstorage): preserve lease across authorized writes; enforce SAS permission + expiry

NitinKumar004:
## Summary

Fixes two HIGH Azure Blob Storage correctness bugs found by the deep end-to-end audit.

### NEW-2 (HIGH, concurrency/data-integrity): lease destroyed on write
An authorized write rebuilt a fresh `blobObject` and dropped all lease fields
(`leaseState`/`leaseID`/`leaseDurationSec`/`leaseExpiresAt`/…). After one write
made with the correct lease id, the blob was silently unleased — a subsequent
`RenewLease` failed `409 LeaseNotPresentWithLeaseOperation`, and any no-lease
writer could immediately overwrite it, defeating exclusive-writer / leader-election.

Fix (provider layer, `providers/azure/blobstorage`): a new `carryOverLease`
helper copies the existing blob's lease bookkeeping onto the replacement object.
Wired into all three overwrite paths that pass through a leased blob:
`PutObject`, `CommitBlockList`, and `CopyObject` (each is lease-gated at the wire
layer via `checkLease` before the write). The wire layer still correctly
requires the lease id for the write itself.

### NEW-1 / #288 (HIGH, security): SAS not enforced
SAS permissions and expiry were never enforced — a read-only SAS (`sp=r`) could
DELETE/overwrite and an expired SAS still read, so SAS-based access control was
untestable against the emulator.

Fix (wire/auth layer, `server/azure/blobstorage/sas.go`): a SAS gate parses the
SAS query and, when a `sig` is present, enforces the validity window (`st`/`se`)
and the permission set (`sp`) against the requested operation
(r→GET/HEAD, w/c→PUT, d→DELETE, a→append, l→list), returning
`403 AuthorizationPermissionMismatch` / `AuthenticationFailed`. Cryptographic
signature verification stays lenient, consistent with the wire layer accepting
any credentials project-wide; permission + expiry scoping is honored so SAS
behavior is testable.

### Deferred
NEW-3 (LOW): GET/DELETE on a blob in a missing container reports `BlobNotFound`
instead of `ContainerNotFound`. Deferred — a correct fix requires the provider
to return a typed error for the missing-container case across GetObject/
DeleteObject/HeadObject, which changes error semantics for portable-layer
consumers (not a trivial adjacent change).

## Tests
Real `azblob`-SDK / wire-harness reproductions:
- `TestSDKLeasedWritePreservesLease` — authorized write keeps the lease active (Renew still works; a later no-lease write is still rejected `LeaseIdMissing`).
- `TestSDKReadOnlySASCannotDeleteOrOverwrite` — `sp=r` reads but is 403'd on DELETE/PUT; blob untouched.
- `TestSDKExpiredSASRejected` — expired `se` → 403 AuthenticationFailed.
- `TestSDKReadWriteSASAllowsDelete` — `sp=rd` DELETE succeeds (positive grant).

## Gates
build · vet · scoped tests · `-race` (provider) · gofmt · golangci-lint (changed pkgs, zero new issues).

### DIFF
--- a/providers/azure/blobstorage/blob_extensions.go
+++ b/providers/azure/blobstorage/blob_extensions.go
@@ -104,6 +104,7 @@ func (m *Mock) CommitBlockList(
 		obj.Data = data
 	}
 
+	m.carryOverLease(ctr, blob, obj)
 	ctr.objects.Set(blob, obj)
 	ctr.staging.Delete(blob)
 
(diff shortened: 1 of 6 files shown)


NitinKumar004:
Review — verified against the pushed commit (bf37ee4b).
- build / vet / gofmt clean; no new lint on changed code.
- Lease fix: PutObject/CommitBlockList/CopyObject preserve lease state across an authorized overwrite, and a subsequent no-lease write is still rejected (CheckBlobLease unchanged). Lease break/release intact; an expired lease still permits a new no-lease write.
- SAS fix: read-only SAS blocked from DELETE/PUT (403), expired and not-yet-valid rejected, read-write grant works; the gate early-returns when no SAS sig is present, so shared-key/anonymous requests are unaffected (no collateral). No regression in existing blobstorage tests.
- -race clean on both packages. Arch: lease fix provider-layer, SAS wire-layer.
- Nits (non-blocking): SAS validity window uses wall-clock rather than config.Clock; PUT ?comp=lease is now permission-gated.
Verdict: approved.

PR #620, opened 2026-08-25
fix(cosmosdb): per-account isolation, resolvable documentEndpoint, indexingPolicy round-trip

NitinKumar004:
## What

Fixes three Cosmos DB correctness findings from the Azure deep audit, all in the wire layer (data-plane account routing + control-plane endpoint derivation). No driver/provider change — the generic database driver already keys tables by opaque name.

### N2 (HIGH, systemic) — no per-account isolation
Two ARM `databaseAccounts` shared one flat database/container namespace, so a client "connected to account B" could read account A's `appdb/users`. The data plane now peels an optional leading `/{account}` path segment (`splitAccount`) and prefixes every driver-table name, databases-set key, and offer key with the account (`qualify`/`dbNS`/`nsPrefix`). Each ARM account gets an isolated namespace.

The default (empty) account keeps the historical `"{db}/{coll}"` table names, so existing single-account and official-emulator (`https://host:port/`) usage — including the driver-level TTL/change-feed tests that address tables by `"db/coll"` — is unchanged.

### N3 (HIGH) — documentEndpoint didn't resolve
ARM returned a synthetic `https://{name}.documents.azure.com:443/` that never routed to the emulator, breaking the canonical create-account-then-connect flow. `documentEndpoint` (plus the per-region endpoints and the connection-string `AccountEndpoint`) is now derived from the request host with the account as a path segment (`https://{host}/{account}/`), so a client built from the returned endpoint reaches its own account. The azcosmos SDK preserves that path prefix on every request and swaps only the host on failover, so isolation holds end-to-end.

### N1 (MED) — indexingPolicy dropped
The data-plane container `indexingPolicy` was decoded off the wire and discarded (nil on read, though TTL + uniqueKeyPolicy round-tripped). It is now tracked in `container_attrs` alongside TTL/unique-keys and round-trips on container read/list.

## Why
Multi-account emulation silently cross-contaminated tenants, and the standard Terraform/armcosmos → `documentEndpoint` → azcosmos flow could not reach the account it just created.

## Tests
Real `armcosmos/v3` + `azcosmos` reproductions:
- `TestSDKCosmosAccountEndpointResolvesAndIsolates` — two accounts built solely from their ARM endpoints; account A's container/item are 404 from account B, each holds independent data, and A survives B's identically-keyed write (covers N2 + N3).
- `TestSDKCosmosIndexingPolicyRoundTrip` — indexingPolicy (mode + included/excluded paths) returns intact on container read (N1).
- Updated the two existing endpoint assertions to the new host-derived form.

## Deferred
- N4 (LOW): `consistencyPolicy` is echoed via the generic overlay but not modeled/enforced. Left as-is to keep this PR cohesive.

### DIFF
--- a/server/azure/cosmos_blob_dispatch_test.go
+++ b/server/azure/cosmos_blob_dispatch_test.go
@@ -1,0 +1,280 @@
+// Regression tests for the Cosmos-vs-Blob dispatch boundary on the shared
+// standalone listener, where NewFromProvider mounts the Cosmos SQL data plane
+// AND the permissive BlobStorage fallback on one server.
+//
+// The Cosmos data plane peels an optional leading /{account} path segment, but
+// it must peel ONLY when that segment names a registered databaseAccount.
+// Otherwise a blob path whose first segment happens to be a virtual-directory
+// prefix like "dbs/" or "offers/" would be stolen by Cosmos (routed to a
+// data-plane collection that answers PUT with 405) instead of served by Blob.
+// These tests drive the fully-assembled server and prove Blob keeps those paths
+// while real Cosmos accounts stay reachable and isolated.
+package azure_test
+
+import (
+	"context"
+	"encoding/json"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/cosmos/armcosmos/v3"
+
+	"github.com/stackshy/cloudemu/v2"
+	azureserver "github.com/stackshy/cloudemu/v2/server/azure"
+)
+
+// dispatchKey is a dummy base64 master key; the data-plane handler ignores auth.
+const dispatchKey = "dGVzdC1rZXk=" // base64("test-key")
+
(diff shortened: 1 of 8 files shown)


NitinKumar004:
Review — verified against the pushed commit (ddf8b6d1). The isolation / endpoint / indexingPolicy work is correct and well-tested (accounts isolated through the data plane, endpoint resolves, default-account backward-compat intact, -race clean).
BLOCKING — cross-handler routing regression: splitAccount (server/azure/cosmosdb/handler.go:141) peels ANY leading path segment as an account. On the shared listener (standalone serve multiplexes blob + cosmos on one port), it now claims blob paths — PUT /mycontainer/dbs and /mycontainer/offers return 405 (served by cosmos), where blob served them before. A blob using a 'dbs/' or 'offers/' virtual-directory prefix breaks whenever blob + cosmos coexist (the default deployment).
Fix: gate account-peeling on the segment being a REGISTERED databaseAccount — the handler already shares dbdriver.Database via AccountTables(), so peel only known account names and let unknown first segments fall through to blob. (Also fixes the edge where an account named literally 'dbs'/'offers' is unreachable.)
Verdict: changes requested.

NitinKumar004:
Re-review — verified against the pushed commit (86ee2de5), after the fix for the prior blocker.
- Blocker resolved: splitAccount now gates peeling on isAccount() (a registered databaseAccount via AccountTables()); an unregistered first segment is left unpeeled → Matches declines → falls through to blob. TestCosmosDoesNotStealBlobPaths passes (PUT /mycontainer/dbs and /offers served by blob, not cosmos 405).
- No new regression: real registered account still peeled + isolated; default/empty-account intra-cosmos routing (/dbs, /offers, /colls) intact; an account named literally 'dbs' is reachable.
- isAccount correct: AccountTables() returns a fresh RLock'd snapshot — no TOCTOU. Original isolation/endpoint/indexingPolicy work intact. -race clean; scoped suites green.
- Non-blocking: a cosmos account and blob container sharing a name can collide only in the narrow case (registered account name == container name AND path second-segment literally dbs/offers) — inherent to the collapsed single-listener, strictly better than pre-fix; worth a doc note. Minor: AccountTables alloc+sort per Matches (negligible at emulator scale).
Verdict: approved.

PR #621, opened 2026-08-25
fix(servicebus): dead-lettering, system properties, TTL and scheduled messages in the data plane

NitinKumar004:
## What

Fixes the Azure Service Bus data-plane correctness bugs found by the deep end-to-end audit. The ARM control plane round-tripped these settings but the REST data plane ignored them, so the canonical reliability patterns (DLQ, correlation, TTL, scheduling) silently broke.

### Fixed findings
- **NEW-1 (BLOCKER) — dead-lettering unwired.** Each queue and subscription now provisions a paired `$DeadLetterQueue` backing store; `createQueue`/`createSubQueue` set `QueueConfig.DeadLetterQueue` with `MaxReceiveCount = maxDeliveryCount`. The data-plane router recognizes the `$DeadLetterQueue` address suffix (queue and subscription forms), so the DLQ is readable from `<entity>/$DeadLetterQueue/messages`. Also fixes the latent data-loss bug: `moveToDLQ` now reports success and the message is dropped from the main queue only once it is safely stored in the DLQ.
- **NEW-3 (HIGH) — system properties dropped.** Caller `MessageId`/`CorrelationId`/`Label`/`SessionId`/`ReplyTo`/`To`/`ReplyToSessionId`/`ContentType` are parsed on send, persisted on the message, and echoed back in the receive `BrokerProperties` header. The sender's `MessageId` is no longer overwritten by the server id.
- **NEW-5 (HIGH) — TTL cosmetic.** Queue/subscription `defaultMessageTimeToLive` and per-message `TimeToLive` are honored; expired messages are not delivered, and dead-letter on expiry when `deadLetteringOnMessageExpiration` is set.
- **NEW-6 (HIGH) — scheduled messages.** `ScheduledEnqueueTimeUtc` delays delivery until that time instead of delivering immediately.
- **NEW-2 (MEDIUM) — renew-lock 405.** `POST` on a locked message re-arms the visibility window by the entity's lock duration.

### 3-layer touch points
- Wire (`server/azure/servicebus`): DLQ provisioning + routing, BrokerProperties parse/echo, TTL/scheduled/renew wiring, ISO-8601 duration parsing, cascade cleanup of DLQ stores.
- Driver (`providers/azure/servicebus`): `SystemProps` on messages, `deadLetterOnExpiration`, guarded `moveToDLQ`, expiry→DLQ.
- Portable interface (`services/messagequeue/driver`): Azure-specific `SystemProperties` (send/receive) and `DeadLetterOnExpiration` (queue config) fields, ignored by non-Azure providers.

## Tests
- `dataplane_dlq_test.go`: dead-letters after maxDeliveryCount and is readable on the DLQ endpoint (queue + subscription), TTL-expiry dead-lettering, renew-lock returns 200.
- `dataplane_message_test.go`: system-property round-trip, expired-TTL message not delivered, scheduled message not delivered before its time.
- `dead_letter_test.go` (driver): missing-DLQ-store data-loss guard, system-property preservation, dead-letter-on-expiration.

## Deferred (remaining Service Bus work)
- **NEW-7** sessions (`requiresSession`) — no session-scoped receive / FIFO / state.
- **NEW-4** custom application-property filters — custom props not parsed off the wire.
- **NEW-8** duplicate detection (`requiresDuplicateDetection`) keyed on MessageId.
- AMQP data plane (`azservicebus` SDK), **NEW-9** LRO namespace create/delete, **NEW-10** resolvable `nextLink` pagination.

## Gates
`go build ./...`, `go vet`, scoped `go test` (server + provider), `go test -race` (provider), `golangci-lint` (0 issues), `gofmt` — all green.

### DIFF
--- a/providers/azure/servicebus/dead_letter_test.go
+++ b/providers/azure/servicebus/dead_letter_test.go
@@ -1,0 +1,103 @@
+package servicebus
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stackshy/cloudemu/v2/services/messagequeue/driver"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestDeadLetterMissingStorePreservesMessage guards the latent data-loss bug:
+// when a message exceeds maxReceiveCount but its dead-letter store is absent, it
+// must stay on the main queue rather than being silently dropped.
+func TestDeadLetterMissingStorePreservesMessage(t *testing.T) {
+	ctx := context.Background()
+	m, clk := newTestMock()
+
+	mainInfo, err := m.CreateQueue(ctx, driver.QueueConfig{
+		Name: "guarded",
+		DeadLetterQueue: &driver.DeadLetterConfig{
+			TargetQueueURL:  "https://missing.example/nope",
+			MaxReceiveCount: 1,
+		},
+	})
+	require.NoError(t, err)
+
+	_, err = m.SendMessage(ctx, driver.SendMessageInput{QueueURL: mainInfo.URL, Body: "keep-me"})
+	require.NoError(t, err)
+
+	msgs, err := m.ReceiveMessages(ctx, driver.ReceiveMessageInput{QueueURL: mainInfo.URL, MaxMessages: 1, VisibilityTimeout: 1})
+	require.NoError(t, err)
+	require.Len(t, msgs, 1)
+
+	clk.Advance(2 * time.Second)
+
+	// Second receive would exceed maxReceiveCount, but the DLQ store is missing.
+	msgs, err = m.ReceiveMessages(ctx, driver.ReceiveMessageInput{QueueURL: mainInfo.URL, MaxMessages: 1})
+	require.NoError(t, err)
+	assert.Empty(t, msgs)
+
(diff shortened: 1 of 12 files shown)


NitinKumar004:
Review — verified against the pushed commit (2f58f4e4).
- build / vet / gofmt clean; AWS SQS + GCP Pub/Sub + messagequeue tests green — new shared-struct fields ignored by non-Azure providers (zero references), no regression.
- Dead-lettering real on <queue>/$DeadLetterQueue and <topic>/subscriptions/<sub>/$DeadLetterQueue; maxDeliveryCount correct.
- Data-loss closed: main-queue removal only when moveToDLQ reports moved==true; missing DLQ store keeps the poison message (TestDeadLetterMissingStorePreservesMessage).
- TTL uses config.Clock (FakeClock); scheduled honored; system props round-trip (sender MessageId preserved); renew-lock via POST; -race clean (qd.mu→dlq.mu order).
- Shared-struct field follows existing driver.go convention, not an optional-interface violation.
Follow-ups (non-blocking): expiry→DLQ path ignores moveToDLQ return; scheduled-delay uses wall-clock at wire layer.
Verdict: approved.

PR #622, opened 2026-08-25
fix(azure-networking): NIC ipConfig <-> LB backend-pool membership

NitinKumar004:
## What

Fixes the Azure **NIC ipConfig ↔ Load Balancer backend-pool membership** bug (deep-audit finding NEW-1, HIGH). Associating a NIC ipConfiguration with an LB backend address pool was a two-way no-op: `GET networkInterface` never showed the pool, and `GET backendAddressPool` never showed the ipConfig — so Terraform's `azurerm_network_interface_backend_address_pool_association`, a NIC ipConfig with `loadBalancerBackendAddressPools`, and `az network nic ip-config address-pool add` all silently dropped the link (perpetual diff; pool membership impossible to reason about).

Root cause: the NIC ipConfig wire struct had no `loadBalancerBackendAddressPools` field (discarded at decode), and the LB backend pool was modeled as `[]string` pool names with no `backendIPConfigurations`.

## Why / approach (3-layer, single source of truth)

The membership is threaded through the driver model rather than blind-echoed, with one source of truth so both sides always agree:

- **Driver:** `AzureIPConfig` gains `LBBackendPoolIDs []string` — the ARM ids of the backend pools an ipConfiguration has joined.
- **NIC wire (forward):** the ipConfig request/response carry `loadBalancerBackendAddressPools`; the value is stored on and echoed from the NIC.
- **LB wire (reverse):** a backend pool projects its read-only `backendIPConfigurations` by reverse-lookup against the NIC side (via the networking driver, wired into the LB handler with `SetNICResolver`). The whole-LB GET, the pool list, and the standalone pool GET all project the same membership.

Because the pool derives its members from the NIC, clearing the NIC (re-PUT without the ref, or deleting the NIC) clears both directions with no sync/drift. When only the LB driver is configured, pools simply report no members.

ARM shapes verified against `armnetwork` (`InterfaceIPConfigurationPropertiesFormat.loadBalancerBackendAddressPools`, `BackendAddressPoolPropertiesFormat.backendIPConfigurations`). No driver interface method changed, so `docs/coverage` is unchanged.

## Tests

New real-`armnetwork` e2e tests in `server/azure/loadbalancer`:

- `TestSDKNICBackendPoolMembershipReflectsBothSides` — after a NIC ipConfig joins a pool, `GET NIC` lists the pool in `loadBalancerBackendAddressPools` **and** `GET pool` / whole-LB GET list the ipConfig in `backendIPConfigurations`.
- `TestSDKNICBackendPoolMembershipClearedBothSides` — re-PUT without the ref clears both sides, and deleting the NIC drops it from the pool.

Gates: build, vet, scoped tests (`server/azure/{networking,loadbalancer}`, `providers/azure/{networking,loadbalancer}`), `-race` on the new cross-service reads, gofmt, and golangci-lint on changed packages all green (zero new issues).

### DIFF
--- a/providers/azure/vnet/network_interface.go
+++ b/providers/azure/vnet/network_interface.go
@@ -326,7 +326,8 @@ func (m *Mock) usedPrivateIPs(subnetID, selfKey string) map[string]bool {
 			continue
 		}
 
-		for _, ipc := range nic.IPConfigs {
+		for i := range nic.IPConfigs {
+			ipc := &nic.IPConfigs[i]
 			if ipc.SubnetID == subnetID && ipc.PrivateIP != "" {
 				used[ipc.PrivateIP] = true
 			}
@@ -355,6 +356,12 @@ func addHost(base net.IP, n int) net.IP {
 func toAzureNIC(n *nicData) driver.AzureNIC {
 	configs := make([]driver.AzureIPConfig, len(n.IPConfigs))
 	copy(configs, n.IPConfigs)
+
+	// Deep-copy each ipConfiguration's backend-pool membership so a caller
+	// mutating the returned slice cannot reach back into stored state.
+	for i := range configs {
+		configs[i].LBBackendPoolIDs = append([]string(nil), n.IPConfigs[i].LBBackendPoolIDs...)
+	}
 
 	return driver.AzureNIC{
 		Name:                   n.Name,
--- a/server/azure/azure.go
+++ b/server/azure/azure.go
@@ -283,7 +283,16 @@ func New(d Drivers) http.Handler {
 	// locations / dnsZones), so registration order relative to them is
 	// unconstrained. Registered before the BlobStorage fallback.
 	if d.LB != nil {
-		srv.Register(lbsrv.New(d.LB))
+		lbHandler := lbsrv.New(d.LB)
+
+		// Project a backend address pool's read-only backendIPConfigurations from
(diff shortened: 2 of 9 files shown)


NitinKumar004:
Review — verified against the pushed commit (d963d6b5).
- build / vet / gofmt clean; PR's own changes lint-clean (the one lll in driver.go:601 is pre-existing, not from this PR).
- Both directions real: NIC ipConfig echoes loadBalancerBackendAddressPools (forward); LB pool projects backendIPConfigurations via reverse-lookup over all NICs (poolMembers), applied to whole-LB GET, list, standalone pool GET/list, and post-PUT read. Single source of truth = the NIC field, so clearing the NIC clears both sides. Author's real-armnetwork e2e (forward, reverse, whole-LB nested, clear-via-rePUT, clear-via-delete, re-associate) all pass under -race.
- Injector safe: SetNICResolver wired via AzureNetworkInterfaces type-assert in server/azure/azure.go (non-nil in production + e2e); nil-resolver guard airtight — all 8 sites funnel through poolMembers which returns nil→empty, no panic.
- -race clean (NIC reads RLock'd + deep-copied, no lock inversion); no collateral (LB child-delete scoping + NIC↔VM intact); docs/coverage no-diff (struct field only).
Non-blocking nits: poolMembers rebuilds the cross-NIC map per LB GET (fine at emulator scale); projected ipConfig id stamps request subscription (harmless single-subscription).
Verdict: approved.

PR #623, opened 2026-08-25
fix(azure-rm): cascade resource-group delete + scope network reads to their RG

NitinKumar004:
## What & why

Deep-audit fixes for the Azure Resource Manager keystone cluster. A resource group is a pure container in ARM, and Microsoft.Network resources are addressed by (subscription, resource group, name) — the emulator honored neither, so a teardown leaked resources and a same-named resource in another group collided. This PR fixes the lifecycle/write path for both.

### N1 [BLOCKER] — RG delete now cascades
Deleting a resource group deleted only the group record; every storage account, virtual network, NSG, VM, etc. created under it stayed alive and globally addressable (a `terraform destroy` "succeeded" while leaking resources; the next `apply` collided).

Fix: a new `resourcegroups.ResourceGroupPurger` interface. Each per-service ARM handler that stores its resources RG-scoped implements `PurgeResourceGroup`; the resource-group handler fans a delete out to the registered purgers (best-effort, matching the async fire-and-forget 202 contract). Wired in `server/azure/azure.go`. Covered today: **compute** (already stored `Instance.ResourceGroup`), **networking** (vnet + its subnets + NSGs), **storage accounts**. Storage accounts now record their RG (`AccountAttributes.ResourceGroup`) on the ARM create so the cascade can find them.

### N5 [HIGH] — network resources scoped to their resource group
`virtualNetworks` / `networkSecurityGroups` lived in the driver's flat namespace, so a GET/DELETE under the **wrong** RG returned 200 and echoed the wrong RG back into the id. Cross-RG reads and same-name collisions across two teams/states corrupted state.

Fix: vnets and NSGs record their RG on create (`cloudemu:azure{VNet,NSG}ResourceGroup` internal tags, mirroring the existing NAT-gateway / public-IP RG-tag pattern). The direct GET / PUT-existing-check / DELETE, and the RG-scoped list, resolve them scoped to the request's RG (case-insensitive, matching ARM) → a read under the wrong group is a 404. Subnets are gated through their (now RG-scoped) parent vnet. Cross-resource references (peering remote vnet, NIC→subnet/NSG, effectiveNSG) stay name-only, since ARM permits cross-RG references.

## Findings fixed
- **N1** RG-delete-no-cascade (BLOCKER) — compute/networking/storage
- **N5** network cross-RG read/delete (HIGH) — vnet, NSG, subnet (via parent), + RG-scoped list

## Deferred (explicitly, per the deferral rule)
The **N2/N3/N4 Resource Graph row-correctness** cluster is NOT in this PR. Those are the read/discovery path (the shared `services/resourcediscovery` walkers, with AWS/GCP blast radius) and are cleanly separable from the write/lifecycle path fixed here:
- **N2** — ARG / generic `.../resources` / exportTemplate still report `resourceGroup: "default"` for networking & storage (compute already carries the real RG). Needs the network/storage walkers + ARN builders to emit the persisted Azure RG.
- **N3** — ARG `location` still shows the AWS default (`us-east-1`) for network/storage resources; the discovery Engine stamps `e.region` rather than each resource's Azure location.
- **N4** — ARG renders storage accounts with a malformed container-path id, drops user tags, and leaks `cloudemu:*` internal tags (the internal RG tags added here are `cloudemu:`-prefixed and already stripped from the ARM handler responses via `stripInternal`, but the ARG renderer needs the same filter).

These were deferred because the proper N2 fix is a systemic rework of the shared cross-provider discovery walkers, too wide to land cleanly alongside the lifecycle fix without risking AWS/GCP collateral. N1/N5 do not depend on it (the cascade queries each driver directly rather than via discovery ARNs).

Also not addressed (out of this PR's scope, from the same audit): N6 ARM-template deployments, N7 provider registration, N8 storage 200-vs-201, R13 Tags RP, R14 locks, R15 moveResources.

## Tests (real SDK, end-to-end)
- `TestResourceGroupDeleteCascades` — real `armresources` + `armstorage` + `armnetwork`: create RG + storage account + vnet, delete the RG, assert both contained resources return 404.
- `TestVNetGetIsResourceGroupScoped` / `TestNSGGetIsResourceGroupScoped` — cross-RG GET is 404; same-named resource in a second RG is independent; RG-scoped list returns only the group's own; deleting one group's copy leaves the other's intact.

## Gates
build ./... · vet · scoped tests (incl. `-race`) · golangci-lint (0 new) · gofmt · `go generate` (no docs/coverage drift — no driver interface method changed).

### DIFF
--- a/server/azure/azure.go
+++ b/server/azure/azure.go
@@ -211,10 +211,39 @@ func New(d Drivers) http.Handler {
 	// before the permissive blob fallback so it isn't swallowed as a blob call.
 	srv.Register(tenants.New(tenantID))
 
+	// Build the per-service handlers that own resource-group-scoped resources up
+	// front so they can be handed to the resource-group cascade below and then
+	// registered at their normal positions. A resource group is a pure
+	// container, so deleting it must delete the resources created under it;
+	// each of these handlers implements ResourceGroupPurger to tear its own
+	// resources down. Other resource types are not cascaded yet.
+	var (
+		vnetHandler    *vnet.Handler
+		vmHandler      *virtualmachines.Handler
+		storageHandler *storageaccountsrv.Handler
+		rgPurgers      []resourcegroups.ResourceGroupPurger
+	)
+
+	if d.Network != nil {
+		vnetHandler = vnet.New(d.Network)
+		rgPurgers = append(rgPurgers, vnetHandler)
+	}
+
+	if d.VirtualMachines != nil {
+		vmHandler = virtualmachines.New(d.VirtualMachines, d.Network)
+		rgPurgers = append(rgPurgers, vmHandler)
+	}
+
+	if d.BlobStorage != nil {
+		storageHandler = storageaccountsrv.New(d.BlobStorage)
+		rgPurgers = append(rgPurgers, storageHandler)
+	}
+
 	// Resource groups have no driver of their own: they are containers, and the
 	// emulator tracks membership by the ids resources already carry. The
-	// discovery engine (nil-safe) lets exportTemplate enumerate that membership.
-	srv.Register(resourcegroups.New(d.ResourceDiscovery))
+	// discovery engine (nil-safe) lets exportTemplate enumerate that membership;
+	// the purgers cascade a group delete into its resources.
+	srv.Register(resourcegroups.New(d.ResourceDiscovery, rgPurgers...))
 
 	// microsoft.insights extension resources (metrics, metricDefinitions,
 	// diagnosticSettings) hang off an arbitrary resource URI, so they must claim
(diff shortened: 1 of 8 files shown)


NitinKumar004:
Review — verified against the pushed commit (4f06c83).
- N1 cascade real + safe: RG delete fans out to per-service ResourceGroupPurger; storage + vnet physically deleted (GET→404), VMs terminated (matches direct ARM VM DELETE). Partial-failure safe — each purger error swallowed, others proceed; RG removed from the map regardless (ARM 202 fire-and-forget). No cross-RG collateral: every purger filters by RG tag with EqualFold, so deleting RG-A can't touch a same-named resource in RG-B.
- N5 cross-RG scope real: wrong-RG vnet/NSG GET→404, same-name-in-two-RGs independent, RG-scoped list own-only, correct RG echoed. Case-insensitive; a data-path resource with no RG tag was never ARM-name-addressable, so no false 404.
- No networking regression: securityRule/peering/NIC↔VM/effective-NSG/LB-child-delete still use unchanged name-only finders; tests pass under -race. Blast radius contained (handler hoisting moves construction only; Register order preserved). Deferral sound — cascade queries drivers directly, not via the discovery/ARN engine, so N1/N5 don't depend on the deferred ARG rework. docs/coverage unaffected (struct field only).
Non-blocking follow-ups: child sub-resource ops (securityRules/peerings/subnet-NSG) still parent-by-name under wrong RG (fits the cross-ref-name-only deferral); VM cascade terminates rather than physically removes; listSubnets returns all (pre-existing).
Verdict: approved.

PR #624, opened 2026-08-25
fix(azure-eventgrid): scope-bound event subscriptions

NitinKumar004:
## What

Adds support for Azure Event Grid **scope-bound / extension-form event subscriptions** — the `Microsoft.EventGrid/eventSubscriptions` extension resource that can hang off any scope. Previously only the direct `.../topics/{t}/eventSubscriptions/{n}` form worked.

Scopes now supported for CreateOrUpdate / Get / Delete / List:

- **Subscription scope** — `/subscriptions/{s}/providers/Microsoft.EventGrid/eventSubscriptions/{n}` (+ `ListGlobalBySubscription`)
- **Resource-group scope** — `/subscriptions/{s}/resourceGroups/{rg}/providers/Microsoft.EventGrid/eventSubscriptions/{n}` (+ `ListGlobalByResourceGroup`)
- **Arbitrary-resource scope** — `.../providers/{ns}/{type}/{res}/providers/Microsoft.EventGrid/eventSubscriptions/{n}` (+ `ListByResource`)
- **Topic extension scope** — `.../providers/Microsoft.EventGrid/topics/{t}/providers/Microsoft.EventGrid/eventSubscriptions/{n}`

## Why

Terraform's `azurerm_eventgrid_event_subscription` and the legacy `armeventgrid.EventSubscriptionsClient` build the scope-bound URL `{scope}/providers/Microsoft.EventGrid/eventSubscriptions/{name}`. Against the emulator this returned **501** at RG/subscription scope (fell through to the BlobStorage fallback) and **400** for the topic extension form, so any Terraform-managed or legacy-SDK event subscription silently failed. Fixes deep-audit finding NEW-1 / DEEP-357.

## How

- `Matches()` now claims the trailing `providers/Microsoft.EventGrid/eventSubscriptions` marker regardless of the outer provider (`azurearm.ParsePath` cannot model the extra nested `providers` segment).
- The **topic** extension form is normalized to the direct-form path and routed through the eventbus driver, so it is the same resource as a direct-form subscription and shares delivery/filter machinery.
- **Non-topic** scopes (subscription / RG / arbitrary resource) have no eventbus topic to hang off, so — mirroring the existing `systemTopics`/`domains` precedent — the wire handler owns their state in memory, keyed by scope+name. Raw ARM properties round-trip verbatim (destination/filter) for Terraform state refresh. No driver interface change; docs/coverage unaffected.

## Tests

New real-SDK reproductions (`scoped_subscriptions_sdk_test.go`) via `armeventgrid.EventSubscriptionsClient`:
- RG-scope create → GET (filter round-trips) → `ListGlobalByResourceGroup` → DELETE → GET 404
- Subscription-scope create → GET → `ListGlobalBySubscription`
- Arbitrary-resource (storage account) scope create → GET → `ListByResource`
- Topic extension form create, visible through the direct `TopicEventSubscriptions` Get + List (unification)

Existing direct-form Event Grid tests stay green.

### DIFF
--- a/server/azure/eventgrid/handler.go
+++ b/server/azure/eventgrid/handler.go
@@ -22,6 +22,8 @@
 //	GET            .../topics/{t}/eventSubscriptions        — TopicEventSubscriptions.List
 //	PUT/GET/DELETE .../domains/{d}/topics/{t}               — DomainTopics CRUD
 //	GET            .../domains/{d}/topics                   — DomainTopics.ListByDomain
+//	PUT/GET/DELETE {scope}/.../eventSubscriptions/{s}      — EventSubscriptions CRUD (subscription/RG/resource scope)
+//	GET            {scope}/.../eventSubscriptions          — EventSubscriptions List (ByResource / Global{BySub,ByRG})
 package eventgrid
 
 import (
@@ -51,6 +53,11 @@ type Handler struct {
 	mu           sync.RWMutex
 	systemTopics map[string]*systemTopicRecord
 	domains      map[string]*domainRecord
+	// scopedSubs holds event subscriptions created as extension resources on a
+	// non-topic scope (subscription, resource group, or an arbitrary resource).
+	// These have no eventbus-driver topic to hang off, so — like systemTopics
+	// and domains — the wire handler owns their state. Keyed by scope+name.
+	scopedSubs map[string]*scopedSubRecord
 }
 
 // New returns an Azure Event Grid handler backed by b.
@@ -59,6 +66,7 @@ func New(b ebdriver.EventBus) *Handler {
 		bus:          b,
 		systemTopics: make(map[string]*systemTopicRecord),
 		domains:      make(map[string]*domainRecord),
+		scopedSubs:   make(map[string]*scopedSubRecord),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


NitinKumar004:
Review — verified against the pushed commit (004a5609).
- build / vet / gofmt / -race clean; full ./server/azure/... suite green (existing eventgrid delivery/filters/domains/systemTopics + dispatch_ordering pass).
- All documented scopes work (author's 4 real-armeventgrid tests): RG (create→GET filter round-trip→ListGlobalByResourceGroup→delete→404), subscription (+ListGlobalBySubscription), arbitrary-resource (+ListByResource), topic-extension (unified with direct form).
- Delivery sound: topic-extension routes through the same bus.PutRule as the direct form (participates in delivery); non-topic scopes live only in scopedSubs (clean CRUD no-op, documented — not a silent drop).
- No forward routing collision: scopedSubMarker fires only on a trailing providers/Microsoft.EventGrid/eventSubscriptions triple; doesn't match direct topic-sub form or topics/domains/systemTopics; eventgrid registered after network/cosmos/dns/lb, before blob fallback.
- Arch: mirrors systemTopics/domains wire-state precedent; no driver change; docs no-diff.
Non-blocking nits (follow-ups): an event-sub hung off a Microsoft.Network/DocumentDB resource is stolen by those handlers (not valid EG event sources, not a regression) — could register eventgrid earlier or guard those Matches(); enrichSubscriptionProperties stamps properties.topic=<scope> for non-topic scopes (harmless).
Verdict: approved.

PR #625, opened 2026-08-25
fix(azure-functions): reject App Service Plan delete while sites assigned

NitinKumar004:
## What

An App Service Plan (`Microsoft.Web/serverfarms`) could be **deleted while a Web App (`Microsoft.Web/sites`) still referenced it**, with no error. The plan vanished and every site was left pointing at a `serverFarmId` that no longer existed — silent state corruption that IaC state files then disagree with.

`DeleteAppServicePlan` now rejects the delete when any site's `ServerFarmID` still targets the plan's ARM id, returning `FailedPrecondition` which the ARM wire layer maps to **409 Conflict** ("Server farm ... cannot be deleted because it has web app(s) assigned to it"). Once the assigned site is deleted, the plan deletes cleanly.

## Why

Real Azure rejects this delete with 409 Conflict (App Service Plans Delete REST API). The previous behavior (`m.plans.Delete(...)` with no reverse reference check) matched neither real Azure nor the emulator's own data-integrity guarantees, and reproduces under any teardown that races app vs plan deletion (`terraform destroy`, `az appservice plan delete`).

## Changes

- `providers/azure/functions/app_service_plan.go` — `DeleteAppServicePlan` checks plan existence, then scans sites for a matching `ServerFarmID` (new `planAssignedSite` helper, scanning the whole subscription since a site's plan may live in another resource group — mirrors the `listPlanWebApps` join). Rejects with `FailedPrecondition` (→ 409) when a site is still assigned.
- Tests:
  - `server/azure/functions/rgscope_sdk_test.go` — `TestSDKAppServicePlanDeleteRejectedWhileSiteAssigned` (real `armappservice/v3`): create plan → create web app on it → DELETE plan → 409 Conflict; delete the web app → DELETE plan succeeds. Updated `TestSDKAppServicePlanDeleteAndListWebApps` to delete the site before the plan.
  - `providers/azure/functions/app_service_plan_test.go` — provider-level guard test covering the reject + delete-after-site-removed path.

### DIFF
--- a/providers/azure/functions/app_service_plan.go
+++ b/providers/azure/functions/app_service_plan.go
@@ -3,8 +3,10 @@ package functions
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	cerrors "github.com/stackshy/cloudemu/v2/errors"
+	"github.com/stackshy/cloudemu/v2/internal/idgen"
 )
 
 // AppServicePlan is an Azure App Service plan (Microsoft.Web/serverfarms) — the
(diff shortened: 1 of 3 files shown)


NitinKumar004:
Review — verified against the pushed commit (b41d981a).
- Guard real: wire deletePlan → DeleteAppServicePlan → planAssignedSite scan → FailedPrecondition → WriteCErr → 409; plan existence checked first, actual Delete only when unreferenced.
- No false positive: a plan with no sites deletes fine; a site referencing a different plan doesn't block this one (verified with throwaway adversarial tests).
- Cross-RG reference caught: planAssignedSite scans all sites and matches the site's ServerFarmID against the plan's full ARM id (which embeds the plan's RG), so a site in rgB pointing at a plan in rgA is caught; the id is byte-identical to the tested listPlanWebApps join.
- Error mapping 409 (PreconditionFailed envelope); test change legitimate (the old test deleted a plan with a site still assigned — that encoded the bug; now deletes the site first); -race clean; no collateral.
- Arch: provider-layer guard protects all run modes; FailedPrecondition matches the dependency-guard convention (vnet in-use, SQL elastic-pool).
Non-blocking nits: real Microsoft.Web returns code 'Conflict' for this case vs the shared mapping's 'PreconditionFailed' (status 409 + envelope correct); bare-name serverFarmId would be missed (mirrors existing join; real clients send full ARM id).
Verdict: approved.

PR #626, opened 2026-08-25
fix(azure-keyvault): certificates data-plane + stop mis-routing to storage

NitinKumar004:
## What

Fixes the Key Vault certificates data-plane bug (deep-audit finding KV-B2, HIGH).

Certificate requests on a vault host (`/certificates/...`, `/deletedcertificates/...`) were not claimed by any Key Vault handler, so in the full emulator they fell through to the permissive Table/Blob/Queue storage fallbacks and returned a wrong-domain error — e.g. `POST /certificates/{name}/create` answered HTTP 400 `odata.error` "PartitionKey and RowKey are required" to the real `azcertificates` SDK. This was a dispatch/routing bug compounded by a missing feature.

### Routing fix (the core bug)
- New `CertsHandler` (`server/azure/keyvault/certs_handler.go`) claims `/certificates` and `/deletedcertificates`, registered in `server/azure/azure.go` **before** the storage fallbacks. Certificate requests now always reach Key Vault, never Table/Blob storage.

### Certificate CRUD data-plane (mirrors the secrets/keys pattern)
- New `KeyVaultCertificates` driver surface (`services/secrets/driver/certificates.go`), a type-asserted optional interface like `KeyVaultKeys`/`KeyVaultSecrets`.
- Provider impl (`providers/azure/keyvault/certs_azure.go`) generates a real self-signed X.509 certificate on create and stores it per-vault (reusing the existing `vaultData` isolation).
- Operations: create (`POST /certificates/{name}/create`, 202 CertificateOperation), get current/version, list, list versions, get policy (round-tripped verbatim), get pending operation, soft-delete, and deletedcertificates get/list/recover/purge — with Key Vault JSON shapes (`cer`, `x5t`, `kid`, `sid`, `attributes`, `policy`) and error code `CertificateNotFound` (404).
- Coverage docs regenerated (`go generate`).

## Why

So real `azcertificates`/raw-REST callers get proper Key Vault certificate responses instead of a baffling storage error, and certificate CRUD round-trips.

## Tests
- `certs_dataplane_test.go` (raw REST — no `azcertificates` in go.mod): full create/get/list/versions/delete round-trip; missing cert → 404 `CertificateNotFound`; **mis-routing regression** against the full server (KeyVault + Table/Blob/Queue) asserting create/list are Key Vault responses, not `odata.error`/XML; per-vault isolation via host.
- `certs_azure_test.go` (provider): create/get/versions/list, X.509 parse of the generated cert, soft-delete/recover/purge, per-vault isolation, empty-name rejection.
- Existing Key Vault secrets/keys suites stay green.

## Deferred (advanced certificate surface)
Recorded for a follow-up; the mis-routing bug is fully fixed regardless:
- Certificate policy **update**, issuers, and contacts (policy is currently round-tripped read-only on GET).
- Certificate **import** and **merge** (CSR signing).
- Independent backing key/secret objects: the bundle returns `kid`/`sid` URLs for shape fidelity, but those key/secret objects are not separately addressable yet.
- Backup/restore for certificates.


NitinKumar004:
Review — verified against the pushed commit (c5b9043c). The mis-routing fix and cert CRUD are correct; one blocking gate stands.
Verified: mis-routing fixed (cert requests reach KeyVault before the Table/Blob/Queue fallbacks, no odata garbage); cert CRUD real (create→get→versions→list→delete→recover→purge; cer parses as valid X.509 DER, x5t/kid/sid shapes correct; missing→404 CertificateNotFound); per-vault isolation holds (host-derived vaultData.certs, SetIfAbsent race-safe); coverage docs correct (only the KeyVaultCertificates block); -race clean; no collateral.

BLOCKING — lint gate (repo requires 0 issues): 11 issues on changed production files —
- dupl x4 without nolint: certs_azure.go GetCertificate/DeleteCertificate/RecoverDeletedCertificate (dups of the Secret equivalents) + certs_operations.go writeCertErr. Add the same explained //nolint:dupl the parallel secret/key routers carry, or refactor.
- gocritic hugeParam: CreateCertificate takes KVCreateCertificateParams (120B) by value → pointer.
- gocritic unnamedResult: generateSelfSigned returns ([]byte, time.Time, time.Time, error) → name them.
- gocyclo: routeNamedCert complexity 11 → split.

Non-blocking: a blob container named 'certificates' is stolen by the cert handler — but this is PRE-EXISTING (the merged secrets/keys handlers steal 'secrets'/'keys' identically; inherent to the single-port path-first-match design), so not grounds to block this PR; tracking as a follow-up to host-gate ALL KV data-plane matchers on .vault.azure.net. kid/sid emitted without a separately-addressable backing key/secret (documented fidelity deferral).
Verdict: changes requested (lint gate only).

NitinKumar004:
Re-review — verified against the pushed commit (bcd81a54), after the lint-only fix. The functional review already passed (mis-routing fixed, cert CRUD correct, per-vault isolation, no collateral); this confirms the gate + behavior-neutrality.
- Lint gate: golangci-lint 0 issues on ./providers/azure/keyvault/..., ./server/azure/keyvault/..., and ./services/secrets/... (the interface-change pkg).
- Behavior-neutral: the nolint:dupl additions to pre-existing files (keys_store_azure.go, keyvault_azure.go, server handler.go) are 2-line doc-comment insertions only, zero logic change; CreateCertificate's *KVCreateCertificateParams is read-only, never stored/mutated, both call sites pass non-nil; generateSelfSigned named results preserve values; routeNamedCert split (routeReservedCertSub) handles the same reserved segments in the same order with identical fallthrough — routing exactly preserved.
- build / vet / gofmt clean; cert CRUD + cross-vault isolation tests green; -race clean; docs/coverage no-diff.
Verdict: approved.

PR #627, opened 2026-08-25
fix(azure-compute): PATCH dataDisk detach + clear disk managedBy on VM delete

NitinKumar004:
## What

Two Azure Microsoft.Compute disk-lifecycle correctness fixes, both localized to the `server/azure/virtualmachines` handler.

1. **PATCH `VirtualMachines.Update` dataDisk detach.** `az vm update` / `armcompute.VirtualMachinesClient.BeginUpdate` remove a data disk by sending a modified (shorter/empty) `storageProfile.dataDisks` array. Previously the PATCH path treated a supplied array as merge-patch and only detached a disk whose entry set `toBeDetached`, so an omitted disk stayed attached forever (managedBy/diskState never cleared). PATCH now distinguishes an **omitted** array (nil → merge-patch, untouched) from a **supplied** array (non-nil, empty included → full replace: detach every attached disk whose LUN is absent), matching the PUT path and real Azure's read-modify-write semantics.

2. **Clear disk `managedBy` on VM delete.** Deleting a VM left each attached managed disk's `managedBy` pointing at the now-deleted VM (default `deleteOption=Detach` keeps the disk). VM delete now detaches every attached volume first, clearing each disk's `managedBy`/`diskState` (back to `Unattached`) so the disk is free and re-attachable.

Both reuse the existing driver `DetachVolume` — the same call `applyDataDisks` uses — rather than duplicating detach logic.

## Why

Verified against the official Microsoft.Compute virtualMachines Update docs: a supplied `dataDisks` array on PATCH replaces the list (omitted LUNs detach), and a deleted VM releases its detach-option disks. The emulator diverged on both, breaking `az vm update`/Terraform disk removal and leaving disks unusably "still-managed" after VM deletion.

## Tests

- New `datadisk_lifecycle_test.go` (real `armcompute/v5` SDK):
  - `TestSDKVMPATCHDataDiskDetachByOmission` — PATCH omitting a LUN detaches that disk (managedBy cleared, diskState Unattached); an explicit empty array detaches all.
  - `TestSDKVMDeleteClearsDiskManagedBy` — delete a VM with an attached disk clears managedBy/diskState and the disk re-attaches to a new VM.
- Updated `TestSDKVMDataDiskAttachDetach` to the correct read-modify-write PATCH shape (the old step encoded the now-fixed merge-patch-attach assumption).

Gates: build, vet, scoped tests, -race, golangci-lint (0 issues), gofmt — all green.

### DIFF
--- a/server/azure/virtualmachines/datadisk_attach_test.go
+++ b/server/azure/virtualmachines/datadisk_attach_test.go
@@ -133,13 +133,20 @@ func TestSDKVMDataDiskAttachDetach(t *testing.T) {
 		t.Errorf("disk-0 diskState=%v, want Attached", diskStateOf(gotDisk1))
 	}
 
-	// PATCH (BeginUpdate) attaches disk-1 at lun 1 — the non-recreating
-	// merge-patch path — without disturbing the existing lun-0 attachment.
+	// PATCH (BeginUpdate) adds disk-1 at lun 1 via the real read-modify-write
+	// pattern: the desired dataDisks list carries BOTH lun 0 (kept) and lun 1
+	// (new). A PATCH's supplied dataDisks array is a full replace, so keeping an
+	// existing attachment means re-listing it — the same way `az vm disk attach`
+	// GETs, appends, then updates.
 	updatePoller, err := vmClient.BeginUpdate(ctx, "rg-1", "vm-disks",
 		armcompute.VirtualMachineUpdate{
 			Properties: &armcompute.VirtualMachineProperties{
 				StorageProfile: &armcompute.StorageProfile{
 					DataDisks: []*armcompute.DataDisk{{
+						Lun:          to.Ptr[int32](0),
+						CreateOption: to.Ptr(armcompute.DiskCreateOptionTypesAttach),
+						ManagedDisk:  &armcompute.ManagedDiskParameters{ID: disk1.ID},
+					}, {
 						Lun:          to.Ptr[int32](1),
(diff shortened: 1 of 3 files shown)


NitinKumar004:
Review — verified against the pushed commit (0d6a8776).
- PATCH detach real: supplying only lun 1 detaches lun 0 (diskState Unattached, managedBy cleared); explicit empty array detaches all.
- Data-loss guard holds (the critical check): verified against the real armcompute SDK — nil DataDisks serializes with NO dataDisks key (merge-patch, no detach); empty non-nil serializes as [] (full replace); a tags-only / properties-only PATCH leaves disks attached. json.Unmarshal preserves nil-vs-empty, so dataDisksPresent() is a sound discriminator. No accidental detach.
- managedBy cleared on delete: detachAttachedVolumes runs before TerminateInstances; diskless VM delete is a no-op; disk re-attachable. -race clean; no collateral (PUT-path detach untouched).
- Updated test legitimate (read-modify-write shape for full-replace PATCH, LUN assertions preserved). Arch: reuses driver DetachVolume, no duplicate bookkeeping.
Follow-ups (non-blocking): RG-cascade VM delete (PurgeResourceGroup) bypasses detachAttachedVolumes → cascade leaves disks dangling (tracked as round-2 R1); stale comment in datadisk_attach_test.go; detachAttachedVolumes detaches every volume incl. a hypothetical OS disk (harmless — VMs built from imageReference).
Verdict: approved.

PR #628, opened 2026-08-25
fix(azure-networking): RG-scope securityRules + peerings sub-resource parent resolution

NitinKumar004:
## What

RG-scope the parent resolution for two Azure Microsoft.Network **sub-resources** so an op is scoped to its parent's resource group.

- **N1 (securityRules / `azurerm_network_security_rule` / `SecurityRulesClient`):** `listSecurityRules` / `getSecurityRule` / `putSecurityRule` / `deleteSecurityRule` resolved the parent NSG via `findNSGByName` (name only). Repointed to `findNSGInGroup(rp.ResourceGroup, rp.ResourceName)`.
- **N2 (virtualNetworkPeerings / `VirtualNetworkPeeringsClient`):** the local-vnet lookups in `list`/`get`/`put`/`delete` (and the remote ref in `resolvePeeringRemote`) resolved via `findVNetByName` (name only). Repointed the local lookups to `findVNetInGroup(rp.ResourceGroup, rp.ResourceName)` and the remote to `findVNetInGroup(remoteRP.ResourceGroup, …)`.

## Why

The whole-NSG / whole-vnet GET/PUT/DELETE paths were already RG-scoped (via `findNSGInGroup` / `findVNetInGroup`, added in #623), but the sub-resource routes were left name-only. With two same-named parents in different resource groups (a legal Azure pattern — the same Terraform module applied to two RGs), a sub-resource op addressed to one group silently operated on the other group's parent (the first-created match): a rule/peering "created in rgB" was readable and mutable under rgA. This routes the sub-resources through the *same* RG-scoped finders the parent GET/PUT/DELETE already use — no new pattern introduced.

Correct Azure behavior: a sub-resource op is scoped to its parent's resource group, and an op under a group holding no such parent is a 404.

## Tests

`server/azure/vnet/subresource_rgscope_test.go` — real `armnetwork` `SecurityRulesClient` / `VirtualNetworkPeeringsClient` against the in-process server, two same-named parents in `rgA`/`rgB`:
- a sub-resource PUT under rgB touches only rgB's parent (rgA's parent unchanged);
- the sub-resource is not visible under rgA's same-named parent (GET + list);
- an op under a group with no such parent → 404;
- a wrong-RG DELETE → 404 and leaves rgB's sub-resource intact.

Both tests were confirmed to fail on the pre-fix code and pass after. Existing securityRule / peering / NSG / vnet tests (incl. the #623 RG-scope tests) stay green.

### DIFF
--- a/server/azure/vnet/security_rule_subresource.go
+++ b/server/azure/vnet/security_rule_subresource.go
@@ -52,7 +52,7 @@ func (h *Handler) routeSecurityRule(w http.ResponseWriter, r *http.Request, rp a
 
 //nolint:gocritic // rp is a request-scoped value
 func (h *Handler) listSecurityRules(w http.ResponseWriter, r *http.Request, rp azurearm.ResourcePath) {
-	info, err := findNSGByName(r.Context(), h.net, rp.ResourceName)
+	info, err := findNSGInGroup(r.Context(), h.net, rp.ResourceGroup, rp.ResourceName)
 	if err != nil {
 		azurearm.WriteCErr(w, err)
 		return
@@ -67,7 +67,7 @@ func (h *Handler) listSecurityRules(w http.ResponseWriter, r *http.Request, rp a
 
 //nolint:gocritic // rp is a request-scoped value
 func (h *Handler) getSecurityRule(w http.ResponseWriter, r *http.Request, rp azurearm.ResourcePath) {
-	info, err := findNSGByName(r.Context(), h.net, rp.ResourceName)
+	info, err := findNSGInGroup(r.Context(), h.net, rp.ResourceGroup, rp.ResourceName)
 	if err != nil {
 		azurearm.WriteCErr(w, err)
 		return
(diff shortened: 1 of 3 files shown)


NitinKumar004:
Review — verified against the pushed commit (44a73ff0).
- RG-scoping real: securityRules + peerings sub-resource handlers repointed from name-only finders to findNSGInGroup / findVNetInGroup(rp.ResourceGroup, name); wrong-RG op → 404; sibling RG's same-named parent untouched.
- Cross-RG peering remote intact (the main risk): resolvePeeringRemote parses the remote vnet id from the request body (which embeds the remote's own RG) and scopes findVNetInGroup to remoteRP.ResourceGroup — NOT the local RG. Verified with adversarial tests: local rgA → remote rgB peering succeeds; a remote id naming a non-existent RG 404s (proving it's genuinely RG-scoped, not name-only fallback).
- No collateral: name-only finders remain in use exactly where claimed (handler/effective_nsg/network_interface/subnet_validate), unaffected; full vnet suites green; -race clean. Arch: reuses #623 finders, no new pattern, no dead code.
Verdict: approved.

PR #629, opened 2026-08-25
fix(azure-servicebus): TTL of scheduled messages counts from enqueue time, not submit

NitinKumar004:
## What

A scheduled Service Bus message's TTL clock started at submit time instead of at the scheduled enqueue time. So a message scheduled for the future with a `TimeToLive` shorter than the schedule delay expired **before it ever became visible** and was silently lost.

Fix: base the message expiry on the effective enqueue time (`visibleAt`) rather than the submit time:

```go
expiresAt := computeExpiry(visibleAt, input.MessageTTLSeconds)
```

For a non-scheduled send `visibleAt == now`, so that path is unchanged.

## Why

Per the official Azure Service Bus docs (message expiration): *"The message expiration time depends on the enqueued time, not on the time at which the message is sent to Service Bus. Therefore, the expires-at-utc is still enqueued time + time-to-live."* Example from the docs: schedule +5 min, TTL 10 min → expires at +15 min.

Before this fix, a message scheduled for +60s with a 30s TTL got `ExpiresAt = submit + 30s = +30s`, which is before its +60s enqueue time, so it was reaped and never delivered (round-2 re-audit finding S1, MED, silent message loss).

## Tests

- `providers/azure/servicebus/scheduled_ttl_test.go` — scheduled message survives until its enqueue time and is delivered; its `ExpiresAt` equals enqueue+TTL; it expires only after enqueue+TTL; a non-scheduled message's TTL still counts from submit.
- `server/azure/servicebus/dataplane_message_test.go` — `TestDataPlaneScheduledMessageTTLFromEnqueue`: end-to-end reproduction over the Brokered-Messaging REST data plane with a FakeClock — a message scheduled +120s with a 30s TTL is delivered after enqueue instead of being lost.

Both new tests fail on the pre-fix code and pass after. Existing Service Bus tests (incl. the #621 DLQ/TTL/scheduled tests) stay green; `-race` clean; lint 0 new issues.

### DIFF
--- a/providers/azure/servicebus/scheduled_ttl_test.go
+++ b/providers/azure/servicebus/scheduled_ttl_test.go
@@ -1,0 +1,100 @@
+package servicebus
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stackshy/cloudemu/v2/services/messagequeue/driver"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestScheduledMessageSurvivesUntilEnqueue confirms a message scheduled for a
+// future enqueue time with a TTL shorter than the schedule delay is still
+// delivered once it becomes active (rather than being silently reaped before it
+// is ever visible). Real Azure computes expires-at-utc from the enqueued time,
+// not the submit time.
+func TestScheduledMessageSurvivesUntilEnqueue(t *testing.T) {
+	ctx := context.Background()
+	m, clk := newTestMock()
+
+	url := createStdQueue(t, m)
+
+	// Scheduled 60s out with a 30s TTL: on real Azure the message materializes at
+	// +60s and lives until +90s. The buggy path expired it at submit+30s (+30s),
+	// before it ever became visible.
+	_, err := m.SendMessage(ctx, driver.SendMessageInput{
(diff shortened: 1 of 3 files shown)


NitinKumar004:
Review — verified against the pushed commit (918af36d).
- Core fix: computeExpiry(now,...) → computeExpiry(visibleAt,...) establishes ExpiresAt = VisibleAt + TTL ≥ VisibleAt for all TTL≥0. Since reapExpired runs before the VisibleAt.After(now) visibility check, the old 'now' anchor could make ExpiresAt < VisibleAt → a not-yet-visible scheduled message reaped early (silent loss); the new anchor makes premature reaping impossible.
- Mutation-verified: reverting the line fails TestScheduledMessageSurvivesUntilEnqueue (0 items) and the e2e returns 204 (lost) pre-fix / 200 post-fix. Expiry asserted at enqueue+TTL (matches Azure expires-at = enqueued + TTL).
- Non-scheduled unchanged (delaySeconds==0 ⇒ visibleAt==now); TTL=0 reaps exactly at visibility (correct, no panic); DLQ dead-letters at enqueue+TTL when deadLetterOnExpiration set; no #621 regression; -race clean. Arch: config.Clock/FakeClock consistent with #621.
Non-blocking nits: the e2e derives the schedule delay from wall-clock while reaping is on FakeClock (~1s margin — works, thin); cancel-before-enqueue N/A (scheduling modeled as delivery delay).
Verdict: approved.

PR #630, opened 2026-08-25
fix(azure-cosmos): account delete fully tears down namespace + prunes accountAttrs

NitinKumar004:
## What

Deleting a Cosmos `databaseAccount` was a **shallow** teardown. The ARM delete path called `DeleteTable(account)`, which dropped only the account's own driver table — it never pruned `accountAttrs`, the account's container tables, or the data-plane handler's in-memory bookkeeping.

Consequences (round-2 re-audit finding C1, MED):
- the deleted account **still appeared in `List`** (a ghost — while `GET` 404'd, an inconsistency real clients trip over);
- its data-plane namespace stayed live (`isAccount` still true);
- recreating a **same-named** account (CI teardown+setup, `terraform destroy`+`apply`) **inherited the previous incarnation's databases/containers/items** — a stale-data leak across lifecycles, especially after the per-account isolation work in #620.

## Why / Fix

Tear the account down fully, across all three layers:

- **Driver** (`providers/azure/cosmosdb`): new `PurgeAccount(account)` drops the account's own table, its `accountAttrs` entry, and every `{account}/…` container table (only that account's prefix — other accounts' namespaces are untouched).
- **Data-plane wire handler** (`server/azure/cosmosdb`): new `PurgeAccount(ctx, account)` prunes its own `databases` set, shared-throughput offers, and container TTL/attrs bookkeeping for the account, then delegates to the driver purge.
- **ARM control plane** (`server/azure/cosmosaccount`): `deleteAccount` delegates to the data-plane purger and returns **204** — the real `armcosmos` `BeginDelete` poller accepts only 202/204 (a 200 fails its client-side validation), so account delete now works via the real SDK.

After delete the account is gone from `List` (no ghost), `GET` 404s, its data plane stops serving, and a same-name recreate starts from an empty namespace.

## Tests

- Real `armcosmos` v3 + `azcosmos` end-to-end: create account → db/container/item → `BeginDelete` → account NOT in `List`, `GET` 404 → recreate same name → its database/container list is EMPTY (no stale data) → a different account's data is untouched.
- Provider-level `PurgeAccount` unit tests (namespace teardown + other-account isolation + unknown-account no-op).
- Existing cosmos tests (incl. #620 isolation + indexingPolicy round-trip) stay green.

### DIFF
--- a/providers/azure/cosmosdb/cosmosdb.go
+++ b/providers/azure/cosmosdb/cosmosdb.go
@@ -195,6 +195,29 @@ func (m *Mock) DeleteTable(_ context.Context, name string) error {
 	return nil
 }
 
+// PurgeAccount fully tears down a Cosmos databaseAccount's driver footprint: the
+// account's own table, every data-plane container table in its namespace (the
+// tables whose name is prefixed "{account}/"), and the account's discovery
+// attributes (accountAttrs). A plain DeleteTable removes only the single named
+// table, which left the deleted account visible in AccountTables and its
+// container tables live; PurgeAccount is the account-delete teardown that makes
+// a deleted account stop listing and a same-name recreate start from an empty
+// namespace.
+func (m *Mock) PurgeAccount(account string) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	delete(m.tables, account)
+	delete(m.accountAttrs, account)
+
+	prefix := account + "/"
+	for name := range m.tables {
+		if strings.HasPrefix(name, prefix) {
+			delete(m.tables, name)
+		}
+	}
+}
+
 // DescribeTable returns the configuration of a container (table).
 func (m *Mock) DescribeTable(_ context.Context, name string) (*driver.TableConfig, error) {
 	m.mu.RLock()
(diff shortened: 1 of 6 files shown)


NitinKumar004:
Review — verified against the pushed commit (512b8482).
- Prefix-safety (the main risk) SAFE: purge matches on account+'/' at every layer (driver HasPrefix(name, account+'/'); wire nsPrefix(account)=account+'/'; database reaping CutPrefix(key, account+'/')). Verified with an adversarial foo/foobar test — purging 'foo' removed foo + foo/db1/coll1 while foobar + foobar/db1/coll1 survived intact.
- Fix real + deep: driver drops account table + accountAttrs + all {account}/… container tables; wire prunes databases set + shared-throughput offers + container TTL/attrs; ARM deleteAccount delegates via AccountPurger. e2e confirms no List ghost, GET 404, clean same-name recreate, keeper account intact.
- 204 correct (armcosmos BeginDelete poller accepts 202/204; no test asserts 200). build/vet/gofmt clean; -race clean (map delete-during-range spec-safe, no lock inversion); no collateral (#620 isolation/indexingPolicy/endpoint tests green). Arch: 3-layer, reuses #620 keying.
Follow-ups (non-blocking, being hardened in a small PR): add a foo/foobar prefix-subset regression test to lock the guarantee into CI; add an empty-account guard in Handler.PurgeAccount (unreachable today but a latent match-all footgun).
Verdict: approved.

PR #631, opened 2026-08-25
fix(azure-rm): complete RG cascade (public IP/NIC/NAT/LB) + detach disks on cascade VM delete

NitinKumar004:
## What

Completes the Azure resource-group cascade delete introduced in #623 and folds in the #627 cascade disk-detach review nit.

- **vnet `PurgeResourceGroup`** now also purges **network interfaces, NAT gateways and public IPs** (RG-tag filtered, `EqualFold`), in addition to the existing vnets + NSGs. Order is chosen so freed references let each delete succeed (NICs → NAT gateways → public IPs → vnets → NSGs).
- **Load balancer handler** now implements `ResourceGroupPurger` and is registered as a purger in `server/azure/azure.go` (it was not previously cascaded at all).
- **Purger order** in `azure.go`: VMs are purged **before** networking, so a terminated VM clears its NICs' `virtualMachine` back-reference and the NIC delete is not blocked by the attached-NIC guard.
- **#627 nit**: the compute cascade purge (`PurgeResourceGroup`) now routes each VM through `detachAttachedVolumes` before terminating, so cascade-deleted VMs clear their data disks' `managedBy`/`diskState` (deleteOption=Detach) — the same release the single-VM `delete()` path already does. Previously only the single-VM path was fixed.

## Why

Deleting an RG left its public IPs, NICs, NAT gateways and load balancers alive and globally addressable (round-2 finding **R1**, MED). Because their lookups are RG-tag scoped by a tag that survives, a GET under the now-deleted RG still returned the orphan — a true ghost that blocks name reuse and shows in subscription-wide lists. Real Azure (`Microsoft.Resources` RG delete) removes every resource contained in the group. Separately, cascade-deleted VMs left their detached data disks with `managedBy` still set.

Resources in a **different** resource group are left untouched (RG-tag `EqualFold` scoping).

## Tests

New real-SDK reproduction `TestResourceGroupDeleteCascadesNetworkChildrenAndDetachesDisks` (armnetwork + armcompute + armresources against the in-process server): creates an RG holding a public IP, a NIC (referencing the public IP), a NAT gateway, a load balancer, and a VM-with-data-disk whose NIC is the created one; deletes the RG; asserts all network children are gone (GET 404) and the data disk survives with `managedBy` cleared / `diskState` Unattached — while a public IP in a different RG is untouched. Existing #623 cascade + cross-RG scope tests stay green.

### DIFF
--- a/server/azure/azure.go
+++ b/server/azure/azure.go
@@ -221,17 +221,35 @@ func New(d Drivers) http.Handler {
 		vnetHandler    *vnet.Handler
 		vmHandler      *virtualmachines.Handler
 		storageHandler *storageaccountsrv.Handler
+		lbHandler      *lbsrv.Handler
 		rgPurgers      []resourcegroups.ResourceGroupPurger
 	)
+
+	// Virtual machines are purged before the networking resources they consume
+	// (NICs, subnets): tearing a VM down first clears its NICs' virtualMachine
+	// back-reference, so the vnet purger's NIC delete is not blocked by the
+	// attached-NIC guard.
+	if d.VirtualMachines != nil {
+		vmHandler = virtualmachines.New(d.VirtualMachines, d.Network)
+		rgPurgers = append(rgPurgers, vmHandler)
+	}
 
 	if d.Network != nil {
 		vnetHandler = vnet.New(d.Network)
 		rgPurgers = append(rgPurgers, vnetHandler)
 	}
 
-	if d.VirtualMachines != nil {
-		vmHandler = virtualmachines.New(d.VirtualMachines, d.Network)
-		rgPurgers = append(rgPurgers, vmHandler)
+	if d.LB != nil {
+		lbHandler = lbsrv.New(d.LB)
+
+		// Project a backend address pool's read-only backendIPConfigurations from
+		// the NIC side of the association, so NIC↔LB pool membership reflects on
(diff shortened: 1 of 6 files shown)


NitinKumar004:
Review — verified against the pushed commit (7f936327).
- Cascade complete: public IP + NIC + NAT gateway + LB + vnet all 404 after RG delete; the VM's data disk survives with managedBy cleared + DiskState Unattached; cross-RG public IP untouched.
- Cascade disks detached: compute PurgeResourceGroup loops detachAttachedVolumes per VM before TerminateInstances (the #627 fix on the cascade path).
- Ordering covers the full dependency graph (traced against the driver guards): VM→(NICs→NAT→publicIPs→vnets→NSGs)→LB→storage. NIC↔VM: TerminateInstances→detachNICs clears the back-ref before NIC delete. NAT→publicIP: NAT delete clears AssociationID before ReleaseAddress. publicIP-on-NIC: NIC create never AssociateAddress, so release always succeeds. LB↔NIC: no guard blocks either; both removed. No orphan chains.
- Dispatch order preserved (only construction hoisted; srv.Register positions byte-identical, each handler registered once). -race clean; no collateral (#623 cascade + cross-RG intact). Arch: reuses #623 ResourceGroupPurger + RG-tag filtering.
Verdict: approved.

PR #632, opened 2026-08-25
fix(azure-eventgrid): exclude resource-scoped subs from global list operations

NitinKumar004:
## What

Regional (resource-scoped) event subscriptions were leaking into the Event Grid **global** list operations. `scopedSubInListScope` (`server/azure/eventgrid/scoped_subscriptions.go`) now excludes records whose own scope carries a nested resource `/providers/` segment from the `ListGlobalBySubscription` and `ListGlobalByResourceGroup` branches.

Repro (3 subs under the same subscription):
- subscription scope (`/subscriptions/{s}`)
- resource-group scope (`/subscriptions/{s}/resourceGroups/{rg}`)
- storage-account resource scope (`.../providers/Microsoft.Storage/storageAccounts/acct1`)

Before: `ListGlobalByResourceGroup(rg)` returned `[at-res, at-rg]` and `ListGlobalBySubscription()` returned `[at-res, at-rg, at-sub]` — the resource-scoped sub leaked into both.
After: global lists return only the global (subscription / RG scoped) subs; the resource-scoped sub is reachable only via `ListByResource` (unchanged).

## Why

Per the official Azure Event Grid REST docs, `ListGlobalBySubscription` / `ListGlobalByResourceGroup` return only **global** event subscriptions; resource-scoped subs are **regional** and belong to `ListByResource` (there are separate `ListRegionalBy*` operations for them). The old code matched every record with the same subscription/RG because `parseScopeContainers` also sets `resourceGroup` on a resource-scoped record. A reconcile/delete loop enumerating "the RG's global subs" could otherwise touch subs it shouldn't.

## Tests

Added `TestSDKEventSubscriptionGlobalListsExcludeResourceScoped` (real `armeventgrid` SDK) asserting the three list operations return the correct disjoint sets. Existing #624 scoped-sub tests remain green.

### DIFF
--- a/server/azure/eventgrid/scoped_subscriptions.go
+++ b/server/azure/eventgrid/scoped_subscriptions.go
@@ -231,12 +231,18 @@ func (h *Handler) listScopedEventSubscriptions(w http.ResponseWriter, sp *scoped
 }
 
 // scopedSubInListScope reports whether rec belongs in the list requested at sp.
-// A resource-extension scope (containing a nested /providers/) matches exactly;
-// a resource-group scope matches subscription+group; a subscription scope
-// matches the subscription (ListGlobalBySubscription).
+// A resource-extension scope (containing a nested /providers/) matches exactly
+// (ListByResource). The global list operations (ListGlobalBySubscription /
+// ListGlobalByResourceGroup) return only global subscription/resource-group
+// scoped subs, never regional resource-scoped ones — so those branches exclude
+// records whose own scope carries a nested resource /providers/ segment.
 func scopedSubInListScope(rec *scopedSubRecord, sp *scopedSubPath) bool {
 	if strings.Contains(sp.scope, "/providers/") {
 		return rec.scope == sp.scope
+	}
+
+	if isResourceScoped(rec.scope) {
+		return false
 	}
 
 	if sp.resourceGroup != "" {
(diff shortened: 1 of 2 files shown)


NitinKumar004:
Review — verified against the pushed commit (438bdad9).
- build / vet / gofmt / lint clean (changed file 0 issues).
- Fix real + load-bearing: isResourceScoped(rec.scope) short-circuits resource-scoped subs in the ListGlobalBySubscription / ListGlobalByResourceGroup branches. Test asserts ListGlobalBySubscription=[at-rg,at-sub], ListGlobalByResourceGroup=[at-rg], ListByResource=[at-res]; without the guard at-res leaks into both global lists.
- Discriminator correct: rec.scope/sp.scope are the scope segments before the eventSubscriptions marker, so a global scope (/subscriptions/{s} or .../resourceGroups/{rg}) has no /providers/ while a resource-extension scope does. An RG literally named 'providers' → /resourceGroups/providers → Contains('/providers/') is false, so no misclassification.
- Sub-level list correctly includes BOTH sub-scope and rg-scope (final branch matches on subscription when sp.resourceGroup==''); ListByResource branch untouched; no collateral to #624; arch minimal wire-layer change.
Verdict: approved.

PR #633, opened 2026-08-25
test(azure-cosmos): lock prefix-safety on account purge + empty-account guard

NitinKumar004:
## What

Hardening follow-up to #630 (Cosmos account-delete full-namespace purge). Adds tests that lock the correctness guarantees into CI, plus a defensive empty-account guard.

- **Prefix-subset regression tests** (foo vs foobar): create accounts `foo` and `foobar`, each with a database + container + item; purge/delete `foo`; assert `foo` and its data are gone while `foobar` and ITS data survive intact.
  - Driver level: `TestPurgeAccountPrefixSubset` in `providers/azure/cosmosdb/purge_account_test.go`.
  - Wire/real-SDK level: `TestSDKCosmosAccountDeletePrefixSubset` in `server/azure/cosmosaccount/account_delete_test.go` (drives armcosmos + azcosmos end-to-end).
- **Empty-account guard**: `Handler.PurgeAccount` (server) and driver `PurgeAccount` return early (no-op) when `account == ""`, with `TestPurgeAccountEmpty` covering it.

## Why

The existing purge tests only used disjoint names (acct1/acct2), so the exact foo-vs-foobar prefix-subset regression class wasn't guarded. `nsPrefix("")==""` makes `HasPrefix(t, "")` always true, so an empty account would match and delete ALL tables — a latent match-all data-loss footgun (unreachable today, but now hardened).

Verified the new foo/foobar test FAILS when the purge is temporarily changed to a bare `HasPrefix(name, account)` without the `/` separator, and passes after reverting — proving it catches the regression class. No other behavior changed.

### DIFF
--- a/providers/azure/cosmosdb/cosmosdb.go
+++ b/providers/azure/cosmosdb/cosmosdb.go
@@ -204,6 +204,13 @@ func (m *Mock) DeleteTable(_ context.Context, name string) error {
 // a deleted account stop listing and a same-name recreate start from an empty
 // namespace.
 func (m *Mock) PurgeAccount(account string) {
+	// Guard against an empty account: prefix "" makes HasPrefix always true, so an
+	// empty account would match and delete every table (a match-all data-loss
+	// footgun). Purging the empty/default account is a no-op.
+	if account == "" {
+		return
+	}
+
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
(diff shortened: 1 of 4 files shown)


NitinKumar004:
Review — CI-gated (test + 7-line defensive guard; mutation-verified: reverting to bare HasPrefix fails TestPurgeAccountPrefixSubset). foo/foobar prefix-subset regression tests at driver + real-SDK levels; empty-account guard in both wire and driver PurgeAccount. Full CI green. Verdict: approved.

PR #634, opened 2026-08-25
fix(azure-blob): fail-closed on unknown PUT comp + SetTags/GetTags (stop blob-body corruption)

NitinKumar004:
## What

Fixes a HIGH data-corruption bug class in the Azure Blob wire handler where a `PUT /{container}/{blob}?comp=<x>` with an unhandled `comp` selector fell through to `putBlob` and wrote the sub-operation's request body **over the blob's content**.

- **SetTags (`?comp=tags`) — HIGH data corruption:** had no case in the blob-PUT dispatch, so `BlobClient.SetTags` silently replaced the blob body with the tags XML. Now implemented (Set/Get Blob Tags) and the body is preserved.
- **UploadPages (`?comp=page`) — same root cause:** page-blob range writes fell through and clobbered the whole blob.

## Fix

- **Fail-closed architecture (core safety fix):** a blob `PUT` carrying a `?comp=` selector is now always treated as a sub-operation. Any unrecognized or unimplemented `comp` returns a proper Azure error (400 `UnsupportedQueryParameter`, or 501 `NotImplemented` for page) instead of falling through to a content write. This kills the entire corruption class (same class as the pre-#619 lease bug).
- **SetBlobTags / GetBlobTags implemented** (`PUT`/`GET ?comp=tags`): parse the `<Tags>` XML body, store per-blob via the existing `PutObjectTagging`/`GetObjectTagging` storage, return 204/200. Set Blob Tags leaves the blob content, ETag, and last-modified untouched (per the Azure spec); an empty `<TagSet>` clears tags. Cross-checked against the official Set/Get Blob Tags REST docs.
- **Page blobs fail closed:** page-blob create (`x-ms-blob-type: PageBlob`) and `UploadPages` (`?comp=page`) now return 501 `NotImplemented` rather than corrupting/misreporting support.

## Tests (real azblob SDK)

- SetTags leaves the blob body unchanged (corruption gone)
- SetTags/GetTags round-trip; empty tag set clears tags without touching the body
- an unknown `comp` PUT fails closed and does not overwrite the blob body
- page-blob create fails closed

## Deferrals

- **Page-blob range semantics (UploadPages) deferred** as a feature gap. It requires a new page-blob type plus create-with-content-length and 512-aligned range writes across the driver interface + provider — out of scope here. It now fails closed (501) with no corruption.

### DIFF
--- a/server/azure/blobstorage/blob_tags.go
+++ b/server/azure/blobstorage/blob_tags.go
@@ -1,0 +1,82 @@
+package blobstorage
+
+import (
+	"bytes"
+	"encoding/xml"
+	"net/http"
+	"sort"
+)
+
+// setBlobTags handles PUT /{container}/{blob}?comp=tags (Set Blob Tags). It
+// parses the <Tags> XML body and stores the tags per blob WITHOUT touching the
+// blob's content, ETag, or last-modified time (per the Azure spec). An empty
+// <TagSet> clears all tags. Returns 204 No Content on success.
+func (h *Handler) setBlobTags(w http.ResponseWriter, r *http.Request, container, blob string) {
+	if h.checkLease(w, r, container, blob) {
+		return
+	}
+
+	body, ok := readLimitedBody(w, r)
+	if !ok {
+		return
+	}
+
+	tags, err := parseTagsXML(body)
+	if err != nil {
+		writeError(w, http.StatusBadRequest, "InvalidXmlDocument", "could not parse Tags body")
+		return
+	}
+
+	if err := h.bucket.PutObjectTagging(r.Context(), container, blob, tags); err != nil {
+		writeErr(w, err)
+		return
+	}
+
+	w.WriteHeader(http.StatusNoContent)
+}
+
+// getBlobTags handles GET /{container}/{blob}?comp=tags (Get Blob Tags),
+// returning the blob's stored tags as a <Tags> XML document with 200 OK.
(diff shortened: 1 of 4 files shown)


NitinKumar004:
Review — verified against the pushed commit (9e50582).
- build / vet / gofmt clean; PR introduces zero NEW lint (the 20 are pre-existing in unchanged blob_extensions.go/handler.go regions).
- Corruption gone: putBlobOp routes any non-empty ?comp= to putBlobComp and returns, so a sub-op can never content-write; SetTags calls PutObjectTagging only, body/ETag/last-modified untouched (TestSDKSetTagsPreservesContent).
- Existing comp handlers intact (the critical check): putBlobComp preserves the exact prior dispatch — compTags→setBlobTags, else dispatchBlobComp (block/blocklist/metadata/properties/tier/snapshot/appendblock/lease unchanged), else compPage→501, else 400. All existing tests green (block/blocklist, metadata, tier, snapshot, append, #619 lease/SAS). Strictly safer: a known comp now 400s instead of falling through to a corrupting putBlob when extensions absent.
- Tags round-trip (204 set/200 get; empty clears; malformed XML→400 InvalidXmlDocument, no panic); page-blob + comp=page fail closed 501 before any write (no partial state); unknown comp→400, body preserved; -race clean; no collateral. Arch: wire-layer routing, reuses existing tagging storage.
Non-blocking nits: comp=page 501 branch not directly tested (reasoned correct); package carries pre-existing lint debt (none introduced).
Verdict: approved.

PR #635, opened 2026-08-25
fix(gcp-compute): instances disks/tags/metadata/actions/zone-scope/dup-409/self-links/pagination

NitinKumar004:
## What & why

The GCP Compute Engine `instances` wire surface dropped almost all instance
detail on read and returned 501 for every update verb, breaking Terraform
round-trips and any VM-update flow. This fixes the full `compute.instances`
bug batch from AUDIT_GCP.md `## compute`.

GCP-specific instance state the portable compute driver model does not carry
(attached `disks[]`, network `tags`, `metadata` incl. `startup-script`, the
network self-link, and the launch zone) is round-tripped through the driver by
encoding it into internal `cloudemu:gcp*` tag entries, then rebuilt into the
real REST read shape. No change to the shared `services/compute/driver`
interface — GCP-specific mutation uses an optional-interface method
(`MutateInstanceGCP`) on the GCE mock, mirroring the existing `RemoveInstance`
pattern — so AWS/Azure are untouched and no docs regen is needed.

`gcprest.ParsePath` now recognizes the `/aggregated/{type}` scope; the GCP vpc
and loadbalancer handlers are guarded to stay unmatched on that scope so their
behavior is unchanged (aggregatedList remains unimplemented there).

## Findings fixed (13/13)

- HIGH get — echo `disks[]` with boot disk (`deviceName/source/boot/autoDelete/diskSizeGb/type`)
- HIGH insert/get — round-trip network `tags.items` + `tags.fingerprint`
- HIGH insert/get — round-trip `metadata.items` (incl. startup-script) + `metadata.fingerprint`
- HIGH — implement `setMetadata`/`setLabels`/`setTags`/`setMachineType` (mutate + DONE Operation) instead of 501
- HIGH — implement `attachDisk`/`detachDisk` (reflected in `disks[]`) instead of 501
- HIGH list — zone-scoped to the requested zone
- HIGH insert — duplicate name in same zone → 409 `alreadyExists`
- HIGH get networkInterfaces — fully-qualified `network` + `subnetwork` self-links
- HIGH get — `creationTimestamp` (RFC3339) from LaunchTime
- MED list — `filter` + `maxResults`/`pageToken` pagination, `nextPageToken` set
- MED get — `labelFingerprint` + `fingerprint` populated
- MED aggregatedList — instances grouped by `zones/{zone}` scope instead of 501
- LOW get — realistic `cpuPlatform`/`scheduling`/`serviceAccounts`/`shieldedInstanceConfig`/`deletionProtection`/`lastStartTimestamp` defaults

## Deferrals

None — all 13 findings fixed.

## Tests

`server/gcp/compute/instances_sdk_test.go` drives the real
`cloud.google.com/go/compute/apiv1` InstancesClient against the in-process GCP
server, one test per finding (disks, tags/metadata, update verbs, attach/detach,
zone-scoped list, duplicate 409, network self-links, timestamp/fingerprints,
defaults, aggregatedList, filter + pagination). Existing GCP compute tests stay
green.

### DIFF
--- a/providers/gcp/compute/gce.go
+++ b/providers/gcp/compute/gce.go
@@ -443,6 +443,37 @@ func (m *Mock) RemoveInstance(ctx context.Context, instanceID string) error {
 	return nil
 }
 
+// MutateInstanceGCP applies a GCP-specific instance mutation used by the GCE
+// wire handler for setLabels/setMetadata/setTags/setMachineType/attachDisk/
+// detachDisk. Unlike ModifyInstance (which requires a stopped instance), these
+// GCP verbs apply to running VMs. set entries are merged into the tag map,
+// remove keys are deleted, and machineType replaces the instance type when
+// non-empty. GCP-specific; reached via a type assertion from the GCE handler.
+func (m *Mock) MutateInstanceGCP(instanceID string, set map[string]string, remove []string, machineType string) error {
+	inst, ok := m.instances.Get(instanceID)
+	if !ok {
+		return cerrors.Newf(cerrors.NotFound, "instance %q not found", instanceID)
+	}
+
+	if inst.Tags == nil {
+		inst.Tags = make(map[string]string)
+	}
+
+	for k, v := range set {
+		inst.Tags[k] = v
+	}
+
+	for _, k := range remove {
+		delete(inst.Tags, k)
+	}
+
+	if machineType != "" {
+		inst.InstanceType = machineType
+	}
+
+	return nil
+}
+
 func (m *Mock) DescribeInstances(
 	_ context.Context, instanceIDs []string, filters []driver.DescribeFilter, _ ...driver.DescribeInstancesOptions,
 ) ([]driver.Instance, error) {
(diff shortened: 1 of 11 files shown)


NitinKumar004:
Review — verified against the pushed commit (f1bcbece).
- build / vet / gofmt clean; PR's own files lint-clean (10 flags are pre-existing in unchanged vpc/lb).
- Shared gcprest blast radius CONTAINED (the main risk): only compute/vpc/loadbalancer call ParsePath; every other GCP service matches a different URL prefix so an aggregated path can't reach them; the ScopeAggregated addition only affects paths whose scope segment literally equals 'aggregated' (previously ok=false→unmatched), so no existing path changes meaning; vpc + lb hard-guard rp.Scope==ScopeAggregated→return false. Full GCP server suite (all 26 pkgs) green.
- 13 fixes real via real cloud.google.com/go SDK: disks[] echo, tags/metadata round-trip + fingerprints, 4 setX verbs mutate + DONE Operation, attach/detach in disks[], list zone-scoped, self-links fully-qualified, creationTimestamp RFC3339, filter+pagination, aggregatedList grouped by zones/{zone}.
- Dup-409 zone-scoped (findInZone filters name AND zone; same-name-different-zone allowed); machineType persists via *instanceData pointer mutation; optional-interface MutateInstanceGCP is GCP-only (AWS/Azure untouched, coverage no-diff); -race clean.
Non-blocking nits: different-zone-allowed dup case not explicitly tested (correct by trace); aggregated/{non-instances} now 501 vs former 404 (more correct).
Verdict: approved.

PR #636, opened 2026-08-25
fix(gcp-iam): testIamPermissions/undelete/uniqueId/etag/keys/pagination

NitinKumar004:
## What

Wire-layer fixes for the GCP IAM (`iam.googleapis.com` v1) audit batch. All eight findings are handled in `server/gcp/iam/` — no driver-interface or provider changes, so `docs/coverage` is unaffected.

## Why

Real `google.golang.org/api/iam/v1` clients hit missing methods (404/405) and behavioral gaps (empty/synthetic identity fields, unenforced etag concurrency, no pagination) that diverge from real GCP.

## Findings fixed

1. **[HIGH]** `ServiceAccounts.testIamPermissions` — `routeSAVerb` had no case → 404. Now returns the held subset of requested permissions (all when no restricting policy; narrowed to a bound custom role's `includedPermissions` when one is set).
2. **[MED]** `ServiceAccounts.undelete` — restores a recently-deleted SA from a tombstone captured at delete time (accepts email or 21-digit uniqueId in the resource name).
3. **[MED]** `Roles.undelete` — `routeRoles` was verb-blind (405); now dispatches the `undelete` verb and restores the soft-deleted custom role.
4. **[MED]** `ServiceAccounts.get/create` — return a stable 21-digit numeric `uniqueId`, matching `oauth2ClientId`, and a populated `etag` (were `uid-…` / empty).
5. **[MED]** `ServiceAccounts.list` — reflects the `disabled` bit (previously only `get` overlaid it).
6. **[MED]** `ServiceAccountKeys.create` — `privateKeyData` is now a base64 `service_account` credentials file (was 15 raw secret bytes); `validAfterTime`/`validBeforeTime` populated.
7. **[LOW]** `ServiceAccounts.setIamPolicy` — enforces optimistic concurrency: request `policy.etag` is compared to the stored etag, stale writes get `409 ABORTED`; each accepted write bumps a per-SA version so successive states get distinct etags (the old `base64(email+bindingCount)` scheme collided).
8. **[LOW]** `Roles.get/list` & `ServiceAccounts.list` — roles carry a populated `etag`; both lists honor `pageSize`/`pageToken` and return `nextPageToken`.

## Tests

Added `server/gcp/iam/sdk_iam_compat_test.go` — nine reproduction tests driving a real `iamv1` client (via `option.WithEndpoint`) against the in-process GCP server, one per finding (plus a custom-role subset check for testIamPermissions and a fresh-then-stale etag flow). Existing GCP IAM tests stay green.

## Deferrals

None.

### DIFF
--- a/server/gcp/iam/handler.go
+++ b/server/gcp/iam/handler.go
@@ -46,26 +46,51 @@ const (
 	serviceAccountsSeg = "serviceAccounts"
 	rolesSeg           = "roles"
 	keysSeg            = "keys"
+
+	undeleteVerb = "undelete"
 )
 
 // Handler serves iam.googleapis.com v1 REST requests against the IAM driver.
 //
-// Service-account resource policies and the enabled/disabled bit have no place
-// in the portable IAM driver, so they're tracked here keyed by SA email.
+// Service-account resource policies, the enabled/disabled bit, policy etag
+// versions and soft-deleted resources have no place in the portable IAM
+// driver, so they're tracked here keyed by SA email / role id. The driver
+// hard-deletes, so undelete is served from these local tombstones.
 type Handler struct {
 	iam iamdriver.IAM
 
-	mu       sync.RWMutex
-	saPolicy map[string]*iamPolicy // SA email -> resource policy
-	disabled map[string]bool       // SA email -> disabled
+	mu            sync.RWMutex
+	saPolicy      map[string]*iamPolicy   // SA email -> resource policy
+	policyVersion map[string]int          // SA email -> monotonic policy etag version
+	disabled      map[string]bool         // SA email -> disabled
+	deletedSA     map[string]*deletedSA   // SA email -> soft-deleted account
+	deletedRole   map[string]*deletedRole // role id -> soft-deleted role
+}
+
+// deletedSA is a tombstone captured at delete time so undelete can restore the
+// account with the same project, metadata and disabled bit.
+type deletedSA struct {
+	project     string
+	displayName string
+	description string
+	disabled    bool
+}
+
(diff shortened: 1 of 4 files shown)


NitinKumar004:
Review — verified against the pushed commit (fcb1cec9).
- build / vet / gofmt clean.
- 8 fixes SDK-verified (real iam/v1): testIamPermissions (full-set when no policy, narrowed to custom-role subset); SA + Role undelete; 21-digit uniqueId + oauth2ClientId==uniqueId + etag; disabled in List; key file base64→service_account JSON; stale-etag→409 ABORTED; pagination + role etag.
- Tombstone thread-safe: all deletedSA/deletedRole/saPolicy/policyVersion/disabled access under h.mu; takeTombstone/undeleteRole remove under lock; undelete of a live/never-deleted SA correctly 404s (no resurrect); recreate-same-id → CreateUser AlreadyExists 409, live SA not clobbered (no data loss). Etag: first write bypasses (create-with-empty allowed, matches GCP); each accepted write bumps policyVersion → distinct etag; whole check+bump under lock. Key file valid; -race clean; no collateral (all 27 server/gcp pkgs green); wire-layer only (docs no-diff).
Follow-ups (non-blocking, narrow): undeleteServiceAccount sets disabled=true BEFORE CreateUser, so on the AlreadyExists path (delete-disabled-SA→recreate-same-id→undelete) the live SA's disabled bit is wrongly flipped + the tombstone is consumed on the failed restore (verbs.go:167-171); heldPermissions resolves only custom roles (predefined-role bindings yield empty — documented emulator limit); unbounded tombstone growth (acceptable).
Verdict: approved.

PR #637, opened 2026-08-25
fix(gcp-gcs): preconditions, generation, checksums, versions, patch, compose, bucket IAM/location/lifecycle

NitinKumar004:
## What

Fixes the GCS JSON API behaviors an unmodified `cloud.google.com/go/storage` client hits, from the AUDIT_GCP.md `## gcs` batch. All GCS-specific state lives behind a new optional `driver.GCSExtensions` capability discovered by type assertion, so the shared `Bucket` interface and the AWS S3 / Azure Blob providers are untouched.

## Why

The GCS handler previously ignored write preconditions, hardcoded object generation/metageneration to `"1"`, returned no checksums, dropped versioned listings, 405'd object patch/compose, 404'd bucket IAM, dropped lifecycle rules on patch, and reported a fixed US/STANDARD location. These break create-if-absent, distributed locks, optimistic concurrency, versioning, download integrity, `google_storage_bucket_iam_*`, and cause perpetual Terraform drift.

## Findings fixed (all 10 in the `## gcs` section)

- objects.insert preconditions — honor `ifGenerationMatch`/`ifGenerationNotMatch`/`ifMetagenerationMatch`/`ifMetagenerationNotMatch`; failed condition → 412 `conditionNotMet` (create-if-absent + locks work).
- objects.insert generation — mint a unique, monotonically increasing generation per write; per-generation metageneration.
- objects.Attrs checksums — return `md5Hash` and `crc32c`.
- objects.list `versions=true` — retain prior generations on a versioning-enabled bucket and list them.
- objects.patch/update — mutate contentType/cacheControl/content-encoding/disposition/language/metadata, bump metageneration (was 405).
- objects.compose — concatenate source bytes into the destination (was 405).
- buckets IAM — `/b/{bucket}/iam` get/set/testPermissions (was 404).
- buckets.patch lifecycle — persist lifecycle rules and round-trip them on read.
- buckets.get location/storageClass — round-trip the create-time values instead of hardcoded US/STANDARD.
- buckets.get metageneration/etag/updated/locationType/iamConfiguration — returned.

## 3-layer placement

- `services/storage/driver`: additive `ObjectInfo` fields + new optional `GCSExtensions` interface, `GCSPrecondition`, `GCSPreconditionError`, `GCSObjectUpdate`, `GCSBucketMeta`.
- `providers/gcp/gcs`: implementation + snapshot round-trip of the new fields.
- `server/gcp/gcs`: wire routing (`/iam`, compose, PATCH/PUT), preconditions → 412, versioned listing, richer bucket/object JSON.

## Tests

Real-SDK reproduction tests (in-process GCP server via `option.WithEndpoint`) proving each finding: `gcs_object_semantics_test.go`, `gcs_bucket_semantics_test.go`. Existing GCS tests stay green.

## Docs

Regenerated `docs/coverage` (`go generate ./...`) — `GCSExtensions` now listed as an optional storage capability.

## Deferrals

None.

### DIFF
--- a/docs/coverage/aws/s3.md
+++ b/docs/coverage/aws/s3.md
@@ -101,6 +101,22 @@ BucketAttributes is an OPTIONAL capability, discovered by type assertion (like
 | --- | --- |
 | `BucketAttributes` |  |
 
+### GCSExtensions
+
+GCSExtensions is an OPTIONAL GCS-specific capability, discovered by type
+
+| Operation | Description |
+| --- | --- |
+| `BucketAttrsGCS` | BucketAttrsGCS returns the bucket's GCS-specific attributes. |
+| `BucketIAMPolicy` |  |
+| `ComposeObjectGCS` | ComposeObjectGCS concatenates the source objects' bytes (in order) into |
+| `ListObjectGenerations` | ListObjectGenerations returns every generation (current + archived) of the |
+| `PutObjectGCS` | PutObjectGCS writes an object honoring pre (a failed condition returns a |
+| `SetBucketAttrsGCS` | SetBucketAttrsGCS records the bucket's location and default storage class |
+| `SetBucketIAMPolicy` | SetBucketIAMPolicy / BucketIAMPolicy persist and return the bucket's IAM |
+| `TouchBucket` | TouchBucket bumps the bucket's metageneration and updated timestamp, |
+| `UpdateObjectGCS` | UpdateObjectGCS mutates an existing object's system properties and/or |
+
 ### ObjectCopier
 
 ObjectCopier is an OPTIONAL capability (discovered by type assertion, like
--- a/docs/coverage/azure/blobstorage.md
+++ b/docs/coverage/azure/blobstorage.md
@@ -101,6 +101,22 @@ BucketAttributes is an OPTIONAL capability, discovered by type assertion (like
 | --- | --- |
 | `BucketAttributes` |  |
 
+### GCSExtensions
+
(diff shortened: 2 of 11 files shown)


NitinKumar004:
Review — verified against the pushed commit (bccfe8c9).
- build / vet / gofmt clean (no copylocks); the one advisory lint (ServeHTTP complexity 13) is report-only, others pre-existing.
- Shared-driver additive-safety CLEARED (the main risk): services/storage/driver changes are 100% additive — new ObjectInfo fields (documented zero for non-GCS) + new GCS-only types + GCSExtensions optional interface (type-asserted in handler.New, compile-asserted in gcs.go; AWS/Azure don't implement it). providers/aws/s3 + providers/azure/blobstorage + services/storage tests green; coverage regen additive-only + deterministic (go generate → zero further diff).
- Preconditions→412 GCS-specific (errors.As *GCSPreconditionError only on the ext path; canonical FailedPrecondition→409 unchanged). Generation atomic monotonic (not hardcoded 1); archiveVersion early-returns when versioning disabled (no growth); versions=true lists archived+current. Snapshot round-trips all new fields (GCS Mock Snapshottable). -race clean (bucketMeta pointer-held, mutex guards versions/location/iamPolicy). All 10 fixes covered by real-SDK tests; no collateral.
Non-blocking nits: ServeHTTP complexity (advisory); compose ignores dest ifGenerationMatch (minor fidelity); tag in-place mutate aliases an archived copy (harmless, tags omitted from versioned listing); no x-goog-hash download header (per scope, SDK integrity works via Attrs).
Verdict: approved.

PR #638, opened 2026-08-25
fix(gcp-pubsub): fan-out, extended sub fields, modifyAckDeadline/seek/snapshots/IAM, ordering key, publishTime, pagination

NitinKumar004:
## What & why

Reworks the GCP Pub/Sub wire handler (`server/gcp/pubsub`) so it models Pub/Sub-native delivery instead of delegating message storage to the SQS-style `messagequeue` driver. Each topic now keeps an append-only message log and each subscription an independent ack cursor, which is what real Pub/Sub fan-out, snapshots and seek require. Topic existence and labels are still backed by the driver.

All 12 findings from the AUDIT_GCP `## pubsub` section are fixed; each has a reproduction test driven by the real `google.golang.org/api/pubsub/v1` SDK against the in-process GCP server (the REST client matching our REST wire protocol — the gRPC `cloud.google.com/go/pubsub` client cannot target an HTTP handler).

## Findings fixed

- **[BLOCKER] subscriptions.pull fan-out** — two subscriptions on one topic each receive an independent copy of every message (was: first consumer drained a shared queue).
- **[HIGH] subscriptions.create/get extended fields** — `pushConfig`/`retainAckedMessages`/`messageRetentionDuration`/`enableMessageOrdering`/`filter`/`expirationPolicy`/`deadLetterPolicy`/`retryPolicy` now round-trip.
- **[HIGH] subscriptions.modifyAckDeadline** — routed; `ackDeadlineSeconds=0` nacks and the message redelivers.
- **[HIGH] topics IAM** — `getIamPolicy`/`setIamPolicy`/`testIamPermissions` served (was 405/mis-routed); subscription IAM added too.
- **[MEDIUM] topics.subscriptions.list** — the 4-segment path lists the topic's subscription names (was mis-routed to getTopic → empty).
- **[MEDIUM] subscriptions.create duplicate** — duplicate name now 409 (was silent overwrite).
- **[MEDIUM] snapshots.create/list/get/delete** — implemented.
- **[MEDIUM] subscriptions.seek** — seek to timestamp or snapshot.
- **[MEDIUM] subscriptions.modifyPushConfig** — implemented.
- **[MEDIUM] topics/subscriptions.list pagination** — `pageSize`/`pageToken` + `nextPageToken` honored.
- **[LOW] pull orderingKey** — populated on the received message.
- **[LOW] pull publishTime** — stamped at publish, not at pull.

## Notes

- Changes are confined to `server/gcp/pubsub`. No shared driver interface changed, so `docs/coverage` needs no regeneration.
- The legacy topic==subscription pull tolerance is preserved (an implicit subscription materializes on pull against a same-named topic) so existing tests stay green.
- Deferrals: none.

### DIFF
--- a/server/gcp/pubsub/handler.go
+++ b/server/gcp/pubsub/handler.go
@@ -1,41 +1,34 @@
 // Package pubsub implements the GCP Pub/Sub v1 REST API as a server.Handler.
-// Real google.golang.org/api/pubsub/v1 clients configured with a custom
-// endpoint hit this handler the same way they hit pubsub.googleapis.com.
+// Real google.golang.org/api/pubsub/v1 and cloud.google.com/go/pubsub clients
+// configured with a custom endpoint hit this handler the same way they hit
+// pubsub.googleapis.com.
 //
-// MVP coverage:
+// Coverage:
 //
-//	PUT    /v1/projects/{p}/topics/{name}                  — Create
-//	GET    /v1/projects/{p}/topics/{name}                  — Get
-//	GET    /v1/projects/{p}/topics                         — List
-//	DELETE /v1/projects/{p}/topics/{name}                  — Delete
-//	POST   /v1/projects/{p}/topics/{name}:publish          — Publish
-//	PUT    /v1/projects/{p}/subscriptions/{name}           — Create
-//	GET    /v1/projects/{p}/subscriptions/{name}           — Get
-//	GET    /v1/projects/{p}/subscriptions                  — List
-//	DELETE /v1/projects/{p}/subscriptions/{name}           — Delete
-//	POST   /v1/projects/{p}/subscriptions/{name}:pull      — Pull
-//	POST   /v1/projects/{p}/subscriptions/{name}:acknowledge — Ack
+//	topics       create/get/list/delete/publish, getIamPolicy/setIamPolicy/testIamPermissions,
+//	             topics.subscriptions.list, topics.snapshots.list
+//	subscriptions create/get/list/delete, pull/acknowledge/modifyAckDeadline/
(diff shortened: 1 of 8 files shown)


NitinKumar004:
Review — verified against the pushed commit (6f74a749).
- Fan-out (BLOCKER) real: topic owns one append-only log; each subState has its own acked map + outstanding leases; a pull delivers indices not in THAT sub's acked/leased sets, ack marks only that sub — so sub A's pull+ack can't consume for sub B. TestSDKPubSubFanOut passes.
- Thread-safe under -race: every topics/subs/snapshots + per-sub cursor access under h.mu; IAM set assigns a fresh policy pointer (no torn read). Reviewer's own adversarial stress test (4 publishers + 6 pullers/ackers on 2 subs + snapshot/seek/IAM/list churn) ran clean under -race — no map-concurrent-write, no torn cursor.
- Ack cursor correct (modifyAckDeadline(0)→immediate redelivery; lease expiry swept; positive deadline extends); seek-to-snapshot/time + snapshots correct; IAM round-trips; backward-compat preserved (implicit topic==subscription pull); no collateral (8 files, all server/gcp/pubsub; messagequeue driver untouched; coverage no-diff).
Follow-ups (non-blocking, tracked): per-topic log + acked/deliveryAttempts maps grow unbounded, messageRetentionDuration accepted but not enforced (doc note); deadLetterPolicy round-trips + deliveryAttempt surfaced but messages not actually routed to the DLQ topic after maxDeliveryAttempts; enableMessageOrdering delivers in publish order but without per-orderingKey head-of-line ack-blocking.
Verdict: approved.

PR #639, opened 2026-08-25
fix(gcp-firestore): value type fidelity, stable timestamps, list paging, transactions, listCollectionIds

NitinKumar004:
## What & why

Fixes the entire Firestore batch from `AUDIT_GCP.md`. These are all real-user-facing correctness gaps in the GCP Firestore wire handler (`server/gcp/firestore`), reproduced and now proven fixed with tests that drive the real `cloud.google.com/go/firestore` REST SDK against the in-process GCP server.

All fixes live at the wire-handler layer; the shared database driver interface is unchanged (no `go generate` needed), so AWS/Azure are untouched.

## Findings fixed

- **[HIGH][WRONG_WIRE] timestamp fidelity** — a `time.Time` field was stored as an RFC3339 string and re-emitted as `stringValue`, decoding back to `string`. Now stored as `time.Time` and emitted as `timestampValue`; round-trips as `time.Time`.
- **[HIGH][WRONG_WIRE] stable commit timestamps** — `createTime`/`updateTime` were regenerated with `time.Now()` on every read (two Gets differed, `createTime==updateTime` always). Now stamped once on write and read back from the stored document; `createTime` is preserved across updates while `updateTime` advances.
- **[HIGH][BEHAVIORAL] bytes/geoPoint/reference fidelity** — these fell through to `nil`. Now round-trip: `[]byte`↔`bytesValue`, `*geoPoint`↔`geoPointValue`, reference↔`referenceValue` (a distinct Go carrier type so it doesn't decay to `stringValue`).
- **[HIGH][BEHAVIORAL] listDocuments paging** — the driver `Scan` default limit silently truncated a 150-doc collection to 100. `listDocuments` now honors `pageSize`/`pageToken` (fetch-all + order + page in the handler) and returns `nextPageToken`, plus `orderBy` and `mask.fieldPaths`.
- **[HIGH][MISSING] transactions** — `beginTransaction`/`rollback` returned 501, making `RunTransaction` impossible. Implemented both (the in-memory store has no MVCC, so the token is an opaque handle and `commit` applies the writes directly); `RunTransaction` now works end-to-end.
- **[MEDIUM][WRONG_WIRE] integer-valued double** — `30.0` was re-encoded as `integerValue` (decoding to `int64`). A float now always encodes as `doubleValue`, preserving `float64`.
- **[MEDIUM][MISSING] listCollectionIds** — returned 501. Implemented; `client.Collections(ctx)` now enumerates the lazily-created collection ids (with paging).

## Tests added

- `firestore_types_test.go` — real-SDK round-trip of timestamp/bytes/reference/integer-double + stable commit timestamps.
- `firestore_convert_test.go` — white-box round-trip of every scalar type incl. geoPoint (the SDK's geoPoint type `latlng.LatLng` is not a direct module dependency, so geoPoint fidelity is asserted at the conversion boundary).
- `firestore_pagination_test.go` — 150-doc collection pages fully via both `DocumentRefs` (ListDocuments) and `Documents` (RunQuery); `Collections` enumeration.
- `firestore_transaction_test.go` — read-modify-write `RunTransaction`.
- Updated `firestore_lifecycle_test.go` to assert the corrected `doubleValue` behavior (previously asserted the now-fixed divergence).

## Deferrals

None — all findings in the `## firestore` section are addressed.

### DIFF
--- a/server/gcp/firestore/firestore_convert_test.go
+++ b/server/gcp/firestore/firestore_convert_test.go
@@ -1,0 +1,77 @@
+package firestore
+
+import (
+	"testing"
+	"time"
+)
+
+// TestValueRoundTripScalars proves each scalar Firestore value type survives a
+// Go<->wire round-trip with its type intact — the fidelity findings from the
+// GCP audit (timestamp, bytes, reference, and integer-valued double).
+func TestValueRoundTripScalars(t *testing.T) {
+	// An integer-valued double must stay a doubleValue, not collapse to
+	// integerValue (which would decode to int64 in the SDK).
+	dv := goValueToFirestore(float64(30))
+	if dv.DoubleValue == nil || dv.IntegerValue != nil {
+		t.Errorf("float64(30) encoded as %+v, want doubleValue", dv)
+	}
+
+	if got, ok := firestoreValueToGo(dv).(float64); !ok || got != 30 {
+		t.Errorf("double round-trip = %v (%T), want float64(30)", firestoreValueToGo(dv), firestoreValueToGo(dv))
+	}
+
+	// bytes -> bytesValue -> []byte.
+	bv := goValueToFirestore([]byte("hello"))
+	if bv.BytesValue == nil {
+		t.Fatalf("[]byte encoded as %+v, want bytesValue", bv)
+	}
+
+	if got, ok := firestoreValueToGo(bv).([]byte); !ok || string(got) != "hello" {
+		t.Errorf("bytes round-trip = %v (%T), want []byte(hello)", firestoreValueToGo(bv), firestoreValueToGo(bv))
+	}
+
+	// reference -> referenceValue -> reference (distinct from a plain string).
+	const path = "projects/p/databases/(default)/documents/c/d"
+
+	rv := goValueToFirestore(reference(path))
+	if rv.ReferenceValue == nil {
+		t.Fatalf("reference encoded as %+v, want referenceValue", rv)
+	}
+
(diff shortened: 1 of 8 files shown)


NitinKumar004:
Review — verified against the pushed commit (f0678ca3).
- build / vet / gofmt clean; no shared-driver/docs change (8 files, all server/gcp/firestore).
- 7 fixes real (real-SDK): time.Time→timestampValue; stable createTime (preserved across updates) + advancing updateTime read from the stored doc; bytes/geoPoint/reference round-trip (correct REST shapes); listDocuments pageSize/pageToken/nextPageToken + orderBy + mask; RunTransaction end-to-end (was 501); integer-valued float always doubleValue; listCollectionIds.
- Transactions safe-though-non-isolated: 16 concurrent RunTransactions on one doc → lost updates (documented, no MVCC) but NO crash/corruption/deadlock; -race clean; rollback discards. Pagination stable + terminating; types round-trip; no collateral.
Follow-ups (non-blocking): :commit doesn't validate the transaction token — a bogus token via raw HTTP returns 200 + silently applies (UNREACHABLE via the real SDK, which threads fresh non-expiring tokens; same endpoint serves tokenless WriteBatch.Commit which must apply) → track issued tokens as a fast-follow; timestamps use wall-clock not config.Clock (non-deterministic write-time); mergeMasked/applyFieldMask handle top-level field paths only (dotted nested a.b not resolved — likely pre-existing).
Verdict: approved.

PR #640, opened 2026-08-25
fix(gcp-networking): routes API, address IP allocation, region-scoped subnets, delete-in-use, pagination, timestamps, custom-mode IPv4Range, JSON 501

NitinKumar004:
## What

Fixes all 9 findings in the `## compute-networking` section of AUDIT_GCP.md, at the GCP compute-networking wire layer (`server/gcp/vpc`, shared `gcprest` codec). No driver-interface changes — GCP-specific state is carried in reserved tags / handler-local stores, so AWS/Azure are untouched.

## Findings fixed

1. **[HIGH][MISSING] Routes API entirely unrouted (501)** — added the global `routes` collection (insert/get/list/delete) as a handler-local store mirroring routers/addresses; responses enriched with `kind`/`id`/`selfLink`/`creationTimestamp`, caller's routing spec (network/destRange/next-hop/priority) preserved.
2. **[HIGH][BEHAVIORAL] Addresses allocated no IP** — reserved addresses now get an allocated IP, `status=RESERVED`, `kind`, numeric `id`, `selfLink`, `region` and `creationTimestamp`; caller fields (purpose/prefixLength/addressType) round-trip. Unblocks PSA / VPC_PEERING range reservation.
3. **[HIGH][WRONG_WIRE] subnetworks.get/delete ignored region** — `findSubnetByName` is now region-scoped (a name shared across two regions resolves to the correct region's CIDR; delete hits the right one). Subnetwork list is region-scoped too.
4. **[HIGH][BEHAVIORAL] networks.delete orphaned live subnets** — delete now scans for attached subnetworks and rejects with `400 resourceInUseByAnotherResource` until they are removed.
5. **[MEDIUM][BEHAVIORAL] list params ignored** — networks/subnetworks/firewalls/addresses/routes lists honor `maxResults`/`pageToken`/`filter` (name filter) and set `nextPageToken`.
6. **[MEDIUM][MISSING] creationTimestamp empty** — networks/subnetworks/firewalls (and routes/addresses) now stamp and return an RFC3339 `creationTimestamp`.
7. **[MEDIUM][WRONG_WIRE] IPv4Range always emitted** — only a legacy network (explicit `IPv4Range`) reports one; auto/custom-mode networks omit it, so a custom network no longer reads back as legacy.
8. **[LOW][MISSING] subnetwork fields** — `gatewayAddress` (first host of the CIDR), `purpose` (default PRIVATE), `stackType` (default IPV4_ONLY), `privateIpGoogleAccess`, and a stable `fingerprint` (needed to patch) are now returned.
9. **[LOW][WRONG_WIRE] bare-text 501 for unmatched /compute/v1** — a compute-space catch-all (`compute.NewFallback`, registered after the real compute/networks/LB handlers) returns a GCP JSON error envelope with the correct Content-Type.

## Why

These are behaviors a real user hits driving the GCP Compute networking REST API with the official `cloud.google.com/go/compute` clients (Terraform `google_compute_route`/`_subnetwork`/`_address`, PSA setup, `gcloud`). Each was verified against `cloud.google.com/compute/docs/reference/rest/v1`.

## Tests

Reproduction tests use the real `cloud.google.com/go/compute/apiv1` networking clients (`Routes`, `GlobalAddresses`, `Subnetworks`, `Networks`) with `option.WithEndpoint` against the in-process GCP server, one per finding: `routes_test.go`, `addresses_test.go`, `subnetworks_test.go`, `networks_semantics_test.go` (delete-in-use, pagination, creationTimestamp, custom-mode IPv4Range, JSON 501). Existing GCP vpc tests stay green.

## Deferrals

None.

### DIFF
--- a/server/gcp/compute/fallback.go
+++ b/server/gcp/compute/fallback.go
@@ -1,0 +1,29 @@
+package compute
+
+import (
+	"net/http"
+	"strings"
+
+	"github.com/stackshy/cloudemu/v2/server/wire/gcprest"
+)
+
+// Fallback claims any /compute/v1/ request that no real compute-space handler
+// (instances, networks, load balancing, …) matched, and answers with a proper
+// GCP JSON error envelope instead of the dispatcher's bare-text 501 with the
+// wrong Content-Type. Register it LAST among the /compute/v1 handlers so
+// first-match-wins keeps every implemented path on its real handler.
+type Fallback struct{}
+
+// NewFallback returns the compute-space catch-all handler.
+func NewFallback() *Fallback { return &Fallback{} }
+
+// Matches reports whether the request targets the compute API URL space.
+func (*Fallback) Matches(r *http.Request) bool {
+	return strings.HasPrefix(r.URL.Path, gcprest.BasePrefix)
+}
+
+// ServeHTTP writes a GCP-style JSON 501 for an unimplemented compute path.
+func (*Fallback) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	gcprest.WriteError(w, http.StatusNotImplemented, "notImplemented",
+		"not implemented: "+r.Method+" "+r.URL.Path)
+}
--- a/server/gcp/gcp.go
+++ b/server/gcp/gcp.go
@@ -171,6 +171,14 @@ func New(d Drivers) *server.Server {
 		srv.Register(lbsrv.New(d.LB))
 	}
 
+	// Compute-space catch-all. Registered AFTER the compute, networks and load-
+	// balancing handlers so first-match-wins keeps every implemented /compute/v1
+	// path on its real handler; this only claims the leftovers, answering with a
+	// GCP JSON error envelope instead of the dispatcher's bare-text 501.
+	if d.Compute != nil || d.Networking != nil || d.LB != nil {
+		srv.Register(compute.NewFallback())
+	}
+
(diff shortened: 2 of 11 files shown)


NitinKumar004:
Review — verified against the pushed commit (977ed72c).
- build / vet / gofmt clean; go generate no diff (no driver change).
- Fallback shadows nothing (the main risk): compute.NewFallback registered strictly after Compute/vpc/LB handlers (first-match-wins); the three claim disjoint resource-type sets, so no real path diverted; TestComputeUnmatchedPathReturnsJSONError exercises targetHttpProxies through production wiring → JSON 501; dispatch_ordering still green.
- Subnet region-scoped: findSubnetByName + listSubnetworks skip subnets whose region != requested; same-name subnets in us-central1/us-east1 stay independent, wrong-region Get returns the right CIDR (not first-found).
- networks.delete in-use scoped to THIS network (subnetOnNetwork matches VPCID==vpcID or the network tag) → 400 resourceInUseByAnotherResource, no partial apply; routes/addresses real + scope-keyed + paginated; IPv4Range emitted only for legacy networks; -race clean; no collateral (all server/gcp; AWS/Azure untouched).
Nit: 5 pre-existing revive findings in vpc/addresses.go/routers.go (byte-identical to development; local golangci-lint version skew vs CI) — zero introduced by this PR.
Verdict: approved.

PR #641, opened 2026-08-25
fix(gcp-cloudsql): request-project connectionName, absolute selfLinks, list filter+pagination, persisted operations

NitinKumar004:
## What & why

Fixes the GCP Cloud SQL bug batch from `AUDIT_GCP.md` (`## cloudsql`). All findings were wire-shape / behavioral mismatches against the real Cloud SQL Admin API; fixes are contained to the wire layer (`server/gcp/cloudsql`) — no driver interface or provider change, so no docs/coverage regeneration was needed.

## Findings fixed

- **[MEDIUM][WRONG_WIRE] connectionName** — was built from the server's configured project (`mock-project:...`), ignoring the request project. Now keyed on the request project as `{project}:{region}:{instance}`, matching real Cloud SQL.
- **[MEDIUM][WRONG_WIRE] selfLink** — instance (and database) selfLinks were relative, wrong-version paths (`/v1/projects/...`). Now emit the absolute `https://sqladmin.googleapis.com/sql/v1beta4/projects/...` URLs the real service returns.
- **[MEDIUM][BEHAVIORAL] instances.list** — `filter`, `maxResults` and `pageToken` were ignored and no `nextPageToken` was returned. Now sorts by name, applies a `field:value`/`field=value` filter (name/state/region/databaseVersion/backendType), paginates with offset tokens, and returns `nextPageToken`.
- **[LOW][BEHAVIORAL] Operations.get** — returned a synthetic `operationType=GET` with empty target fields for any name. Mutating endpoints now persist their Operation, so `Operations.Get` returns the real CREATE/UPDATE/DELETE record pointing at the affected resource; an unknown operation is a 404.
- **[LOW][MISSING] instance insert LRO** — the returned Operation now populates `selfLink`, `insertTime`, `startTime`, `endTime`, `user`, and `targetLink`.

## Tests

Added `instances_sdk_test.go` — reproduction tests using the real `google.golang.org/api/sqladmin/v1` client against the in-process GCP server, one per finding (connectionName, absolute selfLinks for instances+databases, list pagination+filter, persisted operation record + populated LRO fields + 404 on unknown op). Existing Cloud SQL tests stay green.

## Gates

`go build ./...`, `go vet`, `go test` (+ `-race`) on `server/gcp/cloudsql`, `golangci-lint` (0 issues), and `gofmt` all clean.

### DIFF
--- a/server/gcp/cloudsql/handler.go
+++ b/server/gcp/cloudsql/handler.go
@@ -17,7 +17,7 @@
 //	GET    /v1/projects/{p}/instances/{i}/backupRuns             — List backup runs
 //	GET    /v1/projects/{p}/instances/{i}/backupRuns/{id}        — Get backup run
 //	DELETE /v1/projects/{p}/instances/{i}/backupRuns/{id}        — Delete backup run
-//	GET    /v1/projects/{p}/operations/{op}                      — Poll operation (always DONE)
+//	GET    /v1/projects/{p}/operations/{op}                      — Poll operation (recorded record, 404 if unknown)
 //
 // All mutating endpoints return Operation envelopes with status=DONE so SDK
 // pollers terminate on the first response. Start/Stop are emulated via
@@ -32,6 +32,7 @@ package cloudsql
 import (
 	"net/http"
 	"strings"
+	"sync"
 
 	rdsdriver "github.com/stackshy/cloudemu/v2/services/relationaldb/driver"
 )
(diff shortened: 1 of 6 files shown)


NitinKumar004:
Review — verified against the pushed commit (a8fa3290).
- build / vet / gofmt / golangci-lint clean; go generate no diff (no interface change); security flag confirmed false positive (pure cloudsql REST wire).
- 4 fixes spot-checked via real sqladmin/v1 SDK: connectionName request-scoped (project:region:instance from the URL project, no opts.ProjectID leak); absolute v1beta4 selfLinks on instance/database/operation/targetLink; list filter (field:value substring vs field=value exact over name/state/region/version/backendType, unknown field→empty) + stable pagination + terminating nextPageToken + 400 on bad token; Operations.Get returns the persisted record (all 22 mutating call sites route through completeOp/buildOp), unknown op→404.
- Ops persistence thread-safe (the main risk): RWMutex guards every ops access, operation is value-copy so inline LRO and later Get are byte-identical for uniquely-named ops; -race clean incl. a 40-goroutine concurrent Insert+Get+List stress. No collateral (providers green, Handler.New unchanged).
Follow-ups (non-blocking): sub-resource/lifecycle ops use static names (insert-db/clone/failover/…) so consecutive same-verb ops overwrite the ops-map key — later Get returns the most recent for that verb (pollers still terminate; instance ops are uniquely named); ops map has no eviction (unbounded, fine for emulator).
Verdict: approved.

PR #642, opened 2026-08-25
fix(gcp-secretmanager): version lifecycle, patch, IAM, pagination, and etag/state fields

NitinKumar004:
## What

Fixes the entire GCP **Secret Manager** batch from \`AUDIT_GCP.md\` (\`## secretmanager\`). Real \`google.golang.org/api/secretmanager/v1\` clients pointed at the in-process GCP server now drive the full secret + version lifecycle end-to-end.

GCP-specific operations are added through a type-asserted optional interface (\`driver.GCPSecrets\`) so AWS/Azure secret backends are untouched; only new struct fields (\`Etag\`, \`State\`, \`DestroyTime\`) land on the shared driver types (defaults are zero for other providers).

## Findings fixed

- **[BLOCKER] SecretVersions.destroy** — \`:destroy\` now routed; wipes the payload, sets \`state=DESTROYED\` + \`destroyTime\`. \`:access\` on a destroyed version fails FAILED_PRECONDITION.
- **[HIGH] SecretVersions.disable/enable** — \`:disable\`/\`:enable\` routed; toggle ENABLED/DISABLED. \`:access\` on a disabled version returns FAILED_PRECONDITION (HTTP 400).
- **[HIGH] AddSecretVersion version naming** — versions are numbered by monotonic integer (\`1, 2, 3…\`) and addressable by that id (\`.../versions/1\`), not an internal random counter.
- **[HIGH] SecretVersion.state** — real per-version state reported (ENABLED/DISABLED/DESTROYED) with \`destroyTime\`, replacing the hardcoded ENABLED.
- **[HIGH] Secrets.patch** — \`PATCH\` with \`updateMask=labels\` updates the secret's labels and round-trips.
- **[HIGH] Secrets IAM** — \`:getIamPolicy\` / \`:setIamPolicy\` / \`:testIamPermissions\` served and round-trip (policy stored per secret).
- **[MEDIUM] Secrets/Versions.list pagination** — \`pageSize\`/\`pageToken\` honored with \`nextPageToken\`; secrets paged over a stable name ordering, versions newest-first.
- **[MEDIUM] Secret/SecretVersion fields** — \`etag\` on both, plus \`replicationStatus\` and \`destroyTime\` on versions.

## Layering

- Driver (\`services/secrets/driver\`): new optional \`GCPSecrets\` interface + \`Etag\`/\`State\`/\`DestroyTime\` fields.
- Provider (\`providers/gcp/secretmanager\`): integer version numbering, lifecycle/patch/IAM implementations, snapshot round-trip of the new state.
- Wire (\`server/gcp/secretmanager\`): routing for the new verbs/methods, pagination, and the GCP REST JSON shapes.

## Tests

New \`server/gcp/secretmanager/lifecycle_sdk_test.go\` drives the real \`secretmanager/v1\` SDK against the in-process server, one reproduction per finding (integer naming, lifecycle+access gating, patch, IAM, pagination, fields). Existing secretmanager tests stay green.

## Docs

\`docs/coverage\` regenerated (\`go generate ./...\`) for the new \`GCPSecrets\` capability.

## Deferrals

None.

### DIFF
--- a/docs/coverage/aws/secretsmanager.md
+++ b/docs/coverage/aws/secretsmanager.md
@@ -19,6 +19,20 @@ AWS's `secrets` service · portable interface `driver.Secrets` · [AWS index](./
 
 Discovered by type assertion; only some providers implement these.
 
+### GCPSecrets
+
+GCPSecrets is the GCP Secret Manager-specific surface kept off the shared
+
+| Operation | Description |
+| --- | --- |
+| `DestroySecretVersion` | DestroySecretVersion moves a version to DESTROYED, wipes its payload, and |
+| `DisableSecretVersion` | DisableSecretVersion moves a version to DISABLED. It fails on a DESTROYED |
+| `EnableSecretVersion` | EnableSecretVersion moves a version to ENABLED. It is idempotent on an |
+| `GetSecretIAMPolicy` | GetSecretIAMPolicy returns the secret's stored IAM policy (an empty |
+| `PatchSecret` | PatchSecret applies a partial update (labels) to a secret's metadata. |
+| `SetSecretIAMPolicy` | SetSecretIAMPolicy stores the secret's IAM policy and returns it with a |
+| `TestSecretIAMPermissions` | TestSecretIAMPermissions returns the subset of permissions the caller |
+
 ### KeyVaultCertificates
 
 KeyVaultCertificates is the Azure Key Vault-specific certificate data-plane
--- a/docs/coverage/azure/keyvault.md
+++ b/docs/coverage/azure/keyvault.md
@@ -19,6 +19,20 @@ Azure's `secrets` service · portable interface `driver.Secrets` · [Azure index
 
 Discovered by type assertion; only some providers implement these.
 
+### GCPSecrets
+
+GCPSecrets is the GCP Secret Manager-specific surface kept off the shared
+
+| Operation | Description |
+| --- | --- |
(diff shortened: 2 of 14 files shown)


NitinKumar004:
Review — verified against the pushed commit (a728b77f).
- Shared-driver additive-safe: services/secrets/driver adds only zero-default fields (SecretInfo.Etag; SecretVersion.State/DestroyTime/Etag); GCPSecrets is a SEPARATE optional interface (not embedded in Secrets). AWS SecretsManager + Azure KeyVault have zero references to GCPSecrets, tests green; handler type-asserts + nil-checks + 501s a non-GCP driver. Coverage regen additive-only (GCPSecrets rows, same mechanism as KeyVaultKeys); go generate → zero further diff.
- Version state machine correct: monotonic integer ids addressable by id; enable/disable/destroy via shared mutateVersion; destroy wipes Value + stamps DestroyTime + blocks re-destroy (FAILED_PRECONDITION). Access on non-ENABLED → 400 failedPrecondition before payload. Snapshot round-trips verCounter + iam + per-version state (persist/identity_gcp_test passes). IAM get/set/testIamPermissions round-trip (deep-cloned, no aliasing); patch updateMask=labels; pagination stable + terminating; -race clean; no collateral.
Non-blocking nits (both benign): enable refreshes etag even when already ENABLED (real GCP also bumps on enable); accessing 'latest' when newest is DISABLED/DESTROYED returns 400 rather than falling back — matches real GCP (latest = highest version number regardless of state).
Verdict: approved.

PR #643, opened 2026-08-25
fix(gcp-disks): sourceImage/sourceDisk reflection, dup-name 409, resize, aggregatedList, users[], creationTimestamp

NitinKumar004:
## What & why

Fixes the GCP compute **disks** bug batch from the audit — persistent-disk behaviors real users hit against the emulated Compute REST API (disks/images/snapshots). All fixes live at the wire layer plus one GCP-local provider method; no shared driver interface changed.

## Findings fixed

- **[HIGH] snapshots.get sourceDisk** — now echoes the caller's `sourceDisk` URL verbatim (round-tripped via an internal tag) instead of the fabricated driver self-link; also returns `sourceDiskId`.
- **[HIGH] duplicate-name insert → 409** — `disks`, `images`, `snapshots` insert now probe for an existing name and return `alreadyExists` (409) instead of silently creating a second resource over the ID-keyed store.
- **[MEDIUM] images.get sourceDisk/family/diskSizeGb** — reflected on read (`diskSizeGb` resolved from the source disk when the request omits it); also `sourceDiskId`/`sourceSnapshot`.
- **[MEDIUM] disks.get sourceImage/sourceImageId** — reflected on read.
- **[MEDIUM] disks.aggregatedList** — implemented (`GET /aggregated/disks`), grouping disks by `zones/{zone}` scope instead of 501.
- **[MEDIUM] creationTimestamp** — populated on disk/image/snapshot reads from the driver's stored creation time.
- **[LOW] disks.resize** — implemented (`POST /disks/{name}/resize`); grows the disk and returns a DONE Operation (shrink rejected, mirroring GCP). Backed by a GCP-local `ResizeVolumeGCP` provider method reached via an optional interface.

Also, consistent with the instance-side `disks[]` from the attachDisk work: a disk attached to an instance now reports that instance in its `users[]`.

## Tests

Reproduction tests using the real `cloud.google.com/go/compute` clients against the in-process GCP server: disk sourceImage reflection, duplicate-name conflict, creationTimestamp, resize, aggregatedList, users[] attachment; snapshot sourceDisk echo; image sourceDisk/family/diskSizeGb. Existing GCP disk/snapshot/image round-trip tests stay green.

## Deferrals

- The `backendServices`/`forwardingRules` portions of the duplicate-name and `creationTimestamp` findings belong to the compute load-balancer resources and are left to the compute-LB batch to avoid overlap.

### DIFF
--- a/providers/gcp/compute/gce.go
+++ b/providers/gcp/compute/gce.go
@@ -629,6 +629,22 @@ func (m *Mock) DescribeVolumes(_ context.Context, ids []string) ([]driver.Volume
 	return describeResources(m.volumes, ids), nil
 }
 
+// ResizeVolumeGCP grows the disk to sizeGb (a no-op when already that large).
+// GCP forbids shrinking, so the wire handler rejects smaller sizes before this
+// is reached. GCP-specific; reached via a type assertion from the GCE handler.
+func (m *Mock) ResizeVolumeGCP(volumeID string, sizeGb int) error {
+	vol, ok := m.volumes.Get(volumeID)
+	if !ok {
+		return cerrors.Newf(cerrors.NotFound, "disk %q not found", volumeID)
+	}
+
+	if sizeGb > vol.Size {
+		vol.Size = sizeGb
+	}
+
+	return nil
+}
+
 func (m *Mock) AttachVolume(_ context.Context, volumeID, instanceID, device string) error {
 	vol, ok := m.volumes.Get(volumeID)
 	if !ok {
--- a/server/gcp/compute/disks.go
+++ b/server/gcp/compute/disks.go
@@ -15,6 +15,10 @@ import (
 // the underlying compute driver, since the driver indexes by its own ID.
 const gcpDiskNameTag = "cloudemu:gcpDiskName"
 
+// gcpDiskSourceImageTag round-trips the sourceImage URL the caller created the
+// disk from, so a read echoes it (real GCP retains sourceImage/sourceImageId).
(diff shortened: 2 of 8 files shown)


NitinKumar004:
Review — verified against the pushed commit (cd941fd6).
- build / vet / gofmt clean; go generate no diff (no shared-driver change; ResizeVolumeGCP Mock-local via server-local volumeResizer type-assert).
- 7 fixes real (8 author SDK tests green): snapshot sourceDisk echo, dup-409, image/disk source fields, aggregatedList grouped by zone, creationTimestamp, resize.
- disk↔instance consistent both ways: after AttachDisk, disks.Get shows the instance in users[]; after DetachDisk cleared. diskUsersByName derives users[] live from the instance-side disks[] tags, so the sides can't diverge.
- resize grow-only (guarded at wire + Mock; shrink → 400 disk size cannot be reduced); aggregatedList serveAggregated guarded switch (instances+disks only) — full server/gcp suite green, no dispatch shadowing; -race clean; userLabels prefix-stripping collapse is strictly more thorough (no regression).
Follow-up (non-blocking): disk dup-name check is GLOBAL not zonal (cross-zone same-name disk wrongly 409'd) — consistent with the handler's pre-existing non-zone-aware modeling (get/list/delete also scan all zones); the same-zone dup bug IS fixed. Recommend a follow-up zone-scoping disk get/list/delete/dedup together (+ source-disk resolution by name inherits the same ambiguity).
Verdict: approved.

PR #644, opened 2026-08-25
fix(gcp-monitoring): serve timeSeries, metricDescriptors, notificationChannels; opaque alert-policy ids

NitinKumar004:
## What & why

The GCP Cloud Monitoring wire handler only served `/v3/projects/{p}/alertPolicies`; every other Cloud Monitoring collection returned 501. This made GCE auto-metrics unreadable and blocked metric-type discovery, notification channels, and custom-metric ingestion. This PR broadens the handler to the collections a real user hits and fixes two alert-policy correctness bugs.

Fixed at the correct 3-layer location: new GCP-specific raw-series accessors on the provider mock (`providers/gcp/monitoring`) exposed via an **optional interface** the wire handler type-asserts, so the shared `driver.Monitoring` interface is unchanged (no AWS/Azure churn, no coverage regen).

## Findings fixed (AUDIT_GCP.md `## monitoring`)

- **[BLOCKER] timeSeries.list** — was 501; now reads GCE auto-metrics and custom points, grouped per label set into series. This is the only wire read path for every metric.
- **[BLOCKER] timeSeries.create** — was 501; now ingests custom-metric points via `PutMetricData` and returns 200.
- **[HIGH] metricDescriptors.list/get/create** — was 501; descriptors synthesized from live series plus custom-created ones.
- **[HIGH] notificationChannels.create/list/get/delete** — was 501; alert policies can now reference channels that actually exist (backed by the existing driver channel methods).
- **[HIGH] alertPolicies.create (wrong wire)** — name was derived from `displayName`, collapsing duplicates (data loss); now assigns an opaque numeric id.
- **[MEDIUM] alertPolicies.create/get (records)** — `creationRecord`/`mutationRecord` now populated and each condition gets its canonical resource name.
- **[MEDIUM] alertPolicies.patch** — already routed (POST/GET/PATCH/DELETE) and now records a mutation timestamp on patch; verified via the real client.

## Tests

Reproduction tests in `server/gcp/monitoring/client_test.go` use the real `google.golang.org/api/monitoring/v3` REST client with `option.WithEndpoint` against the in-process GCP server, one per finding. Existing alert-policy wire tests updated to address policies by their opaque id (the old displayName-addressing encoded the bug) and stay green.

## Deferrals

None.

### DIFF
--- a/providers/gcp/monitoring/monitoring.go
+++ b/providers/gcp/monitoring/monitoring.go
@@ -70,12 +70,12 @@ func (m *Mock) PutMetricData(_ context.Context, data []driver.MetricDatum) error
 	}
 
 	m.mu.Lock()
-	for _, d := range data {
+	for i := range data {
 		key := metricKey{
-			Namespace:  d.Namespace,
-			MetricName: d.MetricName,
+			Namespace:  data[i].Namespace,
+			MetricName: data[i].MetricName,
 		}
-		m.metrics[key] = append(m.metrics[key], d)
+		m.metrics[key] = append(m.metrics[key], data[i])
 	}
 	m.mu.Unlock()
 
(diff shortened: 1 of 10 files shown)


NitinKumar004:
Review — verified against the pushed commit (2ae2ec8d). Substance is MERGE-ready; one trivial blocking gate.
Verified: 7 fixes real (6 real monitoring/v3 SDK tests pass) incl. 2 BLOCKERs (timeSeries list/create were 501; create→list round-trips a custom point; GCE auto-metrics readable); alertPolicy id now opaque numeric (projects/{p}/alertPolicies/{n}) so duplicate displayNames coexist (was data-loss) — updated wire tests address by opaque id legitimately; timeSeries filter (metric.type eq/starts_with + label clauses, AND-split) + interval + per-label-set grouping; optional GCPSeriesKeys/GCPRawSeries (shared driver.Monitoring unchanged, docs no-diff); -race clean; no collateral (GCE auto-metrics wiring intact).
BLOCKING (gate): server/gcp/monitoring/client_test.go fails gofmt/gci — local-module imports mis-ordered (server/gcp must sort before services/compute/driver). One-line swap. Fixing now.
Non-blocking nits: alertPolicy.Enabled uses omitempty (explicit-disabled serializes without the field, reads back false — minor fidelity); handler-held policies/descriptors maps not covered by snapshot/persist (follow-up).
Verdict: changes (format gate only) — fixing.

PR #645, opened 2026-08-25
fix(gcp-logging): logName-optional list, entry metadata, sinks, log-based metrics, pagination

NitinKumar004:
## What & why

Fixes the 5 GCP Cloud Logging findings from the audit. Each is a behavior a real `google.golang.org/api/logging/v2` client (or `gcloud logging`) hits against the wire server.

## Findings fixed

1. **[HIGH] entries.list logName now optional** — a request with only `resourceNames` (no `logName=` filter) previously returned 400. It now reads across every log in the project (matching `gcloud logging read` / read-all). A single-log `logName=` filter still surfaces not-found for a missing log, as before.
2. **[HIGH] entries.list metadata fields** — entries now return `insertId` (server-assigned when the writer omits one), `resource` (MonitoredResource, defaulting to the `global` resource for the project), and `receiveTimestamp` (ingestion time). A written resource round-trips.
3. **[HIGH] projects.sinks** — full create/get/list/update/delete for export sinks (`google_logging_project_sink`), including a server-assigned `writerIdentity`.
4. **[MEDIUM] entries.list pagination** — `pageSize` is now a page size, not a hard limit; responses carry `nextPageToken` and `pageToken` walks all pages.
5. **[MEDIUM] projects.metrics** — full CRUD for log-based metrics.

## Design

Sinks and log-based metrics have no cross-provider equivalent, so they are served through a new **optional `driver.GCPLogging` interface** (type-asserted by the wire handler) rather than widening the portable `driver.Logging` interface — AWS/Azure backends are untouched. `docs/coverage` regenerated via `go generate` (the new optional interface now shows under "Optional capabilities").

## Tests

Reproduction tests using the real `logging/v2` client against the in-process GCP server: `entries_list_sdk_test.go` (read-all, metadata, default resource, pagination), `sinks_sdk_test.go`, `metrics_sdk_test.go`. Existing logging tests stay green.

## Gates

build, vet, scoped tests, `-race` on the mutated provider, gofmt, and golangci-lint (zero new issues) all green.

### DIFF
--- a/docs/coverage/aws/cloudwatchlogs.md
+++ b/docs/coverage/aws/cloudwatchlogs.md
@@ -22,6 +22,27 @@ AWS's `logging` service · portable interface `driver.Logging` · [AWS index](./
 | `PutMetricFilter` |  |
 | `UpdateLogGroup` | UpdateLogGroup replaces the mutable fields (retention, tags) of an |
 
+## Optional capabilities
+
+Discovered by type assertion; only some providers implement these.
+
+### GCPLogging
+
+GCPLogging is an optional interface implemented by the GCP logging backend for
+
+| Operation | Description |
+| --- | --- |
+| `CreateLogMetric` |  |
+| `CreateSink` |  |
+| `DeleteLogMetric` |  |
+| `DeleteSink` |  |
+| `GetLogMetric` |  |
+| `GetSink` |  |
+| `ListLogMetrics` |  |
+| `ListSinks` |  |
+| `UpdateLogMetric` |  |
+| `UpdateSink` |  |
+
 ## Not in scope
 
 _Not documented yet. See the [emulator boundary](../../../README.md) for cloudemu-wide non-goals._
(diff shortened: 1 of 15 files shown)


NitinKumar004:
Review — verified against the pushed commit (8d74f758).
- build / vet clean; gofmt -l clean on all 11 changed files; new files gci-clean (the only gci finding is a pre-existing scope mis-order in cloudlogging.go, identical on development, not in this diff).
- 5 fixes real (real logging/v2 SDK, -race): logName-optional read-all (no 400, globally timestamp-sorted before paging via sort.SliceStable — verified cross-log out-of-order); insertId/resource/receiveTimestamp on read-back; sinks CRUD + writerIdentity; pageSize/nextPageToken; metrics CRUD. single-log-filter→404 preserved.
- Shared-driver safe: driver.Logging diff empty; GCPLogging separate optional interface (only GCP impl; AWS/Azure build green, 501 via type-assert fallback); coverage regen additive (go generate → zero further diff); no collateral (sinks/metrics Matches don't collide with bigtable/cloudrun which require /instances//locations/).
Follow-ups (non-blocking): opportunistic gci fix of the pre-existing scope import in cloudlogging.go; add a committed cross-log ordering assertion; offset pageToken snapshot-fragile (fine for emulator).
Verdict: approved.

PR #646, opened 2026-08-25
fix(gcp-lb): backendServices patch, healthChecks/targetPools/urlMaps CRUD, forwardingRule fields, dup-name 409

NitinKumar004:
## What & why

Fixes the GCP Compute Load Balancing (`compute-loadbalancer`) bug batch from `AUDIT_GCP.md`. The GCP wire server left three LB collections entirely unrouted (501) and dropped most fields on the two it did serve, so `google_compute_backend_service`, `google_compute_url_map`, `google_compute_target_pool`, health checks and L4/L7 forwarding rules could not be provisioned end-to-end.

All fixes land at the correct layer: opaque GCP-only resources and backend-service patch use new **type-asserted optional driver interfaces** (`GCPComputeResourceStore`, `GCPBackendServicePatcher`) implemented only by the GCP provider — the shared `LoadBalancer` interface is unchanged. New LB resource handlers register **before** the compute-space `NewFallback` (#640), so their real handlers win first-match-wins.

## Findings fixed (8/8)

- **[BLOCKER] backendServices.patch/update** — was 405; now merges non-empty fields onto the target group and returns a DONE Operation.
- **[BLOCKER] healthChecks CRUD** — was 501; full insert/get/list/delete (global).
- **[HIGH] targetPools CRUD** — was 501; full insert/get/list/delete (regional scope).
- **[HIGH] urlMaps CRUD** — was 501; full insert/get/list/delete (global).
- **[HIGH] backendServices.get missing fields** — now emits `loadBalancingScheme`/`timeoutSec`/`sessionAffinity`/`fingerprint`/`creationTimestamp`.
- **[HIGH] forwardingRules.get portRange** — retained even when the rule has no backend service.
- **[MEDIUM] forwardingRules.get IPAddress** — now a real IPv4, not a hostname.
- **[MEDIUM] forwardingRules.get loadBalancingScheme** — exact `EXTERNAL_MANAGED`/`INTERNAL_MANAGED`/… round-trip (no longer collapsed to EXTERNAL/INTERNAL).

## PR #643-deferred LB portions

- Duplicate-name insert on `backendServices` and `forwardingRules` now returns **409 alreadyExists** (was silent success).
- `creationTimestamp` populated on both.

## Implementation notes

- healthChecks/targetPools/urlMaps have no cross-provider model, so they are stored verbatim (decoded body) through `GCPComputeResourceStore` and re-emitted with server-injected identity (`id`/`kind`/`selfLink`/`creationTimestamp`) — every field the client sent round-trips. State lives in the provider mock (memstore), not the handler.
- backendServices/forwardingRules GCP-only fields continue to round-trip through the driver's tag map.

## Tests

Reproduction tests use the real `cloud.google.com/go/compute/apiv1` GAPIC clients against the in-process GCP server (existing harness): `backendservices_sdk_test.go`, `forwardingrules_sdk_test.go`, `healthchecks_sdk_test.go`, `targetpools_sdk_test.go`, `urlmaps_sdk_test.go`. Existing LB tests and the #640 dispatch-ordering/fallback tests stay green.

## Deferrals

None — all 8 findings plus the #643 LB portions are in this PR.

## Gates

`go build ./...`, `go vet`, scoped `go test -count=1` (server/gcp/loadbalancer, providers/gcp/loadbalancer, services/loadbalancer, server/gcp, server/gcp/compute), `go test -race` on the mutated provider, `golangci-lint` on changed packages (0 issues), `gofmt -l` clean, `go generate ./...` (docs/coverage regenerated).

### DIFF
--- a/docs/coverage/aws/elb.md
+++ b/docs/coverage/aws/elb.md
@@ -46,6 +46,25 @@ AzureLoadBalancers is an OPTIONAL, type-asserted capability. The Azure
 | `UpsertAzureLBBackendPool` | UpsertAzureLBBackendPool adds poolName to the load balancer's backend |
 | `UpsertAzureLBNatRule` | UpsertAzureLBNatRule creates or replaces a single inbound NAT rule by |
 
+### GCPBackendServicePatcher
+
+GCPBackendServicePatcher is an OPTIONAL, type-asserted capability implemented
+
+| Operation | Description |
+| --- | --- |
+| `PatchGCPBackendService` | PatchGCPBackendService looks up the target group by its GCP name and |
+
+### GCPComputeResourceStore
+
+GCPComputeResourceStore is an OPTIONAL, type-asserted capability implemented
+
+| Operation | Description |
+| --- | --- |
+| `DeleteGCPResource` | DeleteGCPResource removes the resource, returning NotFound when absent. |
+| `GetGCPResource` | GetGCPResource returns the stored resource, or NotFound. |
+| `ListGCPResources` | ListGCPResources returns every resource in a (collection, scope) bucket. |
+| `PutGCPResource` | PutGCPResource stores res, returning AlreadyExists when a resource with |
+
 ### LBAttributeUpdater
 
 LBAttributeUpdater is implemented by drivers that can apply a partial
--- a/docs/coverage/azure/lb.md
+++ b/docs/coverage/azure/lb.md
@@ -46,6 +46,25 @@ AzureLoadBalancers is an OPTIONAL, type-asserted capability. The Azure
 | `UpsertAzureLBBackendPool` | UpsertAzureLBBackendPool adds poolName to the load balancer's backend |
 | `UpsertAzureLBNatRule` | UpsertAzureLBNatRule creates or replaces a single inbound NAT rule by |
 
+### GCPBackendServicePatcher
+
+GCPBackendServicePatcher is an OPTIONAL, type-asserted capability implemented
+
+| Operation | Description |
+| --- | --- |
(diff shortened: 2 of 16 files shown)


NitinKumar004:
Review — verified against the pushed commit (c1d27a68).
- build / vet / gofmt / gci / golangci-lint clean (no #644-class import issue); full server/gcp suite green incl. #640 dispatch/fallback; -race clean.
- 10 fixes real (5 SDK test files): backendServices.patch merges + DONE op (omitted Protocol preserved, real merge); fields loadBalancingScheme/timeoutSec/sessionAffinity/fingerprint/creationTimestamp; forwardingRules portRange retained w/o backend + real IPv4 (net.ParseIP) + scheme round-trip + creationTimestamp; healthChecks/targetPools(regional)/urlMaps CRUD; dup-name 409.
- Fallback ordering correct: LB registered before compute.NewFallback (first-match); compute Matches claims only instances/ops/disks/snapshots/images/machineTypes → no shadowing; aggregated-scope LB → fallback JSON 501.
- Optional interfaces (GCPComputeResourceStore/GCPBackendServicePatcher) GCP-only, shared LoadBalancer unchanged, AWS/Azure build green; dup-name scoped by type (targetPools regional, others global); opaque verbatim round-trip; PatchGCPBackendService via memstore.Update (no lost updates); coverage regen additive (zero further diff).
Non-blocking nits: forwardingRule with a linked backendService overwrites response portRange with the single listener port (cosmetic, only the linked case); targetPools cross-region scoping correct-by-construction but no explicit SDK test.
Verdict: approved.

PR #648, opened 2026-08-25
fix(gcp-gke): full operation shape, absolute selfLinks, real endpoint, cluster fields, serverConfig

NitinKumar004:
## What & why

Fixes the 5 GKE (Kubernetes Engine / Container API) findings from the GCP audit. These are behaviors a real `google.golang.org/api/container/v1` user hits when driving clusters, node pools, operations, and version discovery.

## Findings fixed

- **[HIGH] operations.get shadowed** — the shared `lro.Handler` was registered ahead of GKE and answered every location-operation poll with only `{name,done,status}`, stripping GKE's richer shape. GKE now registers ahead of the LRO poller and its `Matches` claims a named operation poll **only when the op is in the GKE store**, so foreign location operations (artifactregistry, eventarc, memorystore, …) still fall through to `lro` — no shadowing. `operations.get` now returns `operationType`/`targetLink`/`selfLink`/`zone`/`startTime`/`endTime`.
- **[HIGH] relative selfLinks** — cluster, node-pool, and operation `selfLink` (plus operation `targetLink`) are now absolute `https://container.googleapis.com/v1/...` URLs.
- **[MEDIUM] sentinel endpoint** — a fresh cluster reported `https://GKE-DATAPLANE-NOT-IMPLEMENTED.cloudemu.local`. It now reports a deterministic, dotted-IPv4 control-plane IP (matching real GKE's bare-IP `endpoint` field); when the in-memory Kubernetes data plane is wired, the reachable `/k8s/<uid>` URL still wins.
- **[MEDIUM] empty cluster fields** — `id`, `zone`, `servicesIpv4Cidr`, and `currentNodeCount` are now populated (`clusterIpv4Cidr`/`nodeIpv4CidrSize` were already set).
- **[LOW] getServerConfig 501** — `GET .../locations/{l}/serverConfig` now returns default/valid cluster and node versions so `gcloud container get-server-config` and SDK version validation succeed.

## Tests

New `server/gcp/gke/sdk_fidelity_test.go` drives the real `container/v1` client against the in-process GCP server and proves each finding's repro passes, plus a foreign-operation test proving the LRO fall-through still works. Existing GKE server/provider tests updated for the non-sentinel endpoint and kept green.

## Notes

No shared driver interface changed — GKE uses a concrete provider mock, so no `docs/coverage` regeneration was needed.

### DIFF
--- a/providers/gcp/gke/gke.go
+++ b/providers/gcp/gke/gke.go
@@ -4,13 +4,15 @@
 // Operations they emit) plus a live Kubernetes data plane. When a shared
 // kubernetes.APIServer is wired in, GetCluster's Endpoint + masterAuth CA point
 // at a real in-memory apiserver so `gcloud container clusters get-credentials`
-// yields a working kubeconfig. Without one, GetCluster falls back to a sentinel
-// Endpoint (https://GKE-DATAPLANE-NOT-IMPLEMENTED.cloudemu.local) so kubeconfig
+// yields a working kubeconfig. Without one, GetCluster reports a deterministic
+// control-plane IP (matching real GKE's bare-IP `endpoint` field) so kubeconfig
 // rendering still works syntactically.
 package gke
 
 import (
 	"context"
+	"fmt"
+	"hash/fnv"
 	"sync"
 	"time"
 
@@ -22,10 +24,8 @@ import (
 	mondriver "github.com/stackshy/cloudemu/v2/services/monitoring/driver"
 )
 
-// Stub values used in Cluster responses until the Kubernetes data-plane
-// arrives in Wave 2.
+// Default versions reported by clusters/node pools and getServerConfig.
 const (
-	StubEndpoint    = "GKE-DATAPLANE-NOT-IMPLEMENTED.cloudemu.local"
 	StubMasterVer   = "1.30.0-gke.0"
 	stubNodeVersion = "1.30.0-gke.0"
 )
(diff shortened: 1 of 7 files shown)


NitinKumar004:
Review — verified against the pushed commit (684914b1).
- build / vet / gofmt / gci / golangci-lint clean; no driver/docs change (GKE is a concrete mock).
- LRO reorder SAFE (the main risk): GKE registers before lro.New(), but lro.Matches only claims a named GET op poll (5 segments, GET) — the sole overlap — and for that GKE.Matches gates on HasOperation(name) (exact key lookup in GKE's own store under RLock). server/gcp/fullserver_test.go::TestFullServerLROOperationPolling proves it: artifactregistry (op-r1) + eventarc (op-t1) ops still poll to done via lro while GKE's own op returns richer status:DONE. GKE/AlloyDB clusters collision defended by a panic guard (New panics if both non-nil) + DriversFrom leaves AlloyDB nil.
- 5 fixes real (6 container/v1 SDK tests): full op shape; absolute selfLinks (targetLink prepended to a relative path, no double-prefix); non-sentinel dotted-IPv4 endpoint (data-plane /k8s/<uid> still wins when wired — TestEndpoint_RealURLWhenWired unchanged); cluster id/zone/cidrs/currentNodeCount; getServerConfig valid versions. -race clean; no collateral.
Non-blocking nits: GKE.Matches returns true unconditionally for operations-list (name=='') and operations-cancel (action!='') — harmless today (lro handles neither, no foreign service exposes location op list/cancel via the shared server); a one-line comment would future-proof. defaultServerConfig hardcodes the version list inline.
Verdict: approved.

PR #649, opened 2026-08-25
fix(gcp-cloudrun): Services/Revisions surface + Jobs patch/IAM/executions.list/status fields

NitinKumar004:
## What

Closes the `## cloudrun` findings from the GCP audit. The Cloud Run Admin API v2 handler was Jobs-only; this adds the primary **Services** surface and fills the Jobs gaps a real `run/apiv2` (`google.golang.org/api/run/v2`) user or `google_cloud_run_v2_*` Terraform user hits.

### Services (BLOCKER — was 501, handler was Jobs-only)
- `services.create/get/list/patch/delete` — create reconciles to an Operation carrying the Service with a stable `*.run.app` `uri`, `latestReadyRevision`/`latestCreatedRevision`, default 100% LATEST `traffic` + resolved `trafficStatuses`, `terminalCondition` (Ready/CONDITION_SUCCEEDED) and `conditions`; patch materializes a new revision and bumps generation.
- `services/{svc}/revisions` list/get/delete — every create/patch materializes a `{service}-000NN-{suffix}` revision.
- Services IAM: `:getIamPolicy`/`:setIamPolicy`/`:testIamPermissions`.

### Jobs
- `jobs.patch` (HIGH) — was 405; now returns an LRO with the mutated Job.
- Jobs IAM (HIGH) — `:getIamPolicy` was mis-routed to getJob and `:setIamPolicy` was 405; now stored/round-tripped, plus `:testIamPermissions`.
- `jobs.executions.list` (HIGH) — the bare collection returned the Job; now returns `{executions,nextPageToken}`.
- Template fields (MEDIUM) — `maxRetries`/`serviceAccount`/`timeout`/`parallelism`/`executionEnvironment`/`vpcAccess` now round-trip on create/get.
- Job status fields (MEDIUM) — `terminalCondition`/`etag`/`observedGeneration`/`reconciling`/`latestCreatedExecution` now populated.
- `jobs.list` (LOW) — honors `pageSize`/`pageToken`.
- `jobs.run`/`executions.get` (LOW) — a completed execution reports `succeededCount == taskCount` (distinguishable from failure); covered by a regression test.

## Why

The Services surface is the primary Cloud Run workload type (HTTP ingress); every `run/apiv2 ServicesClient` and `google_cloud_run_v2_service` user was dead on arrival. The Jobs gaps caused 405s, IAM mis-routing, and perpetual Terraform drift.

## Where (3-layer)

[...]

PR #650, opened 2026-08-25
fix(gcp-cloudfunctions): gen2 v2 API, gen1 metadata/versionId, testIamPermissions, list pagination

NitinKumar004:
## What

Fixes the GCP Cloud Functions bug batch from `AUDIT_GCP.md` (`## cloudfunctions`). All at the wire layer (`server/gcp/cloudfunctions`); no shared-driver or provider churn, so no `docs/coverage` regeneration.

### Findings fixed

- **[BLOCKER] gen2 (v2 API)** — the `/v2/...` surface previously 501'd (only gen1 `/v1/` was registered), breaking `functions/apiv2` and `google_cloudfunctions2_function`. Implemented v2 `create/get/list/delete/patch` (each mutation returns an LRO that reconciles synchronously to `state: ACTIVE` with a Cloud Run-backed `serviceConfig.uri`), plus `generateUploadUrl` and the IAM verbs. gen2 functions round-trip `buildConfig`/`serviceConfig`/`eventTrigger` and synthesize the output-only fields real GCP fills in. (Note: v2 has no `functions.call` — gen2 is invoked via the Cloud Run URL — so `call` is intentionally not added to v2.)
- **[MEDIUM] gen1 missing fields + versionId** — `serviceAccountEmail`/`ingressSettings`/`dockerRegistry`/`buildId` were empty and `versionId` was hardcoded `"1"`. Now populated with real defaults (honoring create-body values), and `versionId` increments with a fresh `buildId` on every deploy (PATCH).
- **[MEDIUM] testIamPermissions** — was 405. Added `:testIamPermissions` for v1 and v2; CloudEmu does not enforce IAM (any credential is an owner) so it echoes back the requested permission set, the answer callers use to gate optional UI.
- **[LOW] list pagination** — `pageSize`/`pageToken` were ignored. Now honored via a stable name-sorted base64 offset cursor returning `nextPageToken`; applied to both v1 and v2 list.

## Why

These are behaviors a real user hits immediately: gen2 is the modern default and was entirely unusable; missing metadata and a frozen `versionId` break drift detection; `testIamPermissions` and pagination are standard client/Terraform flows.

## Notes

- gen2 LRO operation polls are namespaced with a `cf2-` id prefix so they stay disjoint from the Cloud Run handler, which also matches `/v2/projects/{p}/locations/{l}/operations/...` (Cloud Functions is registered first).
- GCP-specific state (gen1 metadata, gen2 functions) lives in the wire handler, mirroring the existing IAM-policy store — no portable-driver representation exists for these and gen2 code is not executed.

## Tests

New reproduction tests using the real `google.golang.org/api/cloudfunctions/v1` and `/v2` clients against the in-process GCP server:
- `gen2_sdk_test.go` — gen2 create/get/list/delete, patch, missing-get 404.
- `metadata_sdk_test.go` — gen1 metadata population + versionId bump, body-honoring, testIamPermissions, list pagination.

Existing GCP Cloud Functions tests remain green.

## Deferrals

None.

### DIFF
--- a/server/gcp/cloudfunctions/gen1meta.go
+++ b/server/gcp/cloudfunctions/gen1meta.go
@@ -1,0 +1,208 @@
+package cloudfunctions
+
+import (
+	"crypto/rand"
+	"encoding/base64"
+	"encoding/hex"
+	"net/url"
+	"strconv"
+)
+
+// gen1Meta is the GCP-specific gen1 (v1) output-only metadata a real client
+// reads back from Get/List but that has no portable Serverless-driver field:
+// the runtime service account, ingress policy, docker registry, the current
+// build id and the monotonically increasing deploy generation (versionId).
+type gen1Meta struct {
+	serviceAccountEmail string
+	ingressSettings     string
+	dockerRegistry      string
+	buildID             string
+	versionID           int64
+}
+
+// newGen1Meta builds the initial metadata for a freshly created v1 function,
+// honoring any values the create body set and falling back to the defaults real
+// Cloud Functions assigns.
+func newGen1Meta(body *cloudFunction, project string) *gen1Meta {
+	m := &gen1Meta{
+		serviceAccountEmail: body.ServiceAccountEmail,
+		ingressSettings:     body.IngressSettings,
+		dockerRegistry:      body.DockerRegistry,
+		buildID:             newBuildID(),
+		versionID:           1,
+	}
+
+	applyGen1Defaults(m, project)
+
+	return m
+}
+
+// applyGen1Defaults fills any unset gen1 metadata field with the value real
+// Cloud Functions defaults to.
+func applyGen1Defaults(m *gen1Meta, project string) {
+	if m.serviceAccountEmail == "" {
+		m.serviceAccountEmail = gen1ServiceAccount(project)
+	}
+
+	if m.ingressSettings == "" {
+		m.ingressSettings = defaultIngress
+	}
+
+	if m.dockerRegistry == "" {
+		m.dockerRegistry = defaultDockerRegistry
+	}
+
+	if m.buildID == "" {
+		m.buildID = newBuildID()
+	}
+
+	if m.versionID == 0 {
+		m.versionID = 1
+	}
+}
+
+// putGen1Meta stores freshly created metadata under the function's canonical name.
+func (h *Handler) putGen1Meta(name string, m *gen1Meta) {
+	h.mu.Lock()
+	h.gen1Meta[name] = m
+	h.mu.Unlock()
+}
+
+// bumpGen1Meta advances the deploy generation for an updated function: versionId
+// increments, a new build id is cut, and any gen1 metadata carried in the PATCH
+// body is applied. A function whose metadata is missing (created straight through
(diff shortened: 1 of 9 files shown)


NitinKumar004:
Review — verified against the pushed commit (18def7b0). The gen2 BLOCKER fix is real and well-built, and CF steals NO Cloud Run op — but a data race blocks merge.
Verified real: gen2 create→get→list→delete + LRO→ACTIVE with Cloud Run-backed serviceConfig.uri; CF-vs-CloudRun dispatch disjoint (matchesV2 declines any non-cf2- op / non-functions path; Cloud Run op ids never start with cf2-; lro is /v1/-only — confirmed empirically with both handlers on one server); gen1 metadata + versionId bump; testIamPermissions (v1+v2); pagination terminates; no collateral; coverage no-diff.
BLOCKING — data race in the new gen2 store: getV2 (gen2.go:292) and listV2 (gen2.go:302/311) grab the stored *gen2Function (or a shallow copy) under RLock, RELEASE the lock, then json.Marshal it — but nested ServiceConfig/BuildConfig/EventTrigger pointers + maps are shared with the stored object, while patchV2→mergeServiceConfig (gen2.go:465-493) mutates those same pointers IN PLACE under Lock. Concurrent Get/List + Patch on one function is a data race (verified with -race: write at mergeServiceConfig:475 vs read at getV2→writeJSON:292). finishV2LRO→resourceAsResponseV2 (after mu.Unlock) shares the same pointers and races too. The PR's own -race test misses it (never runs concurrent get+patch on the same function). FIX: marshal while holding the lock, OR deep-copy the function incl. nested structs/maps before releasing; add a concurrent get+patch -race test.
Non-blocking: pre-existing gci diff in sdk_roundtrip_test.go (unrelated, PR trails development — opportunistic fix welcome); unbounded operations cache.
Verdict: changes — fix the data race, then re-review.

NitinKumar004:
Re-review — verified against the pushed commit (0f0d7191), the data race resolved.
- Deep-copy complete: cloneGen2/cloneStringMap clone gen2Function (out := *fn) + Labels map + BuildConfig/ServiceConfig/EventTrigger pointers + BuildConfig.Source/StorageSource + both EnvironmentVariables maps; EventTrigger is all scalars (value copy suffices); there are NO slices anywhere, so nothing is left aliased. Every reader fixed: getV2 clones under RLock, listV2 appends *cloneGen2(fn) under RLock, patchV2 clones under the write Lock, createV2 clones under Lock before Unlock.
- -race clean incl. the new TestSDKGen2ConcurrentGetPatch (1 writer patching serviceConfig + 4 readers Get/List, 40 iters — would trip -race pre-fix) AND a harder reviewer stress (3 functions, 60 iters, concurrent Labels/build-env/service-env patch + 6 readers, -count=3) — clean.
- No field dropped (out := *fn copies all, overrides only pointer/map fields with clones — labels/env/serviceConfig shape intact, no #649-style omission); no regression (gen2/gen1/testIamPermissions/pagination + dispatch-disjoint tests pass); lint 0; gci clean (sdk_roundtrip_test.go reordered); coverage no-diff.
Verdict: approved.

PR #652, opened 2026-08-25
fix(gcp-artifactregistry): sub-collections, repo patch/IAM, numeric format enum, extra fields, list paging

NitinKumar004:
## What

Fixes all 6 GCP Artifact Registry findings from the audit. Real `google.golang.org/api/artifactregistry/v1` and GAPIC `apiv1` clients now drive repositories, their patches, repo IAM, and the packages/versions/tags/files sub-collections end-to-end against the shared container-registry driver.

## Why / findings fixed

- **[BLOCKER] packages/versions/tags/files.list returned the Repository body** — only `dockerImages` was special-cased, so clients silently decoded empty lists. These sub-collections are now routed and return correctly-shaped, populated responses. The driver models docker images by digest without a distinct image-name dimension, so each image is surfaced as one package (id = digest) with a single version and its docker tags; files are the repo blobs.
- **[HIGH][WRONG_WIRE] CreateRepository numeric Format enum** — the GAPIC apiv1 REST client serializes proto enums with `UseEnumNumbers`, sending `format` as a number, which failed to decode into the string field. `format` now unmarshals from either a JSON string or the numeric enum.
- **[HIGH] repositories.patch 404** — PATCH is now supported; labels/description/kmsKeyName update via the update mask (persisted through a GCP-only optional driver extension `UpdateRepository`).
- **[HIGH] repositories IAM colon-verbs 404** — `:getIamPolicy` / `:setIamPolicy` / `:testIamPermissions` on a repository now round-trip (policy stored in the wire handler, matching the cloudfunctions IAM pattern).
- **[LOW] missing fields** — `mode`, `sizeBytes`, `kmsKeyName` are now returned, and `updateTime` is tracked separately from `createTime` (advances on patch) via a new `Repository.UpdatedAt` driver field.
- **[LOW] list paging ignored** — `pageSize`/`pageToken` are honored on the repositories list and the sub-collection lists, with `nextPageToken` emitted.

## Notes

- The shared `ContainerRegistry` driver interface is unchanged (only a struct field was added and a GCP-mock-only optional method); `go run ./internal/coveragegen` produces no doc diff.
- Artifact Registry operations continue to poll through the shared `lro` handler (registered first); this handler's own operation route is only a standalone fallback and does not shadow it.

## Tests

Reproduction tests using the real clients against the in-process GCP server, one per finding: `subcollections_test.go` (packages/versions/tags/files) and `repository_ops_test.go` (patch, repo IAM, numeric format enum, extra fields, list paging). Existing AR + shared LRO tests remain green.

### DIFF
--- a/providers/gcp/artifactregistry/artifactregistry.go
+++ b/providers/gcp/artifactregistry/artifactregistry.go
@@ -99,11 +99,13 @@ func (m *Mock) CreateRepository(_ context.Context, cfg driver.RepositoryConfig) 
 	tags := copyTags(cfg.Tags)
 	uri := fmt.Sprintf("%s-docker.pkg.dev/%s/%s", m.opts.Region, m.opts.ProjectID, cfg.Name)
 	selfLink := idgen.GCPID(m.opts.ProjectID, "repositories", cfg.Name)
+	now := m.opts.Clock.Now().UTC().Format(time.RFC3339)
 
 	info := driver.Repository{
 		Name:       selfLink,
 		URI:        uri,
-		CreatedAt:  m.opts.Clock.Now().UTC().Format(time.RFC3339),
+		CreatedAt:  now,
+		UpdatedAt:  now,
 		Tags:       tags,
 		ImageCount: 0,
 	}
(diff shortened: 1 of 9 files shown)


NitinKumar004:
Review — verified against the pushed commit (d9d5475a).
- build / vet / gofmt / golangci-lint (0 issues) clean; the fmt --diff flags are on non-PR files (standalone formatter default gci sections vs the project's localmodule config); authoritative golangci-lint run is 0.
- 6 fixes real (author SDK tests execute + pass): BLOCKER sub-collections (packages/versions/tags/files return populated correctly-shaped collections, not the Repository body; Docker→AR digest→package mapping coherent); Format enum dual-accept (GAPIC numeric via UseEnumNumbers + string, repoFormat.UnmarshalJSON branches on leading quote); patch mask (applyPatch touches only masked fields, reserved cloudemu: keys survive); IAM round-trips (RWMutex-guarded, stable etag, deleted with repo); repo fields + updateTime.
- Shared field Repository.UpdatedAt additive; UpdateRepository is GCP-only via optional repositoryUpdater assertion (non-GCP → clean 404); AWS ECR + Azure ACR unaffected; coverage no-diff. lro not shadowed (lro@157 before AR@273; lro claims all GET .../operations/{op}, AR's op route is dead fallback); -race clean; no collateral (26 server/gcp pkgs green).
Follow-ups (non-blocking): handler.go doc says PATCH updates 'mode' but code handles kmsKeyName (mode correctly immutable); updateTime RFC3339 second-granularity (create+patch same second → equal, tracked separately though); maskless PATCH treated as '*' would wipe labels (low risk — GAPIC/TF always send a mask); sizeBytes omitempty vs real-AR always.
Verdict: approved.

PR #653, opened 2026-08-25
fix(gcp): bigtable/eventarc/vertexai/fcm/cloudasset fidelity

NitinKumar004:
Fixes real-user fidelity gaps in five small GCP services from the GCP audit. One commit per service so each can be reviewed independently. Bigtable is deferred (see below).

## fcm
- `messages:send` now rejects a message with **no** target (none of token/topic/condition) with **400 INVALID_ARGUMENT**, matching real FCM's "exactly one" rule. Previously such a message was accepted and published to a synthetic topic.

## vertexai
- `endpoints.predict` on an endpoint with **no deployed models** returns **400 FAILED_PRECONDITION** instead of echoing the input instances.
- `endpoints.deployModel` of a model resource that was **never uploaded** returns **404 NOT_FOUND** instead of silently deploying a dangling reference.
- `deployedModels` entries now round-trip `dedicatedResources` (machineSpec, min/max replica counts) and `createTime` on get/deployModel responses.

## eventarc
- `triggers.create` rejects a trigger with **no destination** (400 INVALID_ARGUMENT) rather than storing a dead route.
- `triggers.patch` implements **updateTrigger** with `updateMask` (destination / eventFilters / serviceAccount / labels); previously PATCH 404'd.
- Triggers now carry a server-assigned **uid**, an **etag** (bumps on patch), and an auto-provisioned **Pub/Sub transport** topic.
- `triggers.list` honours `pageSize`/`pageToken` and returns `nextPageToken`.

## cloudasset
- `searchAllResources` now honours the requested **scope**: asset names and the `project` field are rescoped to the queried `projects/{p}` instead of a fixed default project.
- `exportAssets` reflects the caller's **outputConfig** and reports the written destination in `outputResult` instead of the fixed `cloudemu://in-memory` placeholder.

## Deferrals
- **cloudasset.searchAllIamPolicies** — still returns `[]`. CloudEmu has no GCP resource-IAM policy store wired into the asset engine; returning fabricated bindings would be worse than an honest empty result. Needs a resource-IAM policy source wired through the discovery engine first.
- **bigtable data plane (ReadRows/MutateRow)** — deferred entirely. The Bigtable data API is `google.bigtable.v2`, which real clients (`cloud.google.com/go/bigtable`) speak over **gRPC only**; this wire server is HTTP/JSON with no gRPC transport. A REST data route would be unreachable by any real client. Implementing it needs a new gRPC/HTTP2 data-plane server + row storage, out of scope for this small-services pass. The Bigtable **Admin** surface (instances/clusters/tables/backups/appProfiles/IAM) remains fully implemented.

## Tests
Per-service reproduction tests using the real GCP clients (`google.golang.org/api/{fcm,eventarc,cloudasset}` and the vertexai REST round-trip) via `option.WithEndpoint` against the in-process GCP server. Existing tests + shared LRO polling stay green.

## Gates
`go build ./...`, `go vet`, scoped `go test -count=1`, `go test -race` on the mutated vertexai provider, and `golangci-lint` (0 issues) all green on the touched packages. `go generate ./...` produces no docs/coverage diff (only a driver struct field was added, no interface method change).

### DIFF
--- a/providers/gcp/vertexai/endpoints.go
+++ b/providers/gcp/vertexai/endpoints.go
@@ -2,6 +2,7 @@ package vertexai
 
 import (
 	"context"
+	"strings"
 
 	"github.com/stackshy/cloudemu/v2/errors"
 	"github.com/stackshy/cloudemu/v2/services/vertexai/driver"
@@ -63,8 +64,21 @@ func (m *Mock) DeployModel(_ context.Context, endpoint string, dm driver.Deploye
 		return nil, nil, errors.Newf(errors.NotFound, "endpoint %q not found", endpoint)
 	}
 
+	// The model being deployed must exist. Real Vertex AI rejects an unknown
+	// model resource with NOT_FOUND rather than deploying a dangling reference.
+	if dm.Model != "" {
+		base, _, _ := strings.Cut(dm.Model, "@")
+		if !m.models.Has(base) {
+			return nil, nil, errors.Newf(errors.NotFound, "model %q not found", dm.Model)
+		}
+	}
+
 	if dm.ID == "" {
 		dm.ID = m.newID()
+	}
+
+	if dm.CreateTime == "" {
+		dm.CreateTime = m.now()
 	}
 
 	// Deep-copy working set so the stored endpoint is never mutated in place.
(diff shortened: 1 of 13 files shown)


NitinKumar004:
Review — verified against the pushed commit (0628fdd9). True scope = top 4 commits (13 files).
- build / vet / gofmt clean; golangci-lint: 0 NEW (2 reported are pre-existing wsl + gosec G101 on untouched lines; PR net-improves lint by removing a gocyclo-11 via validTarget extraction); -race clean (vertexai); coverage no-diff.
- eventarc: no-destination→400 INVALID_ARGUMENT; PATCH honors updateMask (changes destination, bumps etag); server uid+etag (revision++ on patch) + auto Pub/Sub transport; list pagination. LRO via existing inline /operations/ (no dispatch change).
- vertexai: predict-no-models→400 FAILED_PRECONDITION (handler special-cases since codec maps FP→409); deploy-unknown→404; dedicatedResources+createTime round-trip; predict-with-model still works; updated tests legitimately upload-then-deploy.
- fcm: no-target→400. cloudasset: searchAllResources rescopes name+project to queried project (no engine-project leak); exportAssets echoes outputConfig + outputResult, drops cloudemu:// placeholder.
- Driver field additive (CreateTime on DeployedModel, GCP-only); deferrals sound (bigtable admin complete, only gRPC data plane deferred; searchAllIamPolicies honest empty); no collateral.
Nits: pre-existing wsl/gosec remain (not introduced); computeEtag uses fnv64a (fine for an opaque etag).
Verdict: approved.

PR #654, opened 2026-08-25
fix(gcp-alloydb): cluster labels, createTime/updateTime, real uid + caller displayName

NitinKumar004:
## Summary

Fixes the three AlloyDB metadata findings from the GCP audit. All three are wire-output-shape fixes in an AlloyDB-only server; routing is untouched, so GKE (which shares the `/v1/projects/{p}/locations/{l}/clusters` path and is guarded by the panic invariant + `DriversFrom` leaving AlloyDB nil) is unaffected.

- **Labels dropped (HIGH):** `toWireCluster` never emitted `labels`, causing perpetual Terraform drift. Now round-tripped from the cluster's stored tags (create/get/list).
- **createTime/updateTime empty (MEDIUM):** clusters and instances now report RFC3339 `createTime`/`updateTime`. `updateTime` bumps on cluster patch/promote and instance patch; a freshly created resource reports `updateTime == createTime`, matching real AlloyDB.
- **uid/displayName wrong (LOW):** `uid` is now a stable server-generated GUID (`idgen.SyntheticGUID` seeded on the resource name) instead of the cluster id; `displayName` echoes only the caller-supplied value instead of being forced to the id.

## Notes

- New AlloyDB-native fields flow through `AlloyDBClusterInfo`/`AlloyDBInstanceInfo` and `AlloyDBClusterConfig` (added `DisplayName`), so the portable and SDK-compat layers stay in sync.
- GKE + shared LRO tests remain green.

## Tests

Added `TestSDKAlloyDBClusterMetadata` (real `google.golang.org/api/alloydb/v1` client) asserting label round-trip, populated timestamps for clusters and instances, a server UID distinct from the id (and distinct across clusters), and that displayName is empty when the caller sends none.

### DIFF
--- a/providers/gcp/alloydb/alloydb.go
+++ b/providers/gcp/alloydb/alloydb.go
@@ -26,6 +26,7 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/stackshy/cloudemu/v2/config"
 	cerrors "github.com/stackshy/cloudemu/v2/errors"
@@ -70,6 +71,12 @@ type clusterExtra struct {
 	ContinuousBackup       bool
 	MaintenanceDay         string
 	PrimaryCluster         string // source cluster name, for a SECONDARY cluster
+	// DisplayName is the caller-supplied display name (empty when unset); real
+	// AlloyDB echoes only what the caller sent, never the id.
+	DisplayName string
+	// UpdatedAt is the last-modification time; zero until the first modify, in
+	// which case the cluster's CreatedAt stands in as the update time.
+	UpdatedAt time.Time
 }
 
 // instanceExtra holds AlloyDB instance attributes that don't map onto the
@@ -81,6 +88,9 @@ type instanceExtra struct {
 	AvailabilityType string
 	IPAddress        string
 	GceZone          string
+	// UpdatedAt is the last-modification time; zero until the first modify, in
+	// which case the instance's CreatedAt stands in as the update time.
+	UpdatedAt time.Time
 }
 
 // backupExtra holds AlloyDB backup attributes beyond the portable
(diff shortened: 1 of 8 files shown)


NitinKumar004:
Review — verified against the pushed commit (c537723e).
- build / vet / gofmt / gci clean; golangci-lint 0 NEW (2 pre-existing gocritic in untouched subresources.go).
- 3 fixes real (TestSDKAlloyDBClusterMetadata pins all): Labels round-trip (out.Labels=c.Tags); createTime/updateTime RFC3339Nano (updateTime bumps in ModifyCluster/PromoteCluster/ModifyInstance, falls back to createTime when fresh); uid = idgen.SyntheticGUID(ARN) deterministic + distinct from id; displayName echoes caller only (empty when unset; c1≠c2 uid, c2 empty displayName pinned).
- GKE untouched + collision invariant intact (the central check): diff is alloydb provider/server + relationaldb driver + one test — NO routing/Matches/registration/gcp.go New()/from_provider changes. Panic guard + DriversFrom-leaves-AlloyDB-nil untouched; TestFullServerLROOperationPolling confirms /clusters still routes to GKE (create+GET 200). GKE suite + full server/gcp green.
- Driver fields additive (zero-default, AlloyDB-specific structs only; RDS/Azure SQL/mysqlflex/postgresflex/Cloud SQL green); -race clean; coverage no-diff; no collateral.
Non-blocking nits: instance Uid still emits ID not a SyntheticGUID (outside the 3 cluster-scoped findings, now inconsistent); the test asserts timestamps non-empty but a frozen FakeClock can't observe updateTime advancing (correct by code); detachSecondaries range refactor is behavior-preserving (lint cleanup).
Verdict: approved.

PR #655, opened 2026-08-25
fix(gcp-firestore): nested field-path update (dotted updateMask) no longer silently dropped

NitinKumar004:
## What

A dotted `updateMask` field path (e.g. `profile.age`) in a Firestore write now
correctly addresses a **nested** field. Previously the handler treated the whole
dotted string as a flat top-level map key, so the nested write was **silently
dropped with no error** — a real-user data-loss bug.

**Data-loss repro (real `cloud.google.com/go/firestore` SDK):**

```go
docRef.Set(ctx, map[string]any{"profile": map[string]any{"age": 30, "name": "alice"}})
docRef.Update(ctx, []firestore.Update{{Path: "profile.age", Value: 31}})
snap, _ := docRef.Get(ctx)
// before: profile.age == 30  (the update vanished, no error)
// after:  profile.age == 31
```

## Why

The SDK sends `fieldPaths: ["profile.age"]` alongside a nested body
`fields:{profile:{mapValue:{fields:{age:...}}}}`. The handler looked up
`item["profile.age"]`, which does not exist (the item key is `profile`), fell
into the mask delete branch (a no-op), and never applied the nested value. This
was the residue of the round-1 top-level-only updateMask fix (#639).

## Fix

Server wire-handler only (no driver/provider change):

- `mergeMasked` (used by `:commit` and PATCH `updateDocument`) now splits dotted
  field paths, descends/creates intermediate maps, and sets or deletes the leaf.
  A masked nested path absent from the write deletes just that nested field;
  siblings are preserved. Maps are cloned along the descent so the stored
  document is never mutated in place.
- `applyFieldMask` (the `listDocuments` mask projection) projects nested paths,
  rebuilding the enclosing `mapValue`s so sibling nested fields are excluded.
- Backtick-quoted path segments are honored, so a key literally containing a dot
  can still be addressed (`` `a.b`.c ``). Backslash-escaped backticks inside a
  quoted segment are deferred (uncommon).

Top-level masked updates and full-document sets are unchanged.

## Tests

New real-SDK reproductions (`firestore_fieldpath_test.go`): nested `profile.age`
update now sets the value (was dropped) and Get returns it; sibling nested field
and untouched top-level nested map preserved; a dotted path creates a missing
intermediate map; `firestore.Delete` on a nested path removes only that leaf;
top-level update still works. Verified these fail on the pre-fix handler and pass
after. Existing #639 firestore tests remain green (build/vet/test/-race/lint/gci
all clean; `go generate` produces no coverage diff).

### DIFF
--- a/server/gcp/firestore/firestore_fieldpath_test.go
+++ b/server/gcp/firestore/firestore_fieldpath_test.go
@@ -1,0 +1,187 @@
+// Nested field-path updateMask behavior, driven by the real
+// cloud.google.com/go/firestore SDK (REST client) against the emulator's GCP
+// HTTP server. A dotted field path (e.g. "profile.age") addresses a nested
+// field; before the fix the emulator treated it as a flat top-level key, so the
+// nested write was silently dropped (data loss). These tests assert real
+// Firestore semantics: the nested leaf is written, siblings are preserved, and a
+// masked nested path absent from the write deletes only that leaf.
+package firestore_test
+
+import (
+	"testing"
+
+	gcpfirestore "cloud.google.com/go/firestore"
+)
+
+// nestedMap pulls out a nested map[string]any field from a snapshot payload.
+func nestedMap(t *testing.T, data map[string]any, key string) map[string]any {
+	t.Helper()
+
+	m, ok := data[key].(map[string]any)
+	if !ok {
+		t.Fatalf("field %q is not a nested map: %#v", key, data[key])
+	}
+
+	return m
+}
+
+// TestDatabaseNestedFieldPathUpdate asserts that Update with a dotted field
+// path writes the nested leaf, leaves sibling nested fields untouched, and that
+// updating one nested key does not disturb another top-level nested map.
+func TestDatabaseNestedFieldPathUpdate(t *testing.T) {
+	ctx, client, _ := newDBClient(t, "people")
+
+	doc := client.Collection("people").Doc("u1")
+
+	if _, err := doc.Set(ctx, map[string]any{
+		"name": "root",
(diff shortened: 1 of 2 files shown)


NitinKumar004:
Review — verified against the pushed commit (d3ac089a), exactly 2 files.
- build / vet / gofmt / gci clean; 0 NEW lint (the 6 wsl are pre-existing in the base handler's UnmarshalJSON funcs).
- Data-loss fixed (mutation-verified): reverting handler.go to base fails TestDatabaseNestedFieldPathUpdate (profile.age 30 want 31), CreatesIntermediate, Delete; passes with the fix; TestDatabaseTopLevelUpdateStillWorks green on both (no #639 regression).
- Adversarial edges (reviewer's own tests): deep a.b.c.d update+delete (deep sibling preserved); intermediate-not-map (scalar cleanly overwritten to a map, no panic); delete of non-existent nested path no-op no panic; NO-ALIASING/torn-map — maps cloned along descent (setNestedPath/deleteNestedPath re-clone each level; provider Get/Put only shallow-clone, so descent-cloning is what prevents shared-doc mutation), -race clean incl. 20x concurrent nested update/get; multiple masked paths on the same subtree accumulate (no lost write); backtick literal key  addresses one key not nested.
- Projection: Select(profile.age) returns only that leaf, siblings excluded, enclosing map rebuilt. No regression; coverage no-diff (server-layer only).
Non-blocking nits: backslash-escaped-backtick deferred (no SDK path hits it); minor extra alloc on no-op delete; nested-through-scalar overwrites the scalar (matches real Firestore updateMask).
Verdict: approved.

PR #656, opened 2026-08-25
fix(gcp-lro): operation registry — real response + 404 on unknown ops

NitinKumar004:
## What

The shared GCP long-running-operation poller (`server/gcp/lro`) answered **any** `GET /v1/projects/{p}/locations/{l}/operations/{op}` with a synthetic `{done:true, status:DONE}` and no response body. Two consequences, both flagged in `AUDIT_GCP.md ## lro`:

- A completed operation's **typed response was missing** on poll, so a GAPIC `.Wait()` read a zero-value resource off the operation.
- An operation name that was **never created was masked as `done`** instead of returning `404 NOT_FOUND` as real GCP does.

This adds a thread-safe operation **Registry** the poller consults:

- Services **register** the operations they create (name → typed response) at creation time.
- The poller **replays** a registered operation (done + its response), and **404s** any operation name that was never registered.
- A poller built **without** a registry keeps the legacy always-done behavior, so standalone per-service package servers (which serve their own `/operations/` poll) are unaffected.

## Why

Real GCP returns the typed result in a done operation's `response`, and 404s an unknown operation id. Masking unknown ops as done hides real "not found" errors from clients and breaks `CreateXOperation.Wait()` result decoding.

## Services wired

Only the location-scoped producers whose operations are routed through the shared poller register their ops (via `SetOperationRegistry`):

- **artifactregistry**
- **eventarc**
- **alloydb**
- **memorystore**

**GKE** is unchanged: it registers ahead of the poller and owns its own operations (richer `container.Operation` shape); its `Matches` only claims ops its mock recorded, so foreign/unknown names fall through and now correctly 404.

Non-producers are unaffected (disjoint paths / own handlers): Cloud Run and Cloud Functions poll on `/v2` (and CF gen1 on `/v1/operations`), Cloud SQL on `/v1/projects/{p}/operations`, Bigtable on `/v2/operations`, Cloud Asset on `projects/{p}/operations` — none of which match the poller's 5-segment location path. Vertex AI returns done-on-arrival inline and never polls.

## Tests

- `server/gcp/lro/handler_test.go` (new): registered op → done + response; unknown op → 404; nil-registry legacy done; concurrent register/poll under `-race`.
- `server/gcp/fullserver_test.go`: AR poll now also asserts the response body is present; added an unknown-op → 404 assertion; GKE poll uses the real returned operation name.
- `server/gcp/gke/sdk_fidelity_test.go`: the foreign-operation test now asserts `404 NOT_FOUND` (the corrected behavior) instead of the old masked-done.

Gates run (scoped, `GOMAXPROCS=3 -p=3`): `go build ./...`, `go vet`, `go test ./server/gcp/...`, `go test -race` on `lro` + fullserver, `golangci-lint` on changed packages (0 issues), `gofmt`.

### DIFF
--- a/server/gcp/alloydb/handler.go
+++ b/server/gcp/alloydb/handler.go
@@ -32,6 +32,7 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/stackshy/cloudemu/v2/server/gcp/lro"
 	rdsdriver "github.com/stackshy/cloudemu/v2/services/relationaldb/driver"
 )
 
@@ -60,12 +61,22 @@ const (
 // also implements the AlloyDB optional capability.
 type Handler struct {
 	db rdsdriver.RelationalDB
+
+	// ops records created operations with the shared poller so a client that
+	// polls the returned operation name gets the typed response (and unknown
+	// names 404). Nil in a standalone package server, where this handler serves
+	// its own /operations/ poll.
+	ops *lro.Registry
 }
 
 // New returns an AlloyDB handler backed by db.
 func New(db rdsdriver.RelationalDB) *Handler {
 	return &Handler{db: db}
 }
+
+// SetOperationRegistry wires the shared LRO poller so created operations are
+// resolvable (with their response) through the full server's operations host.
+func (h *Handler) SetOperationRegistry(reg *lro.Registry) { h.ops = reg }
 
 // Matches claims /v1/projects/{p}/locations/{l}/{clusters|backups|operations}
 // paths. Everything else under /v1/projects/ falls through to other GCP
--- a/server/gcp/alloydb/helpers.go
+++ b/server/gcp/alloydb/helpers.go
@@ -62,10 +62,11 @@ func decodeJSON(w http.ResponseWriter, r *http.Request, v any) bool {
 	return true
 }
 
-// doneOperation builds a completed AlloyDB LRO envelope carrying the resource
-// as its response. AlloyDB REST callers receive a terminal operation, so no
-// polling is required.
-func (*Handler) doneOperation(p *alloyPath, verb string, response any) *alloydb.Operation {
+// doneOperation builds a completed AlloyDB LRO envelope carrying the resource as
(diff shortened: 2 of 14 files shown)


NitinKumar004:
Review — verified against the pushed commit (90fbb12). Exactly 14 files, all server/gcp/, no contamination.
- build / vet / gofmt / gci / golangci-lint clean (0 NEW; 4 memorystore findings pre-existing outside the diff); all 26 server/gcp packages green incl. fullserver + lro + every per-service op test; -race clean.
- finding2 fixed: shared poller claims exactly /v1/projects/{p}/locations/{l}/operations/{op}; registered op → done+typed response; unregistered → 404 notFound (no masked done); nil-registry → legacy done (Register nil-guarded, standalone package servers unaffected).
- COMPLETENESS (the whole risk) verified — enumerated every GCP service, each covered: REGISTERED (artifactregistry/eventarc/memorystore/alloydb incl. backups); SELF-OWNED (GKE, registered ahead + store-aware Matches → foreign op falls through to 404); DONE-ON-ARRIVAL never-polls (vertexai, cloudfunctions gen2 — gax .Wait short-circuits on inline done); DISJOINT op-paths (cloudrun /v2, cloudsql /sql/v1beta4, clouddns, compute/vpc/lb /compute/v1, bigtable /v2, cloudasset, CF gen1 /v1/operations); SYNCHRONOUS no-LRO (iam/secretmanager/pubsub/firestore/monitoring/fcm). No service polls the shared handler while unregistered-and-not-self-owned — no broken .Wait().
- Changed TestSDKGKEForeignOperationStillDone→NotFound inverts the exact masked-done bug (real 404), not weakened; GKE real op still returns status:DONE; coverage no-diff.
Non-blocking nits: alloydb op names verb-scoped not resource-unique (registry overwrite across two clusters — harmless, clients short-circuit on inline done); a manual Operations.Get on a done-on-arrival op now 404s (not a .Wait regression — was already shadowed); registry unbounded (emulator-acceptable).
Verdict: approved.

PR #657, opened 2026-08-25
fix(aws-s3): enforce conditional writes (If-None-Match/If-Match→412) + conditional reads (304/412)

NitinKumar004:
## What

Enforce S3 conditional requests on the AWS wire path, closing two HIGH bugs from the deep-e2e audit.

### F1 [HIGH] — Conditional writes on PutObject were silently ignored
`PutObject` with `If-None-Match: *` (create-if-absent) or `If-Match: <etag>` (optimistic replace) used to return 200 and overwrite the object regardless of the condition, breaking create-if-absent and optimistic-lock patterns (e.g. Terraform S3 state locking; AWS added S3 conditional writes in 2024).

Now the condition is enforced and a failed condition returns **412 PreconditionFailed** without overwriting. The check-and-set is done atomically in the provider under `versionsMu` (new optional `driver.ConditionalBucket` capability, mirroring the existing GCS `PutObjectGCS(..., precondition)` pattern) so concurrent create-if-absent writers cannot both win — no Get-then-Put lost-update race.

### F2 [HIGH] — Conditional reads on GetObject/HeadObject were ignored
`If-Match`, `If-Unmodified-Since`, `If-None-Match`, `If-Modified-Since` on GET/HEAD were ignored and always returned 200 + full body. Now they are evaluated (server-only) with RFC 7232 §6 precedence — If-Match beats If-Unmodified-Since (fail → 412), If-None-Match beats If-Modified-Since (match → 304) — returning **304 Not Modified** or **412 PreconditionFailed** as appropriate. Verified against the official AWS S3 GetObject/PutObject docs.

## Why

These are the exact guards clients rely on to prevent clobbering and to revalidate caches; ignoring them is a silent wrong-result / data-loss-adjacent bug.

## Tests

Real `aws-sdk-go-v2` reproductions: PutObject `If-None-Match:*` new-key success / existing-key 412; `If-Match` match success / stale 412 / missing-key 412; GET `If-None-Match` match 304, `If-Modified-Since` future 304, `If-Match` mismatch 412, `If-Unmodified-Since` violated 412; precedence (If-Match overrides If-Unmodified-Since); HEAD 304. Provider-level `-race` test proving exactly one concurrent create-if-absent writer wins. Existing S3 tests (incl. copy-source conditionals) remain green. Coverage docs regenerated.

### DIFF
--- a/docs/coverage/aws/s3.md
+++ b/docs/coverage/aws/s3.md
@@ -101,6 +101,14 @@ BucketAttributes is an OPTIONAL capability, discovered by type assertion (like
 | --- | --- |
 | `BucketAttributes` |  |
 
+### ConditionalBucket
+
+ConditionalBucket is an OPTIONAL capability (discovered by type assertion like
+
+| Operation | Description |
+| --- | --- |
+| `PutObjectConditional` |  |
+
 ### GCSExtensions
 
 GCSExtensions is an OPTIONAL GCS-specific capability, discovered by type
--- a/docs/coverage/azure/blobstorage.md
+++ b/docs/coverage/azure/blobstorage.md
@@ -101,6 +101,14 @@ BucketAttributes is an OPTIONAL capability, discovered by type assertion (like
 | --- | --- |
 | `BucketAttributes` |  |
 
+### ConditionalBucket
+
+ConditionalBucket is an OPTIONAL capability (discovered by type assertion like
+
+| Operation | Description |
+| --- | --- |
+| `PutObjectConditional` |  |
+
 ### GCSExtensions
 
 GCSExtensions is an OPTIONAL GCS-specific capability, discovered by type
--- a/docs/coverage/coverage.json
+++ b/docs/coverage/coverage.json
@@ -10679,6 +10679,15 @@
         ]
       },
       {
+        "name": "ConditionalBucket",
+        "doc": "ConditionalBucket is an OPTIONAL capability (discovered by type assertion like",
+        "operations": [
+          {
+            "name": "PutObjectConditional"
+          }
+        ]
+      },
+      {
         "name": "GCSExtensions",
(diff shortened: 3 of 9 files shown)


NitinKumar004:
Review — verified against the pushed commit (34ce5517).
- build / vet / gofmt clean; golangci-lint --new-from-rev = 0 NEW (27 baseline hits pre-existing in untouched funcs); the 3 fmt hits are pre-existing test files not changed by this PR.
- F1 conditional writes enforced + ATOMIC: driver.ConditionalBucket (type-asserted like VersionedBucket); PutObjectConditional→storeObjectConditional holds versionsMu across BOTH evalPutPrecondition (read) and storeObjectLocked (write) — genuine check-and-set, not Get()+Set(); plain PutObject shares versionsMu (no interleave). If-None-Match:* new→200/existing→412 untouched; If-Match match→200/stale→412 untouched.
- Atomicity: TestPutObjectConditionalConcurrentCreate fires 32 concurrent If-None-Match:* writers → winners==1, -race clean.
- F2 conditional reads + RFC 7232 §6 precedence: matchGroupFails (If-Match beats If-Unmodified-Since→412) before noneMatchGroupNotModified (If-None-Match beats If-Modified-Since→304); wired into getObject+headObject BEFORE the Range branch (304/412 short-circuits ahead of range); 304 emits ETag+Last-Modified (+version-id), no body; LastModified RFC3339 so time conditions fire.
- Capability additive-safe (handler falls back to plain PutObject; Azure Blob + GCS unaffected); coverage regen idempotent (go generate → zero further diff); no collateral (copy/multipart/versioning/pagination green).
Non-blocking nits: multi-value/weak-ETag lists in If-Match/If-None-Match not parsed (single-value compare, mirrors existing copySourceETagMatches); a hypothetical non-AWS driver lacking ConditionalBucket would silently ignore the header (no AWS impact — Mock always implements it).
Verdict: approved.

PR #658, opened 2026-08-25
fix(aws-ec2): terminate detaches EBS volumes + atomic AttachVolume + attach/detach state validation

NitinKumar004:
## What

Fixes a cohesive cluster of AWS EC2/EBS attach-lifecycle bugs found by the deep-e2e audit, including a HIGH data-loss bug and a proven data race.

- **#1 [HIGH data-loss]** `TerminateInstances` now releases every attached EBS volume back to `available` (attachment cleared). Previously a terminated instance stranded its volume `in-use` forever, and the volume could never be deleted (`DeleteVolume` -> `VolumeInUse`). User-attached volumes carry `DeleteOnTermination=false`, so they detach rather than delete.
- **#6 [MED concurrency, -race-proven]** `AttachVolume` is now an atomic check-and-set via `memstore.Update` with copy-on-write (returns a fresh pointer). Two concurrent attaches of the same volume now have exactly one winner; the loser gets `VolumeInUse`. The old check-then-set on the stored pointer both double-attached one volume to two instances and raced with concurrent `DescribeVolumes` reads.
- **#2 [MED]** `AttachVolume` to a terminated/non-running instance is rejected with `IncorrectInstanceState` (was a silent success). A volume can only attach to a running or stopped instance.
- **#3 [MED]** `DetachVolume` validates the supplied `InstanceId`/`Device` against the actual attachment; a mismatch returns `InvalidAttachment.NotFound` (was a silent detach). The wire response now echoes the real attachment, not the caller's request values.

## Why

These match real EC2 semantics for TerminateInstances / AttachVolume / DetachVolume and the EBS volume lifecycle (verified against the AWS EC2 API error reference). The terminate bug orphaned volumes permanently; the attach race could corrupt volume ownership under concurrency.

## Notes

`DetachVolume` gains `instanceID`/`device` params on the shared compute driver. AWS validates them; Azure/GCP accept and ignore them (they detach by disk id), preserving existing behavior.

## Tests

Real `aws-sdk-go-v2` EC2 reproductions in `server/aws/ec2_test.go`:
- terminate an instance with an attached volume -> volume becomes `available` and is deletable
- AttachVolume to a terminated instance -> `IncorrectInstanceState`
- DetachVolume with the wrong InstanceId -> `InvalidAttachment.NotFound` (and a correctly-targeted detach still succeeds)

Plus a `-race` provider test (`providers/aws/ec2/ec2_test.go`): concurrent AttachVolume of one volume to two instances with concurrent DescribeVolumes readers -> exactly one winner, no data race.

### DIFF
--- a/cloudemu_test.go
+++ b/cloudemu_test.go
@@ -7126,7 +7126,7 @@ func TestVolumeLifecycleAWS(t *testing.T) {
 		t.Errorf("expected state 'in-use', got %q", vols[0].State)
 	}
 
-	if err := p.EC2.DetachVolume(ctx, vol.ID); err != nil {
+	if err := p.EC2.DetachVolume(ctx, vol.ID, "", ""); err != nil {
 		t.Fatal(err)
 	}
 
--- a/docs/coverage/aws/ec2.md
+++ b/docs/coverage/aws/ec2.md
@@ -28,7 +28,7 @@ AWS's `compute` service · portable interface `driver.Compute` · [AWS index](./
 | `DescribeSnapshots` |  |
 | `DescribeSpotRequests` |  |
 | `DescribeVolumes` |  |
-| `DetachVolume` |  |
+| `DetachVolume` | DetachVolume detaches a volume. When instanceID or device is non-empty it |
 | `ExecuteScalingPolicy` |  |
 | `GetAutoScalingGroup` |  |
 | `GetLaunchTemplate` |  |
--- a/docs/coverage/azure/virtualmachines.md
+++ b/docs/coverage/azure/virtualmachines.md
@@ -28,7 +28,7 @@ Azure's `compute` service · portable interface `driver.Compute` · [Azure index
 | `DescribeSnapshots` |  |
 | `DescribeSpotRequests` |  |
 | `DescribeVolumes` |  |
-| `DetachVolume` |  |
+| `DetachVolume` | DetachVolume detaches a volume. When instanceID or device is non-empty it |
 | `ExecuteScalingPolicy` |  |
 | `GetAutoScalingGroup` |  |
 | `GetLaunchTemplate` |  |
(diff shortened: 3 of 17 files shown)


NitinKumar004:
Review — verified against the pushed commit (b9587998). All 4 fixes real; one mechanical blocker.
- #1 data-loss fixed: detachTerminatedVolumes returns attached volumes to available on terminate (deletable, was stuck VolumeInUse); multi-volume all detach, no-volume no panic, other-instance untouched.
- #6 atomic attach: memstore.Update runs check(available?)+write(mark attached) under the store lock returning a FRESH pointer (copy-on-write); describeResources derefs the old immutable pointer outside the lock (no torn read); 2-attacker + concurrent-describe single-winner test passes, -race clean; no Get+Set.
- #2 attach-to-terminated → IncorrectInstanceState (state read via readState() under inst.mu, transitionOne under same mu — no race); #3 detach wrong-instance → InvalidAttachment.NotFound (attachment untouched), correct detach echoes the real attachment.
- Shared-driver change safe: DetachVolume gained (instanceID,device); all 5 implementers + wire call sites updated (AWS passes instID+device; 4 Azure pass '',''; no GCP wire site); Azure/GCP accept-and-ignore, detach by disk id (behavior unchanged); no collateral (#508 VolumeInUse/ZoneMismatch preserved); build/vet/gofmt clean, 0 NEW lint.
BLOCKING (mechanical): the DetachVolume doc-comment on the shared driver interface surfaces into docs/coverage/{aws/ec2,azure/virtualmachines,gcp/gce}.md + coverage.json — go generate produces an uncommitted diff → docs-generate gate red. Regenerating + committing those 4 files now.
Nit: AttachVolume now validates instance-existence before volume-existence (matches real EC2, no regression).
Verdict: changes (regenerate coverage) — fixing.

PR #659, opened 2026-08-25
fix(aws-sts): AssumeRole evaluates the trust policy (deny->AccessDenied) + ListEntitiesForPolicy + instance-profile LimitExceeded

NitinKumar004:
## What

Fixes three AWS STS + IAM findings from the deep-e2e audit.

### #1 [MED, security] STS AssumeRole now evaluates the role trust policy
Previously the STS handler was driverless: `AssumeRole` synthesized credentials
for **any** `RoleArn` — a role whose trust policy explicitly *denies*, and even a
**nonexistent** role, both returned valid creds (`err=nil`). Every negative
security test was a false pass.

Now the STS handler receives the IAM driver (provider-factory wiring in
`server/aws/aws.go`). On `AssumeRole` it resolves the target role and evaluates
its `AssumeRolePolicyDocument` for an `sts:AssumeRole` Allow (with explicit Deny
overriding), reusing the IAM backend's policy-evaluation building blocks. A
denying trust policy or a missing role now returns **`AccessDenied` (HTTP 403)**,
matching real STS (which does not disclose which of the two failed). When no IAM
driver is wired (standalone STS-only), `AssumeRole` stays permissive so SDK
init-creds paths keep working. Caller principal is derived as the account-root
identity (`arn:aws:iam::<account>:root`) since cloudemu does not verify SigV4.

### #2 [MED gap] ListEntitiesForPolicy implemented
The standard `iam:ListEntitiesForPolicy` read API was undispatched →
`InvalidAction (400)`. Added the action, an XML shape
(`PolicyUsers`/`PolicyGroups`/`PolicyRoles` with Name+Id), and a backend
reverse-lookup over the existing `userPolicies`/`groupPolicies`/`rolePolicies`
attachment maps, honoring `EntityFilter`, `PathPrefix`, and `MaxItems`/`Marker`
pagination.

### #3 [LOW] AddRoleToInstanceProfile second role → LimitExceeded
Adding a *different* second role to an instance profile returned
`EntityAlreadyExists`; per the AWS one-role-per-profile quota it must be
`LimitExceeded`. Changed the provider to return `ResourceExhausted`
(maps to `LimitExceeded`).

## Why
Trust-policy bypass makes negative security tests silently pass; the missing read
API breaks audit/IaC tooling; the wrong error type misfires SDK retry/branch
logic keyed on `LimitExceededException`.

## Tests
Real `aws-sdk-go-v2` reproductions: AssumeRole with an allowing trust policy →
creds; denying trust policy → AccessDenied 403; nonexistent role → AccessDenied
403; permissive without IAM wired; ListEntitiesForPolicy returns attached
users/roles/groups + EntityFilter + pagination; AddRoleToInstanceProfile second
role → LimitExceeded. Plus provider-level trust-evaluation and reverse-lookup
unit tests. Existing STS/IAM tests remain green.

### DIFF
--- a/providers/aws/iam/entities.go
+++ b/providers/aws/iam/entities.go
@@ -1,0 +1,76 @@
+package iam
+
+import (
+	"context"
+
+	"github.com/stackshy/cloudemu/v2/errors"
+	"github.com/stackshy/cloudemu/v2/services/iam/driver"
+)
+
+// ListEntitiesForPolicy returns the users, groups, and roles the managed policy
+// with the given ARN is attached to (IAM ListEntitiesForPolicy). It reverse-
+// scans the managed-policy attachment maps and resolves each principal's
+// Name/Id/Path from its store. The wire layer applies the EntityFilter,
+// PathPrefix, and pagination filters. AWS-only; not part of the portable driver.
+func (m *Mock) ListEntitiesForPolicy(_ context.Context, policyARN string) (driver.PolicyEntities, error) {
+	if !m.policies.Has(policyARN) {
+		return driver.PolicyEntities{}, errors.Newf(errors.NotFound, "policy %q not found", policyARN)
+	}
+
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	return driver.PolicyEntities{
+		Users:  m.attachedUsersLocked(policyARN),
+		Groups: m.attachedGroupsLocked(policyARN),
+		Roles:  m.attachedRolesLocked(policyARN),
+	}, nil
+}
+
+// attachedUsersLocked returns the users the policy ARN is attached to. Caller
+// must hold m.mu.
+func (m *Mock) attachedUsersLocked(policyARN string) []driver.PolicyEntity {
+	var out []driver.PolicyEntity
+
+	for name, arns := range m.userPolicies {
+		if arns[policyARN] {
+			if u, ok := m.users.Get(name); ok {
+				out = append(out, driver.PolicyEntity{Name: u.Name, ID: u.ID, Path: u.Path})
+			}
+		}
+	}
+
+	return out
+}
+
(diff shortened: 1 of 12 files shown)


NitinKumar004:
Review — verified against the pushed commit (b5efb67). Security-correct.
- Trust eval: EvaluateAssumeRoleTrust→evaluateTrustPolicy returns (allow,deny), result allow && !deny — default-deny + explicit-deny-override; nonexistent role (false,false); wire maps disallow OR missing → AccessDenied 403 (no disclosure). Reviewer ran 12 adversarial cases: empty/{}/non-sts:AssumeRole → DENIED; wrong-account root + specific-user principal → DENIED (principal matching genuinely restricts); Principal:*/{AWS:*}/exact-root/root-in-list/Action:*//sts:* → ALLOWED; Deny scoped to a different action/principal does NOT suppress a matching Allow; explicit Deny+Allow same match → DENIED. No bypass, no over-permissiveness/restriction.
- Fallback deliberate: trustAllows permissive only when h.trust==nil (standalone init-creds, documented); full server NewAWS wires cloud.IAM → enforcing path live (TestSDKAssumeRoleTrustAllows/Denies + nonexistent-role prove 403 end-to-end).
- ListEntitiesForPolicy accurate (reverse-lookup over managed-attachment maps, EntityFilter/PathPrefix/pagination); AddRoleToInstanceProfile 2nd role → LimitExceeded; no interface change (coverage no-diff); -race clean; no collateral (#567 intact).
Nits (informational): caller always account-root (SigV4 unverified) so a trust policy naming a specific user/service principal is denied for direct SDK AssumeRole — deliberate, matches real AWS; malformed doc → deny (correct); a non-nil interface wrapping a nil *Mock would panic (unreachable via the factory — optional hardening).
Verdict: approved.

PR #660, opened 2026-08-25
fix(aws-lambda): resolve Qualifier on Get/GetConfiguration + per-qualifier policy + ESM function existence + principal shape

NitinKumar004:
## What

Fixes four AWS Lambda bugs found by the deep-e2e audit around qualifier resolution, resource policies, and event-source-mapping validation.

**F1 [HIGH] — GetFunction / GetFunctionConfiguration now honor `?Qualifier=`.**
Reading a published version's or alias's configuration previously always returned `$LATEST`. Now a version number resolves to that version's immutable snapshot (its own `Runtime`/`Handler`/`Timeout`/`MemorySize`/`CodeSha256`/`RevisionId` and a version-qualified ARN), and an alias resolves one hop to its target version's config while keeping the alias-qualified ARN. A qualifier that names neither an alias nor a published version returns `ResourceNotFoundException`. `$LATEST`/no-qualifier is unchanged. Mirrors the qualifier resolution Invoke already did.

**F2 [MED] — CreateEventSourceMapping rejects a nonexistent target function.**
Previously an ESM whose `FunctionName` pointed at a missing function was created `Enabled` and silently never fired. Now the target function must exist or the call returns `ResourceNotFoundException` (a documented error for this API). Cross-service DynamoDB/SQS delivery uses `InvokeExternal`, not this path, so it is unaffected.

**F3 [MED] — Resource policy scoped per (function, qualifier).**
`AddPermission`/`RemovePermission`/`GetPolicy` now thread the `Qualifier` from wire to provider and key the policy store by qualifier, so an alias/version grant lands on that qualifier's policy and no longer bleeds into the `$LATEST` policy. `GetPolicy(Qualifier=...)` reads back only that qualifier's statements.

**F4 [MED] — GetPolicy renders the principal per its type.**
A service principal (e.g. `s3.amazonaws.com`) renders as `{"Service": ...}`, an AWS account id as `{"AWS":"arn:aws:iam::<id>:root"}`, an IAM ARN as `{"AWS": <arn>}`, and `*` as a bare `"*"`. The `AddPermission` response echo now reuses the same rendering (pulled back from `GetPolicy`) so the two agree.

## Why

The Lambda surface was half-wired: Invoke resolved qualifiers but Get/GetFunctionConfiguration did not, so any client pinning a version/alias saw `$LATEST` — perpetual Terraform/CDK drift. Alias/version grants were mis-scoped, an account/`*` principal came back mis-shaped, and a typo'd ESM function silently succeeded.

## Tests

New `aws-sdk-go-v2` reproductions in `server/aws/lambda/qualifier_policy_test.go`: version/alias qualifier resolution (distinct CodeSha256/Timeout/ARN), bad qualifier -> RNF, `$LATEST` unchanged; ESM on a nonexistent function -> RNF; per-qualifier `AddPermission`/`GetPolicy`/`RemovePermission`; principal Service vs AWS-account vs `*` shapes. Existing Lambda suites (ESM delivery, alias-invoke, defaults, tagging, resource-policy) stay green.

### DIFF
--- a/providers/aws/lambda/esm.go
+++ b/providers/aws/lambda/esm.go
@@ -28,6 +28,13 @@ func (m *Mock) CreateEventSourceMapping(
 		return nil, cerrors.New(cerrors.InvalidArgument, "event source ARN is required")
 	}
 
+	// The target function must exist. Real Lambda rejects an event-source mapping
+	// whose FunctionName points at a missing function with ResourceNotFoundException
+	// rather than creating a mapping that silently never fires.
+	if _, ok := m.funcs.Get(functionNameFromARN(cfg.FunctionName)); !ok {
+		return nil, cerrors.Newf(cerrors.NotFound, "Function not found: %s", cfg.FunctionName)
+	}
+
 	batchSize := cfg.BatchSize
 	if batchSize == 0 {
 		batchSize = defaultBatchSize
--- a/providers/aws/lambda/lambda.go
+++ b/providers/aws/lambda/lambda.go
@@ -82,8 +82,11 @@ type funcData struct {
 	nextVersion  int
 	aliases      *memstore.Store[*aliasData]
 	concurrency  *driver.ConcurrencyConfig
-	policy       map[string]driver.PermissionStatement
-	urlConfig    *driver.FunctionURLConfig // Lambda Function URL, nil until created
+	// policies is the resource-based policy keyed by qualifier (normalized via
+	// policyKey: "" and "$LATEST" collapse to the unqualified function policy),
+	// then by statement id. AWS keeps a separate policy per version/alias.
+	policies  map[string]map[string]driver.PermissionStatement
+	urlConfig *driver.FunctionURLConfig // Lambda Function URL, nil until created
 	// awsConfig holds the AWS-only settings (VpcConfig/DeadLetterConfig/
 	// TracingConfig) applied through the AWSConfigurable optional interface.
 	awsConfig driver.AWSFunctionConfig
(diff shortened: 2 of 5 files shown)


NitinKumar004:
Review — verified against the pushed commit (ccf9c4ec).
- build / vet / gofmt / gci / golangci-lint clean (the one fmt diff is pre-existing in sdk_roundtrip_test.go, not a PR file); -race clean (policies map + esm); all 5 new SDK tests pass.
- F1 qualifier: resolvedConfiguration — version→toVersionConfiguration (own CodeSha256/Runtime/Timeout, :N ARN); alias→GetAlias→findVersion (alias-qualified ARN); bad→RNF; ''/ unchanged; reuses ListVersions/GetAlias (same alias-first-then-version semantics as Invoke's resolveQualifier), no driver-interface change.
- F3 per-qualifier policy: map[qualifier]map[stmtID]stmt; policyKey('')==policyKey('')=='' (collapse, matches AWS); alias/version separate; AddPermission validates the qualifier (missing→RNF); no cross-qualifier leak; RemovePermission scoped; every mutation followed by funcs.Set.
- F4 principal: *→'*'; arn:→{AWS:arn}; 12-digit→{AWS:...:root}; else {Service:...}; AddPermission echo → addedStatement→GetPolicy so echo and readback agree (and now includes Resource/Condition). F2 ESM: CreateEventSourceMapping funcs.Get→NotFound if missing.
- policyManager AWS-only type-asserted (no driver/coverage change; all prod callers updated); no collateral (#569/#575 ESM delivery, #572 alias-invoke, #512/#552 defaults green).
Verdict: approved.

PR #661, opened 2026-08-25
fix(aws-dynamodb): atomic conditional writes + transactions (close TOCTOU race)

NitinKumar004:
## What

Make AWS DynamoDB conditional writes and transactions **atomic**, closing a data-integrity TOCTOU race found by the deep-e2e audit.

The wire handler used to evaluate a `ConditionExpression` with a standalone `GetItem` (check) and then issue the mutation as a **separate** provider call (act), dropping the table lock in between. The provider exposed no conditional-write primitive, so nothing held a lock across check+write. As a result:

- **#1 conditional-write TOCTOU** — two concurrent `PutItem`/`UpdateItem`/`DeleteItem` with a `ConditionExpression` on the same key (e.g. `attribute_not_exists(pk)` put-if-absent, or optimistic-lock `SET v=:new IF v=:cur`) could **both** succeed. Reproduced deterministically (2 winners in a race).
- **#2 TransactWriteItems TOCTOU** — the transaction evaluated all conditions, then applied all writes, in **separate** lock acquisitions, and bypassed the provider's already-atomic path entirely (the wire layer reimplemented a non-atomic version with hand-rolled rollback). Two conflicting transactions could both commit.

## Why / Fix

Condition evaluation is pushed **down into the provider**, where the check and the write happen under a **single hold of the table lock** (`m.mu`) — the atomic check-and-set pattern, mirroring how S3 conditional writes (#657) and EC2 AttachVolume (#658) were fixed.

- Add `PutItemConditional` / `UpdateItemConditional` / `DeleteItemConditional` to the AWS DynamoDB provider. `PutItem`/`UpdateItem`/`DeleteItem` now delegate to them with an empty condition, so the conditional and unconditional paths share one atomic implementation.
- Add an atomic `TransactWrite` that evaluates **every** `ConditionExpression` and applies **every** write under one lock hold (all-or-nothing), driven from the wire handler. The old handler-level `evaluateTransactConditions` / `applyTransactOps` / rollback machinery is removed (dead code gone).
- These are exposed as a **type-asserted capability** (`conditionalWriter`), like `UpdateThroughput` / `SetPITR`, so the cross-cloud `Database` interface is untouched (no Azure/GCP churn).
- Condition parsing/evaluation reuses `expr.ParseCondition` / `expr.Eval` — no duplicated parser.
- Exact AWS semantics preserved (verified against the official API docs): `ConditionalCheckFailedException` ("The conditional request failed" + optional `Item` for `ReturnValuesOnConditionCheckFailure=ALL_OLD`), and `TransactionCanceledException` with per-op `CancellationReasons` (`None` / `ConditionalCheckFailed`), all-or-nothing.

## Tests

- Provider `-race` + high-iteration concurrency tests asserting the **logical** invariant (winners == 1) that `-race` alone can't catch: concurrent put-if-absent, concurrent optimistic-lock update, transaction all-or-nothing, concurrent conflicting transactions.
- Real `aws-sdk-go-v2` reproductions over the wire: concurrent put-if-absent (exactly one 200, rest `ConditionalCheckFailedException`), optimistic-lock update, `TransactWriteItems` all-or-nothing with `CancellationReasons`, concurrent conflicting transactions.
- Existing DynamoDB tests stay green.

### DIFF
--- a/providers/aws/dynamodb/conditional.go
+++ b/providers/aws/dynamodb/conditional.go
@@ -1,0 +1,446 @@
+package dynamodb
+
+import (
+	"context"
+	"maps"
+	"strings"
+
+	cerrors "github.com/stackshy/cloudemu/v2/errors"
+	"github.com/stackshy/cloudemu/v2/services/database/driver"
+	"github.com/stackshy/cloudemu/v2/services/database/driver/expr"
+)
+
+// This file holds the ATOMIC conditional-write and transaction primitives. The
+// wire handler must NOT evaluate a ConditionExpression with a standalone GetItem
+// and then issue the mutation as a separate call — dropping the table lock in
+// between opens a TOCTOU window where two concurrent conditional writes on one
+// key both succeed. Each method here evaluates the condition and applies the
+// write under a SINGLE hold of m.mu, so no concurrent writer can interleave.
+//
+// PutItem/DeleteItem/UpdateItem delegate to these with an empty Condition, so the
+// unconditional and conditional paths share one implementation.
+
+// evalCondition parses (reusing expr.ParseCondition — no duplicated parser) and
+// evaluates a ConditionExpression against the current stored item. An empty
+// expression always passes. A missing item is evaluated against an empty item so
+// attribute_not_exists is true and attribute_exists is false, matching DynamoDB
+// create-if-absent semantics.
+func evalCondition(cond driver.Condition, existing map[string]any) (bool, error) {
+	if strings.TrimSpace(cond.Expression) == "" {
+		return true, nil
+	}
+
+	node, err := expr.ParseCondition(cond.Expression, cond.Names, cond.Values)
+	if err != nil {
+		return false, err
+	}
+
+	if existing == nil {
+		existing = map[string]any{}
+	}
+
+	return expr.Eval(node, existing)
+}
+
+// condExisting returns the item to evaluate a condition against: the stored item
+// when present, otherwise nil (evalCondition treats nil as an empty item).
+func condExisting(item map[string]any, present bool) map[string]any {
+	if !present {
+		return nil
+	}
+
+	return item
+}
+
+// oldImage returns a defensive copy of the stored item for return to the caller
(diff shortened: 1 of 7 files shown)


NitinKumar004:
Review — verified against the pushed commit (d655712c).
- build / vet / gofmt / gci clean; golangci-lint --new-from-rev 0 NEW (23 baseline in unchanged driver.go/cosmossql); coverage regen no-diff; COMPAT SUITE exit 0 (verified via exit code, no docs/compat drift); -race clean.
- Conditional-write atomicity CONFIRMED: PutItemConditional/UpdateItemConditional/DeleteItemConditional take m.mu.Lock() once, Get→checkConditionLocked→Set/Delete with NO unlock/relock; PutItem/UpdateItem/DeleteItem delegate with empty Condition (one shared atomic impl). winners==1 via reviewer's own 2000×32 put-if-absent + optimistic-lock (final version==bumps, no lost update); author's 300×8 also pass.
- Transaction atomicity: TransactWrite resolve→eval-ALL→plan-ALL→commit under one m.mu (two-phase: a compute error aborts nothing-written); 1000×16 concurrent put-if-absent txns → exactly 1 commit/round; multi-op with one failing condition → nothing written, CancellationReasons [ConditionalCheckFailed]; stream flush deferred outside the lock.
- Dead-code removal safe (gateCondition/checkCondition/evaluateTransactConditions/etc. gone, zero dangling refs; TransactWriteItems interface method still live for portable/Azure/GCP); semantics preserved (ConditionalCheckFailedException + #559 Item echo; CancellationReasons; reuses expr.ParseCondition/Eval, no dup parser); shared types off the Database interface (conditionalWriter capability), Azure cosmosdb/GCP firestore untouched; no collateral.
Nits (all fine): conditional Put/Update now validates keys/size before the condition (ValidationException before ConditionalCheckFailed — matches real DynamoDB better); writer() panics if not the AWS Mock (unreachable via real wiring); upsert via UpdateItemConditional emits MODIFY-shaped stream event (pre-existing, out of scope).
Verdict: approved.

PR #663, opened 2026-08-25
fix(aws-rds): honor DeletionProtection + KmsKeyId on CreateDBInstance; no phantom final snapshot on rejected delete

NitinKumar004:
## What

Fixes three RDS bugs found by the deep-e2e audit, all on the instance create/delete path. Fields now round-trip end-to-end (wire parse -> `InstanceConfig` -> `Instance`/`Cluster` -> Describe XML).

## Why / findings

- **[HIGH] `CreateDBInstance(DeletionProtection=true)` silently dropped the flag.** It read back `false` and the "protected" instance was still fully deletable — defeating the guard behind `aws_db_instance.deletion_protection = true` (real prod-safety bug + Terraform drift). The Modify path (#505) and cluster path already honored it; only instance-create dropped it. `InstanceConfig` had no `DeletionProtection` field, the form parser never read it, and `newInstance` never copied it. Now parsed/stored/emitted, and `DeleteDBInstance` rejects with `InvalidParameterCombination` until it's cleared via `ModifyDBInstance`.

- **[MED] `KmsKeyId` dropped on create and never emitted.** `StorageEncrypted=true` + `KmsKeyId=arn` read back empty → drift/forced-replace loops on encrypted DBs. Added `KmsKeyID` to `InstanceConfig`/`Instance` and `ClusterConfig`/`Cluster`, parse/store/emit `<KmsKeyId>` on Describe, and back-fill the account default (`alias/aws/rds`) when `StorageEncrypted` is set with no key — matching real RDS.

- **[MED] Phantom final snapshot on a rejected delete.** `DeleteDBInstance` created the final snapshot *before* the delete precondition (live read replicas) was enforced inside the driver, so a correctly-rejected delete left an orphan snapshot (and a retry then failed with `DBSnapshotAlreadyExists`). The final snapshot is now rolled back when `DeleteInstance` is rejected, so no snapshot survives a failed delete.

## Tests

New real `aws-sdk-go-v2` reproductions in `server/aws/rds`:
- `CreateDBInstance(DeletionProtection=true)` → Describe shows `true` AND Delete is rejected until `Modify(DeletionProtection=false)`.
- `CreateDBInstance(StorageEncrypted=true, KmsKeyId=arn)` → `KmsKeyId` round-trips on Describe; encrypted-with-no-key back-fills `alias/aws/rds`.
- `DeleteDBInstance` of a source with a live replica → rejected with `InvalidDBInstanceState` AND `DescribeDBSnapshots` is empty (no phantom final snapshot).

Existing RDS tests (#505 deletion-protection-via-Modify, #566 subnet group, #573 replicas, final-snapshot) stay green.

## Gates

`go build ./...`, `go vet`, `go test` (server/aws/rds + providers/aws/rds), `-race`, and `golangci-lint` on changed packages all clean; gofmt clean.

### DIFF
--- a/providers/aws/rds/rds.go
+++ b/providers/aws/rds/rds.go
@@ -48,6 +48,9 @@ const (
 	defaultEngineMode        = "provisioned"
 	auroraAllocatedStorage   = 1
 	resourceIDLen            = 20
+	// defaultKMSKeyAlias is the account-default RDS KMS key real AWS fills in
+	// when StorageEncrypted is set but no KmsKeyId is supplied.
+	defaultKMSKeyAlias = "alias/aws/rds"
 )
 
 // Metric namespaces for engines that share the RDS wire protocol but emit
@@ -467,9 +470,26 @@ func (m *Mock) newInstance(cfg rdsdriver.InstanceConfig) rdsdriver.Instance {
 		PreferredMaintenanceWindow: defaultMaintenanceWindow,
 		CACertificateIdentifier:    defaultCACertIdentifier,
 		StorageEncrypted:           cfg.StorageEncrypted,
+		KmsKeyID:                   resolveKMSKeyID(cfg.StorageEncrypted, cfg.KmsKeyID),
+		DeletionProtection:         cfg.DeletionProtection,
 		CreatedAt:                  m.opts.Clock.Now().UTC(),
 		Tags:                       copyTags(cfg.Tags),
 	}
+}
+
+// resolveKMSKeyID mirrors real RDS: when storage is encrypted and the caller
+// supplied no KmsKeyId, the account's default RDS key is filled in; an explicit
+// key always wins, and an unencrypted resource carries no key.
+func resolveKMSKeyID(encrypted bool, kmsKeyID string) string {
+	if !encrypted {
+		return ""
+	}
+
+	if kmsKeyID == "" {
+		return defaultKMSKeyAlias
+	}
+
+	return kmsKeyID
 }
 
 // planProvision snapshots, under the caller's lock, what runCreateProvision needs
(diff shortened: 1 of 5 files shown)


NitinKumar004 (commented):
Verified against the pushed head (776d4fb0). Clean build of the actual tree; gates green.

- go build ./... — clean
- go vet (server/aws/rds, providers/aws/rds, services/relationaldb) — clean
- go test ./server/aws/rds/... ./providers/aws/rds/... -count=1 — pass
- go test -race ./providers/aws/rds/... — pass
- gofmt -l — clean
- golangci-lint (changed pkgs) — only the two pre-existing dupl hits in parameter_group.go (87↔280), untouched by this PR; no new issues
- go generate ./... — zero diff, so no coverage regen owed (struct fields only, no interface signature change)

All three fixes hold up under independent aws-sdk-go-v2 reproductions beyond the PR's own:

DeletionProtection on create
- create true → Describe true; DeleteDBInstance rejected with InvalidParameterCombination; Modify(false) then delete succeeds.
- Regression clear: unset → false and still deletable. The DP check sits before snapshot creation, so a protected delete also leaves no phantom snapshot.

KmsKeyId round-trip
- explicit key round-trips exactly on both instance and cluster paths.
- StorageEncrypted with no key back-fills alias/aws/rds (instance and cluster).
- Unencrypted → no KmsKeyId element on the wire (omitempty), not the default — verified for instance and cluster.
- The Go field KmsKeyID with tag xml:"KmsKeyId" is unmarshaled correctly by the SDK (confirmed by round-trip, not just by reading the tag).

Phantom snapshot on rejected delete
- delete of a source with a live read replica → InvalidDBInstanceState AND DescribeDBSnapshots empty (rollback fired).
- Happy path intact: SkipFinalSnapshot=false + successful delete → final snapshot survives (rollback does not fire on success).
- SkipFinalSnapshot=true → no snapshot, delete succeeds.
- Retry after a rejected delete (replica removed, same FinalDBSnapshotIdentifier reused) succeeds — no DBSnapshotAlreadyExists, proving the rollback actually removed the snapshot.
- Rollback ordering is safe: finalID is captured before DeleteInstance; all state transitions run under the driver mutex, and a concurrent second delete fails at CreateSnapshot (duplicate id) before reaching DeleteInstance, so there is no double-rollback path. Best-effort DeleteSnapshot on the failure branch is appropriate.

Blast radius checked: persist/export handles compute instances only (not RDS), so the new InstanceConfig/Instance/Cluster fields need no wiring there; no other consumer requires them.

Verdict: LGTM.

PR #664, opened 2026-08-25
fix(aws-sns): FIFO topic→FIFO SQS fanout carries MessageGroupId (no silent drop); numeric filter policy matches string attributes; malformed receipt handle

NitinKumar004:
Fixes three real-user AWS SNS/SQS bugs found by a deep end-to-end audit (real aws-sdk-go-v2 clients over the wire server). Two are silent message loss.

## F1 — [HIGH/BLOCKER] SNS FIFO topic → FIFO SQS fanout delivered ZERO messages
Publishing to a FIFO SNS topic subscribed by a FIFO SQS queue delivered nothing, with no error to the caller. Two compounding gaps:
- The publish wire dropped `MessageGroupId` / `MessageDeduplicationId`, and `PublishInput` had no fields for them.
- Fanout called SQS `SendMessage` with no group id, which a FIFO queue correctly rejects — and the error was swallowed (`_ = DeliverExternal(...)`).

Fix (3-layer): add `MessageGroupID`/`MessageDeduplicationID` to `PublishInput`; parse them in `publish` and `publishBatch`; thread them through a new group-aware `DeliverExternalFIFO` on the SQS backend so the FIFO queue accepts the send; and record each delivery as an SNS metric (`NumberOfNotificationsDelivered` / `NumberOfNotificationsFailed`) so a real misconfiguration is observable instead of a silent drop. The group id is carried onto the delivered SQS message.

## F2 — [HIGH] Numeric FilterPolicy on a message ATTRIBUTE dropped all matches
A policy like `{"price":[{"numeric":[">",100]}]}` never matched an attribute, so subscribers silently lost every message. Attribute values travel the wire as strings, and the matcher's `toFloat` did not parse strings.

Fix: add `eventmatch.MatchLeafAttr`, an attribute-scope numeric path that parses a numeric string, while `MatchLeaf` keeps the stricter body/EventBridge rule (a JSON string never satisfies a numeric operator). MessageBody-scope numeric behavior is unchanged; non-numeric strings still don't match numeric conditions.

## N1 — [LOW] DeleteMessage with a malformed receipt handle returned 200
AWS returns `ReceiptHandleIsInvalid` for a syntactically-invalid handle. `DeleteMessage` now rejects a malformed handle with that code while keeping #515's idempotent 200 for a stale but well-formed handle.

## Tests
Real aws-sdk-go-v2 reproductions: FIFO topic→FIFO queue delivers all messages and carries the group id; numeric attribute filter delivers matching / drops non-matching; body-scope numeric filter unchanged; malformed vs stale receipt handle on DeleteMessage; and an eventmatch unit test pinning the attribute-vs-body numeric-string distinction.

### DIFF
--- a/internal/eventmatch/eventmatch.go
+++ b/internal/eventmatch/eventmatch.go
@@ -10,6 +10,7 @@ import (
 	"encoding/json"
 	"math"
 	"net"
+	"strconv"
 	"strings"
 )
 
@@ -131,22 +132,35 @@ func isReservedOperator(name string) bool {
 // at least one entry of a leaf constraint array. When the event value is itself
 // an array, the constraint matches if any element matches.
 func MatchLeaf(allowed []any, value any, present bool) bool {
+	return matchLeaf(allowed, value, present, false)
+}
+
+// MatchLeafAttr is MatchLeaf for SNS message-attribute matching, where values
+// arrive as strings even when their DataType is Number. It lets a numeric
+// operator parse a numeric string (so a "150" attribute satisfies {"numeric":
+// [">",100]}), while body/EventBridge matching via MatchLeaf keeps the stricter
+// rule that a JSON string never satisfies a numeric operator.
+func MatchLeafAttr(allowed []any, value any, present bool) bool {
+	return matchLeaf(allowed, value, present, true)
+}
+
+func matchLeaf(allowed []any, value any, present, coerceNum bool) bool {
 	// When the event value is an array, the leaf matches if any element does;
 	// otherwise fall through so exists-style operators can evaluate presence.
 	if arr, ok := value.([]any); ok && present {
 		for _, el := range arr {
-			if matchAnyEntry(allowed, el, true) {
+			if matchAnyEntry(allowed, el, true, coerceNum) {
 				return true
 			}
 		}
 	}
 
-	return matchAnyEntry(allowed, value, present)
+	return matchAnyEntry(allowed, value, present, coerceNum)
 }
 
-func matchAnyEntry(allowed []any, value any, present bool) bool {
+func matchAnyEntry(allowed []any, value any, present, coerceNum bool) bool {
 	for _, a := range allowed {
-		if matchEntry(a, value, present) {
+		if matchEntry(a, value, present, coerceNum) {
 			return true
 		}
 	}
@@ -154,10 +168,10 @@ func matchAnyEntry(allowed []any, value any, present bool) bool {
 	return false
 }
 
-func matchEntry(allowed, value any, present bool) bool {
+func matchEntry(allowed, value any, present, coerceNum bool) bool {
 	switch a := allowed.(type) {
 	case map[string]any:
-		return matchOperator(a, value, present)
+		return matchOperator(a, value, present, coerceNum)
 	case nil:
 		return present && value == nil
 	default:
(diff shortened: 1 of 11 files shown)


NitinKumar004 (commented):
**Verdict: MERGE** — verified against pushed head `5f7d3ea` (clean checkout, not the author report).

Gates on the pushed tree: `go build ./...` ok · `go vet` (changed pkgs) ok · `go test` sns/sqs (server+provider) + eventmatch ok · `-race` sns/sqs providers ok · `gofmt -l` clean · `go generate ./...` → zero diff (coverage idempotent). CI golangci-lint is report-only (advisory); `--new-from-rev` against merge-base reports 0 new issues on changed lines.

**F1 — FIFO topic→FIFO SQS fanout.** Confirmed with independent aws-sdk-go-v2 tests:
- FIFO topic→FIFO queue delivers; `MessageGroupId` carries onto the SQS message; same `MessageDeduplicationId` within the window collapses to a single delivery.
- No regression on the standard→standard path (`DeliverExternal` delegating to `DeliverExternalFIFO` with empty ids still delivers).
- Mixed combos are sane: FIFO topic→standard queue delivers; standard topic→FIFO queue keeps `Publish` successful and drops only the downstream delivery (recorded as `NumberOfNotificationsFailed`), no panic/no silent publish failure.
- Partial-fanout failure is isolated: a bad subscription (endpoint ARN with no backing queue) does not block a healthy sibling and does not fail `Publish`. Delivery errors are surfaced as SNS metrics, not returned to the caller — matches real SNS (Publish succeeds even when a downstream delivery fails). Emitting `NumberOfNotificationsDelivered/Failed` per subscription is also a correctness improvement over the previous swallow.

**F2 — numeric filter on attributes.** `coerceNum` is scoped to `MatchLeafAttr` (attribute path) only; `MatchLeaf` (body/EventBridge) is unchanged — existing filter + EventBridge tests green, and a JSON string still never satisfies a numeric operator in body scope. Number attributes now match; non-numeric string ("abc") correctly does not; below-threshold gated.
Two edge deviations from real AWS, both non-blocking and net-positive vs the prior total drop:
1. Because `parseMessageAttributes` reduces attributes to `map[string]string` (DataType is dropped at the wire), a **String-typed** attribute holding a numeric-looking value also satisfies a numeric filter, whereas real SNS keys numeric matching on `DataType == Number`. Proper fix needs DataType threaded through the driver — larger than this PR.
2. `strconv.ParseFloat` over-accepts a few tokens the AWS matcher would not treat as numbers: `"Inf"`, `"Infinity"`, `"1e3"` match `>100` (`"NaN"` does not, comparisons are false). Rare in practice.

**N1 — malformed receipt handle.** Charset-based `isWellFormedReceiptHandle` (not a prefix check) rejects handles with disallowed chars (space/control/punct/non-ASCII) as `ReceiptHandleIsInvalid` while keeping the well-formed-stale handle a 200 idempotent no-op (#515 preserved). Emulator-issued `receipt-…` handles round-trip (send/receive/delete) fine; base64/base64url tokens are accepted. No false accept/reject found on real-shaped handles.

**Blast radius.** EventBridge and S3 keep their own single-method `SQSDeliverer` interfaces (only `DeliverExternal`), so the new `DeliverExternalFIFO` is confined to the SNS→SQS wiring; all implementers compile; no `coerceNum` leak into EventBridge. Coverage regen committed/idempotent.

Minor (non-blocking): the `PublishInput` growth pushes `services/notification.Publish` over gocritic's hugeParam threshold (88 bytes); the author added `//nolint:gocritic` on the provider `Mock.Publish` but not the portable one — inconsistent, and lint is advisory anyway.

PR #665, opened 2026-08-25
fix(aws-vpc): ENI SourceDestCheck/defaults, route+subnet validation, exact error codes

NitinKumar004:
## What

VPC batch-2 fixes from the deep-e2e audit (`aws-e2e/vpc.md`), covering ENI attributes, route/subnet validation, and typed error-code mismatches. Batch-1's IGW/NACL invariants (#662) are untouched.

### ENI (finding #14, #15, #11)
- `CreateNetworkInterface` now auto-assigns a subnet-scoped private IPv4 + MAC address and defaults `SourceDestCheck` to `true` (real EC2 behavior; `aws_network_interface` reads all three). Emitted in the ENI XML and round-tripped through Describe.
- Added `ModifyNetworkInterfaceAttribute` (wire dispatch + `NetworkInterfaceModifier` driver interface + provider impl) for `SourceDestCheck` / `Description` / `Groups` — the required step for NAT-instance / firewall / router VMs.
- Delete-while-attached now maps to `InvalidNetworkInterface.InUse` (was `DependencyViolation`); delete-missing stays `InvalidNetworkInterfaceID.NotFound`.

### Route / subnet validation (finding #12, #13)
- `CreateRoute` rejects a nonexistent gateway / NAT / peering target instead of silently storing a dangling route (`InvalidInternetGatewayID.NotFound` / `InvalidNatGatewayID.NotFound` / `InvalidVpcPeeringConnectionID.NotFound`).
- `CreateSubnet` rejects a CIDR outside the VPC's CIDR block with `InvalidSubnet.Range`.

### Error-code mismatches (findings #4, #5, #8, #9, #10)
- `DeleteRoute` of a missing route on an existing table -> `InvalidRoute.NotFound` (was `InvalidRouteTableID.NotFound`).
- Duplicate `CreateRoute` -> `RouteAlreadyExists` (was `ResourceAlreadyExists`).
- `CreateNatGateway` in a missing subnet -> `InvalidSubnetID.NotFound` (was `NatGatewayNotFound`).
- `ReleaseAddress` with a bad allocation id -> `InvalidAllocationID.NotFound` (was `InvalidVpcID.NotFound`).
- Accept/Reject a non-pending peering -> `InvalidStateTransition` (was `DependencyViolation`).

## Why

Every mismatch was reproduced through the real `aws-sdk-go-v2` EC2 client and each expected code verified against the official EC2 API error reference. Silent successes mask real IaC misconfig; wrong typed codes break SDK/Terraform error handling (e.g. an EIP-release naming an unrelated VPC resource type).

## Where (3-layer)

- Real validation / state invariants: `providers/aws/vpc/*` (ENI defaults + modify, CreateRoute target check, subnet-range check).
- Error-code mapping: `server/aws/ec2/*` `write*Err` helpers, matching the batch-1 marker/message-based pattern.
- Driver interface change regenerated `docs/coverage/` (idempotent).

## Tests

Real `aws-sdk-go-v2` reproductions for each fix (ENI defaults + SourceDestCheck modify/read, delete missing/attached codes; CreateRoute bad gateway/NAT/peering, duplicate route, delete-missing route, subnet out-of-range; release-bad-alloc, NAT-bad-subnet, peering-state) plus provider-level tests. Existing VPC/EC2 suites (incl. #662 IGW/NACL) stay green; a few tests using placeholder `igw-*` route targets were updated to attach a real gateway.

### DIFF
--- a/cloudemu_test.go
+++ b/cloudemu_test.go
@@ -4811,8 +4811,17 @@ func TestRouteTables(t *testing.T) {
 				t.Errorf("expected VPC %s, got %s", vpcInfo.ID, rt.VPCID)
 			}
 
-			// Create route
-			err = p.d.CreateRoute(ctx, rt.ID, "0.0.0.0/0", "igw-12345", "gateway")
+			// Create route. The AWS provider validates that the gateway target
+			// exists, so attach a real internet gateway to point the route at.
+			igw, err := p.d.CreateInternetGateway(ctx, netdriver.InternetGatewayConfig{})
+			if err != nil {
+				t.Fatalf("CreateInternetGateway: %v", err)
+			}
+			if err = p.d.AttachInternetGateway(ctx, igw.ID, vpcInfo.ID); err != nil {
+				t.Fatalf("AttachInternetGateway: %v", err)
+			}
+
+			err = p.d.CreateRoute(ctx, rt.ID, "0.0.0.0/0", igw.ID, "gateway")
 			if err != nil {
 				t.Fatalf("CreateRoute: %v", err)
 			}
--- a/docs/coverage/aws/vpc.md
+++ b/docs/coverage/aws/vpc.md
@@ -317,6 +317,14 @@ NetworkInterfaceCreator is the AWS-specific ENI-creation surface. It's kept
 | --- | --- |
 | `CreateNetworkInterface` |  |
 
+### NetworkInterfaceModifier
+
+NetworkInterfaceModifier is the AWS-specific ENI attribute-modify surface
+
+| Operation | Description |
+| --- | --- |
+| `ModifyNetworkInterfaceAttribute` |  |
+
 ### NetworkInterfaces
 
 NetworkInterfaces is an OPTIONAL capability, discovered by type assertion.
--- a/docs/coverage/azure/vnet.md
+++ b/docs/coverage/azure/vnet.md
@@ -317,6 +317,14 @@ NetworkInterfaceCreator is the AWS-specific ENI-creation surface. It's kept
 | --- | --- |
 | `CreateNetworkInterface` |  |
 
+### NetworkInterfaceModifier
+
+NetworkInterfaceModifier is the AWS-specific ENI attribute-modify surface
+
+| Operation | Description |
+| --- | --- |
+| `ModifyNetworkInterfaceAttribute` |  |
+
 ### NetworkInterfaces
 
 NetworkInterfaces is an OPTIONAL capability, discovered by type assertion.
--- a/docs/coverage/coverage.json
+++ b/docs/coverage/coverage.json
@@ -7036,6 +7036,15 @@
         ]
       },
       {
+        "name": "NetworkInterfaceModifier",
+        "doc": "NetworkInterfaceModifier is the AWS-specific ENI attribute-modify surface",
+        "operations": [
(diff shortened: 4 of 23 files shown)


NitinKumar004 (commented):
Reviewed against the pushed head (63eb19d2), clean-built and exercised with an independent aws-sdk-go-v2 EC2 test suite beyond the PR's own. Verdict: looks good to merge.

Gates (GOMAXPROCS=3): build ./... clean; vet clean; targeted tests (server/aws/ec2, providers/aws/vpc, services/networking, features/topology) pass; -race on providers/aws/vpc and on a concurrent ModifyNetworkInterfaceAttribute clean; root TestVPC*/TestRouteTables/TestNATGateway/TestVPCPeering pass; gofmt clean; `go generate ./...` yields zero further diff (coverage regen idempotent); golangci-lint --new-from-rev = 0 new issues (the 19 reported are all pre-existing in untouched files).

Verified behaviors:
- ModifyNetworkInterfaceAttribute round-trips SourceDestCheck true->false->true, Description, and Groups; a modify on a missing ENI returns InvalidNetworkInterfaceID.NotFound; concurrent modifies are race-clean.
- CreateNetworkInterface defaults: private IPv4 inside the subnet CIDR, well-formed MAC, SourceDestCheck=true; 10 ENIs in one subnet get distinct, in-CIDR addresses (reserved-.0-.3 offset honored).
- CreateRoute target validation: bad igw-/nat-/pcx- map to the specific *.NotFound code; a real attached IGW target works; vgw-/eigw-/tgw- pass through unvalidated (not wrongly rejected); duplicate destination -> RouteAlreadyExists; DeleteRoute miss -> InvalidRoute.NotFound.
- CreateSubnet: out-of-VPC-range CIDR -> InvalidSubnet.Range; in-range and subnet-CIDR==VPC-CIDR boundary both succeed.
- Peering accept/reject on a non-pending connection -> InvalidStateTransition; ReleaseAddress bad alloc -> InvalidAllocationID.NotFound; CreateNatGateway in a missing subnet -> InvalidSubnetID.NotFound.

Error codes were cross-checked against the official EC2 API error reference. All match, including InvalidStateTransition ("the specified VPC peering connection is not in the correct state... accept a request that was rejected"), which is the correct code here rather than IncorrectState/InvalidState.

Test-suite edits (race_test, vpc_test, topology_test, cloudemu_test) swap placeholder igw- strings for a real attached gateway now that CreateRoute validates targets; no assertion is weakened (TraceRoute still asserts the real gateway id) and one case is strengthened (`_ =` -> requireNoError). Topology engine still resolves routes.

Non-blocking observations (no change requested):
1. CreateRoute checks the duplicate-CIDR precondition before target existence, so a request that is both a duplicate and points at a nonexistent target returns RouteAlreadyExists rather than the target code. Acceptable ordering.
2. Standalone-ENI private IPs and RunInstances instance IPs are allocated from independent stores, so a standalone ENI and an instance in the same subnet could in principle share an address; it is not observable through DescribeNetworkInterfaces (instance ENIs are not in that store). Pre-existing gap, not introduced here.
3. Coverage docs list the AWS-specific NetworkInterfaceModifier capability under the azure/gcp/oci networking pages as well; this is the existing generator convention for shared-package optional interfaces and is idempotent.

Issue #319, opened 2026-08-04
AWS parity E2E audit: bugs, gaps & enhancements across all 30 services


thzgajendra:
# AWS parity E2E audit — bugs & enhancements (issue draft)

**Method.** Ran the standalone binary `cloudemu serve --providers aws --aws-port 4566` and drove **real `aws-sdk-go-v2` clients** (`BaseEndpoint=http://127.0.0.1:4566`) through dependency-ordered lifecycles per resource family — create → describe (with field assertions) → modify/associate → list → tag → sub-resources → delete → error paths. Every non-pass was **hand-verified** to separate real defects from bad test inputs.

**Coverage.** **All 30 AWS service handlers** exercised E2E (~250 operation calls): ec2, s3, iam, sts, dynamodb, lambda, rds, redshift, sqs, sns, eventbridge, secretsmanager, ssm, cloudwatch, cloudwatchlogs, ecr, ecs, eks, elbv2, elasticache, memorydb, keyspaces, route53, networkfirewall, resourcegroupstaggingapi, sagemaker, bedrock, bedrockagent, bedrockagentruntime, resourceexplorer2. Depth = representative dependency-ordered lifecycle per service (not literally every one of each service's SDK operations; EC2 additionally got a full 150-action reachability sweep). Runtime data-planes (model inference) not exercised.
**Overall the emulator is strong** — the vast majority of core CRUD + error paths pass, and it correctly enforces dependency violations, resource state, and CIDR overlaps.

### Parity coverage vs full AWS API (implemented ops ÷ total SDK ops)
High-level breadth (denominator = the *entire* AWS API for the service, incl. many niche ops an emulator need not cover, so low % is expected for large services):

| Service | impl/SDK | Service | impl/SDK | Service | impl/SDK |
|---|---|---|---|---|---|
| ec2 | 150/773 | rds | 79/168 | sagemaker | 113/403 |
| iam | 39/176 | ecs | 37/77 | memorydb | 33/45 |
| keyspaces | 18/19 | elbv2 | 18/51 | dynamodb | 15/57 |
| eventbridge | 14/57 | cloudwatch | 10/46 | elasticache | 10/75 |
| sns | 9/42 | redshift | 9/141 | cloudwatchlogs | 9/118 |
| ssm | 11/152 | sqs | 7/23 | secretsmanager | 7/23 |
| ecr | 7/58 | networkfirewall | 20/84 | resourcegroupstaggingapi | 9/9 |
[...]

PR #666, opened 2026-08-26
fix(aws-cloudwatch): exact dimension matching, SetAlarmState history+actions, M-of-N alarm eval + OK recovery, alarm field round-trip, DeleteAlarms + history semantics

NitinKumar004:
## Summary

Fixes six real AWS CloudWatch correctness bugs found by a deep real-SDK e2e audit
(two HIGH). Each is fixed at the correct three-layer location and cross-checked
against the official CloudWatch API docs.

## Findings fixed

1. **[HIGH] Exact dimension matching.** CloudWatch treats each unique dimension
   combination as a separate metric, so a query with fewer/no dimensions must not
   return datapoints published under a superset. `matchDimensions` now requires
   exact set equality, for both metric reads and alarm evaluation (an alarm reads
   only its own dimensioned stream).

2. **[HIGH] SetAlarmState records history and fires actions.** Manually setting a
   state now goes through a shared `transitionAlarm` helper that, on a state
   change, appends a `StateUpdate` history entry and invokes the new state's SNS
   actions — matching AWS, where SetAlarmState behaves like an evaluated
   transition. Previously history was empty and no notification fired.

3. **[MED] Per-period M-of-N alarm evaluation with OK recovery.** Evaluation used
   to fold the whole window into one aggregate, so a de-noised alarm
   (`EvaluationPeriods > 1`) fired on a single breaching period and never
   recovered. It now buckets the window into per-`Period` datapoints, evaluates
   the statistic per bucket, and alarms when `DatapointsToAlarm` of the last
   `EvaluationPeriods` breach (default M=N), returning to OK otherwise. Missing
   periods are counted per `TreatMissingData` (breaching / notBreaching / default
   missing).

4. **[MED] Alarm field round-trip.** `DatapointsToAlarm`, `TreatMissingData`,
   `Unit` and `ExtendedStatistic` are now threaded through the driver, the CBOR
   and query wire structs, and the provider so PutMetricAlarm persists them and
   DescribeAlarms reflects them (fixes Terraform perpetual drift).

5. **[MED] DeleteAlarms tolerates unknown names.** AWS deletes the correctly named
   alarms and ignores incorrect names without error. Both wire paths aborted on
   the first missing name (order-dependent partial delete); they now skip
   not-found names and return success.

6. **[MED] DescribeAlarmHistory ordering + filters.** History was oldest-first and
   dropped the newest on limit; `HistoryItemType`, `StartDate/EndDate` and
   `ScanBy` were ignored. `GetAlarmHistory` now returns newest-first and keeps the
   newest N; the wire layer honors HistoryItemType, the date window and ScanBy,
   truncates to MaxRecords last, and echoes each entry's real HistoryItemType (new
   `AlarmHistoryEntry.HistoryItemType`, defaulting to StateUpdate).

Out of scope (ticketed separately): Dashboards, composite alarms,
GetMetricWidgetImage, and high-res `StorageResolution=1`.

## Tests

Real `aws-sdk-go-v2` reproductions for each fix: exact-dimension exclusion,
per-period M-of-N with OK recovery (FakeClock-backed for determinism), alarm
field round-trip, DeleteAlarms tolerating an unknown name, SetAlarmState
recording history and firing the SNS AlarmAction to a subscribed SQS queue, and
DescribeAlarmHistory newest-first with MaxRecords/ScanBy/HistoryItemType filters.
Existing CloudWatch tests and the #568 alarm/metric-filter e2e stay green.

### DIFF
--- a/cloudemu_test.go
+++ b/cloudemu_test.go
@@ -1170,9 +1170,21 @@ func TestAlarmTriggeredByAutoMetrics(t *testing.T) {
 	clock := config.NewFakeClock(time.Date(2024, 1, 1, 12, 0, 0, 0, time.UTC))
 	p := NewAWS(config.WithClock(clock))
 
-	// Create alarm: CPU > 0 (any CPU should trigger since auto-metrics emit 25.0)
+	// Launch the instance first so we know its id; EC2 auto-metrics are dimensioned
+	// by InstanceId, and CloudWatch matches an alarm to a metric series by its exact
+	// dimension set — so a real user alarms on that instance's InstanceId.
+	instances, err := p.EC2.RunInstances(ctx, computedriver.InstanceConfig{
+		ImageID: "ami-test", InstanceType: "t2.micro",
+	}, 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+	id := instances[0].ID
+
+	// Create alarm: CPU > 0 on that instance (auto-metrics emit 25.0).
 	if err := p.CloudWatch.CreateAlarm(ctx, mondriver.AlarmConfig{
 		Name: "any-cpu", Namespace: "AWS/EC2", MetricName: "CPUUtilization",
+		Dimensions:         map[string]string{"InstanceId": id},
 		ComparisonOperator: "GreaterThanThreshold", Threshold: 0,
 		Period: 600, EvaluationPeriods: 1, Stat: "Average",
 	}); err != nil {
@@ -1185,11 +1197,9 @@ func TestAlarmTriggeredByAutoMetrics(t *testing.T) {
 		t.Errorf("expected INSUFFICIENT_DATA, got %s", alarms[0].State)
 	}
 
-	// Launch instance — auto-metrics should trigger alarm evaluation
-	_, err := p.EC2.RunInstances(ctx, computedriver.InstanceConfig{
-		ImageID: "ami-test", InstanceType: "t2.micro",
-	}, 1)
-	if err != nil {
+	// A lifecycle op re-emits the instance's running auto-metrics (CPU=25),
+	// triggering alarm evaluation the way the next datapoint would in real AWS.
+	if err := p.EC2.RebootInstances(ctx, []string{id}); err != nil {
 		t.Fatal(err)
 	}
 
@@ -1205,16 +1215,8 @@ func TestLifecycleStopEmitsZeroMetrics(t *testing.T) {
 	clock := config.NewFakeClock(time.Date(2024, 1, 1, 12, 0, 0, 0, time.UTC))
 	p := NewAWS(config.WithClock(clock))
 
-	// Create alarm: CPU < 1 (LessThanThreshold)
-	if err := p.CloudWatch.CreateAlarm(ctx, mondriver.AlarmConfig{
-		Name: "low-cpu", Namespace: "AWS/EC2", MetricName: "CPUUtilization",
-		ComparisonOperator: "LessThanThreshold", Threshold: 1,
-		Period: 300, EvaluationPeriods: 1, Stat: "Average",
-	}); err != nil {
-		t.Fatal(err)
-	}
-
-	// RunInstances → CPU=25 → alarm stays OK (25 is not < 1)
+	// Launch first to learn the instance id; the alarm carries that InstanceId
+	// dimension so exact-match evaluation reads this instance's metric series.
 	instances, err := p.EC2.RunInstances(ctx, computedriver.InstanceConfig{
 		ImageID: "ami-test", InstanceType: "t2.micro",
 	}, 1)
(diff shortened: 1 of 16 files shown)


NitinKumar004 (commented):
Reviewed the pushed head (2335c7e). Build, vet, `-race`, `go generate` (no diff), gofmt and the new SDK tests all pass, and the M-of-N rewrite itself is correct — I verified 2-of-3, strict `GreaterThanThreshold` boundary (=threshold does not breach), and `TreatMissingData` breaching/notBreaching independently. But two things block merge.

**1. The pushed tree fails `go test .` (root package) — 4 pre-existing tests now break.**

- `TestAlarmTriggeredByAutoMetrics`, `TestLifecycleStopEmitsZeroMetrics`, `TestLifecycleStartEmitsRunningMetrics`: these create an alarm on `AWS/EC2 CPUUtilization` with **no dimensions**, then rely on RunInstances/Start/Stop auto-metrics (emitted with `{InstanceId: ...}`, ec2.go:340/364) to drive the alarm. With exact-set `matchDimensions`, a no-dimension alarm no longer matches the `InstanceId`-dimensioned stream, so the alarm never evaluates and stays `INSUFFICIENT_DATA`. The new semantics are arguably more AWS-correct, but this is the documented "Auto-Metric Generation drives alarms" behavior and the tests were not updated. Either give these alarms the `InstanceId` dimension or reconcile the intended behavior — but the tree cannot ship red.
- `TestAlarmActionsAWS`: fails purely on ordering — the helper asserts history oldest-first (`history[0] = INSUFFICIENT_DATA->OK`) and now gets newest-first.

**2. Cross-provider divergence (CLAUDE.md: behaviors must be mirrored across AWS/Azure/GCP).**

The `GetAlarmHistory` newest-first change and the exact `matchDimensions` change were made only in `providers/aws/cloudwatch`. `providers/azure/monitor` (monitor.go:455) and `providers/gcp/monitoring` still return oldest-first and still use loose subset dimension matching. Because `testAlarmActions` is a shared helper across all three providers, AWS is now the odd one out — which is exactly why `TestAlarmActionsAWS` fails while the Azure/GCP variants pass. The M-of-N eval, `TreatMissingData`, `DatapointsToAlarm` and SetAlarmState-fires-actions are likewise AWS-only. Either mirror to Azure/GCP + update the shared test, or move the newest-first reordering into the AWS wire layer and keep the portable driver contract stable.

**Non-blocking notes**
- `ExtendedStatistic` (e.g. `p95`) is now stored and round-tripped, but evaluation uses `alarm.Stat` (empty for a percentile alarm), so a `p95` alarm silently evaluates as `Average`; the buckets only retain sum/count/min/max so a true percentile isn't computable without keeping raw values. Fine as a documented approximation, but worth a comment so it isn't mistaken for real percentile evaluation.
- `bucketByPeriod` anchors buckets to `now` rather than epoch-aligning period boundaries as CloudWatch does. Acceptable for the emulator; noting for fidelity.
- SetAlarmState firing only on an actual state change is correct and matches the API docs ("When the updated state differs from the previous value, the action ... is invoked") — no change needed there.

NitinKumar004 (commented):
**Verdict: approve.** Re-reviewed at head `6b4b77c0` — clean-built the actual pushed tree, and both prior CHANGES items are resolved.

**Root suite (was red).** `go test .` green. The 3 EC2 auto-metric tests now exercise the real feature rather than passing trivially: each launches the instance, creates the alarm on that instance's `InstanceId` dimension, and drives evaluation with a lifecycle op that re-emits dimensioned running/zero metrics — so a broken exact-dim match would fail them (the old no-dimension alarm would have matched a superset). Final ALARM/OK assertions are intact, not gutted. `testAlarmActions` now asserts newest-first history for all three providers.

**Shared evaluator.** `services/monitoring/alarmeval` is provider-neutral — it imports only `services/monitoring/driver` + stdlib, so no import cycle (providers → alarmeval → driver). Verified `EvaluateWindow` adversarially beyond the in-tree tests: 2-of-3 M-of-N, strict-GT boundary (value==threshold not a breach under `GreaterThanThreshold`, breaches under GTE), `notBreaching`/`ignore`/default-`missing`/`breaching` policies, future-datum (age<0) and stale-datum exclusion, `evaluated=false` when all periods are missing under a non-breaching policy, and per-bucket stat folding (Sum vs Average). All correct.

**AWS behavior preserved.** All six SDK fidelity tests pass over aws-sdk-go-v2 after the refactor: exact-dim matching, `SetAlarmState` history + SNS action (SNS→SQS delivery e2e green), per-Period M-of-N + OK recovery, field round-trip (`DatapointsToAlarm`/`Unit`/`ExtendedStatistic`/`TreatMissingData` through CBOR + query wire structs), `DeleteAlarms` tolerating unknown names on both wire paths, and newest-first `DescribeAlarmHistory` with `HistoryItemType`/date-window/`ScanBy` honored (MaxRecords applied after the ScanBy reversal — matches AWS first-page semantics).

**Azure / GCP parity.** Both mirror the shared evaluator with identical `alarmParams` projection. Azure keeps its subscription-agnostic `resourceId` compare while now enforcing exact dimension-set equality; GCP uses plain-label exact matching. Fidelity tests (exact-dim, M-of-N + recovery, newest-first + limit, SetAlarmState transition) green on both.

**Concurrency.** `-race` clean on all three providers. The refactor preserves the invariant that `evaluateAlarms` runs after the metrics write lock is released, and `transitionAlarm`→`fireAlarmActions` holds no mock lock while publishing to SNS.

**Lint.** No new debt in the changed packages. The 3 issues golangci-lint reports in `server/aws/cloudwatch/query.go` (lines 46/79/86) predate this PR (last touched in #319) and lie outside its hunks; the two monitor provider packages are clean. `gofmt` clean, `go vet` clean, `go generate ./...` zero diff.

Advisory (pre-existing, not introduced here, non-blocking): `*alarmData` state fields are mutated in `evaluateSingleAlarm`/`transitionAlarm` and read in `DescribeAlarms`/`toAlarmInfo` without holding `m.mu`, so truly concurrent `PutMetricData`+`DescribeAlarms` on the same alarm is a latent race the current tests don't surface. Worth folding those field accesses under the mock lock in a later pass.

PR #667, opened 2026-08-26
fix(aws-kinesis): honor ListStreams ExclusiveStartStreamName + ListShards ShardFilter

NitinKumar004:
## What

Two MED bugs found by the Kinesis deep-e2e real-user audit, plus two low-risk nits — all fixed at the correct 3-layer location (wire → driver → provider).

**F1 [MED] ListStreams ignored `ExclusiveStartStreamName`.** The legacy/CLI pagination cursor was silently dropped, so callers paginating with `ExclusiveStartStreamName` re-read from the top every call (duplicate results / non-terminating loop). Now threaded through wire→driver→provider: results resume strictly *after* the named stream (lexicographic order). The modern opaque `NextToken` paginator still works and takes precedence — both coexist.

**F2 [MED] ListShards ignored `ShardFilter`.** KCL 2.x uses `ShardFilter{Type: AT_LATEST}` to discover the live shard set; we returned *all* shards (including CLOSED parents after a reshard), so KCL could lease already-closed shards. Now honored end-to-end:
- `AT_LATEST` → only currently-open shards
- `AT_TRIM_HORIZON` → shards open at the trim horizon (stream creation, since the emulator never trims)
- `FROM_TRIM_HORIZON` (default) → whole set
- `AFTER_SHARD_ID` → shards after the given id
- `AT_TIMESTAMP` / `FROM_TIMESTAMP` → shards active at / from a timestamp

Implemented via per-shard `createdAt`/`closedAt` timestamps. The active filter is persisted into the `NextToken` so later pages stay consistent with the first.

**Nits (low-risk):**
- CreateStream now rejects an explicit `ShardCount` for `ON_DEMAND` streams (AWS does; on-demand still starts at 4 shards otherwise).
- UpdateShardCount now enforces `TargetShardCount` bounds: `1..10000` and within half-to-double the current shard count in a single call.

Deferred (out of scope, larger): shard-iterator expiry (needs a clock/TTL on iterators).

## Why

Each behavior was cross-checked against the official AWS Kinesis API docs (ListStreams, ShardFilter, CreateStream). These are real divergences a KCL / CLI / boto user hits during pagination and shard discovery.

## Tests

Real `aws-sdk-go-v2` reproductions + provider-level matrix:
- `ListStreams(ExclusiveStartStreamName="s-b")` over s-a..s-d → `[s-c, s-d]`; NextToken paginator still walks all streams with no dupes.
- `ListShards(ShardFilter{AT_LATEST})` after MergeShards → only the open child; unfiltered → all 3.
- Full filter matrix (AT/FROM_TRIM_HORIZON, AFTER_SHARD_ID, AT/FROM_TIMESTAMP), plus timestamp-required validation.
- Nits: ON_DEMAND+ShardCount rejection, UpdateShardCount half-to-double bounds.
- All existing Kinesis tests (#469 / #535 pagination + iterator + resharding) stay green.

Gates: build, vet, `go test`, `-race`, golangci-lint (0 new), gofmt — all clean. Coverage docs regenerated (no diff — operation set unchanged).

### DIFF
--- a/providers/aws/kinesis/kinesis.go
+++ b/providers/aws/kinesis/kinesis.go
@@ -33,13 +33,23 @@ const (
 	shardIDFmt  = "shardId-%012d"
 	hashKeyBits = 128
 	base10      = 10
+	// onDemandStartShards is the shard count an on-demand stream starts with.
+	onDemandStartShards = 4
+	// maxTargetShardCount / shardScaleFactor bound a single UpdateShardCount call:
+	// AWS caps the target at 10000 and permits scaling by at most a factor of two.
+	maxTargetShardCount = 10000
+	shardScaleFactor    = 2
 )
 
-// shardState is a shard plus its stored records and open/closed flag.
+// shardState is a shard plus its stored records and open/closed flag. createdAt
+// and closedAt (zero while open) time-stamp the shard's lifetime so ListShards
+// can honor the timestamp-based ShardFilter types.
 type shardState struct {
-	shard   driver.Shard
-	records []driver.Record
-	closed  bool
+	shard     driver.Shard
+	records   []driver.Record
+	closed    bool
+	createdAt time.Time
+	closedAt  time.Time
 }
 
 // streamData is the full server-side state of a stream.
--- a/providers/aws/kinesis/list_params_test.go
+++ b/providers/aws/kinesis/list_params_test.go
@@ -1,0 +1,270 @@
+package kinesis_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stackshy/cloudemu/v2/config"
+	"github.com/stackshy/cloudemu/v2/providers/aws/kinesis"
+	"github.com/stackshy/cloudemu/v2/services/kinesis/driver"
+)
+
+func newClockMock(t *testing.T, clk config.Clock) *kinesis.Mock {
+	t.Helper()
+
+	return kinesis.New(config.NewOptions(config.WithClock(clk)))
+}
+
+func TestListStreamsExclusiveStartStreamName(t *testing.T) {
+	ctx := context.Background()
+	m := newMock(t)
+
+	for _, n := range []string{"s-a", "s-b", "s-c", "s-d"} {
+		if err := m.CreateStream(ctx, driver.CreateStreamInput{StreamName: n, ShardCount: 1}); err != nil {
+			t.Fatalf("CreateStream(%s): %v", n, err)
+		}
+	}
+
+	out, err := m.ListStreams(ctx, "", "s-b", 0)
+	if err != nil {
+		t.Fatalf("ListStreams: %v", err)
+	}
+
+	if got := out.StreamNames; len(got) != 2 || got[0] != "s-c" || got[1] != "s-d" {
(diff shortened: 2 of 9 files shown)


NitinKumar004 (commented):
**Verdict: approve.** Built and tested against the pushed head (7aeb3d53). Gate matrix green: `go build ./...`, `go vet`, kinesis package tests, `-race` on the provider, `gofmt` clean, `golangci-lint` 0 issues, `go generate ./...` zero diff (coverage regen idempotent).

Verified beyond the PR's own tests with additional adversarial cases:

**ListStreams / ExclusiveStartStreamName**
- Resume strictly after the named stream; `s-d` (last) → `[]`; `s-z` (nonexistent, sorts after all) → `[]`.
- NextToken precedence confirmed: `ListStreams(NextToken="s-a", ExclusiveStart="s-c")` → `[s-b,s-c,s-d]`, i.e. NextToken wins and ExclusiveStart is ignored, matching AWS.
- Multi-page walk driven by ExclusiveStartStreamName + Limit: all streams visited exactly once, no dupes/omissions.

**ListShards / ShardFilter** — each type correct after SplitShard, MergeShards, and UpdateShardCount:
- AT_LATEST → only currently-open shards (2 children after split; 4 open after 2→4 reshard).
- AT_TRIM_HORIZON → only shards open at stream creation (the originals); createdAt is stamped from the same `now` as StreamCreationTimestamp, so the equality holds.
- FROM_TRIM_HORIZON / AFTER_SHARD_ID → full set (AFTER_SHARD_ID additionally seeds the start cursor).
- AT_TIMESTAMP / FROM_TIMESTAMP → open-at / open-or-closed-after semantics via createdAt/closedAt; missing Timestamp rejected; pre-horizon FROM_TIMESTAMP clamps to trim horizon.
- Filter persisted in NextToken across pages: AT_LATEST with MaxResults=2 pages through the exact filtered set with no reset, dupes, or omissions.

**Reshard timestamps** — SplitShard, MergeShards, and UpdateShardCount all stamp `closedAt` on closed parents and `createdAt` on children, so AT_LATEST/AT_TRIM_HORIZON exclude/include the right shards after each op.

**Nits** — ON_DEMAND + explicit ShardCount rejected (PROVISIONED still accepts it); UpdateShardCount bounds enforced (1..10000, and within half→double of current open count); exact boundaries (2x, 0.5x) allowed, just-outside rejected.

**Blast radius** — ListStreams signature change touches only the driver interface, the single provider implementation, and the one server handler; no other caller. Old-shape NextToken (pre-filter fields) decodes safely via omitempty; a cross-type token (e.g. a ListStreams name blob handed to ListShards) errors rather than panics.

Two minor, non-blocking deviations from strict AWS behavior:
1. A **nonexistent** ExclusiveStartStreamName that sorts *before* existing streams returns `[]` (exact-match-then-after) rather than treating the value as a lexicographic cursor and returning all greater names. Only affects invalid input on the deprecated param; NextToken tokens are always real names, so pagination is unaffected.
2. StreamName / ShardFilter sent alongside NextToken are silently ignored (token identity + persisted filter win) instead of raising ValidationException as real AWS does. Lenient, and pre-existing for StreamName.

PR #668, opened 2026-08-26
fix(aws-ecr): content-addressed digest + digest->tags model; by-tag delete untags not nukes; tagStatus filter

NitinKumar004:
## What

Fixes the ECR image/digest model so it matches real ECR content-addressing and tag semantics, plus the tagStatus listing filter. Found by the AWS deep-e2e audit (real aws-sdk-go-v2 client over the wire server).

## Findings addressed

1. **[HIGH] PutImage digest is not sha256(manifest).** The digest was a truncated 16-hex value derived from `repo:tag:now`, so it changed on every push and per tag. It is now the full sha256 of the manifest bytes (64-hex, `sha256:` prefixed), content-addressed — identical manifest content yields the same digest regardless of tag or time. An explicit client-supplied `imageDigest` is still respected.

2. **[HIGH] PutImage same-digest new-tag replaced the image, dropping prior tags.** PutImage rebuilt a fresh `ImageDetail{Tags:[tag]}` and overwrote the entry at the digest. An image is now identified by its digest and carries a SET of tags: pushing the same manifest under a second tag accumulates the tag onto the one manifest (moving it off any other manifest that held it, and preserving the original push time).

3. **[HIGH] BatchDeleteImage/DeleteImage by tag deleted the whole manifest.** Delete is now tag-vs-digest aware: a tag reference removes only that tag (the manifest survives while other tags remain, and is deleted with its last tag); a digest reference removes the manifest and all its tags. The response echoes imageIds the way real ECR does — a by-tag delete returns the digest plus that tag; a by-digest delete returns the digest with every tag it removed.

4. **[MED] ListImages/DescribeImages ignored filter.tagStatus.** The `filter.tagStatus` (TAGGED/UNTAGGED/ANY) is now decoded and applied.

Also: the provider read/mutate paths now take the mutex, so the model is race-clean (verified with `-race`).

## Why

Without content-addressing, pull-by-digest never resolves and DescribeImages dedup is wrong; CI that pushes `:latest` and `:<sha>` for one build, or `docker tag` + re-push, lost tags; and cleanup scripts that list only UNTAGGED images to prune got wrong results.

## Scope

Image/digest model + tagStatus only. Feature gaps (layer/registry-v2 upload ops, registry policy, PutImageTagMutability / PutImageScanningConfiguration dispatch, lifecycle-policy fidelity) are a separate follow-up.

## Tests

New real-SDK reproductions (same manifest under two tags -> one image/two tags/one 64-hex digest; by-tag delete untags and survives; last-tag/by-digest delete removes the manifest; tagStatus subsets) and provider-level model + `-race` concurrency tests. Existing 8 ECR tests stay green.

### DIFF
--- a/providers/aws/ecr/ecr.go
+++ b/providers/aws/ecr/ecr.go
@@ -18,11 +18,10 @@ import (
 )
 
 const (
-	defaultMediaType     = "application/vnd.docker.distribution.manifest.v2+json"
-	mutableTag           = "MUTABLE"
-	immutableTag         = "IMMUTABLE"
-	scanStatusComplete   = "COMPLETE"
-	digestHashFormatByte = 8
+	defaultMediaType   = "application/vnd.docker.distribution.manifest.v2+json"
+	mutableTag         = "MUTABLE"
+	immutableTag       = "IMMUTABLE"
+	scanStatusComplete = "COMPLETE"
 )
 
 // Compile-time check that Mock implements driver.ContainerRegistry.
@@ -81,6 +80,9 @@ func (m *Mock) CreateRepository(_ context.Context, cfg driver.RepositoryConfig) 
 		return nil, errors.New(errors.InvalidArgument, "repository name is required")
 	}
 
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
 	if m.repos.Has(cfg.Name) {
 		return nil, errors.Newf(errors.AlreadyExists, "repository %q already exists", cfg.Name)
 	}
@@ -123,6 +125,9 @@ func (m *Mock) CreateRepository(_ context.Context, cfg driver.RepositoryConfig) 
 
 // DeleteRepository deletes an ECR repository.
 func (m *Mock) DeleteRepository(_ context.Context, name string, force bool) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
 	rd, ok := m.repos.Get(name)
 	if !ok {
 		return errors.Newf(errors.NotFound, "repository %q not found", name)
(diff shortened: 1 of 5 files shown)


NitinKumar004 (commented):
Reviewed the pushed head (0f6aa12). Clean-built the actual tree and ran real aws-sdk-go-v2 flows beyond the PR's own tests. The image/digest model rework is correct and the data-loss class of bug is fixed.

**Verified against a real ECR client**

- Content-addressed digest is *exactly* `sha256:` + hex(sha256(manifest bytes)) — matched byte-for-byte against an independently computed hash. Same manifest under two tags collapses to one image / one 64-hex digest / tags `[v1,v2]`; different manifest bytes yield a different digest; an explicit `imageDigest` is respected.
- Tag move: pushing a tag that currently points at manifest A to manifest B moves the tag to B and leaves A behind as an *untagged, surviving* image — the real "untagged accumulation" semantics, not a duplicate.
- Delete is tag-vs-digest aware. By-tag on `[v1,v2]` untags only (manifest survives tagged `[v2]`); deleting the last tag removes the manifest; by-digest removes the manifest and all tags. Echo shape matches ECR: by-tag returns `{digest,tag}`, by-digest returns one `{digest,tag}` entry per removed tag. Mixed batch (valid tag + ghost) deletes the valid one and returns a single `ImageNotFound` failure.
- `tagStatus` (TAGGED/UNTAGGED/ANY) applied on both ListImages and DescribeImages.
- `PushedAt` is preserved on a same-digest re-push (new tag does not reset the push time); `imageSizeInBytes` still populated.

**Concurrency**: mutex added to all read/mutate paths. No self-deadlock — every locked method delegates only to lock-free helpers (`storeImage`, `findImage`, `updateTagIndex`, `resolveDigest`, `autoScan`, `emitMetric`). A concurrent PutImage/DeleteImage/ListImages stress test under `-race` (repeated) is clean and ImageCount stays consistent with the stored set.

**Gates**: build, vet, `test` (server+provider+containerregistry), `-race` (provider), gofmt, golangci-lint (0 issues), and `go generate` (zero diff — confirms no interface change) all green. Prior 8 ECR tests are untouched and remain valid (each pushes a single manifest per repo, so content-addressing does not collapse them).

**Non-blocking note**: Azure ACR and GCP Artifact Registry still use the old truncated, time-based digest (`repo:tag:now`, 16-hex). Out of this PR's stated AWS-only scope and package-local, so no cross-provider breakage — worth a follow-up to bring them to the same content-addressed model.

Verdict: good to merge.

PR #669, opened 2026-08-26
fix(aws-ecr): dispatch PutImageTagMutability/PutImageScanningConfiguration (enforced); lifecycle policy round-trips with full fidelity

NitinKumar004:
## What

Fixes two AWS ECR bugs found by the deep real-user e2e audit, both of which break Terraform in-place update paths on `aws_ecr_repository` / `aws_ecr_lifecycle_policy`.

### 1. PutImageTagMutability + PutImageScanningConfiguration were undispatched
Both operations returned `UnknownOperationException`, so a `terraform apply` that changed `image_tag_mutability` or `image_scanning_configuration.scan_on_push` on an existing repository failed hard.

- Dispatched both in the ECR wire handler.
- Added AWS-specific provider methods (`PutImageTagMutability`, `PutImageScanningConfiguration`) reached via type assertion — same optional-interface pattern as `GetAuthorizationToken`, keeping them out of the portable ContainerRegistry driver (Azure ACR / GCP Artifact Registry model these differently).
- Responses echo the real AWS shapes: `{registryId, repositoryName, imageTagMutability}` and `{registryId, repositoryName, imageScanningConfiguration:{scanOnPush}}`.
- The mutability change **takes effect**: it updates the value `checkTagMutability` reads, so after setting `IMMUTABLE` a re-push of an existing tag returns `ImageTagAlreadyExistsException`, and after setting back to `MUTABLE` the re-push succeeds. `DescribeRepositories` reflects both new values.

### 2. GetLifecyclePolicy lost fidelity on re-serialization
`parseLifecyclePolicyText` kept only the first pattern (`firstTagPattern`) and collapsed a multi-element `tagPrefixList`/`tagPatternList` into a single `tagPattern`, so `GetLifecyclePolicy` returned a structurally different, semantically lossy document — Terraform saw perpetual drift.

- `driver.LifecyclePolicy` now carries an optional `Document` field preserving the original policy text verbatim.
- The AWS wire layer stores the raw `lifecyclePolicyText` on Put and returns it byte-faithfully on Get/Delete; the structured `Rules` remain available for lifecycle evaluation.

## Why
Terraform (and `aws ecr put-image-tag-mutability` / `put-image-scanning-configuration`) drive these operations on repository updates; the gaps caused apply failures and perpetual-diff drift.

## Tests
Real `aws-sdk-go-v2` reproductions:
- `PutImageTagMutability(IMMUTABLE)` -> DescribeRepositories shows IMMUTABLE AND re-push existing tag -> `ImageTagAlreadyExistsException`; set MUTABLE -> re-push succeeds; missing-repo -> `RepositoryNotFoundException`.
- `PutImageScanningConfiguration(scanOnPush=true)` -> DescribeRepositories reflects it.
- `PutLifecyclePolicy` with a 2-rule policy whose first rule has `tagPrefixList:["prod","stg"]` -> `GetLifecyclePolicy` returns both rules with the full lists intact (verbatim + structural check).

Merged #668 image-model tests and the prior ECR SDK tests stay green.

### DIFF
--- a/providers/aws/ecr/ecr.go
+++ b/providers/aws/ecr/ecr.go
@@ -123,6 +123,57 @@ func (m *Mock) CreateRepository(_ context.Context, cfg driver.RepositoryConfig) 
 	return &result, nil
 }
 
+// PutImageTagMutability updates a repository's image tag mutability setting.
+// This is AWS-specific (not part of the portable ContainerRegistry driver), so
+// the ECR wire handler reaches it via type assertion. The new value takes effect
+// immediately: it is read by checkTagMutability on subsequent pushes.
+func (m *Mock) PutImageTagMutability(
+	_ context.Context, repository, mutability string,
+) (*driver.Repository, error) {
+	if mutability != mutableTag && mutability != immutableTag {
+		return nil, errors.Newf(errors.InvalidArgument,
+			"invalid imageTagMutability %q; expected MUTABLE or IMMUTABLE", mutability)
+	}
+
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	rd, ok := m.repos.Get(repository)
+	if !ok {
+		return nil, errors.Newf(errors.NotFound, "repository %q not found", repository)
+	}
+
+	rd.tagMutability = mutability
+	rd.info.ImageTagMutability = mutability
+
+	result := rd.info
+	result.ImageCount = rd.images.Len()
+
+	return &result, nil
+}
+
+// PutImageScanningConfiguration updates a repository's scan-on-push setting.
+// AWS-specific; reached by the ECR wire handler via type assertion.
+func (m *Mock) PutImageScanningConfiguration(
+	_ context.Context, repository string, scanOnPush bool,
+) (*driver.Repository, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	rd, ok := m.repos.Get(repository)
+	if !ok {
+		return nil, errors.Newf(errors.NotFound, "repository %q not found", repository)
+	}
+
+	rd.scanOnPush = scanOnPush
+	rd.info.ScanOnPush = scanOnPush
+
+	result := rd.info
+	result.ImageCount = rd.images.Len()
+
+	return &result, nil
+}
+
 // DeleteRepository deletes an ECR repository.
 func (m *Mock) DeleteRepository(_ context.Context, name string, force bool) error {
 	m.mu.Lock()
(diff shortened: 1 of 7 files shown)


NitinKumar004 (commented):
Reviewed at head 38a9e22 (clean build/vet/test/-race, gofmt clean, golangci-lint 0 issues on changed packages, go generate produces zero diff). Verdict: looks good to merge.

**Mutability enforcement — verified live and correct**
- `checkTagMutability` reads the live `rd.tagMutability` off the pointer stored in memstore, and `PutImageTagMutability` mutates that same field, so a mid-life MUTABLE→IMMUTABLE change takes effect on the next push (re-push of an existing tag → `ImageTagAlreadyExistsException` via the `FailedPrecondition`→`writePutImageErr` mapping), and IMMUTABLE→MUTABLE re-enables overwrite.
- A repo *created* IMMUTABLE (`CreateRepository` sets `rd.tagMutability` from cfg) enforces immediately with no explicit Put — confirmed with a fresh SDK test.
- A *new* tag on an IMMUTABLE repo is still allowed; only overwriting an existing tag is blocked — matches real ECR.
- Invalid `imageTagMutability` value → `InvalidArgument` → `InvalidParameterException`.

**Optional-interface / type-assertion wiring — correct**
`h.registry` is the `crdriver.ContainerRegistry` interface whose dynamic value is `*providers/aws/ecr.Mock` (wired via `ecr.New(d.ECR)` ← `p.ECR`), which implements `PutImageTagMutability`/`PutImageScanningConfiguration`. The assertion therefore always succeeds for AWS; the "not supported" `ServerException` branch is a dead guard for AWS and only trips for a provider genuinely lacking the method — it does NOT trigger on a missing repo. Missing-repo goes through the setter → `NotFound` → `RepositoryNotFoundException` (confirmed for both new operations). Same pattern as the existing `GetAuthorizationToken` surface; portable driver untouched (Azure/GCP unaffected).

**Scanning config** round-trips both true and false through `DescribeRepositories` and the echoed `imageScanningConfiguration`.

**Lifecycle fidelity** — the optional `Document` field is stored verbatim on Put and returned byte-faithfully on both Get and Delete (shared `writeLifecyclePolicyResult`/`marshalLifecyclePolicyText`, `copyLifecyclePolicy` now propagates `Document`); structured `Rules` remain populated for evaluation, so eval is unaffected. Multi-rule + multi-element `tagPrefixList` survives intact. Invalid policy JSON is rejected before storage (`parseLifecyclePolicyText` error → `InvalidParameterException`) and nothing is persisted (subsequent Get → `LifecyclePolicyNotFoundException`).

**Blast radius**: `driver.LifecyclePolicy.Document` is a purely additive optional field; no consumer break, coverage regen idempotent.

PR #670, opened 2026-08-26
fix(aws-secretsmanager): explicit per-version staging labels — honor ClientRequestToken/VersionStages, add UpdateSecretVersionStage, ARN suffix

NitinKumar004:
## What

Reworks AWS Secrets Manager staging from a `Current bool` + slice-adjacency heuristic to an **explicit per-version stage-label model** (each secret maps a staging label → the version id that holds it), fixing three HIGH and several supporting bugs in the rotation state machine. Every behavior was cross-checked against the official AWS Secrets Manager API docs.

## Why / bugs fixed

- **[HIGH F1] ClientRequestToken ignored.** The token is now threaded through Create/PutSecretValue and used as the VersionId. A repeat with the same token + same content is an idempotent no-op (same VersionId, no new version); the same token + different content is `ResourceExistsException`. `GetSecretValue(VersionId=token)` resolves. Version ids are now 36-char UUIDs (fixes F9).
- **[HIGH F2] PutSecretValue VersionStages ignored → clobbered AWSCURRENT.** Passing `VersionStages` (e.g. `[AWSPENDING]`) now attaches exactly those labels and does NOT move AWSCURRENT — a default `GetSecretValue` still returns the prior value. Only an empty/unset `VersionStages` promotes the new version to AWSCURRENT (demoting the prior current to AWSPREVIOUS and evicting the old AWSPREVIOUS label).
- **[MED-HIGH F4] UpdateSecretVersionStage added** (rotation `finishSecret` step). Moving AWSCURRENT to a version auto-demotes the old AWSCURRENT to AWSPREVIOUS; validates target versions and `RemoveFromVersionId`.
- **[MED F6] UpdateSecret now returns the new VersionId** when a value change creates a version.
- **[MED F8] Secret ARNs now carry the 6-char `-XXXXXX` suffix** (deterministic per secret, FakeClock-stable). Lookup by the bare friendly name OR the suffixed ARN both resolve.
- **[MED F5] GetRandomPassword now enforces RequireEachIncludedType** — guarantees ≥1 of each included class and rejects a length shorter than the class count.

Out of scope (separate follow-up): F3 (resource-policy write path) and F7 (BatchGetSecretValue).

## Notes

- The portable `driver.Secrets` interface is unchanged; the new behaviors live on the AWS-specific `secretStager`/`secretMutator` optional interfaces the wire handler already type-asserts, so Azure/GCP are unaffected. `SecretConfig` gains a `ClientRequestToken` field (ignored by other providers).
- Staging labels are persisted through snapshot/restore.
- New `idgen.UUID()` (crypto/rand v4) and `idgen.SecretARNSuffix()` helpers.

## Tests

Real `aws-sdk-go-v2` reproductions for F1/F2/F4/F6/F8/F5 plus provider-level tests for the explicit model, idempotency, stage promotion, and snapshot/restore of stages. Existing tests (AWSPREVIOUS retrieval, binary round-trip, delete/restore, #567 KmsKeyId, staging label counts) stay green.

### DIFF
--- a/internal/idgen/idgen.go
+++ b/internal/idgen/idgen.go
@@ -2,6 +2,7 @@
 package idgen
 
 import (
+	"crypto/rand"
 	"fmt"
 	"hash/fnv"
 	"sync/atomic"
@@ -9,6 +10,62 @@ import (
 
 // guidNodeMask isolates the low 48 bits used as a GUID's node field.
 const guidNodeMask = 0xffffffffffff
+
+// uuidByteLen is the number of random bytes in a version-4 UUID.
+const uuidByteLen = 16
+
+// RFC 4122 version/variant bit masks for a version-4 UUID.
+const (
+	versionMask  = 0x0f
+	version4Bits = 0x40
+	variantMask  = 0x3f
+	variantBits  = 0x80
+)
+
+// UUID returns a random RFC 4122 version-4 UUID as a 36-character string
+// (8-4-4-4-12 hex with hyphens). AWS Secrets Manager version ids take this
+// shape, and callers can pin one via a client request token. It draws from
+// crypto/rand; a random source failure falls back to a zero-filled UUID so the
+// function never panics.
+func UUID() string {
+	var b [uuidByteLen]byte
+	if _, err := rand.Read(b[:]); err != nil {
+		// Degrade to an all-zero UUID rather than panic; still 36 chars.
+		b = [uuidByteLen]byte{}
+	}
+
+	// RFC 4122: set the version to 4 and the variant to 10xx.
+	b[6] = (b[6] & versionMask) | version4Bits
+	b[8] = (b[8] & variantMask) | variantBits
+
+	return fmt.Sprintf("%08x-%04x-%04x-%04x-%012x",
+		b[0:4], b[4:6], b[6:8], b[8:10], b[10:16])
+}
+
+// suffixAlphabet is the 6-character random-suffix character set AWS appends to a
+// Secrets Manager ARN's resource segment (:secret:<name>-<suffix>).
+const suffixAlphabet = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+
+// secretARNSuffixLen is the length of the ARN suffix AWS appends.
+const secretARNSuffixLen = 6
+
+// SecretARNSuffix derives a deterministic 6-character alphanumeric suffix from
+// seed, matching the trailing "-XXXXXX" AWS adds to a Secrets Manager ARN. It is
+// deterministic per seed so the same secret keeps a stable ARN across a run
+// (and under a FakeClock), while distinct secrets get distinct suffixes.
+func SecretARNSuffix(seed string) string {
+	h := fnv.New64a()
+	_, _ = h.Write([]byte(seed))
+	n := h.Sum64()
+
+	out := make([]byte, secretARNSuffixLen)
+	for i := range out {
+		out[i] = suffixAlphabet[n%uint64(len(suffixAlphabet))]
+		n /= uint64(len(suffixAlphabet))
+	}
+
+	return string(out)
+}
 
 // SyntheticGUID derives a deterministic GUID-shaped string from seed. The value
 // is synthetic (a stand-in for an Azure principal/tenant id), not a real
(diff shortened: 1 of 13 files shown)


NitinKumar004 (commented):
Reviewed against the pushed head (8f093de). Clean-built the actual tree and ran a real aws-sdk-go-v2 rotation cycle plus provider-level and -race checks. Gates are green: build, vet, targeted tests, `-race`, gofmt, golangci-lint (0 new issues on changed packages), and `go generate ./...` yields no diff.

Most of the rework holds up under adversarial testing:

- ClientRequestToken idempotency: same token+content is a no-op (version count unchanged), same token+different content is ResourceExistsException, token is the VersionId, and it is correctly isolated per-secret (same token reused on a different secret does not collide).
- Empty-stages puts keep exactly one AWSCURRENT and at most one AWSPREVIOUS; the old AWSPREVIOUS holder is evicted (verified across two rotation cycles — the version that loses AWSCURRENT becomes AWSPREVIOUS and the prior previous drops to no labels, still retrievable by VersionId and listed only with IncludeDeprecated).
- PutSecretValue(VersionStages=[AWSPENDING]) does not clobber AWSCURRENT; default GetSecretValue still returns the prior value.
- UpdateSecretVersionStage edges reject a non-existent MoveToVersionId and a RemoveFromVersionId that does not hold the label, leaving AWSCURRENT intact.
- ARN suffix is deterministic (stable across calls and snapshot/restore) and resolves by both bare name and full suffixed ARN; a prefix-named pair ("foo" vs "foo-bar") does not mis-resolve.
- Snapshot/restore preserves the multi-version stage map and the suffixed ARN.
- GetRandomPassword with RequireEachIncludedType always includes each class over 50 iterations, rejects a length shorter than the class count, and honors ExcludePunctuation.

One blocking correctness gap, in the headline path:

**AWSPENDING is not auto-removed when AWSCURRENT is promoted onto that version.** In a real rotation, finishSecret calls `UpdateSecretVersionStage(VersionStage=AWSCURRENT, MoveToVersionId=<pending>, RemoveFromVersionId=<oldCurrent>)`. Real Secrets Manager then automatically removes AWSPENDING from the newly-current version, so `DescribeSecret.VersionIdsToStages` shows the current version as `[AWSCURRENT]` only. Here it reports `[AWSCURRENT, AWSPENDING]` — the AWSPENDING label rides forward onto whatever version is current after every rotation. This is observable via DescribeSecret and GetSecretValue.VersionStages, and can confuse rotation logic that inspects for the AWSPENDING version.

Repro (real SDK): CreateSecret (v1=AWSCURRENT); PutSecretValue(VersionStages=[AWSPENDING]) -> v2; UpdateSecretVersionStage(AWSCURRENT, MoveTo=v2, RemoveFrom=v1); DescribeSecret -> v2 stages = [AWSCURRENT, AWSPENDING], expected [AWSCURRENT].

Fix locus: `promoteToCurrent` in providers/aws/secretsmanager/staging.go — when setting AWSCURRENT on a version, delete the AWSPENDING label if it currently points to that same version (matching the documented "automatically removes the AWSPENDING label from that version" behavior).

NitinKumar004 (commented):
Re-review of the delta fix (97186331) — verified on the pushed tree.

**Verdict: approve.** The reported bug (AWSPENDING riding forward onto AWSCURRENT across rotations) is fixed correctly and introduces no regression.

The fix in `promoteToCurrent` is minimal and correctly scoped:
```go
if pending, ok := sd.stages[stagePending]; ok && pending == versionID {
    delete(sd.stages, stagePending)
}
```
The `pending == versionID` guard means AWSPENDING is only stripped from the version actually being promoted — never from another version that happens to hold it.

Adversarially verified with real aws-sdk-go-v2 clients:
- Full two-rotation cycle: after finishSecret (UpdateSecretVersionStage moving AWSCURRENT + RemoveFrom old current), the newly-current version shows exactly `[AWSCURRENT]`, prior shows `[AWSPREVIOUS]`; AWSPENDING never accumulates on any version across both rotations.
- Guard scoping: a version holding AWSPENDING that is NOT the promotion target keeps AWSPENDING when a different version is promoted.
- No spurious deletion: promoting a target with no AWSPENDING works and preserves the AWSPREVIOUS shuffle.
- Custom labels: a non-AWSPENDING custom stage label on the promoted version is preserved (only AWSPENDING is auto-cleared, matching the real service).
- Invariants hold: exactly one AWSCURRENT, at most one AWSPREVIOUS, including under concurrent rotations (-race clean).

Gates on the pushed SHA: build, vet, targeted tests (server/providers/services/persist), -race, gofmt, golangci-lint (0 issues), and go generate (zero diff) all pass.

PR #671, opened 2026-08-26
fix(aws-eventbridge): scope matching to the event's own bus; validate EventPattern; TestEventPattern; ListEventBuses NamePrefix; clean error messages

NitinKumar004:
## Summary

Deep real-user (aws-sdk-go-v2) e2e audit of AWS EventBridge found one blocker plus four correctness/coverage gaps. Each is fixed at its correct 3-layer location and reproduced with the real SDK.

## What changed

- **[BLOCKER] Event-bus isolation** — `MatchedRules` scanned every bus, so a rule on a custom bus fired for events published to the default bus (and vice-versa), defeating the whole point of custom buses (isolation/routing) and inflating the `MatchedEvents` metric. Matching is now scoped to the event's own bus (empty → default).
- **[HIGH] TestEventPattern** — was `UnknownOperationException`. Added the wire op: it validates the pattern and evaluates the sample event with the **same matcher PutEvents delivery uses**, so its verdict is consistent with actual delivery. Returns `{Result: bool}`.
- **[MED] Invalid EventPattern accepted** — `PutRule` stored a structurally invalid pattern verbatim (`{"source":"not-an-array"}`, malformed JSON), which then silently matched nothing so targets never fired. Now validated (leaf values must be arrays / content-operators; nested objects recurse) and rejected with `InvalidEventPatternException` at both PutRule and TestEventPattern. Valid array/content-operator/nested-detail patterns still pass.
- **[MED] ListEventBuses ignored NamePrefix** — the handler never decoded it, returning every bus. Now decoded and filtered by prefix.
- **[LOW] Error messages leaked internal code prefix** — `writeErr` used `err.Error()` (`"InvalidArgument: ..."`), corrupting the deliberately-verbatim DeleteRule wording. Now uses `cerrors.Message(err)`, mirroring ec2/iam/lambda.

Out of scope (deferred separately): PutPermission/RemovePermission, ListRuleNamesByTarget.

## Why

Cross-bus leakage breaks multi-tenant/multi-app isolation; silently-accepted bad patterns are the worst failure mode for event-driven apps (deploy succeeds, nothing ever fires); TestEventPattern is the canonical way users debug patterns before deploying.

## Tests

Real aws-sdk-go-v2 reproductions: bus isolation (both directions, asserted via SQS delivery), TestEventPattern (match/no-match + consistency with real delivery + invalid-pattern rejection), PutRule invalid/valid patterns, ListEventBuses NamePrefix, and the clean DeleteRule error message. Existing EventBridge tests remain green. No driver-interface change, so no coverage regeneration needed.

### DIFF
--- a/internal/eventmatch/eventmatch.go
+++ b/internal/eventmatch/eventmatch.go
@@ -446,6 +446,54 @@ func numericValue(value any, coerceNum bool) (float64, bool) {
 	return toFloat(value)
 }
 
+// PatternError describes why an event pattern is structurally invalid. Its
+// message is safe to surface verbatim to a client (no internal code prefix).
+type PatternError struct {
+	Reason string
+}
+
+func (e *PatternError) Error() string {
+	return "Event pattern is not valid. Reason: " + e.Reason
+}
+
+// ValidatePattern reports whether raw is a structurally valid EventBridge event
+// pattern / SNS filter policy. A pattern is a JSON object whose every leaf value
+// is an array (of match values or content-operator objects); a value that is a
+// nested object is validated recursively. A scalar leaf (string/number/bool), a
+// non-object top level, or malformed JSON is invalid — real EventBridge rejects
+// these with InvalidEventPatternException rather than silently matching nothing.
+func ValidatePattern(raw string) error {
+	if strings.TrimSpace(raw) == "" {
+		return &PatternError{Reason: "event pattern must not be empty"}
+	}
+
+	var p map[string]any
+	if err := json.Unmarshal([]byte(raw), &p); err != nil {
+		return &PatternError{Reason: "event pattern is not valid JSON"}
+	}
+
+	return validatePatternObject(p)
+}
+
+// validatePatternObject enforces the array/object leaf rule over every field of
+// a pattern object, recursing into nested objects.
+func validatePatternObject(obj map[string]any) error {
+	for key, val := range obj {
+		switch v := val.(type) {
+		case []any:
+			continue
+		case map[string]any:
+			if err := validatePatternObject(v); err != nil {
+				return err
+			}
+		default:
+			return &PatternError{Reason: "\"" + key + "\" must be an object or an array"}
+		}
+	}
+
+	return nil
+}
+
 // ParsePattern decodes a JSON event pattern / filter policy into the object form
 // MatchEvent and MatchLeaf consume. It returns false when the JSON is not a
 // well-formed object.
--- a/internal/eventmatch/validate_test.go
+++ b/internal/eventmatch/validate_test.go
@@ -1,0 +1,40 @@
+package eventmatch_test
+
+import (
+	"testing"
+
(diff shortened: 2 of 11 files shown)


NitinKumar004 (commented):
**Verdict: approve.** Verified against the pushed head `ff8f770`, clean-built and exercised with real aws-sdk-go-v2 (build/vet/test/-race all green).

**Bus isolation (blocker) — correct, both directions.** `MatchedRules` now resolves `event.EventBus` (empty→default) and scans only that bus's rules. Verified: default-bus event does not fire a custom-bus rule and vice-versa; a multi-bus setup with the *same* source pattern on default+two custom buses delivers an event only to its own bus (no cross-bus leak); a no-explicit-bus rule still matches an empty-`EventBus` event via default resolution. `-race` with concurrent `PutEvents` across 6 distinct buses stays per-bus consistent (each queue receives exactly 1).

**TestEventPattern — consistent with delivery.** Both the wire op and the `PutEvents` delivery path go through the same `eventmatch.ParsePattern` + `eventmatch.MatchEvent`, so the verdict is structurally (not just test-observed) consistent with actual delivery. Match→true, non-match→false, invalid pattern→`InvalidEventPatternException`, all confirmed.

**Pattern validation — no SNS regression.** `eventmatch.ValidatePattern` is a purely additive addition; the diff removes/modifies no existing `eventmatch` function. It is called only from the three EventBridge sites (provider `PutRule`, wire `putRule`, wire `testEventPattern`). SNS filter policies use only the pre-existing `ParsePattern`/`MatchEvent`/`MatchLeafAttr`/`IsOrClause` and are untouched — the SNS filter-policy and #664 numeric-attribute suites pass unchanged. EventBridge cases verified: non-array leaf and malformed JSON rejected; array / content-operator (`prefix`) / nested-`detail` / `$or` accepted.

**ListEventBuses NamePrefix / clean error messages — correct.** `NamePrefix` decoded and filtered (prefix returns only matches; absent returns all incl. default); `writeErr` now uses `cerrors.Message`, so DeleteRule-with-targets arrives as the verbatim `Rule can't be deleted since it has targets.` with no internal code prefix.

`go generate ./...` produces zero diff (no interface change).

**Non-blocking notes:**
- Lint delta vs development (same binary): `ServeHTTP` gocyclo 18→19 from the new `TestEventPattern` dispatch case (already-flagged dispatcher, unavoidable), and one new wsl "range cuddled with assignment not used in the iteration" at `operations.go:78` — the inserted `NamePrefix` `if…continue` is now the first statement in the loop, so the cuddled `entries := make(...)` is no longer used on the first line; a blank line before the `for` clears it. Both are advisory (CI lint is report-only) and mainline already carries such findings, so neither blocks.
- Empty pattern `{}` is accepted and matches all events; this matches prior behavior (unchanged by this PR) but differs from real EventBridge, which rejects an empty pattern. Minor, pre-existing, out of scope here.

PR #672, opened 2026-08-26
fix(aws-secretsmanager): resource-policy ops (Put/Get/Delete/Validate) + BatchGetSecretValue

NitinKumar004:
## What

Adds the AWS Secrets Manager resource-policy control-plane ops and the batch read, all previously returning `UnknownOperationException`:

- **PutResourcePolicy** — stores the JSON policy on the secret; `BlockPublicPolicy=true` rejects a policy that grants public access (wildcard `Principal` with no `Condition`) as `PublicPolicyException`; malformed JSON is `MalformedPolicyDocumentException`.
- **GetResourcePolicy** — returns `{ARN, Name, ResourcePolicy}`; `ResourcePolicy` is omitted (nil) when none is set, which the SDK reads as "no policy" (not an error).
- **DeleteResourcePolicy** — clears the policy.
- **ValidateResourcePolicy** — checks JSON syntax + broad access without storing; returns `PolicyValidationPassed` with `ValidationErrors` (passed=false + errors on malformed/public).
- **BatchGetSecretValue** — resolves `SecretIdList` (or `Filters`), returning `SecretValues` for the secrets that exist and a per-item `Errors` entry (`ResourceNotFoundException`) for missing ones, rather than failing the whole batch. Supports `MaxResults`/`NextToken` via the existing offset paginator.

## Why

`PutResourcePolicy`/`DeleteResourcePolicy` being unimplemented broke Terraform's `aws_secretsmanager_secret_policy` (create/update/destroy). `BatchGetSecretValue` is a common startup path for apps fetching many secrets. Found by the AWS real-user e2e deep audit (findings F3 [HIGH], F7 [MED]).

## How

Resource policies are stored per-secret in the AWS provider and exposed through a type-asserted `secretPolicyManager` optional interface (same pattern as the existing `secretStager`/`secretMutator`), so the portable `driver.Secrets` interface stays stable and Azure/GCP are untouched. The policy survives snapshot/restore.

## Tests

Real `aws-sdk-go-v2` reproductions: resource-policy Put/Get/Delete round-trip; Get-after-delete returns empty (not error); ValidateResourcePolicy valid→passed / malformed→failed+errors; BlockPublicPolicy rejects a public policy; BatchGetSecretValue over `[a,b,missing]` yields values for a,b + an error entry for missing; empty list handled. Provider-level round-trip + snapshot tests. Existing SecretsManager tests remain green.

### DIFF
--- a/providers/aws/secretsmanager/resource_policy.go
+++ b/providers/aws/secretsmanager/resource_policy.go
@@ -1,0 +1,71 @@
+package secretsmanager
+
+import (
+	"context"
+
+	"github.com/stackshy/cloudemu/v2/errors"
+	"github.com/stackshy/cloudemu/v2/services/secrets/driver"
+)
+
+// PutResourcePolicy attaches (or replaces) the JSON resource-based policy on a
+// secret, the write path Terraform's aws_secretsmanager_secret_policy uses. The
+// BlockPublicPolicy check lives in the wire layer, which rejects a public policy
+// before calling this. A secret scheduled for deletion is rejected.
+func (m *Mock) PutResourcePolicy(_ context.Context, name, policy string) (*driver.SecretInfo, error) {
+	sd, ok := m.secrets.Get(name)
+	if !ok {
+		return nil, errors.Newf(errors.NotFound, "secret %q not found", name)
+	}
+
+	sd.mu.Lock()
+	defer sd.mu.Unlock()
+
+	if !sd.deletedAt.IsZero() {
+		return nil, errors.New(errors.FailedPrecondition,
+			"secret is scheduled for deletion, so this operation is not allowed")
+	}
+
+	sd.resourcePolicy = policy
+	info := sd.info
+
+	return &info, nil
+}
+
+// GetResourcePolicy returns the secret's metadata and its resource policy JSON
+// (empty when none is set). It reads leniently so DescribeSecret-style reads
+// keep working; only a missing secret is a NotFound.
+func (m *Mock) GetResourcePolicy(_ context.Context, name string) (*driver.SecretInfo, string, error) {
+	sd, ok := m.secrets.Get(name)
+	if !ok {
+		return nil, "", errors.Newf(errors.NotFound, "secret %q not found", name)
+	}
+
+	sd.mu.RLock()
+	defer sd.mu.RUnlock()
+
+	info := sd.info
+
+	return &info, sd.resourcePolicy, nil
+}
+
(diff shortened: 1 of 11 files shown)


NitinKumar004 (commented):
Reviewed against the pushed head (8918cbe). Clean-built the actual tree and ran the gate matrix plus an independent aws-sdk-go-v2 probe suite. Verdict: looks good to merge.

**Gates (pushed SHA):** build, vet, targeted tests (server+provider secretsmanager, services/secrets, persist), `-race` on the provider, golangci-lint on both changed packages (0 issues), gofmt, and `go generate ./...` (zero diff) all pass.

**Verified behaviors (independent SDK tests):**
- Resource policy Put -> Get returns the exact JSON byte-for-byte; a second Put overwrites; Delete -> Get returns nil `ResourcePolicy` (omitempty), not an error and not stale; Put on a missing secret -> `ResourceNotFoundException`.
- BlockPublicPolicy detector is correct across the tricky cases: `Principal:"*"`, `{"AWS":"*"}`, and `{"AWS":["*"]}` are all caught; `Effect:Deny` + wildcard is not public; a wildcard principal *with* a Condition is allowed; a specific principal is allowed; without the flag the public policy is accepted; and it parses `Statement` as either an object or an array.
- ValidateResourcePolicy: valid -> passed=true and stores nothing (Get still empty afterward); malformed JSON -> passed=false with a ValidationErrors entry.
- BatchGetSecretValue: missing id yields a per-item `Errors{ResourceNotFoundException}` without failing the batch; MaxResults/NextToken pagination walks all items with no dupes or omissions; correct AWSCURRENT VersionId + `[AWSCURRENT]` stages (consistent with the #670 staging surface).
- Snapshot/restore preserves the resource policy.
- Optional-interface pattern is clean: `secretPolicyManager` is type-asserted in the wire layer, the portable driver is untouched, and Azure/GCP compile and fall through to the not-supported path. The `respondSecretRef` dedup keeps the RestoreSecret/DeleteResourcePolicy `{ARN,Name}` shape intact.

**Non-blocking observations (real-AWS parity edge cases, safe to defer):**
1. `BatchGetSecretValue` with an empty `SecretIdList` and no `Filters` returns an empty result; real AWS returns `InvalidParameterException` (one of the two must be supplied). Same for supplying both — AWS rejects it, here the id list silently wins.
2. The public-policy check treats any non-empty `Condition` as narrowing, so a degenerate `"Condition":{}` alongside a wildcard Allow is deemed non-public; AWS's Zelkova analysis would still flag it. Reasonable simplification for the emulator.

None of these affect the claimed fixes.

PR #673, opened 2026-08-26
fix(aws-elbv2): enforce deletion_protection; model redirect/fixed-response actions + certs + rule conditions round-trip; SetRulePriorities uniqueness

NitinKumar004:
## What

Fixes six AWS ELBv2 correctness bugs found by a deep real-user (aws-sdk-go-v2) e2e audit. All were empirically reproduced and each is verified against the official ELBv2 API reference.

- **deletion_protection honored** — `DeleteLoadBalancer` ignored `deletion_protection.enabled`, silently wiping a protected load balancer (data loss; the exact scenario a `terraform destroy` hits). It now rejects the delete with `OperationNotPermitted` while a missing ARN stays idempotent. Delete also cascades the rules under the LB's listeners and drops stored LB attributes (was an in-memory leak).
- **Redirect / fixed-response default actions round-trip** — the listener stored only a single forward `TargetGroupArn`, so a `redirect` or `fixed-response` default action was silently dropped. The HTTP→HTTPS redirect (the most common ALB pattern) came back with an empty `default_action` → permanent Terraform drift. The listener now stores the full default-action list and the wire layer parses/echoes forward, redirect and fixed-response actions verbatim.
- **HTTPS `SslPolicy` + `Certificates` round-trip** — both were parsed off the request but never stored; Describe returned them empty. Now stored and echoed, with the create certificate marked `IsDefault=true` as real ELBv2 does.
- **`SetRulePriorities` uniqueness** — the reorder path had no collision check, so two rules on one listener could share a priority (silent routing corruption). It now validates each requested priority against the listener's other rules (and within the batch) before any write, returning `PriorityInUse`; a same-batch swap stays allowed.
- **Typed rule conditions round-trip** — a condition stored only `Field` + a flat `Values` list, so `http-header` / `query-string` / `source-ip` conditions lost their typed config on Create→Describe. `RuleCondition` now carries the full typed config for host-header, path-pattern, http-header, query-string, source-ip and http-request-method.
- **LoadBalancer `VpcId` populated (F6 fixed, not deferred)** — `VpcId` came back empty. The ELB mock now derives it from the member subnets via a `SubnetResolver` wired to the VPC mock in the AWS factory — the same cross-service pattern RDS/ElastiCache/EC2 already use. Unwired, the field stays empty rather than failing.

Extras included: DeleteLoadBalancer rule + attribute orphan cleanup (F11).

## Why

Each bug causes real data loss or permanent IaC drift against unmodified SDK/Terraform code. Behaviors cross-checked against the official AWS ELBv2 API docs (DeleteLoadBalancer → OperationNotPermitted; RedirectActionConfig required StatusCode HTTP_301/HTTP_302; SetRulePriorities → PriorityInUse).

## Notes on the shared driver

The `loadbalancer` driver is shared by Azure/GCP. To keep this PR AWS-scoped and non-breaking, `ListenerConfig`/`ListenerInfo` keep the simple single-forward `TargetGroupARN` field (used by the Azure/GCP providers) and gain the full `DefaultActions` list that AWS uses.

## Tests

Real aws-sdk-go-v2 reproductions for every fix (deletion-protection block+clear, redirect + fixed-response round-trip, HTTPS certs/SslPolicy, SetRulePriorities PriorityInUse + swap, http-header/query-string/source-ip conditions, VpcId from subnets) plus provider-level tests. Existing ELB tests remain green. Coverage docs regenerate with zero diff (interface methods unchanged).

### DIFF
--- a/providers/aws/aws.go
+++ b/providers/aws/aws.go
@@ -251,6 +251,8 @@ func New(opts ...config.Option) *Provider {
 	p.RDS.SetSubnetResolver(p.VPC)
 	p.ElastiCache.SetSubnetResolver(p.VPC)
 	p.EC2.SetSubnetResolver(p.VPC)
+	// A load balancer's VpcId is derived from its subnets, matching ELBv2.
+	p.ELB.SetSubnetResolver(p.VPC)
 	// An IamInstanceProfile passed to RunInstances resolves through IAM so the
 	// role->profile->instance chain reads back on DescribeInstances.
 	p.EC2.SetInstanceProfileResolver(p.IAM)
--- a/providers/aws/elbv2/elb.go
+++ b/providers/aws/elbv2/elb.go
@@ -169,16 +169,49 @@ func canonicalHostedZoneID(region string) string {
 // balancer does not exist or has already been deleted, the call succeeds." So a
 // second delete (or a delete of an ARN that never existed) returns no error,
 // which keeps idempotent teardown flows (terraform destroy retries) working.
+//
+// A load balancer with deletion_protection.enabled=true cannot be deleted: real
+// ELBv2 rejects the call with OperationNotPermitted. This is the exact scenario
+// deletion protection exists to prevent, so honoring it stops an errant delete
+// (or a terraform destroy) from wiping a load balancer the user protected.
 func (m *Mock) DeleteLoadBalancer(_ context.Context, arn string) error {
-	m.lbs.Delete(arn)
+	// Only an existing load balancer can be protected; a missing ARN stays
+	// idempotent (no error), matching the API reference.
+	if _, ok := m.lbs.Get(arn); ok {
+		m.attrsMu.RLock()
+		protected := m.attrs[arn].DeletionProtection
+		m.attrsMu.RUnlock()
 
-	// Delete all listeners associated with this load balancer.
-	all := m.listeners.All()
-	for key, li := range all {
-		if li.LBARN == arn {
-			m.listeners.Delete(key)
+		if protected {
+			return errors.Newf(errors.FailedPrecondition,
+				"load balancer %q cannot be deleted because deletion protection is enabled", arn)
 		}
 	}
+
+	m.lbs.Delete(arn)
+
+	// Delete listeners for this load balancer and, with them, the rules under
+	// those listeners. Leaving the rules behind would leak them for the life of
+	// the process (their parent listener is gone, so nothing ever reaches them).
+	for key, li := range m.listeners.All() {
+		if li.LBARN != arn {
+			continue
+		}
+
+		for rkey, r := range m.rules.All() {
+			if r.ListenerARN == li.ARN {
+				m.rules.Delete(rkey)
+			}
+		}
+
+		m.listeners.Delete(key)
+	}
+
+	// Drop the stored load-balancer attributes so a recreated ARN does not
+	// inherit a prior deletion-protection flag.
+	m.attrsMu.Lock()
+	delete(m.attrs, arn)
+	m.attrsMu.Unlock()
 
 	return nil
 }
--- a/providers/aws/elbv2/elb.go
+++ b/providers/aws/elbv2/elb.go
@@ -728,10 +728,16 @@ func (m *Mock) ModifyRule(_ context.Context, input driver.ModifyRuleInput) (*dri
 
 // SetRulePriorities reassigns the priorities of the named rules and returns the
 // updated rules.
+//
+// A listener can't have two rules with the same priority, and the reorder path
+// must enforce that just as the create path does: real ELBv2 rejects a target
+// priority already held by another rule on the same listener with PriorityInUse.
+// Every rule is resolved and validated before any write, so a conflict leaves
+// the existing priorities untouched rather than half-applied.
 func (m *Mock) SetRulePriorities(
 	_ context.Context, pairs []driver.RulePriorityPair,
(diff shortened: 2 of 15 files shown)


NitinKumar004 (commented):
Reviewed against the pushed head (04539bc). Clean-built the actual tree and drove every fix through real aws-sdk-go-v2 (elasticloadbalancingv2) round-trips. All seven behavioral fixes are correct; one new lint-gate issue is the only blocker.

## Verified correct

- **deletion_protection** — a protected LB delete returns `OperationNotPermitted` and the LB survives; a missing ARN stays idempotent (no error); a new LB (default false) is deletable; clearing the flag lets the delete through. Error mapping FailedPrecondition -> `failedPreconditionCode` -> OperationNotPermitted is right, and the "deletion protection" substring match is disjoint from the "priority" branch.
- **redirect / fixed-response default actions** — redirect (HTTP_301 + Protocol/Port/Host/Path/Query) and fixed-response (StatusCode/ContentType/MessageBody) round-trip verbatim on Describe, and persist through ModifyListener. A multi-action default (forward Order 2 + redirect Order 1) preserves both type and Order.
- **SslPolicy + Certificates** — round-trip on an HTTPS listener with the create cert marked `IsDefault=true`; ModifyListener replaces both.
- **SetRulePriorities uniqueness** — a target priority already held by another rule on the listener returns `PriorityInUse`, and the pre-write validation (all pairs resolved + `checkPriorityConflicts` before any `Set`) means a rejected batch leaves both priorities untouched. A same-batch swap is allowed. In-batch duplicate slots are also caught.
- **Rule conditions** — host-header, path-pattern, http-header (name+values), query-string (key/value pairs), source-ip and http-request-method all round-trip with their typed config intact; the deprecated flat Values backfill for host/path is sensible.
- **VpcId via SubnetResolver** — `SetSubnetResolver(p.VPC)` in the AWS factory mirrors the RDS/ElastiCache/EC2 pattern; VpcId resolves from the member subnet's VPC (confirmed across two distinct VPCs and multi-subnet-same-VPC). Unwired or unresolvable subnets leave VpcId empty with no panic. `subnet_resolver.go` imports only the networking driver, so no import cycle.
- **Orphan cleanup** — DeleteLoadBalancer now drops the listeners, their rules, and the stored LB attributes; a describe against the dead listener returns nothing.

## Dual representation — safe

Keeping `TargetGroupARN` on ListenerConfig/ListenerInfo for Azure/GCP while AWS uses the new `DefaultActions` list is sound and well-documented on the driver types. No AWS production path still reads `li.TargetGroupARN`: `checkTargetGroupNotInUse` was updated to scan DefaultActions, and DescribeTargetHealth reads target-group registrations independently of the listener representation. Azure and GCP loadbalancer provider + server packages compile and pass unchanged. providers/aws/elbv2 is `-race` clean.

## Blocker (lint gate)

- `providers/aws/elbv2/elb.go:202` — new `gocritic rangeValCopy`: the DeleteLoadBalancer cleanup loop `for key, li := range m.listeners.All()` now copies a 136-byte `ListenerInfo` each iteration (the struct grew past the 128-byte threshold with DefaultActions/Certificates). This is the only issue `--new-from-rev` reports for the changed packages. Iterate by key + `Get`, index, or annotate. The dedicated "satisfy linters" commit covered the CreateListener/ModifyListener hugeParam cases but missed this one.

## Minor (non-blocking, pre-existing / out of scope)

- `describeRules` honors only `ListenerArn`, not the `RuleArns` filter — pre-existing, unrelated to this PR.
- `resolveVPCID` returns the first non-empty subnet VPC and does not error on subnets spanning different VPCs (real ELBv2 rejects that). Acceptable emulator behavior and documented.

Coverage docs regenerate with zero diff (interface methods unchanged). Fix the one rangeValCopy and this is good to merge.

NitinKumar004 (commented):
Delta re-review of the last commit (586a761b, the per-iteration ListenerInfo copy fix). Verified on the pushed head SHA.

**Lint**: the previously-flagged gocritic `rangeValCopy` at `elb.go` is cleared. golangci-lint on `providers/aws/elbv2`, `server/aws/elbv2`, and `services/loadbalancer` reports the same 18 issues as the development merge-base (dupl 4, funlen 1, goconst 3, gocyclo 1, lll 1, nolintlint 1, revive 4, wsl 3) — all pre-existing, zero introduced by this PR.

**No behavior change in the rewrite**: `memstore.Store.All()` returns a fresh snapshot map, so ranging over its keys while calling `m.listeners.Delete(key)` on the backing store is safe (no concurrent-map mutation, no skipped/off-by-one entries), and `listeners[key].Field` reads the identical struct the prior value-copy did. The two loops visit the same listeners and delete the same set as before.

- `checkTargetGroupNotInUse`: a TG referenced by a listener forward `DefaultAction` still returns ResourceInUse; deleting the listener frees it. A redirect-only listener (no forward TG) does not falsely block an unrelated TG. Both confirmed with real aws-sdk-go-v2 calls.
- `DeleteLoadBalancer` orphan cleanup: delete drops the listener and its rules; DescribeRules on the dead listener errors and the TG becomes deletable. Confirmed via SDK.

**Gates on the pushed SHA**: `go build ./...`, `go vet`, `go test` (server/providers elbv2 + loadbalancer), `go test -race` (providers/aws/elbv2), gofmt, and `go generate ./...` (zero diff) all pass.

Verdict: LGTM — the delta is a mechanical value-copy to key+map-index transform with no semantic change and clears the lint.

PR #674, opened 2026-08-26
fix(aws-route53): NoSuchHostedZone on missing-zone change; DNS reversed-label ordering; FQDN normalization; persist CallerReference/Comment; reject apex CNAME; exact-match DELETE

NitinKumar004:
Fixes 7 AWS Route 53 correctness bugs (1 HIGH + 6 MED) found by a deep real-SDK
end-to-end audit (aws-sdk-go-v2 against an in-process wire server). Each fix is
at the correct 3-layer location and cross-checked against the official Route 53
API docs. One commit per bug.

## What & why

- **[HIGH] NoSuchHostedZone on a change to a missing zone.**
  `ChangeResourceRecordSets` against a non-existent zone returned
  `InvalidChangeBatch` (400) because batch validation hit the driver's
  zone-not-found first. The zone is now checked before the batch is validated,
  so a missing zone maps to `NoSuchHostedZone` (404) as real Route 53 does. The
  existing test only asserted `err != nil` (masking the code) — strengthened to
  assert the 404/code.

- **[MED-HIGH] Reversed-label DNS ordering + pagination.**
  `ListResourceRecordSets` sorted forward-lexical, putting the zone apex in the
  middle of its subdomains. Route 53 sorts by DNS name with the labels reversed
  (apex first: `com.order.` < `com.order.a.`), then by type, and
  `NextRecordName`/`StartRecordName` pagination walks that order. Added a
  reversed-label comparator applied to both the sort and the paging start
  position, so apex NS+SOA lead and one-at-a-time paging never skips or repeats.

- **[MED] FQDN normalization.** Zone and record names are normalized to a
  trailing dot on create/change and returned as FQDNs, matching Route 53.
  Lookups match whether the client sends the dot or not; the topology `Resolve`
  engine now compares names trailing-dot-insensitively so it still resolves.

- **[MED] Persist CallerReference.** Get/List returned the zone name as the
  `CallerReference`. Added it to the DNS driver `ZoneConfig`/`ZoneInfo`,
  persisted on create, and returned verbatim.

- **[MED] Persist HostedZoneConfig.Comment.** Create/Get always returned an
  empty comment (Terraform `aws_route53_zone` perpetual-diff). Now persisted and
  emitted.

- **[MED] Reject apex CNAME.** A CNAME at the zone apex was accepted; Route 53
  rejects it (`InvalidChangeBatch`) since the apex carries SOA/NS. Now rejected.

- **[MED] Exact-match DELETE.** A DELETE with a stale/wrong TTL or values still
  removed the record. The DELETE target's TTL and value set (or alias target) is
  now validated against the stored record; a mismatch is `InvalidChangeBatch`.

Out of scope (deferred): `AssociateVPCWithHostedZone` (private-zone VPC) and the
4 low nits from the audit.

## Tests

Real aws-sdk-go-v2 reproductions added for every fix: missing-zone change ->
NoSuchHostedZone; reversed-label ordering (apex NS+SOA first, then a/m/z) and
one-record-at-a-time pagination walk; FQDN normalization with dotted/undotted
lookup; CallerReference persisted on Get/List; Comment persisted on Create/Get;
apex CNAME rejected (non-apex accepted); mismatched DELETE rejected + exact
DELETE succeeds. Existing route53 and topology Resolve tests stay green
(#467/#532/#553).

## Notes

`CallerReference`/`Comment` are additive fields on the shared DNS driver types;
Azure/GCP DNS providers carry them as zero-value pass-throughs (no behavior
change there). `go generate ./...` is idempotent (no coverage-doc diff).

### DIFF
--- a/features/topology/resolve.go
+++ b/features/topology/resolve.go
@@ -25,7 +25,10 @@ func (e *Engine) Resolve(ctx context.Context, hostname string) ([]string, error)
 		}
 
 		for _, rec := range records {
-			if rec.Name == hostname && isResolvableType(rec.Type) {
+			// Route 53 stores record names as FQDNs (trailing dot) while a query
+			// hostname may omit it, so compare trailing-dot-insensitively.
+			if strings.TrimSuffix(rec.Name, ".") == strings.TrimSuffix(hostname, ".") &&
+				isResolvableType(rec.Type) {
 				return rec.Values, nil
 			}
 		}
--- a/providers/aws/route53/route53.go
+++ b/providers/aws/route53/route53.go
@@ -133,6 +133,7 @@ func (m *Mock) CreateZone(_ context.Context, cfg driver.ZoneConfig) (*driver.Zon
 		RecordCount:     0,
 		Tags:            tags,
 		CallerReference: cfg.CallerReference,
+		Comment:         cfg.Comment,
 		Scope:           cfg.Scope,
 	}
 
--- a/providers/aws/route53/route53.go
+++ b/providers/aws/route53/route53.go
@@ -127,12 +127,13 @@ func (m *Mock) CreateZone(_ context.Context, cfg driver.ZoneConfig) (*driver.Zon
 	}
 
 	zone := driver.ZoneInfo{
-		ID:          id,
-		Name:        cfg.Name,
-		Private:     cfg.Private,
-		RecordCount: 0,
-		Tags:        tags,
-		Scope:       cfg.Scope,
+		ID:              id,
+		Name:            cfg.Name,
+		Private:         cfg.Private,
+		RecordCount:     0,
+		Tags:            tags,
+		CallerReference: cfg.CallerReference,
+		Scope:           cfg.Scope,
 	}
 
 	m.zones.Set(id, zone)
--- a/server/aws/route53/apex_cname_test.go
+++ b/server/aws/route53/apex_cname_test.go
@@ -1,0 +1,62 @@
+package route53_test
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	awsr53 "github.com/aws/aws-sdk-go-v2/service/route53"
+	r53types "github.com/aws/aws-sdk-go-v2/service/route53/types"
+)
+
+// TestSDKApexCNAMERejected locks that a CNAME at the zone apex is rejected with
+// InvalidChangeBatch, as real Route 53 does — the apex must use A/AAAA or an
+// ALIAS, since it carries the SOA and NS records.
+func TestSDKApexCNAMERejected(t *testing.T) {
+	client := newRoute53Client(t)
+	ctx := context.Background()
+
+	created, err := client.CreateHostedZone(ctx, &awsr53.CreateHostedZoneInput{
+		Name:            aws.String("apex-cname.com."),
+		CallerReference: aws.String("apex-ref"),
+	})
+	if err != nil {
+		t.Fatalf("CreateHostedZone: %v", err)
+	}
+	zoneID := aws.ToString(created.HostedZone.Id)
+
+	_, err = client.ChangeResourceRecordSets(ctx, &awsr53.ChangeResourceRecordSetsInput{
+		HostedZoneId: aws.String(zoneID),
+		ChangeBatch: &r53types.ChangeBatch{Changes: []r53types.Change{{
+			Action: r53types.ChangeActionCreate,
+			ResourceRecordSet: &r53types.ResourceRecordSet{
+				Name:            aws.String("apex-cname.com."),
+				Type:            r53types.RRTypeCname,
+				TTL:             aws.Int64(300),
+				ResourceRecords: []r53types.ResourceRecord{{Value: aws.String("target.example.com.")}},
+			},
+		}}},
+	})
+
+	var badBatch *r53types.InvalidChangeBatch
+	if !errors.As(err, &badBatch) {
+		t.Fatalf("apex CNAME: got %v, want InvalidChangeBatch", err)
+	}
+
+	// A CNAME below the apex is fine.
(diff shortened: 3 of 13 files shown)


NitinKumar004 (commented):
## Review: fix/aws-route53-dns-correctness (verified against pushed SHA 461d469)

Built and ran the pushed tree in isolation: `go build ./...`, `go vet`, and the target suites (`server/aws/route53`, `providers/aws/route53`, `services/dns`, `features/topology`) all pass; `-race` on `providers/aws/route53` clean; `gofmt -l` clean; `golangci-lint --new-from-rev=origin/development` reports 0 new issues on every changed package (the pre-existing 14 in the package are baseline); `go generate ./...` produces zero coverage diff. Six of seven fixes verified correct with real aws-sdk-go-v2 tests: missing-zone → NoSuchHostedZone(404), valid-zone/bad-batch → InvalidChangeBatch, FQDN dot/undot UPSERT collapses to one record and DELETE by either form works, apex CNAME rejected while non-apex accepted, DELETE exact-match enforces TTL + order-insensitive value set, CallerReference/Comment persisted and round-tripped, topology Resolve still resolves after normalization.

### Blocking: ListResourceRecordSets order diverges from AWS for names with characters below `.` (0x2e)

`compareDNSName`/`reversedLabels` (server/aws/route53/operations.go) compare the reversed name **label-by-label**, which is equivalent to treating the label separator as lower than any character. Real Route 53, per the ListResourceRecordSets docs, sorts by "DNS name with the labels reversed, for example `com.example.www.`" — i.e. the reversed name is compared as a **flat ASCII string including the inter-label dots and the trailing dot**, and the docs explicitly warn that "the trailing dot ... can change the sort order when the record name contains characters that appear before `.` (decimal 46) in the ASCII table ... `! " # $ % & ' ( ) * + , -`". `-` (0x2d) and `*` (0x2a) are in that set.

Consequence on ubiquitous inputs: with `api.order.com.` and `api-v2.order.com.` in a zone, AWS returns `api-v2` **before** `api` (reversed `com.order.api-v2.` < `com.order.api.`, because `-`=0x2d < `.`=0x2e), but this PR returns `api` first. Verified with an SDK test against the running server:

```
got:  [order.com./NS order.com./SOA api.order.com./A     api-v2.order.com./A]
want: [order.com./NS order.com./SOA api-v2.order.com./A  api.order.com./A]
```

Same divergence for a nested subdomain vs a hyphenated sibling (`x.svc.wc.com.` vs `svc-a.wc.com.`). Pagination is internally self-consistent (it walks the same comparator, no dup/skip against the emulator), so this is a fidelity gap vs real AWS rather than a data-loss bug — but it defeats the PR's headline "match AWS's exact order" on hyphenated hostnames (`api` vs `api-v2`, `prod` vs `prod-1`), which are everywhere. The PR's own `dns_order_test`/pagination tests only use single-char alphanumeric labels, so they never exercise the documented tricky case.

Fix is contained: build the reversed name as a flat dotted string (reverse labels, join with `.`, append trailing `.`, lowercase) and compare with `strings.Compare` instead of label-by-label. Apex-first and NS-before-SOA type tiebreak still hold under that comparison. Recommend adding a hyphen/wildcard ordering case to `dns_order_test`.

Everything else is solid; this one item is what stands between the PR and its stated goal.

NitinKumar004 (commented):
Delta re-review of the reversed-DNS-name ordering fix (commit 38eafe28).

**Ordering fix is AWS-correct.** `compareDNSName` now compares `reversedDotted(a)` vs `reversedDotted(b)` as flat ASCII strings with the trailing dot kept in place. This matches the Route 53 ListResourceRecordSets spec verbatim: results are sorted "by DNS name with the labels reversed, for example: `com.example.www.` — Note the trailing dot, which can change the sort order when the record name contains characters that appear before `.` (decimal 46)". Verified each adjacency in the tricky zone independently against the flat reversed key:

- `*.order.com.` -> `com.order.*.` : `*` (0x2a) < `a` (0x61), so wildcard sorts first. Correct.
- `a.order.com.` -> `com.order.a.` : after `com.order.`, `a` then `.`; `.` (0x2e) < `p`, so `a.` sorts before any `api*`. Correct.
- `api-v2.order.com.` (`com.order.api-v2.`) vs `api.order.com.` (`com.order.api.`): at the divergence `-` (0x2d) < `.` (0x2e), so api-v2 sorts before api. Correct — this is the exact case the old per-label compare got wrong.
- `x.api.order.com.` -> `com.order.api.x.` : `com.order.api.` is a prefix of it, so it sorts after `api.order.com.`. Correct.
- Extra cases checked: `_dmarc` (`_` 0x5f) sorts between `*` and `a`; case-insensitive (`API.` folds to `api.`); `z.com.` after `a.com.`. All correct.

**Pagination consistent.** `recordsFrom` uses the same comparator, so the MaxItems=1 walk reproduces the full-list order across the api-v2/api boundary (where naive per-label diverges). Pagination-walk tests pass.

**Blast radius: none.** `compareDNSName`/`reversedDotted` are unexported and referenced only inside `operations.go` (sort + pagination cursor). The topology engine has its own DNS resolution and does not consume this comparator, so Resolve ordering is unaffected.

**Gates (pushed SHA 38eafe28):** build, vet, targeted tests (server/providers route53, services/dns, features/topology), `-race` on providers/aws/route53, gofmt, and `go generate` (zero diff) all clean. No new lint on the changed package (the 14 existing findings are all in untouched lines: types.go, change.go, operations.go:338/441). The other 6 fixes (NoSuchHostedZone, FQDN normalization, CallerReference/Comment, apex-CNAME reject, exact-match DELETE) still pass.

**One observation (non-blocking, pre-existing, orthogonal to this commit):** the apex type tiebreak sorts types by plain ASCII (`NS` < `SOA`), and the new dns_order_test.go locks NS-before-SOA. The Route 53 API-reference example, however, returns `example.com. SOA` as the first record under `maxitems=1` with `NextRecordType=NS`, i.e. SOA sorts before NS at the apex. Worth a real-account confirmation in a follow-up; it does not affect the name-ordering fix under review here.

PR #675, opened 2026-08-26
fix(aws-ssm): persist PutParameter Tags; reject Overwrite+Tags and invalid Type; GetParametersByPath ParameterFilters

NitinKumar004:
## What

Fixes 4 real AWS SSM Parameter Store bugs found by a deep real-SDK e2e audit (repros with `aws-sdk-go-v2/service/ssm`).

### [HIGH] PutParameter dropped create-time Tags → Terraform tag drift
`PutParameter` accepted a `Tags` array but the wire request and driver `PutConfig` had no `Tags` field, so tags supplied at create were silently discarded. Terraform/CDK/CloudFormation set parameter tags inline on create and read them back with `ListTagsForResource` → cloudemu returned none → perpetual plan diff. Threaded `Tags` through the wire request → `PutConfig` → provider and store them on the parameter at create; `ListTagsForResource` now returns them.

### [MED] Overwrite + Tags accepted (AWS rejects)
`PutParameter` with both `Overwrite=true` and `Tags` is now rejected with `ValidationException` ("The Tags and Overwrite parameters can't be used at the same time."), matching real Parameter Store. Tags are only settable on create.

### [MED] Invalid Type silently coerced to String
An unrecognized `Type` (a typo like `Strng`) was coerced to `String` and stored, masking client bugs. An explicitly set `Type` is now validated against `{String, StringList, SecureString}` and an invalid value is rejected with `UnsupportedParameterType` (HTTP 400). Omitted `Type` still defaults to String on create and retains the existing type on Overwrite (unchanged).

### [MED] GetParametersByPath ignored ParameterFilters
`GetParametersByPath` returned the whole subtree regardless of `ParameterFilters`, so callers filtering a hierarchy by `Type`/`Label` received parameters they had excluded. Threaded `ParameterFilters` through the wire request → `GetByPathInput` and apply them. Supports the `Type`, `KeyId`, and `Label` keys AWS allows for this operation (Equals/BeginsWith); an unsupported key or option is rejected with `InvalidFilterKey` / `InvalidFilterOption`.

## Why

Each behavior was cross-checked against the official AWS SSM API reference (PutParameter, GetParametersByPath). The tag-drift bug in particular breaks IaC round-tripping against cloudemu.

## Tests

Real `aws-sdk-go-v2` SDK reproductions at the wire layer plus provider-level unit tests:
- Tags persisted on create; Overwrite+Tags → ValidationException
- Invalid Type → UnsupportedParameterType; valid types accepted; Overwrite with omitted Type retains type
- ByPath Type=String filter returns only String params; Label filter; unsupported key → InvalidFilterKey; unsupported option → InvalidFilterOption

Existing SSM tests (versioning/overwrite/selectors/history/labels/path/pagination) stay green. Deferred (not in scope): AllowedPattern enforcement, SecureString KeyId/ciphertext-at-rest, parameter-name validation, Intelligent-Tiering echo.

### DIFF
--- a/providers/aws/ssm/path_filter.go
+++ b/providers/aws/ssm/path_filter.go
@@ -1,0 +1,106 @@
+package ssm
+
+import (
+	"strings"
+
+	"github.com/stackshy/cloudemu/v2/services/parameterstore/driver"
+)
+
+// GetParametersByPath filter keys and options. GetParametersByPath supports the
+// Type, KeyId, and Label keys only (unlike DescribeParameters, which also
+// supports Name/Tier/DataType/tag).
+const (
+	byPathKeyType  = "Type"
+	byPathKeyKeyID = "KeyId"
+	byPathKeyLabel = "Label"
+
+	filterOptionEquals     = "Equals"
+	filterOptionBeginsWith = "BeginsWith"
+)
+
+// validateByPathFilters checks each ParameterFilters entry uses a key and option
+// GetParametersByPath supports, returning the AWS-distinct error otherwise.
+func validateByPathFilters(filters []driver.ParameterStringFilter) error {
+	for _, f := range filters {
+		switch f.Key {
+		case byPathKeyType, byPathKeyKeyID, byPathKeyLabel:
+		default:
+			return driver.ErrInvalidFilterKey
+		}
+
+		switch f.Option {
+		case "", filterOptionEquals, filterOptionBeginsWith:
+		default:
+			return driver.ErrInvalidFilterOption
+		}
+	}
+
+	return nil
+}
+
+// matchesByPathFilters reports whether version v of a parameter satisfies every
+// filter (filters are AND'd; a filter's Values are OR'd).
+func matchesByPathFilters(v *version, filters []driver.ParameterStringFilter) bool {
+	for i := range filters {
+		if !matchByPathFilter(v, &filters[i]) {
+			return false
+		}
+	}
+
+	return true
+}
+
+func matchByPathFilter(v *version, f *driver.ParameterStringFilter) bool {
+	switch f.Key {
+	case byPathKeyType:
+		return fieldMatchesFilter(v.typ, f.Option, f.Values)
+	case byPathKeyLabel:
+		return labelsMatchFilter(v.labels, f.Option, f.Values)
+	default:
+		// KeyId is a valid key but cloudemu doesn't model a per-parameter KMS
+		// KeyId (SecureString isn't encrypted), so it can't constrain results.
+		return true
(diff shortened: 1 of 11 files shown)


NitinKumar004 (commented):
Verified against the pushed head (37db0763) in a clean checkout. Gate matrix green: build, vet, gofmt, go generate (zero diff — confirms the new PutConfig.Tags / GetByPathInput.ParameterFilters are struct fields, not interface method-signature changes), golangci-lint (new-only, 0 issues), package tests, and -race on providers/aws/ssm.

Wrote independent aws-sdk-go-v2 SSM tests beyond the PR's; all pass:

Tags
- Inline Tags on create persist and both are returned by ListTagsForResource; a tagless create returns an empty tag list (no phantom tag).
- AddTagsToResource after a create merges (doesn't clobber inline tags).
- Tags survive an Overwrite update that passes no Tags (AWS keeps existing tags). This is the highest-value case and wasn't in the PR's own suite — it works because tags live on paramData and overwriteParameter never touches them. Good.

Overwrite + Tags / invalid Type
- Overwrite=true + Tags -> ValidationException; Overwrite=true with no Tags -> succeeds (v2); create with Tags and no Overwrite on a new name -> succeeds with tags. The SetIfAbsent race-retry path is safe: it's only reachable with Overwrite already true, and Overwrite+Tags is rejected up front, so the recursive call never carries tags.
- Type="Bogus" -> UnsupportedParameterType (400) and the param is not stored (GetParameter -> ParameterNotFound); String/StringList/SecureString all accepted; Overwrite with omitted Type retains the stored type (verified StringList stays StringList).

GetParametersByPath ParameterFilters
- Type Equals filter isolates String vs StringList vs SecureString; Label Equals and BeginsWith both work; unsupported keys (Name, tag:Env) -> InvalidFilterKey; bad option (Contains) -> InvalidFilterOption. Supported key set (Type/KeyId/Label) matches the AWS GetParametersByPath reference; KeyId accepted-but-non-constraining is a reasonable modeling choice given SecureString isn't encrypted here.
- Filtering runs in the driver (full sorted set) and pagination in the handler over that set, so Recursive + MaxResults/NextToken paginates the filtered set with no dupes or omissions (verified across pages).

Minor, non-blocking:
- The Label filter matches only the latest version's labels; real SSM can match a label attached to any version while still returning latest. Fine for the modeled surface.
- Error message strings are close paraphrases of the AWS text rather than byte-exact, but the wire error codes (ValidationException / UnsupportedParameterType / InvalidFilterKey / InvalidFilterOption) are what SDK clients match, and all resolve correctly.

No blocking issues.

PR #676, opened 2026-08-26
fix(aws-kms): implement GetPublicKey + advertise Signing/Mac/asymmetric-Encryption algorithms on KeyMetadata

NitinKumar004:
## What

Two fixes to the AWS KMS asymmetric-key surface found by the deep real-user e2e audit.

### [HIGH] Implement `GetPublicKey`
`GetPublicKey` was unrouted, so the wire server returned `UnknownOperationException`. Every real user of an asymmetric KMS key (SIGN_VERIFY, or RSA ENCRYPT_DECRYPT) calls it to download the public key for offline signature verification / offline encryption / hybrid crypto (JWT signers, mTLS, `data.aws_kms_public_key` in Terraform).

- Added `GetPublicKey` to the driver `Crypto` interface, a provider impl (`crypto_asym.go`) that marshals the existing `keyData.privKey` to DER `SubjectPublicKeyInfo` via `x509.MarshalPKIXPublicKey`, and a wire op registered in the handler.
- Response carries `KeyId`, `PublicKey` (DER SPKI), `KeySpec`/`CustomerMasterKeySpec`, `KeyUsage`, and the spec-matched `SigningAlgorithms` (SIGN_VERIFY) or `EncryptionAlgorithms` (ENCRYPT_DECRYPT).
- A symmetric key is rejected with `UnsupportedOperationException` (new `ErrUnsupportedOperation` sentinel).
- `Decrypt` now also accepts **raw RSA-OAEP ciphertext** produced offline from the downloaded public key (asymmetric path keyed on the supplied `KeyId`), so offline-encrypt then KMS `Decrypt` round-trips. The existing self-describing wrapped-blob path and every confirmed behavior (IncorrectKey, InvalidCiphertext, encryption-context binding) are unchanged.

### [MED] KeyMetadata algorithm lists
`CreateKey`/`DescribeKey` returned empty algorithm lists for asymmetric and HMAC keys. Real KMS always advertises, per KeySpec+KeyUsage:

- RSA SIGN_VERIFY -> `SigningAlgorithms` (RSASSA_PSS + RSASSA_PKCS1_V1_5, SHA 256/384/512)
- ECC P256/P384/P521 -> `SigningAlgorithms` (ECDSA_SHA_256/384/512)
- RSA ENCRYPT_DECRYPT -> `EncryptionAlgorithms` (RSAES_OAEP_SHA_1, RSAES_OAEP_SHA_256)
- SYMMETRIC_DEFAULT -> `EncryptionAlgorithms` (SYMMETRIC_DEFAULT)
- HMAC_* -> `MacAlgorithms` (HMAC_SHA_*)

Added `EncryptionAlgorithms`/`SigningAlgorithms`/`MacAlgorithms` to the driver `KeyMetadata` and wire shape, populated on `CreateKey` from one spec->algorithms mapping in the driver (reused by `GetPublicKey`), and rendered on `CreateKey`/`DescribeKey`. `ReplicateKey` carries them via its metadata copy.

## Why

Both were flagged in the KMS deep real-user e2e audit as the only remaining gaps in an otherwise strong real-crypto KMS. Without `GetPublicKey` and the algorithm lists, offline crypto workflows and any SDK/app that inspects the advertised algorithms break against the emulator.

## Tests

Real `aws-sdk-go-v2/kms` reproductions plus provider-level tests:
- GetPublicKey on RSA SIGN_VERIFY -> DER parses to `*rsa.PublicKey`; offline `VerifyPSS` of a KMS `Sign` signature succeeds.
- GetPublicKey on RSA ENCRYPT_DECRYPT -> offline RSA-OAEP encrypt then KMS `Decrypt` recovers the plaintext.
- GetPublicKey on ECC -> correct spec + `*ecdsa.PublicKey` + single ECDSA algorithm.
- GetPublicKey on a symmetric key -> `UnsupportedOperationException`.
- KeyMetadata algorithm lists for RSA SIGN_VERIFY / RSA ENCRYPT_DECRYPT / ECC / HMAC / SYMMETRIC on both `CreateKey` and `DescribeKey`.

All existing KMS tests stay green. Coverage docs regenerated via `go generate`.

### DIFF
--- a/docs/coverage/README.md
+++ b/docs/coverage/README.md
@@ -36,7 +36,7 @@ capability the code does not implement. Machine-readable: [`coverage.json`](./co
 | `kafka` | [Kafka](./aws/kafka.md) | — | — | — | 59 |
 | `keyspaces` | [Keyspaces](./aws/keyspaces.md) | — | — | — | 18 |
 | `kinesis` | [Kinesis](./aws/kinesis.md) | — | — | — | 39 |
-| `kms` | [KMS](./aws/kms.md) | — | — | — | 45 |
+| `kms` | [KMS](./aws/kms.md) | — | — | — | 46 |
 | `loadbalancer` | [ELB](./aws/elb.md) | [LB](./azure/lb.md) | [LB](./gcp/lb.md) | — | 19 |
 | `logging` | [CloudWatchLogs](./aws/cloudwatchlogs.md) | [LogAnalytics](./azure/loganalytics.md) | [CloudLogging](./gcp/cloudlogging.md) | — | 14 |
 | `managedcassandra` | — | [ManagedCassandra](./azure/managedcassandra.md) | — | — | 15 |
--- a/docs/coverage/aws/README.md
+++ b/docs/coverage/aws/README.md
@@ -24,7 +24,7 @@ Services cloudemu emulates for AWS, by native name. Back to the [cross-provider 
 | [Glue](./glue.md) | `glue` | 299 |
 | [GuardDuty](./guardduty.md) | `guardduty` | 87 |
 | [IAM](./iam.md) | `iam` | 40 |
-| [KMS](./kms.md) | `kms` | 45 |
+| [KMS](./kms.md) | `kms` | 46 |
 | [Kafka](./kafka.md) | `kafka` | 59 |
 | [Keyspaces](./keyspaces.md) | `keyspaces` | 18 |
 | [Kinesis](./kinesis.md) | `kinesis` | 39 |
--- a/docs/coverage/aws/kms.md
+++ b/docs/coverage/aws/kms.md
@@ -3,7 +3,7 @@
 
 AWS's `kms` service · portable interface `driver.KMS` · [AWS index](./README.md)
 
-## Operations (45)
+## Operations (46)
 
 | Operation | Description |
 | --- | --- |
@@ -29,6 +29,7 @@ AWS's `kms` service · portable interface `driver.KMS` · [AWS index](./README.m
 | `GetKeyPolicy` | Key policies. |
 | `GetKeyRotationStatus` |  |
 | `GetParametersForImport` |  |
+| `GetPublicKey` |  |
 | `ImportKeyMaterial` |  |
 | `ListAliases` |  |
 | `ListGrants` |  |
@@ -71,6 +72,7 @@ Crypto is the cryptographic surface of KMS. It is embedded in the KMS
 | `GenerateDataKeyWithoutPlaintext` |  |
 | `GenerateMac` |  |
 | `GenerateRandom` |  |
+| `GetPublicKey` |  |
 | `ReEncrypt` |  |
 | `Sign` |  |
 | `Verify` |  |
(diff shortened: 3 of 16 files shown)


NitinKumar004 (commented):
Reviewed the pushed head (`deaf0da`) with a clean build and independent aws-sdk-go-v2 reproductions, with focus on the new raw-asymmetric `Decrypt` fallback. Verdict: looks good to merge.

## Gates (pushed SHA)
- `go build ./...`, `go vet` (kms pkgs), `gofmt -l`: clean
- `go test ./server/aws/kms/... ./providers/aws/kms/... ./services/kms/...`: green
- `go test -race ./providers/aws/kms/...` and race on the server kms pkg: clean
- `golangci-lint` on the three changed packages: 0 issues
- `go generate ./...`: zero diff — coverage regen is idempotent, KMS op count reflects the new `GetPublicKey`

## GetPublicKey
- RSA SIGN_VERIFY: DER parses to `*rsa.PublicKey`; offline `rsa.VerifyPSS` of a KMS `Sign` signature passes; 6 signing algs, no encryption algs.
- RSA ENCRYPT_DECRYPT: 2 encryption algs; offline `EncryptOAEP` with the returned key then KMS `Decrypt(KeyId)` recovers plaintext.
- ECC: correct spec, single ECDSA alg, `*ecdsa.PublicKey` parses.
- Symmetric: `UnsupportedOperationException` (new `ErrUnsupportedOperation` maps to the exact wire code).

## Decrypt fallback — security
Wrote adversarial SDK tests; all pass:
- Wrong/mismatched KeyId on a raw offline ciphertext -> `InvalidCiphertextException` (OAEP integrity fails; does not decrypt).
- Random/garbage ciphertext -> `InvalidCiphertextException`, no panic, no partial plaintext.
- Fallback is gated: reached only after the self-describing wrapped path declines, requires an explicit KeyId, and only proceeds for `*rsa.PrivateKey` + `ENCRYPT_DECRYPT` (raw ciphertext under a SIGN_VERIFY RSA key -> `InvalidKeyUsageException`; symmetric key has nil private key -> rejected, so the symmetric AEAD path cannot be bypassed).
- Symmetric AEAD binding unchanged: wrong or missing `EncryptionContext` -> failure; wrong KeyId on a symmetric wrapped blob -> `IncorrectKeyException`.
- Wrapped-blob round-trip intact for both symmetric and asymmetric (KMS Encrypt -> KMS Decrypt).
- Errors are sentinels; no key material leaked in error text. Encrypt path untouched.

## KeyMetadata algorithm lists
Populated at `CreateKey` from a single `spec+usage` mapping (reused by `GetPublicKey`); symmetric ENCRYPT_DECRYPT still advertises `SYMMETRIC_DEFAULT` (no regression), and `ReplicateKey` carries them via the metadata copy.

## Minor, non-blocking
- `GetPublicKey` runs `requireUsable`, so a Disabled key returns `DisabledException`. Real KMS returns the public key for a disabled key (it is metadata retrieval, not a crypto op) — worth confirming, but harmless.
- Asymmetric `Decrypt` ignores the requested `EncryptionAlgorithm` and tries both OAEP hashes; more lenient than real KMS (which rejects a mismatch) but not a correctness/security concern.

PR #678, opened 2026-08-26
fix(aws-cwlogs): timestamp-order GetLogEvents + deterministic FilterLogEvents interleave; unset default retention; metric-filter fields; orderBy; missing-stream error

NitinKumar004:
## What

Fixes six CloudWatch Logs ordering / retrieval / retention / metric-filter divergences found by the AWS deep-e2e real-user audit (driven with the real `aws-sdk-go-v2/cloudwatchlogs` client). Each is committed separately.

- **Timestamp-order GetLogEvents** — events are returned in timestamp order (stable on ties), not ingestion order. The stream's backing slice is kept sorted on insert and the multi-stream merge is sorted. Fixes out-of-order puts returning `[late, early]`.
- **Deterministic FilterLogEvents interleave** — matches across streams are sorted by `(timestamp, stream)` after collection, so results are ascending and identical run-to-run instead of grouped per stream in random map order.
- **Unset default log-group retention** — a group created without a retention policy stays unset (never expire) instead of defaulting to 30 days; `DescribeLogGroups` omits `retentionInDays`, and the phantom 30-day event-drop cutoff no longer runs when retention is unset. `PutRetentionPolicy` still sets and reflects it.
- **Metric-filter round-trip** — `MetricTransformation` `DefaultValue` / `Unit` / `Dimensions` are added to the logging driver config/info, persisted, and echoed back on `DescribeMetricFilters` (previously dropped).
- **DescribeLogStreams orderBy + descending** — decode `orderBy` (`LogStreamName` default, or `LastEventTime`) and the `descending` flag; the newest-stream lookup now returns the right stream.
- **Missing-stream error** — `GetLogEvents` on a nonexistent stream returns `ResourceNotFoundException` instead of an empty page.

## Why

These are real behavioral divergences from AWS (verified against the official CloudWatch Logs API reference): log-tail tooling and Terraform data sources depend on timestamp ordering; `aws_cloudwatch_log_group` with no `retention_in_days` and `aws_cloudwatch_log_metric_filter` with `default_value`/`unit` were showing perpetual drift; the `orderBy=LastEventTime, descending, limit=1` newest-stream pattern returned the wrong stream.

Subscription filters and CWL filter-pattern grammar are intentionally out of scope (separate follow-ups).

## Tests

Real `aws-sdk-go-v2` wire reproductions plus provider-level tests for each fix. Existing CW Logs tests (metric-filter increment, pagination, retention table, DeleteLogGroup cascade, `-race` concurrent PutLogEvents) stay green; two existing assertions that encoded the old buggy behavior (default retention 30, missing-stream empty) were updated to the AWS-correct expectations. Coverage docs regenerated (no diff).

### DIFF
--- a/providers/aws/cloudwatchlogs/cloudwatchlogs.go
+++ b/providers/aws/cloudwatchlogs/cloudwatchlogs.go
@@ -456,9 +456,20 @@ func (m *Mock) FilterLogEvents(
 		results = m.filterStreamEvents(g, input.LogStream, input)
 	} else {
 		for name, s := range g.streams.All() {
-			events := m.matchEvents(s, name, input)
-			results = append(results, events...)
+			results = append(results, m.matchEvents(s, name, input)...)
 		}
+
+		// CloudWatch Logs interleaves matches across streams in ascending
+		// timestamp order; sort by (timestamp, stream) so the result is both
+		// time-ordered and deterministic run-to-run regardless of the random
+		// map-iteration order above. Ties break on stream name.
+		sort.SliceStable(results, func(i, j int) bool {
+			if !results[i].Timestamp.Equal(results[j].Timestamp) {
+				return results[i].Timestamp.Before(results[j].Timestamp)
+			}
+
+			return results[i].LogStream < results[j].LogStream
+		})
 	}
 
 	// A negative limit means "no cap" — the wire handler fetches the full
--- a/providers/aws/cloudwatchlogs/cloudwatchlogs.go
+++ b/providers/aws/cloudwatchlogs/cloudwatchlogs.go
@@ -558,6 +558,9 @@ func (m *Mock) PutMetricFilter(
 		MetricName:      cfg.MetricName,
 		MetricNamespace: cfg.MetricNamespace,
 		MetricValue:     cfg.MetricValue,
+		DefaultValue:    cfg.DefaultValue,
+		Unit:            cfg.Unit,
+		Dimensions:      maps.Clone(cfg.Dimensions),
 		CreatedAt:       m.opts.Clock.Now().UTC(),
 	}
 
--- a/providers/aws/cloudwatchlogs/cloudwatchlogs.go
+++ b/providers/aws/cloudwatchlogs/cloudwatchlogs.go
@@ -4,6 +4,7 @@ package cloudwatchlogs
 import (
 	"context"
 	"maps"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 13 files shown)


NitinKumar004 (commented):
Reviewed the pushed head (d22df75) in a clean checkout: built, vetted, ran the CW Logs suites plus an independent adversarial aws-sdk-go-v2 pass. Verdict: looks good to merge.

**HIGH — GetLogEvents timestamp order**
- Out-of-order ingestion (put +5s before +1s) returns ascending; a 250-event set put in reverse comes back fully ascending. Equal timestamps keep ingestion order (stable sort verified with three same-ts puts).
- Tail-window default preserved: with no `startFromHead` the latest window is returned ascending; the service default cap is 10,000 (correct — not 100), and an explicit `Limit=100` over 250 events returns the tail slice `[150,250)` ascending.
- Forward and backward `nextToken` walks (page size 40/50) each cover the full set exactly once — no dupes, no skips, and terminate on the stable token.

**HIGH — FilterLogEvents deterministic interleave**
- Matches across 4 streams at mixed timestamps come back ascending `[t0..t5]`; the same query run 10x is byte-identical every time (the random map-iteration nondeterminism is gone). Substring `filterPattern` still filters.

**Retention**
- CreateLogGroup without a policy → DescribeLogGroups omits `retentionInDays` (nil, not 0/30); a 40-day-old event is retained.
- PutRetentionPolicy(7) reflected; the latent `AddDate` cutoff bug is fixed — retention=1 drops a 2-day-old event while keeping a fresh one, and an unset group drops nothing.

**Metric-filter fields**
- `DefaultValue` is `*float64` and round-trips `0` as `0` (distinguished from unset, which stays nil); `Unit` and `Dimensions` round-trip through DescribeMetricFilters. Note: the emitted metric datum in `evaluateMetricFilters` does not carry Unit/Dimensions onto the published CloudWatch metric — round-trip and the #568 increment are intact, but the emitted custom metric is still dimensionless. Minor fidelity gap, not a regression.

**DescribeLogStreams orderBy**
- `orderBy=LastEventTime, descending=true` puts the newest-last-event stream first; a no-event stream sorts sanely (empty LastEvent → epoch 0). Default `LogStreamName` ascending unchanged.

**Missing resources**
- GetLogEvents on a missing stream and on a missing group both return ResourceNotFoundException.

**Blast radius**
- The two updated existing assertions encoded the old bugs (default-30 retention, missing-stream-returns-empty) — correctly flipped to real-AWS behavior.
- Lock discipline is sound: the insert-time sort runs under the stream write lock; multi-stream/filter sorts operate on local slices. `-race` with concurrent PutLogEvents + GetLogEvents/FilterLogEvents on the same stream is clean (3x).
- Build/vet clean; go generate zero diff; gofmt clean; golangci-lint introduces zero new issues (the 8 reported all pre-exist at the merge-base).

PR #679, opened 2026-08-26
fix(aws-sfn): require an updatable field on UpdateStateMachine + roleArn on CreateStateMachine

NitinKumar004:
## What

Two AWS Step Functions state-machine validation fixes found by the real-user e2e audit (F3, F5). Scoped to CreateStateMachine / UpdateStateMachine request validation — the ASL execution engine is untouched (separately deferred).

### [MED F3] UpdateStateMachine with no updatable field silently succeeded
`UpdateStateMachine` accepted a request carrying neither `definition`, `roleArn`, nor logging/tracing/encryption config, returned HTTP 200, and bumped `revisionId` on a no-op. Real SFN rejects this with `MissingRequiredParameter` ("This error occurs if both definition and roleArn are not specified" — API_UpdateStateMachine).

Fix: reject an update supplying none of the updatable fields with `MissingRequiredParameter`, before any mutation, so the revision is not bumped. A valid update (definition/roleArn/logging/tracing/encryption present) still works and bumps the revision.

### [LOW F5] roleArn not required/validated on CreateStateMachine
`CreateStateMachine` accepted an empty or malformed `roleArn` and stored it. AWS marks `roleArn` required (a valid IAM role ARN). The aws-sdk-go-v2 client-side check is a nil-check, so an explicit empty string reaches the server.

Fix: require a valid IAM role ARN shape (`arn:<partition>:iam::<account>:role/<name>`); empty or malformed → `InvalidArn` (the error real SFN returns for a bad role ARN), validated after name/definition checks so those errors still take precedence.

## Why

Both are real deviations from the AWS Step Functions API verified against the official API reference (API_CreateStateMachine, API_UpdateStateMachine) with a real aws-sdk-go-v2 sfn client. Errors surface as their own exception types via the existing `driver.APIError` wire mapping (new `MissingRequiredParameter` exception constant added).

## Tests

- Server (real aws-sdk-go-v2): empty/malformed roleArn → `InvalidArn`, valid → create; empty update → `MissingRequiredParameter` with unchanged `revisionId`, definition update → new `revisionId`.
- Provider: direct-driver equivalents asserting the exact SFN exception.
- Existing SFN suite (create/idempotency/versions/aliases/error-codes/pagination) green, `-race` clean.

Coverage docs regenerated — no diff (no driver interface change).

### DIFF
--- a/providers/aws/sfn/errors.go
+++ b/providers/aws/sfn/errors.go
@@ -78,6 +78,12 @@ func invalidName(msg string) error {
 	return &driver.APIError{Exception: driver.ExInvalidName, Err: errors.New(errors.InvalidArgument, msg)}
 }
 
+// missingRequiredParameter builds a MissingRequiredParameter-tagged error,
+// returned when UpdateStateMachine supplies none of its updatable fields.
+func missingRequiredParameter(msg string) error {
+	return &driver.APIError{Exception: driver.ExMissingRequiredParameter, Err: errors.New(errors.InvalidArgument, msg)}
+}
+
 // invalidExecutionInput builds an InvalidExecutionInput-tagged error, returned
 // when a StartExecution Input is not valid JSON.
 func invalidExecutionInput(msg string) error {
--- a/providers/aws/sfn/sfn.go
+++ b/providers/aws/sfn/sfn.go
@@ -30,6 +30,8 @@ const (
 	arnScheme = "arn"
 	// statesService is the SFN service segment of a states ARN.
 	statesService = "states"
+	// iamService is the IAM service segment of a role ARN.
+	iamService = "iam"
 	// emptyJSON is the default output/input payload when none is supplied.
 	emptyJSON = "{}"
 )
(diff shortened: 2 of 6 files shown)


NitinKumar004 (commented):
Reviewed against the pushed head (c827fd81), clean-built and re-verified with independent aws-sdk-go-v2 and direct-driver tests.

**Verdict: approve.** Both fixes are correct and match the real Step Functions API.

Gates on the pushed tree: build, vet, `go test` (server+provider+driver), `-race` on the provider, gofmt, and golangci-lint on all three changed packages — all clean (0 lint). `go generate ./...` produces zero diff (no driver-interface change).

Behavior verified with my own tests:
- **F3** — UpdateStateMachine with only the ARN → `MissingRequiredParameter`, and the guard runs before the mutation so `revisionId` is unchanged. A definition update and a roleArn-only update both succeed and bump the revision.
- **F5** — empty and malformed roleArn → `InvalidArn`; valid IAM role ARN creates. `validRoleARN` correctly rejects an S3 ARN, a states ARN, a `user/` ARN, and an empty role name; it accepts other partitions (`aws-cn`).
- **Precedence** — invalid definition + bad roleArn → `InvalidDefinition`; missing name + bad roleArn → `InvalidName`. Name/definition validation keeps precedence as claimed.
- **Ordering** — duplicate name + valid roleArn → `StateMachineAlreadyExists`; duplicate name + invalid roleArn → `InvalidArn`. roleArn validation precedes the uniqueness claim, which is the right order (request-parameter validation ahead of resource-state conflict). `TestCreateDuplicateNameFails` is not weakened — the added roleArn is required to reach the uniqueness check, and the collision assertion is preserved.

One non-blocking observation for a follow-up (not introduced by this change, but surfaced by it):

An SDK `UpdateStateMachine` carrying **only** `loggingConfiguration` / `tracingConfiguration` / `encryptionConfiguration` (no definition/roleArn) now returns `MissingRequiredParameter`, whereas real SFN returns 200 and applies the config. Root cause is a pre-existing gap in the server handler: `server/aws/sfn/operations.go` never wires `LoggingConfigJSON`/`TracingConfigJSON`/`EncryptionCfgJSON` into the driver input on create or update (describe returns defaults), so `hasUpdatableField` can never see them set over the wire. The provider/driver layer treats those fields as updatable correctly — only the wire path can't reach it. Before this change that path was a silent 200 no-op; it is now a hard error. Worth a follow-up to wire those three fields through the create/update handlers.

Minor: `validRoleARN` does not require a non-empty account segment, so `arn:aws:iam:::role/x` is accepted. Harmless in practice (SDKs emit fully-qualified ARNs).

PR #680, opened 2026-08-26
fix(aws-efs): derive mount-target VpcId/AZ from subnet; honor Backup + One Zone AZ; FileSystemAlreadyExists carries id; access-point ClientToken idempotency

NitinKumar004:
## Summary

Five real AWS EFS bugs found by a deep end-to-end audit (real `aws-sdk-go-v2/service/efs` client over the wire), each fixed at the correct 3-layer location and verified against the official EFS API docs.

### Fixes (one commit each)

- **FileSystemAlreadyExists carries the existing FileSystemId** — a duplicate `CreationToken` now returns a `FileSystemAlreadyExists` error that includes the existing file system's id, so an idempotent retry can recover it (the whole point of the idempotency contract). Driver `ResourceError` grew an optional `ResourceID`; the restJson1 error body emits `FileSystemId`.
- **Honor `CreateFileSystem` Backup + One Zone default** — the `Backup` flag was ignored, leaving every file system's backup policy DISABLED. `Backup=true` now enables it, and One Zone file systems (AvailabilityZoneName set) default to ENABLED per real EFS. `PutBackupPolicy` still overrides.
- **Set AvailabilityZoneId on One Zone file systems** — a One Zone file system now reports the consistent `AvailabilityZoneId` (e.g. `us-east-1b` → `use1-az2`) derived from the zone name; DescribeFileSystems emits it.
- **Derive mount-target VpcId and AZ from the subnet** — `CreateMountTarget` fabricated a fresh random `VpcId` per call and hardcoded the region's first AZ, so mount targets of one file system reported different VPCs and a constant zone. A `SubnetResolver` (wired to VPC, mirroring RDS/ELB/ElastiCache) now derives the VpcId and AZ from the subnet: all mount targets of a file system share a VpcId and each reflects its subnet's zone, with the one-per-AZ conflict keyed on AZ. When the subnet is unresolvable it falls back to a deterministic per-file-system VpcId instead of a random one.
- **Enforce CreateAccessPoint ClientToken idempotency** — a repeated `ClientToken` created a brand-new access point every call. It now atomically claims the token and a repeat returns the existing access point, mirroring the `CreateFileSystem` CreationToken idempotency.

A final refactor commit extracts small helpers so the fixes stay under the lint complexity/receiver/cuddle rules.

### Tests

Real aws-sdk-go-v2 EFS reproductions added: duplicate-token error carries the id; Backup=true and One Zone default → ENABLED; One Zone AvailabilityZoneId; two mount targets in subnets of one VPC → same VpcId + subnet-matched AZ; two CreateAccessPoint with one ClientToken → one fsap id. Existing EFS tests (CreationToken idempotency, per-subnet MountTargetConflict, NumberOfMountTargets, FileSystemInUse guard, mode round-trip, tags) stay green.

### Gates

`go build ./...`, `go vet`, scoped `go test` (server/providers/services efs), `go test -race` (providers/aws/efs), `golangci-lint` (0 issues), `gofmt` all green. Coverage docs regenerated with no diff (no driver method signatures changed).

### DIFF
--- a/providers/aws/aws.go
+++ b/providers/aws/aws.go
@@ -253,6 +253,9 @@ func New(opts ...config.Option) *Provider {
 	p.EC2.SetSubnetResolver(p.VPC)
 	// A load balancer's VpcId is derived from its subnets, matching ELBv2.
 	p.ELB.SetSubnetResolver(p.VPC)
+	// EFS mount targets derive their VpcId and AZ from the subnet, so all mount
+	// targets of a file system share a VpcId and each reflects its subnet's zone.
+	p.EFS.SetSubnetResolver(p.VPC)
 	// An IamInstanceProfile passed to RunInstances resolves through IAM so the
 	// role->profile->instance chain reads back on DescribeInstances.
 	p.EC2.SetInstanceProfileResolver(p.IAM)
--- a/providers/aws/efs/efs.go
+++ b/providers/aws/efs/efs.go
@@ -38,6 +38,10 @@ type Mock struct {
 	// via SetIfAbsent so concurrent same-token CreateFileSystem calls can't both
 	// create (EFS idempotency).
 	tokenIndex *memstore.Store[string]
+	// apTokenIndex maps a CreateAccessPoint ClientToken to its access-point id,
+	// claimed atomically via SetIfAbsent so a repeated ClientToken returns the
+	// existing access point rather than creating a duplicate (EFS idempotency).
+	apTokenIndex *memstore.Store[string]
 
 	// accountPref is the account-level resource-id preference ("LONG_ID" |
 	// "SHORT_ID"); empty until PutAccountPreferences is called.
@@ -55,11 +59,12 @@ type Mock struct {
 // New creates a new EFS mock with the given configuration options.
 func New(opts *config.Options) *Mock {
 	return &Mock{
-		fileSystems: memstore.New[*fsData](),
-		mtIndex:     memstore.New[string](),
-		apIndex:     memstore.New[string](),
-		tokenIndex:  memstore.New[string](),
-		opts:        opts,
+		fileSystems:  memstore.New[*fsData](),
+		mtIndex:      memstore.New[string](),
+		apIndex:      memstore.New[string](),
+		tokenIndex:   memstore.New[string](),
+		apTokenIndex: memstore.New[string](),
+		opts:         opts,
 	}
 }
 
--- a/providers/aws/efs/efs.go
+++ b/providers/aws/efs/efs.go
@@ -44,6 +44,11 @@ type Mock struct {
 	accountPref string
 	prefMu      sync.RWMutex
 
+	// subnetResolver derives a mount target's VpcId and AZ from its subnet. Real
+	// EFS infers both from the subnet; without a resolver a deterministic
+	// per-file-system fallback is used instead of a random per-call VpcId.
+	subnetResolver SubnetResolver
+
 	opts *config.Options
 }
 
--- a/providers/aws/efs/efs_test.go
+++ b/providers/aws/efs/efs_test.go
@@ -2,6 +2,7 @@ package efs_test
 
 import (
 	"context"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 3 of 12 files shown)


NitinKumar004 (commented):
Reviewed against the pushed head `0b14055f`, clean-built and tested on that exact tree.

## Verified working (real aws-sdk-go-v2 + provider tests)
- **F1** duplicate CreationToken → `FileSystemAlreadyExists` whose `FileSystemId` member equals the existing fs id. The SDK deserializes it from the `FileSystemId` JSON body member added to `errorBody`, gated on `kind == KindFileSystem`. Confirmed via `TestSDKDuplicateTokenCarriesFileSystemId`. No object-lookup on the loser path here, so no race.
- **F2** `CreateFileSystem(Backup=true)` → `DescribeBackupPolicy` ENABLED; One Zone (`AvailabilityZoneName` set) → ENABLED default even without Backup; regular fs → DISABLED; `PutBackupPolicy(DISABLED)` overrides. `backupOnCreate` logic correct.
- **F3** One Zone fs → `AvailabilityZoneId` set (`us-east-1b`→`use1-az2`); regional fs → empty. `azIDFromName` ordinal mapping correct.
- **F4** (cross-service) mount targets in different subnets of one VPC share the derived `VpcId`, each reports its subnet's AZ name + id; different VPCs → different VpcId; one-per-AZ conflict keyed on AZ when the subnet resolves, per-subnet fallback otherwise. `SetSubnetResolver(p.VPC)` wired in `providers/aws/aws.go`, no import cycle, nil-safe when unwired (falls back to deterministic per-fs `vpc-<fsid-suffix>` + region az1). Deterministic fallback VpcId confirmed stable across calls.

## Blocking issue — CreateAccessPoint ClientToken idempotency has a TOCTOU race (F5)
`providers/aws/efs/mount_targets.go` `CreateAccessPoint`: the winner claims the token via `apTokenIndex.SetIfAbsent(token, id)` and only afterwards takes `fd.mu.Lock()` and stores the AP object + `apIndex.Set(id, fsID)`. A concurrent same-token caller that loses `SetIfAbsent` reads the winner's id from `apTokenIndex.Get` and calls `accessPointByID(existingID)`, which does `apIndex.Get(existingID)` — but the winner may not have reached `apIndex.Set` yet, so the loser gets a spurious `AccessPointNotFound` instead of the existing access point.

Reproduced reliably: 40 concurrent same-token `CreateAccessPoint` calls, run under `-race -count=200`, fail with `NotFound: access point "fsap-..." not found`. The "exactly one created" invariant holds (SetIfAbsent guarantees that), but the idempotency contract — a racing retry returns the existing AP, never errors — is violated. This is the exact scenario ClientToken idempotency exists to protect. Real EFS returns the same AP to both callers.

Fix direction: ensure the AP object/index is observable before (or atomically with) the token claim, or bound-retry the `accessPointByID` lookup on the SetIfAbsent-loss path until the winner's store lands.

## Gates (on pushed SHA)
- `go build ./...` PASS; `go vet` PASS
- `go test ./server/aws/efs/... ./providers/aws/efs/... ./services/efs/...` PASS
- root `go test .` PASS; `go test -race ./providers/aws/efs/...` PASS (existing suite)
- golangci-lint new-only: clean (the one gocyclo hit is pre-existing in `providers/aws/rds_discovery.go`, untouched here)
- gofmt clean; `go generate ./...` zero diff

Everything except the F5 concurrency defect is solid. Recommend fixing the CreateAccessPoint idempotency race before merge.

NitinKumar004 (commented):
Delta re-review of the concurrency fix (commit 96925661, CreateAccessPoint ClientToken TOCTOU) on the pushed head — verdict: fix is correct, no regressions.

Race fix — sound. The reorder now publishes the AP (fd.accessPts + apIndex.Set) BEFORE apTokenIndex.SetIfAbsent, and a loser recovers the winner via accessPointByID(existingID) — by id, not by object lookup. Happens-before holds because both indexes are mutex-guarded memstore.Store: winner.apIndex.Set → winner.apTokenIndex.SetIfAbsent (program order) → loser.apTokenIndex.SetIfAbsent (same apTokenIndex mutex, and the false return proves the winner's write already landed) → loser.apIndex.Get + fd.accessPts read (program order). Transitively the loser is guaranteed to observe the winner's published AP and its accessPts entry — the spurious AccessPointNotFound window is closed. SetIfAbsent is an atomic CAS under Lock, so exactly one caller wins (no both-win path).

Rollback — clean. The loser deletes only its own id from accessPts and apIndex; the winner's id is untouched, no orphan, and the loser's id is never leaked into apTokenIndex (its SetIfAbsent returned false, so it never claimed the key). DeleteAccessPoint also evicts apTokenIndex for tokened APs, keeping the token index consistent.

Verified:
- go test -race -count=200 ./providers/aws/efs/... — clean, no data race, no AccessPointNotFound.
- Added 64-goroutine same-token stress + different-token + sequential-retry + empty-token tests under -race: exactly one fsap id, all callers receive it, zero errors, DescribeAccessPoints returns 1; different tokens produce distinct APs; empty token never dedups.
- Other 4 fixes intact (F1 FileSystemAlreadyExists id, F2 Backup/One Zone, F3 AvailabilityZoneId, F4 mount-target VpcId/AZ from subnet) — efs scoped + wire SDK + root integration tests pass.
- build, vet, gofmt, go generate (zero diff), golangci-lint (0 new issues) all green on the pushed SHA.

Minor, non-blocking: a losing racer's AP is briefly visible in accessPts/apIndex between publish and rollback, so a concurrent DescribeAccessPoints could momentarily observe duplicates before converging. This is the inherent cost of publish-then-claim, and it is the right tradeoff — claiming first would reintroduce the original TOCTOU. Not worth changing.

PR #681, opened 2026-08-26
feat(aws-cwlogs): subscription filters (Put/Describe/Delete) with Lambda delivery on matching PutLogEvents

NitinKumar004:
## What

Implements AWS **CloudWatch Logs subscription filters** end-to-end, with real-time delivery of matching log events to a Lambda destination.

- **PutSubscriptionFilter / DescribeSubscriptionFilters / DeleteSubscriptionFilter** — previously unrouted (`UnknownOperationException`), so Terraform `aws_cloudwatch_log_subscription_filter` failed outright. Now fully wired through the 3-layer stack:
  - `driver.Logging` gains `SubscriptionFilterConfig` / `SubscriptionFilterInfo` and the three CRUD methods; all three providers (AWS/Azure/GCP) implement store-backed CRUD; the portable `services/logging` wrapper forwards them.
  - Wire dispatch added in `server/aws/cloudwatchlogs` (new `subscription_filters.go`).
- **Delivery (the real point)** — on `PutLogEvents`, events whose message matches a subscription filter's pattern are delivered to the filter's **Lambda** destination via the shared `InvokeExternal` choke point (recursion-guarded), mirroring the existing S3 → Lambda and DynamoDB-Stream → Lambda wiring. Wired only on AWS via the optional `SetLambdaInvoker` pattern in `providers/aws/aws.go`.
  - Payload matches the real CloudWatch Logs subscription shape: gzipped + base64 `awslogs.data` wrapping `{messageType, owner, logGroup, logStream, subscriptionFilters, logEvents[]}`.
  - The stream lock is released before delivery so a subscriber Lambda that writes back into CloudWatch Logs cannot deadlock.
- **Two-filters-per-group quota** enforced (`LimitExceededException`); updating an existing filter name is not counted against the quota.

## Why

Deferred HIGH Terraform-breaker from the AWS deep-e2e audit (`cloudwatchlogs.md` BUG D). Subscription filters are the CloudWatch Logs analogue of the X → Lambda delivery already present for S3/DynamoDB/EventBridge/SNS; the audit found them entirely unimplemented.

## Deferred

- **Kinesis / Firehose delivery.** The destination ARN type is detected and only the Lambda path delivers; a Kinesis/Firehose `destinationArn` is stored and returned by DescribeSubscriptionFilters but does not yet stream records (cross-service Kinesis/Firehose deliverer wiring is a larger lift). Filter matching remains substring-level (richer CWL filter-pattern grammar is separately deferred — audit BUG C).

## Tests (real aws-sdk-go-v2)

- Provider: CRUD round-trip, validation, two-filter limit, **Lambda delivery on matching PutLogEvents** (decodes the gzipped awslogs payload and asserts logGroup/logStream/logEvents), non-matching → not delivered, delete stops delivery, Kinesis destination → not delivered (deferred). Passes under `-race`.
- Wire (SDK): PutSubscriptionFilter(Lambda ARN) → DescribeSubscriptionFilters round-trip (pattern/destinationArn/roleArn/distribution/creationTime), filterNamePrefix, DeleteSubscriptionFilter, ResourceNotFound + LimitExceeded errors.
- Existing CW Logs tests (#568 metric filter, #678 ordering/retention, pagination, cascade delete) stay green.

Coverage docs regenerated (logging 14 → 17 ops).

### DIFF
--- a/docs/coverage/README.md
+++ b/docs/coverage/README.md
@@ -38,7 +38,7 @@ capability the code does not implement. Machine-readable: [`coverage.json`](./co
 | `kinesis` | [Kinesis](./aws/kinesis.md) | — | — | — | 39 |
 | `kms` | [KMS](./aws/kms.md) | — | — | — | 46 |
 | `loadbalancer` | [ELB](./aws/elb.md) | [LB](./azure/lb.md) | [LB](./gcp/lb.md) | — | 19 |
-| `logging` | [CloudWatchLogs](./aws/cloudwatchlogs.md) | [LogAnalytics](./azure/loganalytics.md) | [CloudLogging](./gcp/cloudlogging.md) | — | 14 |
+| `logging` | [CloudWatchLogs](./aws/cloudwatchlogs.md) | [LogAnalytics](./azure/loganalytics.md) | [CloudLogging](./gcp/cloudlogging.md) | — | 17 |
 | `managedcassandra` | — | [ManagedCassandra](./azure/managedcassandra.md) | — | — | 15 |
 | `memorydb` | [MemoryDB](./aws/memorydb.md) | — | — | — | 33 |
 | `messagequeue` | [SQS](./aws/sqs.md) | [QueueStorage](./azure/queuestorage.md) | [PubSub](./gcp/pubsub.md) | — | 14 |
--- a/docs/coverage/aws/README.md
+++ b/docs/coverage/aws/README.md
@@ -11,7 +11,7 @@ Services cloudemu emulates for AWS, by native name. Back to the [cross-provider 
 | [BedrockAgentRuntime](./bedrockagentruntime.md) | `bedrockagentruntime` | 3 |
 | [CloudTrail](./cloudtrail.md) | `cloudtrail` | 60 |
 | [CloudWatch](./cloudwatch.md) | `monitoring` | 12 |
-| [CloudWatchLogs](./cloudwatchlogs.md) | `logging` | 14 |
+| [CloudWatchLogs](./cloudwatchlogs.md) | `logging` | 17 |
 | [Config](./config.md) | `configservice` | 102 |
 | [DynamoDB](./dynamodb.md) | `database` | 24 |
 | [EC2](./ec2.md) | `compute` | 37 |
--- a/docs/coverage/aws/cloudwatchlogs.md
+++ b/docs/coverage/aws/cloudwatchlogs.md
@@ -3,7 +3,7 @@
 
 AWS's `logging` service · portable interface `driver.Logging` · [AWS index](./README.md)
 
-## Operations (14)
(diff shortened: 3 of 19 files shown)


NitinKumar004 (commented):
**Verdict: approve / mergeable.**

Verified against the pushed head (`41522134`) in an isolated checkout: `go build ./...`, `go vet`, scoped `go test`, `go test -race` on the AWS provider, `gofmt`, and `go generate ./...` (zero coverage diff — the 14→17 op bump is committed and idempotent) all pass.

**Correctness / real-logic (the hard part) checks out:**
- **No deadlock / no torn state.** Wrote a 50-goroutine concurrent `PutLogEvents` test with an active filter under `-race`: completes cleanly, exactly one delivery per matching batch. Releasing the stream lock before `deliverToSubscriptions` is the right call — a re-entrant `PutLogEvents` from a subscriber acquires the same `s.mu` fresh.
- **Recursion bounded.** Simulated a subscriber that writes back into the subscribed group following the real `lambda.InvokeExternal` depth contract: the loop terminates at `recursionguard.MaxDepth` (no stack overflow / hang). `deliverToSubscriptions` correctly seeds `context.Background()` with `Depth(callerCtx)` so the chain keeps counting across the re-entry, and the increment/drop lives in `InvokeExternal`.
- **Delivery failure never fails ingestion.** `deliverSubscription` discards the invoker error (`_ = m.lambda.InvokeExternal(...)`); confirmed `PutLogEvents` returns nil and events persist even when the invoker errors.
- **Payload shape is faithful.** gunzip+base64 → JSON round-trips `messageType=DATA_MESSAGE`, `logGroup`, `logStream`, `subscriptionFilters`, and matched `logEvents`; non-matching events are excluded; a fully non-matching batch delivers nothing; delete stops delivery.
- **Error surface correct.** Missing group → ResourceNotFoundException; the 2-filter quota → LimitExceededException (ResourceExhausted mapped in `writeErr`); update-in-place at the limit still succeeds.
- **Kinesis/Firehose deferral is a clean no-op** — stored, returned by Describe, no invoke, no panic (guarded by the `:lambda:` ARN check).

**Non-blocking notes (follow-ups, not merge gates):**
- Subscription-filter methods are promoted through the embedded `driver.Logging` in the chaos wrapper, so these calls bypass chaos/latency injection — a completeness gap, not a correctness one.
- Delivery is synchronous in-process (matching the existing S3→Lambda / DynamoDB-Streams pattern), though the doc comments describe it as "asynchronous"/"best-effort"; consider aligning the wording.
- Azure/GCP are CRUD-only (no delivery), which is the documented and correct choice for those backends.

The lint findings in the touched files are all on pre-existing lines (imports, CreateLogGroup/ListLogGroups/UpdateLogGroup); the new subscription-filter code is clean.

PR #683, opened 2026-08-26
feat(aws-elasticache): parameter-group contents — DescribeCacheParameters + ModifyCacheParameterGroup ParameterNameValues

NitinKumar004:
## What

Implements the **contents** of AWS ElastiCache cache parameter groups so
`aws_elasticache_parameter_group` with `parameter { name=… value=… }` blocks
works end-to-end.

- **DescribeCacheParameters** — previously unimplemented (fell through to the EC2
  catch-all → `InvalidAction`). Now returns a group's parameter list: engine-family
  defaults with any user overrides applied. Supports the `Source` filter
  (`user` / `system` / `engine-default`). A missing group returns
  `CacheParameterGroupNotFound`.
- **ModifyCacheParameterGroup** — previously ignored `ParameterNameValues` (no-op
  beyond existence check). Now parses `ParameterNameValues.ParameterNameValue.N`
  and applies each name→value (Source becomes `user`), reflected on the next
  DescribeCacheParameters. Unknown / non-modifiable parameter name →
  `InvalidParameterValue`.
- **ResetCacheParameterGroup** — newly dispatched. `ResetAllParameters=true` clears
  all overrides; a named list resets only those parameters back to their engine
  default (Source → `system`).
- Each parameter group now carries a per-family default parameter set (curated,
  not the full real hundreds) for `redis*` and `memcached*` families, including the
  parameters commonly set via Terraform (notably `maxmemory-policy` for Redis).
  Each parameter exposes ParameterName / ParameterValue / Source / DataType /
  AllowedValues / IsModifiable / MinimumEngineVersion / Description.

## Why

Deferred real-Terraform-breaker from the AWS deep-e2e audit (elasticache finding
#8). `aws_elasticache_parameter_group` with any non-empty `parameter {}` block
calls DescribeCacheParameters on refresh — which failed with `InvalidAction` — and
applied parameters never persisted. End-to-end broken for any non-empty parameter
group; now converges.

## Layers

- Driver interface (`services/cache/driver`) is **unchanged** — parameter groups are
  an AWS-only surface asserted via the wire handler's `parameterGroupManager`
  interface, so coverage docs are unaffected (no regeneration).
- Provider (`providers/aws/elasticache/parameters.go`): default parameter sets +
  per-group override store + the three ops.
- Wire (`server/aws/elasticache`): dispatch for DescribeCacheParameters /
  ResetCacheParameterGroup, ParameterNameValues parsing, XML response shapes.

## Tests

Real `aws-sdk-go-v2/service/elasticache` round-trips plus provider-level tests:
default list non-empty (incl. maxmemory-policy=volatile-lru, Source=system),
Modify applies `maxmemory-policy=allkeys-lru` with Source=user, unknown param →
InvalidParameterValue, missing group → CacheParameterGroupNotFound, Source filter,
ResetAllParameters and named reset. Existing ElastiCache tests (#516/#468/#677) stay
green.

## Notes

ResetCacheParameterGroup was implemented (not deferred) as it was low-risk given the
override-store model. Node-type-specific parameters (`CacheNodeTypeSpecificParameters`)
are returned empty, matching a simple single-node emulator.

### DIFF
--- a/providers/aws/elasticache/elasticache.go
+++ b/providers/aws/elasticache/elasticache.go
@@ -160,12 +160,17 @@ type Mock struct {
 }
 
 // ParameterGroup is an ElastiCache cache parameter group — a named, engine-family
-// set of engine parameters. The emulator stores its identity so IaC that creates
-// and references a custom group succeeds.
+// set of engine parameters. The emulator stores its identity plus any user
+// overrides (name→value) applied via ModifyCacheParameterGroup, so IaC that
+// creates a group, sets `parameter { … }` blocks, and reads them back on refresh
+// converges. The engine-family defaults themselves are synthesized on demand
+// (see defaultCacheParameters); Overrides holds only the parameters the user has
+// changed from their default.
 type ParameterGroup struct {
 	Name        string
 	Family      string
 	Description string
+	Overrides   map[string]string
 }
 
 // SetMonitoring sets the monitoring backend for auto-metric generation.
--- a/providers/aws/elasticache/parameters.go
+++ b/providers/aws/elasticache/parameters.go
@@ -1,0 +1,247 @@
+package elasticache
+
+import (
+	"context"
+	"sort"
+	"strings"
+
+	cerrors "github.com/stackshy/cloudemu/v2/errors"
+)
+
+// Parameter is one entry in a cache parameter group's detailed parameter list,
+// as returned by DescribeCacheParameters. Source is "system" for an
+// engine-default value and "user" once the parameter has been modified via
+// ModifyCacheParameterGroup.
+type Parameter struct {
+	Name                 string
+	Value                string
+	Source               string
+	DataType             string
+	AllowedValues        string
+	IsModifiable         bool
+	MinimumEngineVersion string
+	Description          string
+}
+
+// ParameterUpdate is one name→value pair from a ModifyCacheParameterGroup /
+// ResetCacheParameterGroup ParameterNameValues list. For a reset the Value is
+// ignored — only the Name identifies the parameter to restore to its default.
+type ParameterUpdate struct {
+	Name  string
+	Value string
+}
+
+const (
+	paramSourceSystem = "system"
+	paramSourceUser   = "user"
+
+	paramSourceEngineDefault = "engine-default"
+)
+
+// defaultCacheParameters returns a curated set of engine-default parameters for
+// a cache parameter group family (e.g. "redis7", "memcached1.6"). It is not the
+// full real list of hundreds — it is the representative subset IaC reads and the
+// parameters commonly set through Terraform (notably maxmemory-policy for
+// Redis). Every entry carries Source "system".
+func defaultCacheParameters(family string) []Parameter {
+	if isMemcachedFamily(family) {
+		return memcachedDefaultParameters()
+	}
+
+	return redisDefaultParameters()
+}
+
+// isMemcachedFamily reports whether a parameter-group family names the Memcached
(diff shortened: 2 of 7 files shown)


NitinKumar004 (commented):
Reviewed at pushed SHA `543ddfc4`.

**Gates (all green):**
- `GOMAXPROCS=3 go build ./...` clean
- `go vet ./...` clean
- `go test ./server/aws/elasticache/... ./providers/aws/elasticache/... ./services/cache/...` — all pass
- `go test -race ./providers/aws/elasticache/...` — clean, no data races (the `map[string]string` Overrides field is copied/reassigned through `memstore.Get`/`Set`, no shared mutable state escapes)
- `gofmt -l` — no diffs on any touched file
- `go generate ./...` — zero diff (confirms the "no driver interface change" claim; docs/coverage untouched, correctly no `CacheParameterGroup`/`Parameter` types on the portable driver surface)
- `golangci-lint run ./providers/aws/elasticache/... ./server/aws/elasticache/...` — 14 pre-existing findings, all outside this diff's hunks (`ListCaches`/`UpdateCache` wsl nits, `Mock` struct `typeDefFirst`, `describeCacheEngineVersions` unused receiver, etc. — verified by line-content diff against `origin/development`). Zero new findings in `parameters.go`, `parameters_test.go`, or the new functions added to `handler.go`/`tags.go`.

**Claims verified independently** (own read of the diff + ran the PR's own SDK round-trip and provider tests, all passing):
- `DescribeCacheParameters`: family defaults returned (redis7 non-empty incl. `maxmemory-policy=volatile-lru`/Source=system; memcached1.6 exposes `max_item_size` and correctly excludes redis-only params). `Source` filter (`user`/`system`/`engine-default`→system) works via `matchesParameterSource`. Missing group → `CacheParameterGroupNotFound` (via `notFoundCode`'s `"parameter group"` substring match on the error message).
- `ModifyCacheParameterGroup`: parses `ParameterNameValues.ParameterNameValue.N.{ParameterName,ParameterValue}` (via the existing `awsquery.CollectIndices` helper, no new duplicate parsing helper introduced), applies as Source=user, reflected on next Describe. Unknown/non-modifiable name → `InvalidParameterValue`, and validation is atomic — both loops in `ModifyCacheParameterGroup`/`ResetCacheParameterGroup` validate every entry *before* mutating, so a rejected batch applies zero overrides (verified by reading `parameters.go:214-244`).
- `ResetCacheParameterGroup`: `ResetAllParameters=true` clears all overrides; named list resets only the given names (unknown name in the list → `InvalidArgument`/`InvalidParameterValue`, not separately tested by the PR's suite but the code path is symmetric with Modify's validation and is simple).
- Dispatch fix: `DescribeCacheParameters` and `ResetCacheParameterGroup` are correctly added to the `elastiCacheActions` allow-map consulted by `Matches`, and the SDK round-trip tests use a client wired with both ElastiCache *and* EC2 handlers specifically to catch fallthrough — all pass, confirming these actions no longer leak to the EC2 catch-all/`InvalidAction`.
- Existing ElastiCache suites (endpoints/RG/subnet-group families, #516/#468/#677) all still pass in the full package run.

**Minor, non-blocking observations:**
- `matchesParameterSource` silently returns an empty result for an unrecognized `Source` value (e.g. a typo) rather than an error; real ElastiCache would likely reject an invalid `Source` with `InvalidParameterValue`. Not exercised by any current caller/IaC path, low risk.
- Package-wide coverage for `providers/aws/elasticache` (64.9%) and `server/aws/elasticache` (78.0%) sits under the repo's 90% target, but that's pre-existing package debt — the new `parameters.go` itself is 75-100% covered per function.

No blocking issues found.

PR #684, opened 2026-08-26
feat(aws-route53): private hosted zone VPC association (Associate/Disassociate/CreateHostedZone VPC)

NitinKumar004:
## What

Implements AWS Route 53 private-hosted-zone VPC association — the real Terraform gap (`aws_route53_zone` with a `vpc {}` block, `aws_route53_zone_association`) deferred from the deep-e2e audit (finding F8).

- **CreateHostedZone with a `<VPC>` block** → the zone becomes PRIVATE (`Config.PrivateZone=true`) and is associated with that VPC. The create response echoes the `<VPC>` element; a zone created without a VPC stays public.
- **AssociateVPCWithHostedZone** (`POST /2013-04-01/hostedzone/{id}/associatevpc`) → adds the VPC to the private zone's association list and returns a `ChangeInfo` (INSYNC). Idempotent for an already-associated VPC. A public zone is rejected with `PublicZoneVPCAssociation`.
- **DisassociateVPCFromHostedZone** (`POST .../disassociatevpc`) → removes the VPC. Removing a VPC that isn't associated → `VPCAssociationNotFound`; removing the last VPC → `LastVPCAssociation`.
- **GetHostedZone** → emits the `<VPCs>` list for a private zone (empty for a public zone).
- **ListHostedZonesByVPC** (`GET /2013-04-01/hostedzonesbyvpc?vpcid=&vpcregion=`) → returns the private zones a VPC is associated with, as `HostedZoneSummaries` with an `OwningAccount` owner.

## Why

Previously a POST to `{id}/associatevpc` fell through `serveZone` (GET/DELETE only) and returned `404 NoSuchHostedZone: unrecognized Route 53 path`, and the driver had no VPC storage — so private hosted zones and `aws_route53_zone_association` could not work against the emulator. All request/response shapes and error codes were cross-checked against the official Route 53 API reference (CreateHostedZone, AssociateVPCWithHostedZone, DisassociateVPCFromHostedZone, GetHostedZone, ListHostedZonesByVPC).

## Design (3-layer)

- **Driver** (`services/dns/driver`): new `VPCAssociation` type + `VPCs []VPCAssociation` on `ZoneConfig`/`ZoneInfo` (shared struct fields; Azure/GCP DNS leave them nil — zero-value pass-through, no re-implementation of their private-zone semantics). No new methods on the shared `driver.DNS` interface, so generated `docs/coverage` is unchanged (`go generate` is a no-op).
- **Provider** (`providers/aws/route53`): `CreateZone` stores the VPCs; new `AssociateVPC`/`DisassociateVPC` mutators (idempotent add / exact remove, deep-copied to detach from caller memory).
- **Wire** (`server/aws/route53`): routes `associatevpc`/`disassociatevpc` POSTs and the `hostedzonesbyvpc` GET; all state-dependent validation (privacy, last-VPC, not-associated) is done at the wire layer from the zone returned by `GetZone`, via an AWS-only `vpcAssociator` extension interface (mirroring the existing `recordSetDeleter` pattern). `ServeHTTP` tail dispatch extracted into a helper to keep complexity in check.

The existing zone CRUD and the #674 correctness fixes are untouched; existing route53 + topology tests stay green.

## Tests

Real `aws-sdk-go-v2/service/route53` reproductions: create-with-VPC → private + VPC on Get; associate second VPC → both listed + ChangeInfo; disassociate one → removed; disassociate last → `LastVPCAssociation`; disassociate unassociated → `VPCAssociationNotFound`; associate on a public zone → `PublicZoneVPCAssociation`; idempotent re-associate; `ListHostedZonesByVPC` filtering. Plus provider-level unit tests for store/idempotency/deep-copy.

### DIFF
--- a/providers/aws/route53/route53.go
+++ b/providers/aws/route53/route53.go
@@ -134,6 +134,7 @@ func (m *Mock) CreateZone(_ context.Context, cfg driver.ZoneConfig) (*driver.Zon
 		Tags:            tags,
 		CallerReference: cfg.CallerReference,
 		Comment:         cfg.Comment,
+		VPCs:            copyVPCs(cfg.VPCs),
 		Scope:           cfg.Scope,
 	}
 
@@ -170,6 +171,7 @@ func (m *Mock) GetZone(_ context.Context, id string) (*driver.ZoneInfo, error) {
 	}
 
 	result := zone
+	result.VPCs = copyVPCs(zone.VPCs)
 
 	return &result, nil
 }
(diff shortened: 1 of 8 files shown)


NitinKumar004 (commented):
Reviewed at pushed SHA `11e21533aba93d001fbc33769ce489021236bc23`.

**Gates (GOMAXPROCS=3/-p=3):**
- `go build ./...` — clean
- `go vet ./...` — clean
- `go test ./server/aws/route53/... ./providers/aws/route53/... ./services/dns/... ./features/topology/...` — all green (existing #467/#532/#553/#674 tests + new tests pass)
- `go test -race ./providers/aws/route53/...` — clean
- `gofmt -l` / `gofmt -d` on changed files — clean
- `go generate ./...` — zero diff (confirms no driver.DNS method changed, docs/coverage untouched, as claimed)
- `golangci-lint run ./server/aws/route53/... ./providers/aws/route53/... ./services/dns/...` — 23 findings, all pre-existing debt (confirmed via `git blame`) except one new one: `server/aws/route53/types.go:183`, `hostedZoneSummaryXML.HostedZoneId` trips `revive:var-naming` (wants `HostedZoneID`). It matches the file's existing house style for other Route53 XML fields (`hostedZoneXML.Id`, `NextHostedZoneId`, etc., all already-violating), and this repo's CI runs golangci-lint in report-only/advisory mode, so non-blocking — flagging for awareness only.

**Correctness verification** (read provider + wire code, plus wrote and ran extra ad-hoc SDK tests, deleted afterward — not part of the PR):
- Create-with-VPC → private + VPC on Get: correct. Associate 2nd VPC → both present + ChangeInfo(INSYNC): correct. Re-associate an already-associated VPC → idempotent, no duplicate (`containsVPCAssoc`): correct. Disassociate one → removed, siblings intact: correct. Disassociate last VPC → `LastVPCAssociation`: correct (checked before the mutator call, using the VPC count from `GetZone`). Disassociate not-associated VPC → `VPCAssociationNotFound`: correct. Associate on a public zone → `PublicZoneVPCAssociation`, checked before calling the mutator: correct. Missing zone on Associate/Disassociate → `NoSuchHostedZone` via the shared `writeErr`/`cerrors.IsNotFound` path: verified independently, correct.
- `ListHostedZonesByVPC`: filters correctly by VPCID+VPCRegion. Minor gap — the response never truncates or sets `NextToken`/`IsTruncated` despite parsing a `maxitems` query param and carrying a `MaxItems` field, so a caller with more matching zones than `maxitems` gets everything back in one page. Out of the review's explicit checklist but worth a follow-up.
- Deep-copy: `CreateZone`/`GetZone` both defensively copy `VPCs` on the way in/out (`copyVPCs`), and `AssociateVPC`/`DisassociateVPC` also copy-then-append/filter rather than mutating the stored slice in place. Independently verified: mutating the `VPCs` slice on a struct returned from one `GetHostedZone` call does not corrupt what a later `GetHostedZone` call returns. Note that `CreateZone`'s returned `*ZoneInfo` still shares its `VPCs` backing array with the just-stored value (only `GetZone` re-copies on the way out) — this mirrors the exact same pre-existing pattern already used for `Tags` in this file, so it's consistent with house style, not a new gap.
- Routing refactor (`serveHostedZonePath`, extracted from `ServeHTTP`'s tail dispatch): the switch over `sub` (`""`, `rrsetSeg`, `associateVPCSeg`, `disassociateVPCSeg`, default 404) is behaviorally equivalent to the prior if-chain; confirmed by the full existing route53 test suite passing unchanged.
- `driver.DNS` interface unchanged (no new methods) — `VPCAssociation`/`VPCs` are additive struct fields only, matching the existing pattern for other AWS-only fields (`CallerReference`, `Comment`, `AliasTarget`, etc.). Azure/GCP DNS backends are unaffected and don't implement the AWS-only `vpcAssociator` extension interface.
- `defaultOwningAccount = "123456789012"` in `vpc.go` is hardcoded rather than threaded from `config.Options.AccountID`, but the route53 `Handler` doesn't carry `*config.Options` at all today — this is a pre-existing limitation of the package, not something newly introduced here.

No blocking issues found.

PR #686, opened 2026-08-26
fix(aws-acm): wildcard DNS-validation record rooted at base domain + deduped; support EC/RSA-4096 KeyAlgorithm; validate DomainName; ListCertificates default-RSA filter + pagination

NitinKumar004:
## What

Fixes four AWS ACM correctness issues found by a real-user aws-sdk-go-v2 e2e audit, at the correct 3-layer locations (driver / provider / wire).

**F2 (real bug) — Wildcard DNS-validation record contained a literal `*`; wildcard+apex not deduplicated.**
For `*.example.com`, the DNS validation `ResourceRecord.Name` was `_acm-validations.*.example.com.` — Route53 rejects `*` as a non-leftmost label, breaking the common `aws_acm_certificate_validation` + `aws_route53_record` wildcard flow. Now the record is rooted at the base domain (`_acm-validations.example.com.`, no `*`), and a wildcard together with its apex SAN collapse to a single shared validation record, matching real ACM.

**F1 (Terraform-breaker) — RequestCertificate rejected every KeyAlgorithm except RSA_2048.**
Terraform `aws_acm_certificate { key_algorithm = "EC_prime256v1" }` failed against cloudemu. Now the documented EC/RSA set is accepted (`RSA_1024/2048/3072/4096`, `EC_prime256v1/secp384r1/secp521r1`); real key material is generated for the requested algorithm, and `DescribeCertificate` reflects the requested `KeyAlgorithm` plus a matching `SignatureAlgorithm`. A genuinely-unsupported value is rejected with `InvalidParameterException`.

**F4 — RequestCertificate accepted malformed DomainName.**
`DomainName`/`SubjectAlternativeNames` are now validated as FQDNs (mirroring ACM's `(\*\.)?(label\.)+tld` pattern, leading `*.` wildcard allowed); malformed input returns `InvalidParameterException`.

**F5 — Default ListCertificates returned non-RSA_2048 certs and ignored pagination/Includes.**
Per AWS, the default (no `Includes.keyTypes`) list returns only RSA_2048 certificates. Now default-filters to RSA_2048, honors `Includes.keyTypes`, and honors `MaxItems`/`NextToken` pagination over a stable ARN ordering.

Deferred (out of scope, tracked): F3 (EMAIL ValidationEmails) and F7 (DeleteCertificate in-use guard — currently unreachable, nothing populates `InUseBy`).

## Why

These diverge from real AWS ACM and break common Terraform flows (ECDSA certs, wildcard DNS validation). Verified against the official ACM API reference for RequestCertificate (KeyAlgorithm valid values, DomainName pattern) and ListCertificates (default RSA filtering, Includes.keyTypes, MaxItems/NextToken).

## Tests

New real aws-sdk-go-v2 reproductions plus provider-level unit tests: EC/RSA-4096 key algorithms issue and reflect correctly (bogus rejected); wildcard record rooted with no `*` and wildcard+apex sharing one record; non-wildcard record unchanged; malformed DomainName/SAN rejected and valid FQDN+wildcard accepted; default list excludes an imported EC cert while `Includes.keyTypes` includes it; MaxItems/NextToken paginate. Existing ACM tests (lifecycle, import, export guards, tags) stay green.

### DIFF
--- a/providers/aws/acm/acm.go
+++ b/providers/aws/acm/acm.go
@@ -24,6 +24,13 @@ const (
 	maxDomains = 100
 	// maxTags is ACM's cap on tags per certificate.
 	maxTags = 50
+	// maxDomainNameLen / maxLabelLen bound a valid FQDN (RFC 5280 / ACM pattern).
+	maxDomainNameLen = 253
+	maxLabelLen      = 63
+	// minTLDLen is the minimum length of the final (top-level) label; minLabels is
+	// the minimum number of dot-separated labels an FQDN must have.
+	minTLDLen = 2
+	minLabels = 2
 )
 
 // Mock is an in-memory implementation of AWS ACM.
--- a/providers/aws/acm/certificate_gen.go
+++ b/providers/aws/acm/certificate_gen.go
@@ -1,6 +1,7 @@
 package acm
 
 import (
+	"crypto"
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/rand"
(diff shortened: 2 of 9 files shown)


NitinKumar004 (commented):
**Verdict: MERGE** — verified against the pushed head (662a4f4), clean-built and adversarially tested.

**Gates (pushed SHA, isolated worktree)**
- `go build ./...`, `go vet ./...` clean
- `go test ./server/aws/acm/... ./providers/aws/acm/... ./services/acm/...` green
- `go test -race ./providers/aws/acm/...` green
- gofmt clean; `go generate ./...` zero diff; golangci-lint (new-only, changed pkgs) 0 issues

**F2 — wildcard DNS-validation record (the real bug): confirmed fixed.**
`validationOptions()` strips the leading `*.` and roots the record at the base domain (`_acm-validations.example.com.`), so `ResourceRecord.Name`/`Value` carry no literal `*` and are valid DNS names Route53 accepts. Wildcard + apex SAN dedupe on base domain into ONE shared DVO. A multi-SAN cert with distinct base domains emits one record per distinct base (verified `example.com` + `*.example.com`/`other.com`/`*.other.com`/`www.other.com` -> exactly 3 records). Non-wildcard single-domain record unchanged.

**F1 — KeyAlgorithm / cert-gen: confirmed.**
All seven algorithms (RSA_1024/2048/3072/4096, EC_prime256v1/secp384r1/secp521r1) issue real, parseable X.509 material whose key type and curve match the request; Describe reflects the requested KeyAlgorithm with the correct SignatureAlgorithm pairing (RSA*->SHA256WITHRSA, P256->ECDSAWITHSHA256, P384->ECDSAWITHSHA384, P521->ECDSAWITHSHA512), and the reported sig matches the cert's actual `SignatureAlgorithm` (x509 default per key). Bogus value -> InvalidParameterException. RenewCertificate preserves the key algorithm and re-issues fresh, parseable material (verified EC P-384 survives renew with a new serial). No panic on any supported value.

**F4 — DomainName/SAN validation: confirmed.**
FQDN pattern `(\*\.)?(label\.)+tld` enforced on DomainName and each SAN; malformed input (`not a domain!!`, single-label, `a.b` short TLD, leading-hyphen label, multi-`*`) -> InvalidParameterException; valid FQDN and single-leading-`*.` wildcard accepted.

**F5 — ListCertificates: confirmed.**
Default (no Includes) filters to RSA_2048 and excludes an imported EC cert; `Includes.keyTypes=[EC_prime256v1]` surfaces it; MaxItems/NextToken paginate via `PaginateSorted` over stable ARN order — 5-cert set at MaxItems=2 spans multiple pages with no dupes/omissions (limit<=0 falls back to page size 100). SearchCertificates shares the same filter path.

**Blast radius:** the only consumers of the ACM driver are the ACM provider, the ACM server handler, and `server/aws/aws.go` (wiring) — no external caller depends on the old always-RSA_2048 list or single-algorithm behavior. Coverage docs derive from driver method signatures; adding `ListFilter.KeyTypes` + KeyAlg constants produced zero `go generate` diff. Existing ACM tests (lifecycle, import, export guards, tags) stay green. Diff touches only the 9 ACM files.

**Non-blocking note:** the DNS validation record uses a static `_acm-validations.<base>.` name and a static `_<base-dashed>.acm-validations.aws.` value rather than real ACM's per-certificate `_<hash>` label — so two different certs on the same base domain would surface identical records. Harmless for an auto-validating emulator (records are informational and no real DNS challenge occurs); the Route53-breaking `*` is what mattered and is gone.

PR #687, opened 2026-08-26
fix(aws-eks): tag nodegroup/fargate/addon ARNs; nodegroup taints round-trip; clusterSecurityGroupId+vpcId on DescribeCluster; clean errors + scaling validation

NitinKumar004:
## What

Fixes a batch of AWS EKS control-plane parity gaps found by a deep-e2e real-SDK audit (`aws-sdk-go-v2/service/eks`).

- **[HIGH] Tagging works on every EKS resource ARN, not just clusters.** `TagResource`/`UntagResource`/`ListTagsForResource` previously only resolved `:cluster/` ARNs; nodegroup/Fargate-profile/addon ARNs 404'd. The provider now parses the ARN resource segment and routes tag read/write to the matching store.
- **[MED] Nodegroup taints round-trip end-to-end.** `taints` is threaded through the driver, provider, and wire: stored on `CreateNodegroup`, returned on `DescribeNodegroup`, and `UpdateNodegroupConfig` applies `addOrUpdateTaints`/`removeTaints`. Label updates now merge `addOrUpdateLabels` and honor `removeLabels` (was a wholesale replace that ignored removes).
- **[MED] `DescribeCluster` reports `resourcesVpcConfig.clusterSecurityGroupId` and `vpcId`.** A deterministic `sg-` cluster security group is synthesized on create and the VPC is derived from the cluster's subnets via the shared `SubnetResolver` (`SetSubnetResolver(p.VPC)`, mirroring EFS/ELB).
- **[LOW] Clean error messages.** Generic error branches used `err.Error()`, leaking the internal `NotFound:`/`AlreadyExists:` prefix; they now use `wireMessage(err)` (the `X-Amzn-ErrorType` header still carries the typed code).
- **[LOW] Nodegroup mutation faithfulness.** `modifiedAt` advances on every mutating op, inconsistent scaling (`minSize>maxSize`, desired out of range) is rejected with `InvalidParameterException`, and managed-nodegroup defaults (`diskSize=20`, `instanceTypes=[t3.medium]`, `amiType=AL2_x86_64`) are applied when omitted.

## Why

These break real-user flows: terraform tag updates on `aws_eks_node_group`/`aws_eks_fargate_profile`/`aws_eks_addon` hard-fail; `aws_eks_node_group` `taint{}` blocks show perpetual drift; node-group modules reference `aws_eks_cluster.vpc_config[0].cluster_security_group_id` and `vpc_id`, which were empty.

## Tests

Real-SDK reproductions added: tagging across all four ARN types; taints create/describe + add/remove via `UpdateNodegroupConfig`; `clusterSecurityGroupId`+`vpcId` derivation from real subnets; clean error message; plus provider-level modifiedAt/label-merge/scaling-validation/defaults. Existing EKS tests (OIDC, delete guards, DescribeUpdate/ListUpdates, partial-scaling merge, pagination, data plane) stay green.

## Findings source

Deep-e2e audit `eks.md`: findings #1 (tagging), #2 (taints), #3 (clusterSG/vpcId), #5 (modifiedAt), #6 (error prefix), #7 (scaling validation), #8 (defaults), #9 (removeLabels). Deferred as feature gaps: #4 (logging/kubernetesNetworkConfig/accessConfig blocks) and #9's addon health block.

### DIFF
--- a/providers/aws/aws.go
+++ b/providers/aws/aws.go
@@ -270,6 +270,9 @@ func New(opts ...config.Option) *Provider {
 	p.ECS.SetTargetRegistrar(p.ELB)
 	p.Redshift.SetMonitoring(p.CloudWatch)
 	p.EKS.SetMonitoring(p.CloudWatch)
+	// An EKS cluster's resourcesVpcConfig.vpcId is derived from its subnets,
+	// matching real EKS (which auto-creates the cluster SG and infers the VPC).
+	p.EKS.SetSubnetResolver(p.VPC)
 	p.SageMaker.SetMonitoring(p.CloudWatch)
 	// CloudWatch alarm -> SNS: an alarm state transition fires its configured
 	// SNS-topic actions, fanning a notification out to the topic's subscribers.
--- a/providers/aws/eks/driver/driver.go
+++ b/providers/aws/eks/driver/driver.go
@@ -107,6 +107,15 @@ type NodegroupScalingConfig struct {
 	DesiredSize int
 }
 
+// Taint is a Kubernetes taint applied to a managed node group's nodes. Effect
+// is one of NO_SCHEDULE, PREFER_NO_SCHEDULE, or NO_EXECUTE. A taint is
+// identified by its Key+Effect pair.
+type Taint struct {
+	Key    string
+	Value  string
+	Effect string
+}
+
 // NodegroupConfig configures a new managed node group.
 type NodegroupConfig struct {
 	ClusterName    string
@@ -121,6 +130,7 @@ type NodegroupConfig struct {
 	ReleaseVersion string
 	ScalingConfig  NodegroupScalingConfig
 	Labels         map[string]string
+	Taints         []Taint
 	Tags           map[string]string
 }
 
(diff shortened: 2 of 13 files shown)


NitinKumar004 (commented):
Adversarial review of the pushed head (b1ad8a1). Checked out the PR head SHA and clean-built the actual tree; wrote and ran an independent aws-sdk-go-v2/eks suite in addition to the PR's own.

**Verdict: LGTM.** All gates green on the pushed commit and every claimed fix reproduces against the real SDK.

Gates (GOMAXPROCS=3): `go build ./...` clean; `go vet` clean; `go test ./server/aws/eks/... ./providers/aws/eks/...` green; `go test -race ./providers/aws/eks/...` green; golangci-lint `--new-from-rev=development` 0 issues on the three changed trees; `gofmt -l` clean; `go generate ./...` zero diff (EKS driver isn't consumed by coveragegen, so no stale doc — confirmed).

Independently verified (my own tests, not the PR's):
- Tag-by-ARN across cluster/nodegroup/fargateprofile/addon, including a nodegroup ARN carrying the trailing `.../<uuid>` segment real EKS appends — `parseResourceRef` ignores the uuid and routes correctly; the write is visible via DescribeNodegroup.Tags.
- Malformed/unknown/empty-path ARNs return a clean ResourceNotFound, no panic (the too-few-segments case falls through to the bare-name cluster fallback).
- 24 concurrent TagResource writes on one nodegroup ARN under `-race`: race-clean and no lost writes (read-modify-write under `m.mu`).
- Taints round-trip on create+describe; UpdateNodegroupConfig add/update (value change on an existing Key+Effect) + remove touches only the named taint; label addOrUpdate + removeLabels both honored (no wholesale replace); modifiedAt strictly advances past createdAt after update (FakeClock), createdAt stable.
- Scaling: min>max on create and desired>max on update both rejected as InvalidParameterException.
- Cross-service vpcId: cluster created with a real VPC subnet reports resourcesVpcConfig.vpcId = the subnet's VPC and a deterministic sg- clusterSecurityGroupId; unwired resolver leaves vpcId empty with no panic (sg still synthesized). No import cycle (build proves it); SetSubnetResolver(p.VPC) wired in aws.go.
- Signature change (UpdateNodegroupConfig → NodegroupConfigUpdate): sole non-test caller is server/aws/eks/operations.go, updated; whole-repo build confirms no stragglers.

Non-blocking observations:
1. CreateCluster calls resolveVpcID (cross-service DescribeSubnets into the VPC mock) while holding the EKS write lock `m.mu`. The EFS pattern it claims to mirror resolves the subnet *before* taking its lock (mount_targets.go). Safe today because VPC never calls back into EKS (no cycle), but it serializes all EKS ops for the duration of the subnet lookup and diverges from the mirrored pattern; consider resolving before Lock().
2. resolveVpcID relies on VPC.DescribeSubnets, which errors if *any* id is unknown → vpcId ends up empty for a partially-valid subnet set. Fine for the mock; just not maximally faithful.
3. Scaling defaults are not synthesized on create (omitted → 0/0/0 on the wire) whereas real EKS defaults to 2/2/2. Out of scope for this PR's stated defaults (disk/instanceTypes/amiType) — noting only.

PR #689, opened 2026-08-26
fix(aws-opensearch): DescribeDomainConfig {Options,Status} envelope; VPC domain Endpoints["vpc"]; VPCId enrichment

NitinKumar004:
## What

Fixes three real-user bugs in AWS OpenSearch Service found by a deep real-SDK (aws-sdk-go-v2) e2e audit. The primary `DescribeDomain` (DomainStatus) round-trip was already correct and is left unchanged.

### F1 (HIGH) — DescribeDomainConfig/UpdateDomainConfig config-view envelope
`domainConfigToWire` emitted every raw option block (EBSOptions, VPCOptions, EncryptionAtRestOptions, NodeToNodeEncryptionOptions, DomainEndpointOptions, AdvancedSecurityOptions, AccessPolicies, …) **flat**, so the SDK deserialized `Options == nil` and `Status == nil` on the config view. AWS `DomainConfig` wraps every block in an `{Options, Status}` envelope (`OptionStatus`). Now every block — modeled and raw-passthrough alike — is wrapped, with a proper `OptionStatus` (`State: Active`, `PendingDeletion: false`, `CreationDate`, `UpdateDate`, `UpdateVersion`). `UpdateDomainConfig` returns the updated block in the same envelope.

### F2 (MED) — VPC-access domains
`CreateDomain` always set a public `Endpoint` and never populated `Endpoints`. AWS: a domain created with `VPCOptions` has **no** public `Endpoint`; it exposes `Endpoints["vpc"]` instead (Terraform reads `endpoint` from there). Now, when `VPCOptions` carries subnets, `Endpoint` is left empty and `Endpoints["vpc"]` is set; public domains keep their single `Endpoint`.

### F3 (LOW) — VPCOptions enrichment
The echoed `VPCOptions` omitted the AWS-derived `VPCId`/`AvailabilityZones`. Now enriched on create: a deterministic `VPCId` (hash of subnets, since the emulator has no EC2 backend to resolve the real VPC) and `AvailabilityZones` derived from the subnets. `DescribeDomain.VPCOptions.VPCId` is now populated.

## Why
The config-view API (`aws opensearch describe-domain-config`, Terraform reading EBS/VPC/encryption off the config view) returned unreadable `null` Options for every unmodeled block. VPC-domain users read an empty vpc endpoint. These match real AWS now.

## Tests
Real aws-sdk-go-v2 reproductions added: `TestSDKDescribeDomainConfigEnvelope` (EBS/VPC/encryption `.Options` populated + `.Status.State==Active`; UpdateDomainConfig EBS=50 in the envelope) and `TestSDKVPCDomainEndpoints` (VPC domain → `Endpoints["vpc"]` set, `Endpoint` empty, `VPCId` enriched; public domain → `Endpoint` set, no vpc key). Existing lifecycle/tags/update-persist tests stay green; `-race` clean.

## Notes
- No driver interface ops changed (only a `DomainConfig.CreatedAt` struct field); `docs/coverage` and `docs/compat` are zero-diff.
- The pre-existing `handler.go:156` nolintlint drift is untouched by this PR.

### DIFF
--- a/providers/aws/opensearch/domains.go
+++ b/providers/aws/opensearch/domains.go
@@ -3,6 +3,8 @@ package opensearch
 import (
 	"context"
 	"encoding/json"
+	"fmt"
+	"hash/fnv"
 	"sort"
 	"strings"
 
@@ -61,6 +63,20 @@ func (m *Mock) CreateDomain(_ context.Context, in driver.CreateDomainInput) (*dr
 
 	now := m.now()
 	cfg := copyClusterConfig(in.ClusterConfig)
+	raw := copyRaw(in.RawOptions)
+
+	// A VPC-access domain (created with VPCOptions carrying subnets) has no
+	// public Endpoint; instead it exposes Endpoints["vpc"], and AWS enriches the
+	// returned VPCOptions with a derived VPCId and AvailabilityZones.
+	endpoint := m.endpointFor(in.DomainName)
+
+	var endpoints map[string]string
+
+	if vpcEndpoints, enriched, ok := m.deriveVPC(raw, in.DomainName); ok {
+		endpoint = ""
+		endpoints = vpcEndpoints
+		raw["VPCOptions"] = enriched
+	}
 
 	status := driver.DomainStatus{
 		DomainID:               m.opts.AccountID + "/" + in.DomainName,
@@ -70,14 +86,15 @@ func (m *Mock) CreateDomain(_ context.Context, in driver.CreateDomainInput) (*dr
 		Deleted:                false,
 		Processing:             false,
 		EngineVersion:          engine,
-		Endpoint:               m.endpointFor(in.DomainName),
+		Endpoint:               endpoint,
+		Endpoints:              endpoints,
 		DomainProcessingStatus: driver.ProcessingActive,
 		ClusterConfig:          cfg,
 		AccessPolicies:         in.AccessPolicies,
 		AdvancedOptions:        copyTags(in.AdvancedOptions),
 		IPAddressType:          in.IPAddressType,
 		EngineMode:             in.EngineMode,
-		RawOptions:             copyRaw(in.RawOptions),
+		RawOptions:             copyRaw(raw),
 		CreatedAt:              now,
 	}
 
@@ -89,7 +106,8 @@ func (m *Mock) CreateDomain(_ context.Context, in driver.CreateDomainInput) (*dr
 			AccessPolicies:  in.AccessPolicies,
 			AdvancedOptions: copyTags(in.AdvancedOptions),
 			IPAddressType:   in.IPAddressType,
-			RawOptions:      copyRaw(in.RawOptions),
+			RawOptions:      copyRaw(raw),
+			CreatedAt:       now,
 			UpdatedAt:       now,
 		},
 		tags:     copyTags(in.Tags),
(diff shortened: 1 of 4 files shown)


NitinKumar004 (commented):
Fast-lane review at pushed SHA 7bbbad34.

**Gates (GOMAXPROCS=3, -p=3, pushed SHA):**
- `go build ./...`, `go vet ./...` — clean.
- `go test ./server/aws/opensearch/... ./providers/aws/opensearch/... ./services/opensearch/...` — all pass, including the two new `TestSDKDescribeDomainConfigEnvelope` / `TestSDKVPCDomainEndpoints` and every pre-existing OpenSearch test (lifecycle, tags, update-persist, cascades, pagination, dup-name).
- `go test -race ./providers/aws/opensearch/...` — clean.
- `gofmt -l` — no diffs.
- `go generate ./...` — zero diff (docs/coverage untouched, as expected: driver interface gained only a struct field).
- `docs/compat` — OpenSearch isn't part of the AWS SDK-compat suite (`compat/aws/` has no opensearch test file, `docs/compat/compat.json` has no opensearch entries), so the "compat zero-diff" claim is trivially true, not something this PR could regress.
- `golangci-lint run ./providers/aws/opensearch/... ./server/aws/opensearch/... ./services/opensearch/...` — 1 finding, but it's the pre-existing `handler.go:156` nolintlint drift called out in the PR description; `handler.go` isn't in this PR's diff, confirmed via `git diff --name-only development...HEAD`. Zero new lint issues.
- Coverage on the new code (measured with `-coverpkg` so cross-package calls from the server-level SDK tests attribute correctly): `deriveVPC` 80%, `vpcEndpointHost`/`deterministicVPCID`/`azsForSubnets` 100%, `marshalJSON` 75% (only the JSON-marshal-error fallback branch untested, unreachable in practice). Good.

**Claim verification:**
- F1: confirmed `domainConfigToWire` now wraps every modeled block *and* every `RawOptions` entry (EBS/VPC/encryption/endpoint/advanced-security/…) in `statusEnvelope` with `State/PendingDeletion/CreationDate/UpdateDate/UpdateVersion`. `domainStatusToWire`/`mergeRaw` (used by `DescribeDomain`) is untouched — flat raw blocks confirmed still flat. `UpdateDomainConfig` returns the changed raw block through the same envelope path (verified: EBS VolumeSize=50 round-trips enveloped).
- F2: `CreateDomain` now clears `Endpoint` and sets `Endpoints["vpc"]` only when `VPCOptions` carries subnets; public domains keep `Endpoint` and get no `Endpoints["vpc"]` key. Verified no aliasing risk — `raw` is a copy of `in.RawOptions` (via `copyRaw`, which deep-copies each `json.RawMessage`), so mutating `raw["VPCOptions"]` doesn't touch the caller-owned input.
- F3: `deterministicVPCID` hashes sorted subnet IDs (fnv32a) so ordering doesn't matter and output is stable across calls; the enriched VPCOptions raw block is stored into both `DomainStatus.RawOptions` and `DomainConfig.RawOptions`, so `DescribeDomain.VPCOptions.VPCId` is populated as claimed.

**Non-blocking observations (not required for this fix's scope):**
- `deriveVPC` only runs in `CreateDomain`. `UpdateDomainConfig` merges a raw `VPCOptions` patch via `applyConfigPatch` without going through `deriveVPC`, so updating a domain's VPCOptions post-create won't flip `Endpoint`/`Endpoints["vpc"]` or enrich `VPCId`/`AvailabilityZones`. Given VPC options aren't in the PR's stated scope (CreateDomain-only) and this isn't a regression (update never handled VPCOptions specially before), it's fine to leave as a follow-up rather than block this fix.
- `statusEnvelope`'s `UpdateVersion` is hardcoded to `1` regardless of how many times a domain's config has actually changed. Real AWS increments it per change. Cosmetic — unlikely any client asserts on the exact number — but flagging in case a future consumer keys off it.

No blocking issues found. Verdict: MERGE.

PR #691, opened 2026-08-26
fix(aws-sesv2): emit event-destination sub-blocks (SNS/CloudWatch/Kinesis) + GetEmailIdentity Policies

NitinKumar004:
## What

Two AWS SES v2 wire-serialization fixes found by the deep-e2e real-user audit. In both cases the driver + provider already **store** the data correctly — only the server wire response dropped it.

### [HIGH] GetConfigurationSetEventDestinations dropped the destination sub-block
`aws_sesv2_configuration_set_event_destination` reads the `event_destination` block on refresh (e.g. `sns_destination.topic_arn`). The wire response only emitted `Name`/`Enabled`/`MatchingEventTypes`, so the destination came back `nil` → perpetual Terraform drift for every event-destination user.

- `GetConfigurationSetEventDestinations` now emits the configured destination sub-object: `SnsDestination{TopicArn}`, `KinesisFirehoseDestination{IamRoleArn, DeliveryStreamArn}`, `CloudWatchDestination{DimensionConfigurations[]}`, and `PinpointDestination{ApplicationArn}`.
- The create/update input side now captures the Kinesis `IamRoleArn`, the CloudWatch `DimensionConfigurations` (name / value-source / default-value), and the Pinpoint `ApplicationArn` — previously discarded (dimensions were dropped and the namespace hardcoded).
- Driver `EventDestination` / `EventDestinationInput` gained the matching fields so each type round-trips.

### [MED] GetEmailIdentity omitted the Policies map
`GetEmailIdentity` never serialized `Policies`, even though `GetEmailIdentityPolicies` (and the stored `Identity.Policies`) had them.

- Added `Policies` to the `GetEmailIdentity` response, populated from the stored identity.
- `copyIdentity` now deep-copies the `Policies` map (avoids aliasing / data races).

## Why
Both are wire-only round-trip bugs that break IaC refresh/plan stability. Additive wire changes only — no driver interface method added; coverage docs and the compat matrix are zero-diff.

## Tests
Real `aws-sdk-go-v2/service/sesv2` reproductions:
- `TestSDKEventDestinationSubBlocksRoundTrip` — SNS, CloudWatch (with a dimension configuration), and Kinesis Firehose destinations each round-trip through create → get.
- `TestSDKGetEmailIdentityIncludesPolicies` — `CreateEmailIdentityPolicy(p1)` then `GetEmailIdentity.Policies` contains `p1`.

Existing SES v2 tests (DKIM tokens, config-set round-trip, SendEmail verified-sender, GetAccount, tags) stay green, including `-race`.

### DIFF
--- a/providers/aws/sesv2/configset_options.go
+++ b/providers/aws/sesv2/configset_options.go
@@ -65,13 +65,13 @@ func (m *Mock) DeleteConfigurationSetEventDestination(_ context.Context, configS
 	err := m.withConfigSet(configSet, func(cs *driver.ConfigurationSet) {
 		kept := cs.EventDestinations[:0]
 
-		for _, ed := range cs.EventDestinations {
-			if ed.Name == name {
+		for i := range cs.EventDestinations {
+			if cs.EventDestinations[i].Name == name {
 				found = true
 				continue
 			}
 
-			kept = append(kept, ed)
+			kept = append(kept, cs.EventDestinations[i])
 		}
 
 		cs.EventDestinations = kept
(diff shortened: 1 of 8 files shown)


NitinKumar004 (commented):
Verified against pushed SHA `64839b66` (matches `headRefOid`).

**Build/vet/tests**: `go build ./...`, `go vet ./...` clean. `go test ./server/aws/sesv2/... ./providers/aws/sesv2/... ./services/sesv2/...` — all green including the two new tests (`TestSDKEventDestinationSubBlocksRoundTrip`, `TestSDKGetEmailIdentityIncludesPolicies`). `-race` on `./providers/aws/sesv2/...` clean.

**F1 (event destinations)**: Confirmed SNS/CloudWatch(with dimension)/Kinesis round-trip via the PR's own test. Independently wrote and ran extra coverage the PR didn't include:
- `PinpointDestination` round-trip (untested by the PR) — passes: `ApplicationArn`, `Enabled`, `Name`, `MatchingEventTypes` all preserved.
- Multiple destinations on one config set — already covered (3 destinations in `TestSDKEventDestinationSubBlocksRoundTrip`).

`driver.go` interface is untouched — confirmed via `git diff` on `services/sesv2/driver/driver.go` (zero diff), matching the "additive, no interface method added" claim.

**F2 (GetEmailIdentity.Policies)**: Confirmed `copyIdentity`/`copyTags` deep-copies (`GetEmailIdentity` holds the identity's `RLock` during copy). Independently verified with a scratch test: two policies both present, plus a concurrent-access stress test (20 concurrent `GetEmailIdentity` readers mutating their returned map + a writer looping `UpdateEmailIdentityPolicy`) run under `-race` — clean, and the store was not corrupted by mutating the returned map.

**Incidental gocritic fix** (`DeleteConfigurationSetEventDestination` range-value → index-based loop): correct, behavior-preserving.

**Gates**: `gofmt -l` empty on all 8 changed files. `golangci-lint run` on the 3 touched packages: 0 issues. `go generate ./...` zero diff. `go run ./internal/compatgen` exits 0 (both AWS/Azure/GCP compat suites `ok`, no red-cell abort) and `git diff --exit-code docs/compat/` is zero-diff.

PR #692, opened 2026-08-26
fix(gcp-compute): zone-scope disk get/delete/resize/list/insert — no cross-zone resolution or data-loss

NitinKumar004:
## Summary

A confirming re-audit surfaced residue in GCP Compute Engine disk handling: disks were keyed by name **project-wide** instead of **per-zone**, even though the zone is stored on `VolumeInfo.AvailabilityZone` (and already used correctly by `aggregatedList`). `findDiskByName` scanned `DescribeVolumes` matching only the disk-name tag, never the zone.

Root cause: zone-blind `findDiskByName`. It produced 3 real HIGH bugs:

- **[HIGH, DATA LOSS]** `get`/`delete`/`resize` on zone A resolved a disk living in zone B — a `DELETE` in the wrong zone destroyed another zone's disk, and responses mislabeled zone/selfLink as the request zone. GCP-correct: **404 notFound** when the disk isn't in the requested zone.
- **[HIGH]** `disks.list` returned **every** zone's disks, each mislabeled with the requested zone's selfLink/zone. GCP-correct: only the requested zone's disks, each with its own zone.
- **[HIGH]** A duplicate disk name across zones falsely returned **409**. GCP-correct: disk names are unique **per-zone** — the same name in a different zone must succeed.

## Changes

- `findDiskByName` now takes a zone and matches **both** the disk-name tag **and** `AvailabilityZone` (mirrors instances' `findInZone`); a disk with no recorded zone still matches any zone for driver-created disks.
- `get`/`delete`/`resize`/insert-conflict pass the request zone → 404 (GCP error shape) when the disk isn't in that zone.
- `toDiskResponse` derives zone / selfLink / type-link from the disk's own `AvailabilityZone`, not the request scope.
- `listDisks` filters to the requested zone; each response carries the disk's own zone.
- `images.go` source-disk size lookup extracts the zone from the disk self-link.
- `aggregatedList` still spans zones (unchanged).

Verified against GCP Compute disks REST semantics: disks are zonal and unique per-zone; get/delete/resize are zone-scoped; list is per-zone; aggregatedList spans zones.

## Tests

New real google-cloud-go DisksClient reproductions:
- insert in `us-west1-b`, then get/delete/resize via `us-central1-a` → 404 (west disk untouched)
- list(`us-central1-a`) with disks in two zones → only the central disk, correct selfLinks
- same disk name in two different zones → both insert OK, distinct sizes
- aggregatedList still spans both zones

Existing GCP compute tests remain green.

### DIFF
--- a/server/gcp/compute/disks.go
+++ b/server/gcp/compute/disks.go
@@ -71,7 +71,7 @@ func (h *Handler) insertDisk(w http.ResponseWriter, r *http.Request, rp gcprest.
 		return
 	}
 
-	if _, err := findDiskByName(r.Context(), h.compute, req.Name); conflictIfExists(w, err, "disk "+req.Name+" already exists") {
+	if _, err := findDiskByName(r.Context(), h.compute, req.Name, rp.ScopeName); conflictIfExists(w, err, "disk "+req.Name+" already exists") {
 		return
 	}
 
@@ -96,7 +96,7 @@ func (h *Handler) insertDisk(w http.ResponseWriter, r *http.Request, rp gcprest.
 
 //nolint:gocritic // rp is a request-scoped value
 func (h *Handler) getDisk(w http.ResponseWriter, r *http.Request, rp gcprest.ResourcePath) {
-	vol, err := findDiskByName(r.Context(), h.compute, rp.ResourceName)
+	vol, err := findDiskByName(r.Context(), h.compute, rp.ResourceName, rp.ScopeName)
 	if err != nil {
 		gcprest.WriteCErr(w, err)
 		return
@@ -121,6 +121,10 @@ func (h *Handler) listDisks(w http.ResponseWriter, r *http.Request, rp gcprest.R
 	out := make([]diskResponse, 0, len(vols))
 
 	for i := range vols {
+		if !diskInZone(&vols[i], rp.ScopeName) {
+			continue
+		}
+
 		scope := rp
 		name := tagOr(vols[i].Tags, gcpDiskNameTag, vols[i].ID)
 		scope.ResourceName = name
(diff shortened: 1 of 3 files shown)


NitinKumar004 (commented):
**Verdict: approve.** Verified against the actual pushed head (1f066ff), not the description.

**Root cause confirmed real.** Pre-fix `findDiskByName` matched only the `cloudemu:gcpDiskName` tag and ignored zone, so a `DELETE` addressed to zone A resolved and destroyed a same-named disk in zone B — a genuine HIGH cross-zone data-loss path. The zone is already stored on `VolumeInfo.AvailabilityZone` (set on insert, used by `aggregatedList`), so scoping the lookup is the correct minimal fix.

**Independently reproduced (own google-cloud-go DisksClient tests, separate from the PR's):**
- Insert `payload` (42GB) in `us-west1-b`; GET/DELETE/RESIZE via `us-central1-a` → 404, and the west disk is fully intact afterwards (still 42GB, resizable to 100, then deletable in its own zone). The data-loss delete is gone.
- Two same-named disks (`twin`) in two zones both insert (no false 409); deleting the central one leaves the west one alive with its own size/zone/selfLink.
- `List(us-central1-a)` returns only central disks, each carrying its own zonal `zone`, `selfLink`, and `diskTypes` link — not the request scope stamped on all.
- `aggregatedList` still spans both zones.
- `zoneFromDiskURL` unit-tested against a full self-link, a bare name, empty, and malformed inputs (`zones/`, trailing `zones`) — correct extraction, no panic.

**Blast radius clean.** `findDiskByName` has exactly two callers (`disks.go`, `images.go`), both updated; the build proves the new signature. Azure's like-named `findDiskByName` is a separate package function, untouched. `attachDisk`/`detachDisk` record the disk source URL as an instance tag and never call `findDiskByName`, so they're unaffected and still resolve correctly. `toDiskResponse` deriving identity from `vol.AvailabilityZone` (falling back to request scope only when unset) is consistent because the lookup already guarantees zone match for wire-created disks.

**Gates (GOMAXPROCS=3, -p=3):** build ./..., vet, `go test` on server/gcp/compute + providers/gcp/compute + services/compute, `-race` on providers/gcp/compute, gofmt, `go generate ./...` zero-diff, golangci-lint on the changed package — all green. Coverage on new code solid (diskInZone 100%, getDisk 100%, findDiskByName 88.9%, toDiskResponse 88.9%).

**Non-blocking note (pre-existing, out of scope):** `diskUsersByName` keys attached disks by name only, so two same-named disks in different zones would share a `users[]` list. Latent, not introduced here, and unrelated to the data-loss fix.

PR #693, opened 2026-08-26
fix(azure-compute): VM PATCH applies vmSize/tags; in-place disk update (no dup, stable uniqueId); VMSS delete + RG cascade

NitinKumar004:
Confirming re-audit residue for Azure compute: fixes 2 HIGH data-loss bugs, a MED disk-identity churn, and a HIGH broken op, each verified against the official Azure ARM REST docs (VirtualMachines Update, Disks CreateOrUpdate, VirtualMachineScaleSets Delete) and covered by real azure-sdk-for-go (armcompute) reproduction tests.

## What changed

**VM PATCH Update dropped vmSize/tags (HIGH, silent drift).** `update()` only reconciled `storageProfile.dataDisks` and never applied the modeled mutable fields. It now applies the fields the PATCH body supplies — `hardwareProfile.vmSize` (resize), `tags` (RFC 7386 merge into the existing set), `identity` — via a new merge-aware driver `PatchInstance`, leaving omitted fields (priority, licenseType, existing tags) untouched rather than routing through `UpdateInstance`'s full cfg-replace. `az vm resize` / `az vm update --set tags.*` now take effect; no Terraform perpetual diff.

**Disk CreateOrUpdate on an attached disk duplicated the volume + dropped the update (HIGH).** The old path did `DeleteVolume` (which returns FailedPrecondition for an attached disk — the error was discarded) then `CreateVolume` anyway, yielding two volumes sharing the same disk-name tag with the VM's LUN pointing at the stale one. Replaced with an in-place `AzureDiskUpdater.UpdateVolume` that mutates the existing volume by ID.

**Disk uniqueId + timeCreated churned on every re-PUT (MED).** Because delete+create regenerated the volume ID (and thus the derived uniqueId) and CreatedAt. The in-place update preserves both, so `uniqueId` + `timeCreated` stay stable across updates.

**VMSS DELETE returned 501 (HIGH, broken op).** `serveScaleSet` handled only PUT/GET. Added a provider `DeleteScaleSet`, a DELETE branch returning 202 + async poller, and included scale sets in `PurgeResourceGroup` so an RG cascade tears them down. `terraform destroy` of a scale set now succeeds and a VMSS does not survive its resource group.

## Tests (real armcompute SDK)
- VM `BeginUpdate({Tags, VMSize})` → Get shows the resize + added tag; a pre-existing omitted tag is preserved.
- Disk: create → attach via VM PATCH → re-PUT with a new tag → exactly ONE backing volume, tag applied, `uniqueId` + `timeCreated` unchanged from the first PUT, still Attached.
- VMSS `BeginDelete` → 202 → DONE → Get NotFound; a VMSS in an RG is gone after the RG is deleted.

Existing compute tests (VM lifecycle, disk attach/detach, VMSS create/scale/scale-to-zero, cross-RG isolation) stay green, including `-race`.

Notes: `go generate` regenerated `docs/coverage` for the two new optional driver capabilities; the AWS compat suite passes (compatgen exit 0, `docs/compat` clean).

### DIFF
--- a/docs/coverage/aws/ec2.md
+++ b/docs/coverage/aws/ec2.md
@@ -58,6 +58,14 @@ AzureDiskAccessor is an optional Azure-only capability for the managed-disk
 | `GrantDiskAccess` | GrantDiskAccess issues a time-bounded SAS URI granting the requested |
 | `RevokeDiskAccess` | RevokeDiskAccess revokes any SAS access previously granted to the disk. |
 
+### AzureDiskUpdater
+
+AzureDiskUpdater is an optional Azure-only capability for an in-place managed
+
+| Operation | Description |
+| --- | --- |
+| `UpdateVolume` | UpdateVolume mutates the existing volume id in place from cfg (size, sku/ |
+
 ### AzureSSHKeyUpdater
 
 AzureSSHKeyUpdater is an optional Azure-only capability for the sshPublicKeys
@@ -74,6 +82,7 @@ AzureVMController is an optional Azure-only capability supporting the ARM
 | --- | --- |
 | `Deallocate` | Deallocate stops the guest and releases the allocated compute |
 | `GeneralizeInstance` | GeneralizeInstance marks an instance as generalized (Azure Generalize |
+| `PatchInstance` | PatchInstance applies a merge-patch (ARM PATCH Update / BeginUpdate) to an |
 | `PowerOff` | PowerOff stops the guest OS while keeping the VM allocated |
 | `UpdateInstance` | UpdateInstance overwrites the mutable configuration of an existing |
 
--- a/docs/coverage/azure/virtualmachines.md
+++ b/docs/coverage/azure/virtualmachines.md
@@ -58,6 +58,14 @@ AzureDiskAccessor is an optional Azure-only capability for the managed-disk
 | `GrantDiskAccess` | GrantDiskAccess issues a time-bounded SAS URI granting the requested |
 | `RevokeDiskAccess` | RevokeDiskAccess revokes any SAS access previously granted to the disk. |
 
+### AzureDiskUpdater
+
+AzureDiskUpdater is an optional Azure-only capability for an in-place managed
+
+| Operation | Description |
+| --- | --- |
+| `UpdateVolume` | UpdateVolume mutates the existing volume id in place from cfg (size, sku/ |
+
 ### AzureSSHKeyUpdater
 
 AzureSSHKeyUpdater is an optional Azure-only capability for the sshPublicKeys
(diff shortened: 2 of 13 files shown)


NitinKumar004 (commented):
Adversarial review — verdict: MERGE (green on the pushed head 917d8692).

Verified on a clean checkout of the PR head (not the fixer's report):

VM PATCH (HIGH)
- update() now applies the PATCH body's modeled mutable fields via the new merge-aware driver PatchInstance (locked instances.Update callback), not UpdateInstance's full cfg-replace. Confirmed by real armcompute BeginUpdate:
  - resize + new tag applied; a pre-existing omitted tag preserved (RFC 7386 merge).
  - a PATCH that omits tags entirely (nil Tags) does NOT wipe existing tags — the original full-replace bug is gone.
  - a tags-only PATCH leaves vmSize untouched, and vice-versa.
  - identity update path exercised (existing identity tests green).
- omitted keys (priority/licenseType/internal ARM-name tag) left intact.

Disk in-place (HIGH + MED)
- createOrUpdate on an existing disk routes through updateExistingDisk -> AzureDiskUpdater.UpdateVolume (type-asserted optional), replacing the old DeleteVolume(discarded err)+CreateVolume that duplicated an attached disk.
- Attached re-PUT: exactly ONE backing volume, tag applied, uniqueId + timeCreated identical to first PUT, attachment (State/AttachedTo/LUN) intact. Detached re-PUT resize: single volume, size updated, uniqueId/timeCreated stable.
- UpdateVolume preserves ID/CreatedAt/State/AttachedTo/Device and mutates only the cost fields. The unlocked Get-pointer-mutate is the established provider pattern (AttachVolume/DetachVolume do the same); -race clean.

VMSS (HIGH)
- serveScaleSet gained a DELETE branch -> provider DeleteScaleSet -> 202 + async poller; poller settles and follow-up Get is NotFound.
- ScaleSet now records ResourceGroup; PurgeResourceGroup calls purgeScaleSets so an RG cascade removes the VMSS. RG comparison case-insensitive.
- PUT-capacity scaling and scale-to-zero not regressed (existing SDK tests green). InstanceView/named-scale-action remain deferred (out of scope).

Blast radius
- PatchInstance added to Azure-only AzureVMController; AzureDiskUpdater is a new optional interface. No AWS/GCP implementer of either — no shared-interface break. Both consumed via type-assert. build ./... and vet ./... clean.

Gates (GOMAXPROCS=3, -p=3): build ok; vet ok; targeted tests ok; -race on providers/azure/virtualmachines ok; golangci-lint on the 4 changed pkg trees 0 issues; gofmt clean; go generate zero diff (coverage committed); compatgen exit 0 with docs/compat clean.

No blocking issues.

PR #694, opened 2026-08-26
fix(azure-networking): in-place subnet update + NSG/NAT (dis)association; peering-delete disconnects reciprocal; RG-scoped NSG validation

NitinKumar004:
## Summary

A confirming re-audit of the Azure networking wire layer surfaced residue from the first round. This lands three fixes, all reproduced with the real `azure-sdk-for-go/armnetwork` SDK.

### 1. [HIGH] Subnet CreateOrUpdate was create-only (one root, three symptoms)
`upsertSubnet`'s existing-subnet branch was merge-only, so an in-place update was a silent no-op. Real ARM's `Subnets.CreateOrUpdate` is a full replacement. Fixed so the existing-subnet path:
- **updates the address prefix** when the body's `addressPrefix` differs (via a new optional `SubnetCIDRUpdater` driver capability, type-asserted; AWS/GCP model an immutable subnet CIDR and do not implement it);
- **replaces** the NSG / NAT gateway associations from the request body — an omitted `networkSecurityGroup` / `natGateway` now **clears** the association (the `azurerm_subnet` in-place update + `azurerm_subnet_network_security_group_association` delete path).

The same in-place update is now applied on the whole-VNet PUT path for inline subnets (changed CIDR + replaced NSG). The subnet-in-use guard and full-replace-on-whole-VNet-PUT semantics (an omitted subnet is still deleted) are preserved.

### 2. [MED] Peering delete leaves the surviving side stuck 'Connected'
Deleting one side of a two-way VNet peering now transitions the orphaned reciprocal side to `Disconnected` (wiring the previously-dead `AzurePeeringStateDisconnected` constant), matching real ARM.

### 3. [LOW] Cross-RG NSG reference leak
`resolveSubnetNSG` and the NIC-create NSG validation were name-only. They now resolve within the NSG id's own resource group (`findNSGInGroup`), mirroring how NAT gateways are scoped — a NIC/subnet in one RG can no longer reference an NSG that exists only in another.

## Tests
New real-SDK reproductions (`server/azure/vnet/subnet_update_test.go`): subnet CIDR re-PUT, NSG disassociation, peering-delete reciprocal disconnect, and cross-RG NSG rejection (NIC + subnet). First-round tests remain green.

### DIFF
--- a/docs/coverage/aws/vpc.md
+++ b/docs/coverage/aws/vpc.md
@@ -364,6 +364,14 @@ SubnetAttributes is an OPTIONAL capability, discovered by type assertion.
 | --- | --- |
 | `ModifySubnetAttribute` |  |
 
+### SubnetCIDRUpdater
+
+SubnetCIDRUpdater is an OPTIONAL capability, discovered by type assertion. It
+
+| Operation | Description |
+| --- | --- |
+| `UpdateSubnetCIDR` |  |
+
 ### TrafficMirroring
 
 TrafficMirroring is an OPTIONAL AWS capability (type-asserted).
--- a/docs/coverage/azure/vnet.md
+++ b/docs/coverage/azure/vnet.md
@@ -364,6 +364,14 @@ SubnetAttributes is an OPTIONAL capability, discovered by type assertion.
 | --- | --- |
 | `ModifySubnetAttribute` |  |
 
+### SubnetCIDRUpdater
+
+SubnetCIDRUpdater is an OPTIONAL capability, discovered by type assertion. It
+
+| Operation | Description |
+| --- | --- |
+| `UpdateSubnetCIDR` |  |
+
 ### TrafficMirroring
 
 TrafficMirroring is an OPTIONAL AWS capability (type-asserted).
--- a/docs/coverage/coverage.json
+++ b/docs/coverage/coverage.json
@@ -7117,6 +7117,15 @@
         ]
       },
       {
+        "name": "SubnetCIDRUpdater",
+        "doc": "SubnetCIDRUpdater is an OPTIONAL capability, discovered by type assertion. It",
+        "operations": [
+          {
+            "name": "UpdateSubnetCIDR"
+          }
+        ]
+      },
+      {
         "name": "TrafficMirroring",
         "doc": "TrafficMirroring is an OPTIONAL AWS capability (type-asserted).",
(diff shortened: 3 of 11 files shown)


NitinKumar004 (commented):
Adversarial review of the pushed head (8c7a551c), clean-built in an isolated worktree — not the fixer's report.

Verdict: LGTM / merge-ready. All three fixes verified end-to-end with the real azure-sdk-for-go/armnetwork SDK, including the regression surfaces the PR's own tests don't cover.

Gates (GOMAXPROCS=3, -p=3, one build at a time)
- go build ./... , go vet (changed pkgs): clean
- go test ./server/azure/vnet/... ./providers/azure/vnet/... ./services/networking/...: pass
- go test -race ./providers/azure/vnet/...: pass
- go generate ./...: zero diff (committed coverage matches)
- go run ./internal/compatgen + git diff docs/compat/: exit 0, zero diff
- gofmt -l changed files: clean
- golangci-lint (changed pkgs): the one lll finding (driver.go:618 CreateNetworkInterface, 142 chars) predates this PR (commit 6fc67825d); the added SubnetCIDRUpdater interface and all new code is lint-clean. New-only: clean.

Behavior verified with my own real-SDK reproductions (beyond the PR's tests)
- Subnet CIDR: standalone re-PUT 10.0.1.0/24 -> 10.0.5.0/24 lands; whole-VNet PUT with an inline subnet's changed CIDR lands.
- Full-replace intact: a whole-VNet re-PUT that OMITS a subnet still DELETES it; subnet-in-use guard still blocks an in-use standalone delete.
- NSG (dis)association: omitted networkSecurityGroup on re-PUT clears it; a body still carrying the NSG PRESERVES it (not wrongly cleared); a different NSG REPLACES it. NAT gateway association shares the same generic replaceSubnetAssociation path.
- Peering: deleting one side transitions the surviving reciprocal to Disconnected; a clean two-sided delete still removes both.
- RG-scoped NSG: NIC and subnet in rg-a referencing an NSG only in rg-b -> 404; same-RG reference works for both. findNSGByName is still used by effective_nsg.go (no dead code).

Blast radius: UpdateSubnetCIDR is an optional type-asserted capability implemented only by the Azure mock; AWS/GCP are unaffected and compile clean. No dangling references to the renamed createMissingSubnets.

No blocking issues.

PR #695, opened 2026-08-26
fix(gcp-networking): persist subnetwork secondaryIpRanges/description + network routingMode/mtu/description; reject firewall on missing network; expandIpCidrRange; aggregatedList

NitinKumar004:
## Summary

Fixes GCP Compute networking (VPC/subnet/firewall) gaps found by a confirming re-audit against the official GCP Compute REST API. Builds on the first-round fixes (#635–#656: IPv4Range, dup-409, delete-in-use, region-scoping) without regressing them.

### Fixes
- **[HIGH] Subnetwork `description` + `secondaryIpRanges` dropped** — alias ranges (GKE VPC-native pods/services, Terraform `google_compute_subnetwork.secondary_ip_range`) and the subnet description were silently discarded on insert and read back empty. Now persisted (secondary ranges JSON-encoded verbatim in a reserved tag, mirroring the firewall spec) and reconstructed on Get/List.
- **[HIGH] Network `description`, `routingConfig.routingMode`, `mtu` dropped** — Terraform `google_compute_network` sets these; the silent drop produced a perpetual plan diff. Now persisted in VPC tags and emitted on read.
- **[MED] Firewall referencing an unknown network fabricated a phantom VPC** — `insertFirewall` swallowed the network-resolve error and created a synthetic VPC that then leaked into `networks.list`. A supplied-but-missing network now returns 404 (matching real GCP); no phantom VPC is created. Firewall inserts with no network keep the default-network fallback.
- **[MED] `subnetworks.expandIpCidrRange` returned 405 (unrouted)** — the verb is now routed: the new range is validated as a strict superset of the current range (subset/equal rejected) and the stored subnet CIDR is widened via a new provider capability (`ExpandSubnetCIDR`, discovered by type assertion; kept out of the portable Networking interface since GCP's expand verb has no cross-cloud analog).
- **[MED] `aggregatedList` returned zero items for subnetworks/addresses** — now emits `compute#subnetworkAggregatedList` / `compute#addressAggregatedList` with items grouped into `regions/{region}` buckets (plus a warned `global` bucket), as gcloud and Terraform data sources expect.

### Tests
Real `cloud.google.com/go/compute/apiv1` reproductions: subnetwork insert with description + 2 secondary ranges round-trips; network insert with description + `routingMode=GLOBAL` + `mtu=1500` round-trips; firewall insert against a missing network → error with no phantom in `networks.list`; `ExpandIpCidrRange` superset widens (subset errors); `SubnetworksClient.AggregatedList` / `AddressesClient.AggregatedList` return items grouped by region. First-round tests (region-scoping, routes, addresses) stay green, including `-race`.

### Gates
build · vet · test · `-race` · gofmt · golangci-lint (0 new) · `go generate` zero diff · compatgen zero diff. No driver-interface or compat-matrix change.

### DIFF
--- a/providers/gcp/vpc/vpc.go
+++ b/providers/gcp/vpc/vpc.go
@@ -166,6 +166,21 @@ func (m *Mock) DescribeSubnets(_ context.Context, ids []string) ([]driver.Subnet
 	return describeResources(m.subnets, ids, toSubnetInfo), nil
 }
 
+// ExpandSubnetCIDR widens a subnetwork's primary IP range, backing GCP's
+// subnetworks.expandIpCidrRange. The wire layer validates that cidr is a
+// superset of the current range before calling; here it just replaces the
+// stored block under the store lock.
+func (m *Mock) ExpandSubnetCIDR(_ context.Context, id, cidr string) error {
+	if !m.subnets.Update(id, func(s *subnetData) *subnetData {
+		s.CIDRBlock = cidr
+		return s
+	}) {
+		return cerrors.Newf(cerrors.NotFound, "subnet %q not found", id)
+	}
+
+	return nil
+}
+
 // CreateSecurityGroup creates a new firewall rule group.
 func (m *Mock) CreateSecurityGroup(_ context.Context, cfg driver.SecurityGroupConfig) (*driver.SecurityGroupInfo, error) {
 	if cfg.Name == "" {
--- a/server/gcp/vpc/addresses.go
+++ b/server/gcp/vpc/addresses.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"net/http"
 	"sort"
+	"strings"
 	"sync"
 
 	"github.com/stackshy/cloudemu/v2/internal/pagination"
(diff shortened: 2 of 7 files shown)


NitinKumar004 (commented):
Reviewed at head 1de7100. Clean-built the actual pushed tree (detached checkout of the head SHA, not the fixer's report) and ran an independent real-`cloud.google.com/go/compute/apiv1` adversarial suite alongside the gate matrix. Verdict: MERGE.

Gate matrix (GOMAXPROCS=3, -p=3, one go cmd at a time):
- go build ./... clean; go vet (server/gcp/vpc, providers/gcp/vpc, services/networking) clean.
- go test -p=3 ./server/gcp/vpc/... ./providers/gcp/vpc/... ./services/networking/... green.
- go test -race ./providers/gcp/vpc/... AND ./server/gcp/vpc/... green.
- go generate ./... zero diff; go run ./internal/compatgen + git diff docs/compat/ zero diff.
- gofmt -l clean; golangci-lint on changed pkgs = 10 issues, all on pre-existing lines outside the diff hunks (vpc.go AddIngress/Egress/Remove, addresses.go key/getAddress, routers.go scope/getRouter/listRouters) — 0 NEW.

Adversarial verification (my own tests, beyond the PR's):
- Subnet secondaryIpRanges/description: exact round-trip (order + RangeName/IpCidrRange preserved); a subnet with NO secondary ranges fabricates none and returns empty description.
- Network description + routingConfig.routingMode=GLOBAL + mtu=1500 all round-trip; a bare custom-mode network returns empty desc/routingMode, mtu=0, and correctly OMITS IPv4Range (first-round behavior intact).
- Phantom VPC: firewall insert referencing a missing network returns an error AND networks.list stays empty (no phantom leaked); no-network insert keeps the default-network fallback; firewall against a real network works.
- expandIpCidrRange: strict superset widens and GET shows the new CIDR; EQUAL, subset, and unrelated ranges are all rejected with the stored CIDR left unchanged; Operation envelope shape correct.
- aggregatedList (subnetworks + addresses): each resource lands in its OWN regions/{region} bucket; empty scope handled with a NO_RESULTS_ON_PAGE global warning bucket, no crash.

Blast radius:
- ExpandSubnetCIDR added only to providers/gcp/vpc Mock, discovered by type assertion; services/networking/driver untouched — no portable interface change, compat matrix unchanged.
- Reserved tags (cloudemu:gcpNetDesc/RoutingMode/Mtu, cloudemu:gcpSubnetDesc/Secondary) are distinct keys with no collision; GCP VPC responses expose no labels surface, so tags cannot leak to users.
- resolveNetwork now propagates its not-found error only when a network ref was supplied (empty ref still returns "",nil), preserving the fallback path.

No blocking issues.

PR #696, opened 2026-08-26
fix(gcp-gcs): enforce delete/read/update/compose preconditions; versioned-delete archival; insert-time system props; generation-addressed reads; storageClass

NitinKumar004:
Confirming re-audit residue for GCS object semantics — 2 HIGH (incl. a versioned-delete data-loss bug) plus 5 MED/LOW, concentrated in the read/delete/update/compose precondition surface, versioning, and insert-time system properties. Every behavior cross-checked against the official GCS JSON API docs.

## Bugs fixed

- **[HIGH — data-integrity] Delete precondition not enforced.** `deleteObject` ignored `ifGenerationMatch`/`ifMetagenerationMatch` (+`?generation`). Now routed through a preconditioned delete; mismatch → 412 `conditionNotMet` with the object untouched.
- **[HIGH — DATA LOSS] Versioned delete lost data.** A live `DELETE` on a versioning-enabled bucket unconditionally removed the key (archival only happened on overwrite). Now the current generation is archived as noncurrent, and `list?versions=true` surfaces archived generations of keys that are no longer live. A generation-addressed delete stays permanent (matching GCS).
- **[MED] Insert-time object system properties dropped.** `cacheControl`/`contentEncoding`/`contentDisposition`/`contentLanguage` are now parsed in the multipart metadata and threaded through `storeObject` → `PutObjectGCS` (new `GCSObjectAttrs`) so they persist at INSERT.
- **[MED] `?generation` ignored on read.** `Attrs` and media download now honor `?generation`, reading the addressed revision from live/archived generations.
- **[MED] Conditional read preconditions ignored.** Read handlers now evaluate `ifGenerationMatch`/`ifMetagenerationMatch` (412) and the NotMatch variants (304), consulting both query params (JSON) and `x-goog-if-*` headers (media download).
- **[MED] Object update/patch precondition ignored.** `UpdateObjectGCS` now enforces the precondition → 412.
- **[LOW] Compose destination precondition + source generations ignored.** Compose threads the destination precondition (DoesNotExist/`ifGenerationMatch`) and honors each source's pinned generation.
- **[LOW] Object `storageClass` hardcoded STANDARD.** Objects now store a storage class, defaulting to the bucket's default class, and surface it on read.

## Layering
Driver `GCSExtensions` extended (new `DeleteObjectGCS`/`GetObjectGCS`/`HeadObjectGCS`; `PutObjectGCS`/`UpdateObjectGCS`/`ComposeObjectGCS` take preconditions/attrs/pinned generations). Provider implements archival + generation addressing; wire handler parses params/headers and maps `GCSPreconditionError` → HTTP 412. Prior-round fixes (write preconditions, generation minting, checksums, patch-merge, versioned-list-on-overwrite, bucket attrs/IAM) preserved and still green.

## Tests
Real `cloud.google.com/go/storage` reproductions added: delete-with-mismatched-`ifGenerationMatch` (412, survives), versioned live delete retains 2 noncurrent generations, insert-time system props round-trip, `Generation(gen1).NewReader` → v1 bytes, conditional read mismatch → 412, update bad `MetagenerationMatch` → 412, compose DoesNotExist over existing dst → 412, NEARLINE bucket → object `StorageClass=NEARLINE`.

Gates: build, vet, `-race`, target tests, `golangci-lint` (0 new), gofmt, `go generate` (coverage docs regenerated), and compatgen (no `docs/compat` diff) all green.

### DIFF
--- a/docs/coverage/aws/s3.md
+++ b/docs/coverage/aws/s3.md
@@ -118,6 +118,9 @@ GCSExtensions is an OPTIONAL GCS-specific capability, discovered by type
 | `BucketAttrsGCS` | BucketAttrsGCS returns the bucket's GCS-specific attributes. |
 | `BucketIAMPolicy` |  |
 | `ComposeObjectGCS` | ComposeObjectGCS concatenates the source objects' bytes (in order) into |
+| `DeleteObjectGCS` | DeleteObjectGCS deletes an object honoring pre and optional generation |
+| `GetObjectGCS` | GetObjectGCS returns an object's bytes+info, selecting a specific |
+| `HeadObjectGCS` | HeadObjectGCS returns an object's info, selecting a specific generation |
 | `ListObjectGenerations` | ListObjectGenerations returns every generation (current + archived) of the |
 | `PutObjectGCS` | PutObjectGCS writes an object honoring pre (a failed condition returns a |
 | `SetBucketAttrsGCS` | SetBucketAttrsGCS records the bucket's location and default storage class |
--- a/docs/coverage/azure/blobstorage.md
+++ b/docs/coverage/azure/blobstorage.md
@@ -118,6 +118,9 @@ GCSExtensions is an OPTIONAL GCS-specific capability, discovered by type
 | `BucketAttrsGCS` | BucketAttrsGCS returns the bucket's GCS-specific attributes. |
 | `BucketIAMPolicy` |  |
 | `ComposeObjectGCS` | ComposeObjectGCS concatenates the source objects' bytes (in order) into |
+| `DeleteObjectGCS` | DeleteObjectGCS deletes an object honoring pre and optional generation |
+| `GetObjectGCS` | GetObjectGCS returns an object's bytes+info, selecting a specific |
+| `HeadObjectGCS` | HeadObjectGCS returns an object's info, selecting a specific generation |
 | `ListObjectGenerations` | ListObjectGenerations returns every generation (current + archived) of the |
 | `PutObjectGCS` | PutObjectGCS writes an object honoring pre (a failed condition returns a |
 | `SetBucketAttrsGCS` | SetBucketAttrsGCS records the bucket's location and default storage class |
--- a/docs/coverage/coverage.json
+++ b/docs/coverage/coverage.json
@@ -10728,6 +10728,18 @@
             "doc": "ComposeObjectGCS concatenates the source objects' bytes (in order) into"
           },
           {
+            "name": "DeleteObjectGCS",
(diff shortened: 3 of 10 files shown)


NitinKumar004 (commented):
Reviewed at head 42b96895 (clean checkout, clean build/vet/tests/-race/generate/compat). Verified against the real `cloud.google.com/go/storage` SDK end-to-end.

## Blocking

**Compose per-source pinned generation is broken over the wire.** `composeSource.Generation` (server/gcp/gcs/types.go:111) is declared `int64` with tag `json:"generation,omitempty"`. The GCS JSON API — and the Go SDK — encode a source object's `generation` as a JSON **string** (matching `objectResource.Generation string`). Any compose that pins a source generation therefore fails to unmarshal:

```
c := bkt.Object("dst2").ComposerFrom(bkt.Object("a").Generation(gen1), bkt.Object("b"))
c.Run(ctx)
// googleapi: Error 400: json: cannot unmarshal string into Go struct field
//   composeSource.sourceObjects.generation of type int64, invalid
```

This is the exact "honors each source's pinned generation" behavior the PR claims to add — it hard-400s on the real SDK path. The added compose test only exercises the `DoesNotExist`-over-existing-dst 412 case and never pins a source generation, so the gap is uncovered.

Fix: `Generation int64 \`json:"generation,omitempty,string"\``.

## Minor

**Non-canonical header literals (new-only lint, canonicalheader x4).** readPrecondition (handler.go:625-628) reads `h.Get("x-goog-if-generation-match")` etc. Functionally correct (`http.Header.Get` canonicalizes its argument, and the 304/412 media path is verified working) but flagged as a new lint issue on lines this PR adds. Use the canonical `X-Goog-If-Generation-Match` form.

## Verified correct

- Versioned live delete archives the current generation; `Objects{Versions:true}` still returns the noncurrent generations (the data-loss HIGH), while a non-versioned delete fully removes the object.
- Delete precondition: mismatched `ifGenerationMatch` → 412, object untouched; matching → deletes.
- Conditional read: media path returns 304 for satisfied `x-goog-if-generation-not-match` (header and query) and 412 for a failed match; JSON metadata honors query params.
- Generation-addressed read returns the v1 bytes and v1 attrs (size/generation).
- Update bad `ifMetagenerationMatch` → 412; insert-time system props (CacheControl/ContentEncoding/ContentDisposition/ContentLanguage) round-trip; storageClass defaults to the bucket class (NEARLINE) and honors an explicit per-object class (COLDLINE).
- GCSExtensions remains an optional type-asserted extension; AWS/Azure unaffected; build ./..., -race, go generate zero-diff, compatgen zero-diff all green.

PR #697, opened 2026-08-26
fix(aws-wafv2): correct ARN scope-path/region; block delete of associated WebACL; List pagination

NitinKumar004:
## What

Three real-user WAFv2 fixes found by the deep-e2e audit (repros driven by the real `aws-sdk-go-v2/service/wafv2`).

**F1 — ARN shape (MED).** WebACL/IPSet/RuleGroup/RegexPatternSet ARNs used the raw uppercase Scope constant (`REGIONAL`/`CLOUDFRONT`) as the resource-path segment and put the literal word `global` in the region slot for CloudFront. Real WAFv2 uses the lowercase scope word — `regional` for REGIONAL, `global` for CLOUDFRONT — and anchors CloudFront-scoped ARNs in `us-east-1` (never `global` as the region). REGIONAL ARNs use the configured region. Fixed in `providers/aws/wafv2/wafv2.go` `arn()`.

**F2 — delete guards (MED).** `DeleteWebACL` succeeded while the ACL was still associated with a resource, silently orphaning the association. It now returns `WAFAssociatedItemException` while any association points at the ACL. Likewise `DeleteIPSet`/`DeleteRuleGroup`/`DeleteRegexPatternSet` now reject with `WAFAssociatedItemException` when the set/group is still referenced by a WebACL's rules (reference statements embed the referenced ARN verbatim in the rule JSON). New `WAFAssociatedItemException` exception + `associated()` error helper.

**F3 — List pagination (LOW).** `ListWebACLs`/`ListIPSets`/`ListRuleGroups`/`ListRegexPatternSets` ignored `Limit` and `NextMarker` and returned the full list in one page. The server list path now paginates with stable name ordering via the shared `internal/pagination` helper, returning `NextMarker` when more remain and resuming from it.

## Why

The ARN diverged from real AWS on both casing and CloudFront region, breaking clients/IaC that parse scope out of the ARN or assert `us-east-1` on CloudFront ACLs. Out-of-order deletes silently orphaned associations instead of erroring like AWS. Paginating clients that set `Limit` got everything regardless and never advanced a marker.

LockToken enforcement, verbatim rule round-trip, scope partitioning, and association resolution are unchanged.

## Tests

Real `aws-sdk-go-v2` reproductions added: ARN shape (REGIONAL → `regional/webacl/` + request region; CLOUDFRONT → `global/webacl/` + `us-east-1`, parsed and asserted); associate → delete → `WAFAssociatedItemException`, disassociate → delete succeeds; referenced-item delete guard; and `Limit=2` pagination over 5 ACLs (multiple pages, `NextMarker` resume, no dupes, full set). Existing WAFv2 tests remain green.

### DIFF
--- a/providers/aws/wafv2/associations.go
+++ b/providers/aws/wafv2/associations.go
@@ -1,11 +1,48 @@
 package wafv2
 
 import (
+	"bytes"
 	"context"
 	"strings"
 
 	"github.com/stackshy/cloudemu/v2/services/wafv2/driver"
 )
+
+// webACLAssociated reports whether any protected resource is still associated
+// with the given web ACL ARN. Deleting an associated web ACL is rejected with
+// WAFAssociatedItemException, matching real WAFv2.
+func (m *Mock) webACLAssociated(webACLARN string) bool {
+	m.assocMu.RLock()
+	defer m.assocMu.RUnlock()
+
+	for _, aclARN := range m.assoc {
+		if aclARN == webACLARN {
+			return true
+		}
+	}
+
+	return false
+}
+
+// itemReferencedByWebACL reports whether any web ACL's rules reference the given
+// resource ARN. IPSet/RuleGroup/RegexPatternSet reference statements embed the
+// referenced ARN verbatim in the rule JSON, so a substring scan of each web
+// ACL's stored rules detects an in-use set/group.
+func (m *Mock) itemReferencedByWebACL(arn string) bool {
+	needle := []byte(arn)
+
+	for _, wd := range m.webACLs.All() {
+		wd.mu.RLock()
+		referenced := bytes.Contains(wd.acl.Rules, needle)
+		wd.mu.RUnlock()
+
+		if referenced {
+			return true
+		}
+	}
+
+	return false
+}
 
 // webACLByARN finds a stored web ACL by its ARN.
 func (m *Mock) webACLByARN(arn string) (*webACLData, bool) {
--- a/providers/aws/wafv2/errors.go
+++ b/providers/aws/wafv2/errors.go
@@ -24,3 +24,9 @@ func staleLock(format string, args ...any) error {
 func invalidParameter(format string, args ...any) error {
 	return &driver.APIError{Exception: driver.ExInvalidParameter, Err: errors.Newf(errors.InvalidArgument, format, args...)}
 }
+
+// associated builds a WAFAssociatedItemException-tagged error, returned when a
+// resource is deleted while still associated with or referenced by another.
+func associated(format string, args ...any) error {
+	return &driver.APIError{Exception: driver.ExAssociatedItem, Err: errors.Newf(errors.FailedPrecondition, format, args...)}
+}
--- a/providers/aws/wafv2/ipset.go
+++ b/providers/aws/wafv2/ipset.go
@@ -133,6 +133,10 @@ func (m *Mock) DeleteIPSet(_ context.Context, ref driver.Ref, lockToken string) 
 		return staleLock("stale lock token for IP set %q", ref.ID)
 	}
 
+	if m.itemReferencedByWebACL(sd.set.ARN) {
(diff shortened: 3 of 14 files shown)


NitinKumar004 (commented):
Verified against pushed head `db6bbbf5` in an isolated worktree (real aws-sdk-go-v2 wafv2 client, not just the driver-level tests already in the diff).

**F1 — ARN shape: confirmed.** REGIONAL → `.../regional/webacl/...` + configured region; CLOUDFRONT → `.../global/webacl/...` + `us-east-1` (never the literal word `global` in the region slot). Verified for WebACL, IPSet, RuleGroup, RegexPatternSet via both the driver-level `TestARNShape` and SDK-level `TestSDKWebACLARNShape`, both green.

**F2 — delete guards: confirmed, including ordering.** `DeleteWebACL` rejects with `WAFAssociatedItemException` while associated, succeeds after `DisassociateWebACL`. `DeleteIPSet` rejects while referenced by a WebACL rule (substring scan on `Rules` JSON), succeeds once the referencing WebACL is deleted. Confirmed the stale-lock check precedes the association guard in all 4 Delete funcs (`ipset.go`, `regexset.go`, `rulegroup.go`, `webacl.go`) — a stale lock token still returns `WAFOptimisticLockException`, not masked by the new guard. `WAFAssociatedItemException` propagates correctly to the wire via the generic `errors.As(*APIError)` path in `handler.go` (no per-exception wiring needed) — confirmed with the real SDK's `*waftypes.WAFAssociatedItemException`.

**F3 — pagination: confirmed, including the untested bad-marker path.** All 4 List ops (`ListWebACLs`, `ListIPSets`, `ListRuleGroups`, `ListRegexPatternSets`) route through the shared `listOp`/`PaginateSorted` helper. `Limit=2` over 5 WebACLs pages correctly with `NextMarker` resume, no dupes (covered by the diff's own `TestSDKListWebACLsPagination`). Additionally verified end-to-end with the real SDK (not in the diff) that an invalid `NextMarker` (`"not-valid-base64!!!"`) returns `WAFInvalidParameterException` — `pagination.DecodeToken` error is correctly wrapped in `helpers.go`.

**Driver/coverage/compat: zero-diff confirmed.** `services/wafv2/driver/driver.go` is byte-identical; only `ExAssociatedItem` was added to `errors.go`. `go generate ./...` produces no diff, `compatgen` exits 0 with no diff under `docs/compat/`.

**Gates:** `go build ./...`, `go vet ./...` clean. Scoped tests (`server/aws/wafv2`, `providers/aws/wafv2`, `services/wafv2`) all green, including the new tests. `go test -race ./providers/aws/wafv2/...` clean. `gofmt -l` clean on changed files.

**Blocking: golangci-lint introduces 4 NEW `dupl` issues**, not pre-existing baseline. I diffed lint output at the merge-base (`c5ba180`) for these same packages — **0 issues** — versus the PR head, which reports 4:

```
providers/aws/wafv2/ipset.go:123: duplicate of regexset.go:120-140 (dupl)
providers/aws/wafv2/regexset.go:120: duplicate of rulegroup.go:126-146 (dupl)
providers/aws/wafv2/rulegroup.go:126: duplicate of webacl.go:143-163 (dupl)
providers/aws/wafv2/webacl.go:143: duplicate of ipset.go:123-143 (dupl)
```

The 4 `Delete*` functions were already structurally similar before this PR, but each was under the `dupl` threshold (100 tokens). Adding the same 4-line association-guard block to each pushed all 4 over the threshold simultaneously. Since the guard call itself differs only by which lookup function is invoked (`webACLAssociated` vs. `itemReferencedByWebACL`) and which field is passed, this is a good candidate for a small shared helper (e.g. a `deleteGuarded(getData, arn, associated-check)` wrapper, or extracting the lock-check + guard-check pair into one function reused by all 4 sites) rather than a `//nolint:dupl` suppression, given the existing codebase convention of avoiding blanket nolints without strong justification.

Everything else here is solid — recommend fixing the lint regression (dedupe or justified `//nolint`) before merge.

PR #699, opened 2026-08-26
fix(azure-blob): enforce conditional headers on all mutating ops + conditional reads; Copy Blob metadata override

NitinKumar004:
## Summary

Residue from a confirming re-audit of Azure Blob Storage conditional-header handling. The precondition evaluator existed but was wired into `putBlob` only, so optimistic-concurrency writes on a stale ETag silently succeeded on every other mutating op — a lost-update hazard.

### HIGH — silent lost-update: conditional WRITE preconditions enforced on all mutating ops
`If-Match`/`If-None-Match` are now evaluated on every mutating data-plane op, not just plain Put Blob: `CommitBlockList`, `SetMetadata`, `SetProperties`, `SetTier`, `AppendBlock`, `Snapshot`, create Append Blob, `Copy`, and `DELETE`. Lease checks are preserved (both lease AND conditional headers apply). A mismatched `If-Match` returns 412 `ConditionNotMet`; `If-None-Match:*` on an existing blob returns 409 `BlobAlreadyExists` for creates (Put Blob / create Append Blob) and 412 for other writes, per the Azure write-operations response-code table.

### MED — Copy Blob honors x-ms-meta-* override
Per the Azure Copy Blob spec: if the copy request supplies any `x-ms-meta-*`, the destination gets exactly that metadata (full replace, none inherited); with none supplied, the destination inherits the source's. Routed through the existing `ObjectCopier`/`CopyObjectRequest.ReplaceMetadata` capability (mirroring S3); the Azure provider now implements `CopyObjectV2`, sharing a single copy path with `CopyObject`.

### MED — conditional READS
`getBlob`/`headBlob` now evaluate read conditionals before writing a body: `If-None-Match` matching the current ETag (including `*`) returns 304 Not Modified; a mismatched `If-Match` returns 412.

## Tests
Real azblob reproductions: CommitBlockList/SetMetadata/SetProperties/SetTier/AppendBlock/DELETE with a mismatched `If-Match` → 412 (blob unchanged) and a matching `If-Match` → success; Copy with fresh metadata over a source → destination has only the fresh set; Copy with no metadata → destination inherits the source's; GET/HEAD `If-None-Match:<current etag>` → 304; GET `If-Match:<wrong>` → 412. Provider-level `CopyObjectV2` replace/inherit tests added. First-round tests (lease carry-over, snapshot immutability, SAS, fail-closed comp, -race) stay green.

Gates: build, vet, scoped tests, -race, gofmt, golangci-lint (0 new), go generate (zero diff), compatgen (zero diff) all green.

### DIFF
--- a/providers/azure/blobstorage/blobstorage.go
+++ b/providers/azure/blobstorage/blobstorage.go
@@ -569,26 +569,53 @@ func (m *Mock) ListObjects(_ context.Context, bucket string, opts driver.ListOpt
 	}, nil
 }
 
-// CopyObject copies a blob from one location to another.
+// CopyObject copies a blob from one location to another, inheriting the source
+// blob's metadata (the default Azure Copy Blob behavior when the request carries
+// no x-ms-meta-* headers).
 func (m *Mock) CopyObject(ctx context.Context, dstBucket, dstKey string, src driver.CopySource) error {
+	_, err := m.copyBlobInternal(ctx, dstBucket, dstKey, src, nil, false)
+
+	return err
+}
+
+// CopyObjectV2 implements the ObjectCopier capability so the Azure Blob wire
+// layer can express Copy Blob's metadata override: when ReplaceMetadata is set
+// the destination takes exactly req.Metadata instead of inheriting the source's.
+// Content properties are always inherited from the source (Azure Copy Blob does
+// not let the caller override them).
+func (m *Mock) CopyObjectV2(ctx context.Context, req *driver.CopyObjectRequest) (*driver.CopyObjectResult, error) {
+	info, err := m.copyBlobInternal(ctx, req.DstBucket, req.DstKey, req.Src, req.Metadata, req.ReplaceMetadata)
+	if err != nil {
+		return nil, err
+	}
+
+	return &driver.CopyObjectResult{ETag: info.ETag, LastModified: info.LastModified}, nil
+}
+
+// copyBlobInternal is the shared copy path for CopyObject and CopyObjectV2. When
+// replaceMeta is true the destination's metadata is exactly overrideMeta;
+// otherwise it inherits the source blob's metadata.
+func (m *Mock) copyBlobInternal(
+	ctx context.Context, dstBucket, dstKey string, src driver.CopySource, overrideMeta map[string]string, replaceMeta bool,
+) (*driver.ObjectInfo, error) {
 	srcCtr, ok := m.containers.Get(src.Bucket)
 	if !ok {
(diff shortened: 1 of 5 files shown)


NitinKumar004 (commented):
Adversarial review of the pushed head (d79a5a3) — clean-built from a fresh checkout, not the fixer's report.

**Verdict: approve / mergeable.** The lost-update fix is correct and complete; every gate is green.

**Gate matrix (GOMAXPROCS=3, -p=3):**
- `go build ./...`, `go vet` (changed pkgs): clean
- `go test` azure blobstorage (server+provider) + services/storage + aws/s3 (server+provider): PASS
- `go test -race` providers/azure/blobstorage AND server/azure/blobstorage: PASS
- gofmt: clean; `go generate ./...`: zero diff; `go run ./internal/compatgen`: exit 0, `docs/compat/` zero diff
- golangci-lint `--new-from-rev=origin/development` on both changed pkgs: 0 issues

**Independently verified (my own raw-wire + azblob tests, beyond the PR's suite):**
- If-None-Match:* on an existing blob correctly forks on isCreate: Put Blob AND create Append Blob → 409 BlobAlreadyExists (content unchanged); Set Metadata (non-create write) → 412 ConditionNotMet. This is the key branch the PR adds and it behaves per the write-operations response-code table.
- Snapshot honors the conditional write (mismatched If-Match → 412, no snapshot; matching → 201) — an op the PR's own suite did not cover.
- Lease AND conditional are both enforced: a correct lease id + a stale If-Match still → 412; correct lease + matching If-Match → 200.
- Conditional reads: GET/HEAD If-None-Match:<etag> and If-None-Match:* → 304; wrong If-Match → 412.
- Copy metadata: override (x-ms-meta-*) → destination has exactly the new set (source dropped); no metadata → inherits source. Verified at both provider (CopyObjectV2/CopyObject) and wire (StartCopyFromURL) levels.

**No regressions:** Azure's new CopyObjectV2 is a method on the Azure Mock type, independent of the S3 Mock that also implements ObjectCopier — S3 copy tests pass unchanged. Prior-round behaviors (lease carry-over, snapshot immutability, full-replace Put Blob) all green.

**Non-blocking notes:**
- copyBlob passes isCreate=false, so If-None-Match:* against an existing copy destination yields 412 rather than 409 — this matches real Azure Copy Blob (only Put Blob / create Append Blob return 409), so it is correct, not a defect.
- A specific-ETag If-None-Match on a write (as opposed to `*`) is handled by evalWriteConditions' third branch; the read path's If-Match+If-None-Match combined ordering (If-Match 412 first, then If-None-Match 304) is a reasonable interpretation. Neither affects the lost-update guarantee.

PR #700, opened 2026-08-26
fix(aws-rgta): ResourceTypeFilters from ARN; ElastiCache ARN; broaden TagResources coverage; pagination

NitinKumar004:
## What

Fixes bugs in the AWS Resource Groups Tagging API found by a real-SDK deep e2e audit. The cross-service aggregation was already correct; type-filtering, ElastiCache ARNs, tagging coverage, and pagination had gaps.

- **ResourceTypeFilters from ARN (F1):** `GetResources` derives the AWS service token from each resource's own ARN (segment 3 of `arn:aws:<service>:…`) rather than a 4-entry portable→AWS name map. Filters like `secretsmanager`, `sqs`, `sns`, `elasticache`, `rds`, `ecr` now select their resources instead of silently returning nothing. Both `service` and `service:resourceType` filter forms are supported. Falls back to the portable→AWS map only for rows whose identifier is not a real ARN.
- **ElastiCache / ECR ARN (F2):** `walkCache` emits `CacheInfo.ARN` (fallback endpoint) instead of the `host:port` endpoint string; `walkContainerRegistry` emits the ECR ARN (fallback registry URI). These walkers also feed Resource Explorer / Cloud Asset, which now report real ARNs too.
- **Broaden TagResources coverage (F3):** `TagResources`/`UntagResources` now round-trip on `elasticache`, `ecr`, `logs`, `rds`, and `iam` (role/user) via narrow tagger capability interfaces (the existing pattern). RDS routes through the discovery adapter to the RDS `Tagging` capability. Services with no tag store (e.g. CloudWatch alarms, Route 53 zones, IAM policies/groups) still return a `FailedResourcesMap` entry.
- **Pagination (F4):** `ResourcesPerPage` + `PaginationToken` are honored on `GetResources`, and `PaginationToken` on `GetTagKeys`/`GetTagValues` (opaque base64 offset tokens); a token is returned when more remain and resumed without duplicates.
- **Nits:** an empty `ResourceARNList` is rejected with a `ValidationException`; `FailedResourcesMap` error messages no longer leak internal `cerrors` phrasing.

Deferred (out of scope): compliance/report operations + `IncludeComplianceDetails` (F5).

## Why

`ResourceTypeFilters` worked for only 4 services, ElastiCache surfaced an endpoint where an ARN was expected (breaking any consumer that parses `ResourceARN`), most aggregated services could not be tagged through the same API that lists them, and the documented paging contract was ignored.

## Tests

Real `aws-sdk-go-v2/resourcegroupstaggingapi` reproductions: per-service `ResourceTypeFilters`, ElastiCache ARN shape, `TagResources` round-trip on rds/elasticache/ecr (visible on each service's own read + a subsequent `GetResources`), genuinely-unsupported service still failing, empty-list rejection, and `ResourcesPerPage=2` paging with no duplicates. Existing RGTA / resourcediscovery / Resource Explorer / Cloud Asset / Resource Graph tests stay green (`-race` on resourcediscovery too).

### DIFF
--- a/providers/aws/rds_discovery.go
+++ b/providers/aws/rds_discovery.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"context"
 
+	cerrors "github.com/stackshy/cloudemu/v2/errors"
 	rdsdriver "github.com/stackshy/cloudemu/v2/services/relationaldb/driver"
 	"github.com/stackshy/cloudemu/v2/services/resourcediscovery"
 )
@@ -20,6 +21,27 @@ type redshiftClusters interface {
 type rdsDiscovery struct {
 	m        rdsdriver.RelationalDB
 	redshift redshiftClusters
+}
+
+// AddTagsToResource lets the Resource Groups Tagging API apply tags to an RDS
+// resource by its ARN, delegating to the RDS mock's optional Tagging capability.
+func (a rdsDiscovery) AddTagsToResource(ctx context.Context, arn string, tags map[string]string) error {
+	tg, ok := a.m.(rdsdriver.Tagging)
+	if !ok {
+		return cerrors.New(cerrors.FailedPrecondition, "relational database driver does not support tagging")
+	}
+
+	return tg.AddTagsToResource(ctx, arn, tags)
+}
+
+// RemoveTagsFromResource removes tags from an RDS resource by its ARN.
+func (a rdsDiscovery) RemoveTagsFromResource(ctx context.Context, arn string, keys []string) error {
+	tg, ok := a.m.(rdsdriver.Tagging)
+	if !ok {
+		return cerrors.New(cerrors.FailedPrecondition, "relational database driver does not support tagging")
+	}
+
+	return tg.RemoveTagsFromResource(ctx, arn, keys)
 }
 
 func (a rdsDiscovery) DiscoverDatabases(ctx context.Context) ([]resourcediscovery.DiscoveredDatabase, error) {
--- a/server/aws/resourcegroupstaggingapi/handler.go
+++ b/server/aws/resourcegroupstaggingapi/handler.go
@@ -28,6 +28,10 @@ const (
 // resource-not-found.
 const errInvalidParameter = "InvalidParameterException"
 
+// tagPageSize is the internal page size for GetTagKeys/GetTagValues, which
+// (unlike GetResources) carry no client-supplied page-size parameter.
+const tagPageSize = 100
+
 // Handler serves Resource Groups Tagging API JSON-RPC requests.
 type Handler struct {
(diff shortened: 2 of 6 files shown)


NitinKumar004 (commented):
**Verdict: MERGE** — verified against the pushed head (`c58e4bc5`), clean checkout.

Gate matrix on the pushed tree: `go build ./...` clean, `go vet` clean, `-race` on `services/resourcediscovery` clean, `go generate ./...` zero-diff, `go run ./internal/compatgen` exit 0 + `docs/compat/` zero-diff, `golangci-lint --new-from-rev` on the changed packages reports 0 new issues (the table-dispatch refactor keeps `matchesTypeFilter`/`parseSupportedARN`/`dispatchTag` under the complexity gate).

Verified adversarially with independent real `aws-sdk-go-v2/resourcegroupstaggingapi` tests beyond the ones in the PR:

- **F1 type-filter discrimination** — `secretsmanager`/`sqs`/`elasticache`/`rds`/`ecr` each select exactly their own resource; `ec2:instance` vs `ec2:vpc` discriminate on resourceType under the same service token; plain `ec2` returns both; a non-matching filter (`kinesis`, `ec2:natgateway`) returns empty rather than everything. Service and `service:type` forms both work.
- **F2 ARNs** — ElastiCache `ResourceARN` is `arn:aws:elasticache:…:cluster:…` (not `host:port`); ECR ARN correct. The shared walker change is ARN-with-fallback (`firstNonEmpty(Arn/ARN, URI/Endpoint)`), so Azure/GCP registries/caches that model no ARN are unchanged — Resource Explorer, Cloud Asset, and Resource Graph tests stay green.
- **F3 tagging breadth** — round-trip confirmed on elasticache/ecr/logs/rds/iam-role/iam-user: tag via RGTA is visible on each service's native read (`ListLogGroupTags`, `ListRoleTags`/`ListUserTags`, `ListTagsForResource`, etc.) and on a subsequent `GetResources`; untag propagates back. Genuinely-unsupported ARNs (elbv2, route53, eks, iam-policy, iam-group) land in `FailedResourcesMap` with no crash, and the message no longer leaks internal `cerrors` phrasing.
- **F4 pagination** — page sizes 1 and 2 cover every resource exactly once with no dupes/omissions and no empty trailing page at the size boundary; `GetTagKeys`/`GetTagValues` paginate.

Two non-blocking notes (not gating):
1. A malformed `PaginationToken` silently resets to offset 0 rather than returning an error; real RGTA rejects an invalid token. The "never wedge pagination" choice is defensible, just a semantic divergence.
2. The tokens are plain slice offsets, so resuming assumes a stable walker ordering between calls (fine for the deterministic in-memory walker; not cursor-stable if the set mutates mid-iteration).

PR #701, opened 2026-08-26
fix(aws-memorydb): TLSEnabled/NumReplicasPerShard default via pointers; typed *NotFoundFault; node port

NitinKumar004:
## What

Fixes four AWS MemoryDB create-time fidelity bugs surfaced by a real aws-sdk-go-v2 deep e2e audit.

- **TLSEnabled default** — `CreateClusterConfig.TLSEnabled` is now `*bool`. Omitted (nil) defaults to `true` (matching AWS); an explicit `false` is preserved. Previously it defaulted to `false`, causing perpetual Terraform replacement (ForceNew) and TLS-in-transit silently disabled.
- **NumReplicasPerShard default** — now `*int32`. Omitted (nil) defaults to `1` (2 nodes/shard + MultiAZ, per AWS API ref); an explicit `0` is preserved (1 node/shard + SingleAZ). Previously defaulted to `0`, understating HA topology and `AvailabilityMode`.
- **Typed *NotFoundFault** — a dangling `ACLName` / `SubnetGroupName` / `ParameterGroupName` now returns the SDK-modeled `ACLNotFoundFault` / `SubnetGroupNotFoundFault` / `ParameterGroupNotFoundFault` (via a kind-tagged `driver.NotFoundError` mapped in `writeErr`), instead of a generic `InvalidParameterValueException` that typed `errors.As` matches never caught.
- **Node endpoint port** — per-node `Endpoint.Port` now reflects the cluster's configured `Port` (threaded into `buildShards`), matching the cluster endpoint; previously hardcoded to the default 6379.

## Why

All four are create-time default / error-fidelity gaps confirmed against the official AWS MemoryDB `CreateCluster` API reference (NumReplicasPerShard "default value is 1"; ACL/SubnetGroup/ParameterGroup NotFoundFault are documented errors). The pointer distinction is required so nil-vs-explicit-zero is not lost across the handler->driver boundary. The MemoryDB provider is the sole implementer of `CreateClusterConfig`, so the signature change is contained.

## Tests

Real aws-sdk-go-v2 reproductions plus provider-level coverage:
- Omitted TLSEnabled/NumReplicasPerShard -> TLS true, 2 nodes/shard, MultiAZ, custom Port on node + cluster endpoints.
- Explicit TLSEnabled=false / NumReplicasPerShard=0 preserved -> 1 node/shard, SingleAZ.
- Bad ACLName/SubnetGroupName/ParameterGroupName -> typed `*NotFoundFault` via `errors.As`.
- Existing cluster/ACL/user/snapshot/subnet-group/parameter-group/pagination/-race tests remain green.

Gates: build, vet, scoped tests, -race, golangci-lint (0 issues), gofmt, `go generate` (no coverage diff), compatgen (no compat diff) all green.


NitinKumar004 (commented):
Fast-lane review of `e30d1af` (PR head). Verified all four claims independently against the real `aws-sdk-go-v2` MemoryDB client (temp tests, not committed — removed after review), plus read the full diff.

**F1 — TLSEnabled `*bool`**: confirmed. `tlsOrDefault(nil)=true`, explicit `false` preserved. `applySnapshotShape` inherits from the snapshot only when `cfg.TLSEnabled == nil`. Verified via SDK: omit → `Describe` shows `TLSEnabled=true`; explicit `false` → stays `false`.

**F2 — NumReplicasPerShard `*int32`**: confirmed. `replicasOrDefault(nil)=1` → 2 nodes/shard + `AvailabilityMode=MultiAZ` (via `availabilityMode(replicas)`); explicit `0` → 1 node/shard + `SingleAZ`. Restore path (`applySnapshotShape`) inherits only on `nil`. Verified via SDK for both branches, multi-shard.

**F3 — typed NotFoundError → `<Kind>NotFoundFault`**: confirmed. `validateClusterRefs` returns `*mdbdriver.NotFoundError{Kind, Name}` per ref kind; `writeErr` checks `errors.As` for it *before* the generic `cerrors.IsNotFound` switch, so it can't be shadowed by the generic branch. Verified via SDK + `errors.As` against the concrete `mdbtypes.ACLNotFoundFault` / `SubnetGroupNotFoundFault` / `ParameterGroupNotFoundFault` types (not just string/HTTP-code matching). The rewritten `TestSDKErrorFaults` genuinely corrects prior (buggy) behavior — old test asserted `InvalidParameterValueException` for a dangling ACL ref, new test asserts the typed fault. Not a weakening.

**F4 — node port threading**: confirmed. `buildShards` takes `port` and applies `portOr(port)` per node; both `reserveCluster` and `reconfigureShards` (shard-count and replica-count branches) now pass the cluster's port through. Verified via SDK across 3 shards x replicas that every node's `Endpoint.Port` matches a custom `ClusterEndpoint.Port`.

**Blast radius**: only `server/aws/memorydb/clusters.go` constructs `mdbdriver.CreateClusterConfig` (grepped repo-wide; the other `CreateClusterConfig{` hits are unrelated Azure/GCP driver packages of the same name). Wire handler maps `in.TLSEnabled`/`in.NumReplicasPerShard` (SDK's own `*bool`/`*int32`) straight through — no lossy `aws.ToBool`/`aws.ToInt32` reintroduced. `go build ./...` clean, confirming no other caller broke.

**Minor (non-blocking) observation**: `reserveCluster` calls `portOr(cfg.Port)` when passing to `buildShards`, and `buildShards` itself also calls `portOr(port)` internally — redundant double-default, harmless (idempotent) but could just pass `cfg.Port` raw and let `buildShards` be the single place that defaults it. Not blocking.

**Gates (pushed SHA, GOMAXPROCS=3/-p=3)**: `go build ./...` clean, `go vet ./...` clean, `go test ./server/aws/memorydb/... ./providers/aws/memorydb/... ./services/memorydb/...` PASS, `go test -race ./providers/aws/memorydb/...` PASS, `golangci-lint run` on the 3 changed packages: 0 issues, `gofmt -l`/`-d` on changed files: no diff, `go generate ./...`: zero diff, `go run ./internal/compatgen` exit 0, `git diff docs/compat/` zero diff.

Verdict: MERGE. No blocking issues found.

PR #702, opened 2026-08-26
fix(azure-cosmos): reject partition-key mutation on Replace (no orphan) + 404 on replace-missing

NitinKumar004:
Confirming re-audit residue in the Cosmos DB data-plane replace path (server/azure/cosmosdb/handler.go). Two localized data-integrity/correctness fixes:

- **[MED — data integrity] Partition-key change on ReplaceItem silently orphaned the document.** A PUT /docs replace whose body carried a different partition-key value than the request's `x-ms-documentdb-partitionkey` header was PutItem'd under the NEW key without removing the original, leaving the document duplicated under both partitions. The partition key is immutable in real Cosmos, which rejects the mutation with 400 BadRequest. The PUT branch now compares the body's partition-key value against the addressed key before writing and returns 400 on mismatch (no orphan written).
- **[LOW] ReplaceItem on a non-existent id created it (200) instead of 404.** `replaceDocument`'s PutItem was an unconditional upsert. It now gates the write behind an existence check under the write lock already held (no new race), returning NotFound so the wire maps to 404.

Documented-as-designed divergences are unchanged (DeleteItem-on-missing 204, PATCH 405, ReplaceItem ignores IfMatchEtag). Round-1 fixes (isolation, custom-pk routing, throughput, indexing/uniqueKeys/TTL, multi-region) are not touched.

Tests: real azcosmos SDK reproduction (TestCosmosReplaceSemantics) asserting the 400 + no-orphan point read + unchanged original + same-pk replace still succeeds + replace-missing 404. All Cosmos gates green (build, vet, test, -race, lint 0, gofmt, go generate zero diff, compatgen zero diff).

### DIFF
--- a/server/azure/cosmosdb/cosmos_lifecycle_test.go
+++ b/server/azure/cosmosdb/cosmos_lifecycle_test.go
@@ -428,6 +428,54 @@ func TestCosmosWriteDivergences(t *testing.T) {
 	}
 }
 
+// TestCosmosReplaceSemantics pins the two replace-path invariants real Cosmos
+// enforces: the partition key is immutable (a replace that would move the
+// document to a new partition is rejected 400 and leaves no orphan), and a
+// replace of a non-existent id is a 404, not a silent create.
+func TestCosmosReplaceSemantics(t *testing.T) {
+	ctx := context.Background()
+	env := newCosmosEnv(t)
+	cc := env.container(ctx, t, "repldb", "items")
+
+	createDoc(ctx, t, cc, "books", map[string]any{"id": "i1", "pk": "books", "title": "Dune"})
+
+	// Replacing the document with a body that changes the partition key value
+	// is rejected 400 — the partition key is immutable.
+	moved, _ := json.Marshal(map[string]any{"id": "i1", "pk": "movies", "title": "Dune"})
+
+	_, err := cc.ReplaceItem(ctx, azcosmos.NewPartitionKeyString("books"), "i1", moved, nil)
+	wantRespErr(t, err, 400, "ReplaceItem with mutated partition key")
+
+	// No orphan landed under the attempted new partition.
+	_, err = cc.ReadItem(ctx, azcosmos.NewPartitionKeyString("movies"), "i1", nil)
+	wantRespErr(t, err, 404, "point read under new partition (no orphan)")
+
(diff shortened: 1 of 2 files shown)


NitinKumar004 (commented):
Verified against the pushed head (391938f), clean-built and re-tested independently with the real azcosmos SDK.

Gate matrix (GOMAXPROCS=3, -p=3): build, vet, test (cosmosdb / cosmosaccount / providers cosmos / services database) all green; `-race` on the two cosmos packages clean; golangci-lint on server/azure/cosmosdb 0 issues; gofmt clean; `go generate ./...` and `go run ./internal/compatgen` both zero-diff.

Adversarial reproduction (my own tests, on top of TestCosmosReplaceSemantics):
- pk-mutation (/pk container, body books->movies) -> 400 BadRequest; point-read under the new partition -> 404 (no orphan); original under books UNCHANGED. Confirmed.
- same-partition replace (other fields change) -> succeeds, not wrongly 400. Confirmed.
- container partitioned by /id: replace keeping id -> succeeds (pk pinned by URL id, not flagged). Confirmed.
- replace-missing id on empty/existing container -> 404, not a silent create; existing-doc replace still succeeds. Confirmed.

Concurrency/blast radius: the new existence check in replaceDocument runs under the already-held per-container writeMu and reads via the provider's GetItem (separate m.mu RLock), so it does not re-acquire writeMu -> no deadlock, and -race is clean. partitionKeyMutated is a pure body-vs-header compare with no store access, so no race with the subsequent locked write. Documented divergences intact (DeleteItem-missing 204, PATCH 405, IfMatch ignored) - TestCosmosLifecycle and TestCosmosWriteDivergences unchanged and green.

One non-blocking note on the omitted-partition-key edge: a replace whose body omits the partition-key field now returns 404 (existence probe looks the doc up under a nil pk value) with the original left untouched - a safe outcome and an improvement over the prior orphan-under-nil-key write. But the partitionKeyMutated docstring still states such a body "is left to the normal write path (it stores under the same identity)"; that sentence no longer describes the actual behavior after the replaceDocument existence gate. Real Cosmos returns 400 here rather than 404, so the emulator remains divergent on this obscure path either way. Comment-only; recommend tightening the docstring on a future touch.

Verdict: MERGE.

PR #703, opened 2026-08-26
fix(gcp-firestore): implement field transforms (serverTimestamp/increment/arrayUnion/arrayRemove) — stop transform-only writes wiping the document; != absent-field; updateTime/delete preconditions

NitinKumar004:
## Summary

A confirming re-audit surfaced a CRITICAL data-loss residue in the GCP Firestore wire handler: **field transforms were entirely unimplemented**, and a transform-only write **destroyed the stored document**.

The Go SDK sends transform fields (serverTimestamp / increment / arrayUnion / arrayRemove) in `Write.updateTransforms` with empty `update.fields` and an empty `updateMask`. The handler had no `UpdateTransforms` field on `writeOp` and never read/applied them, so such a write fell into the wholesale-REPLACE path and wiped the document to `{}`.

Repro (now fixed): `Set({count:10, keep:"me"})` then `Update([{count, Increment(5)}])` → `Get` previously returned `{}` (count not incremented AND keep wiped).

## Changes

- **Field transforms implemented** (`transform.go`, new): `setToServerValue=REQUEST_TIME` (commit timestamp), `increment` (numeric add, int/double typing preserved, absent field starts at 0), `appendMissingElements` (arrayUnion), `removeAllFromArray` (arrayRemove), `maximum`/`minimum`. Applied against the existing stored value, with results echoed in `writeResults[].transformResults`.
- **Transform-carrying writes MERGE, never replace** — a transform-only write layers onto the stored document and preserves untouched fields. Wired through single `:commit`, `WriteBatch`, and `RunTransaction`.
- **`setToServerValue` enum decoding** — accepts the SDK's protobuf number (REQUEST_TIME=1) and the string name.
- **`!=` (NOT_EQUAL)** now excludes absent/null-field docs — built as `AND(present-and-not-null, <> value)`, matching the existing `NOT_IN` shape.
- **Preconditions** — `currentDocument.updateTime` optimistic-concurrency support added (mismatch → FAILED_PRECONDITION), and Delete now honors its `currentDocument` precondition (guarded exists=true on a missing doc → NOT_FOUND).

Deferred (noted, out of scope for this PR): `showMissing`, `IS_NAN`/`IS_NOT_NAN`.

## Tests

Real `cloud.google.com/go/firestore` SDK reproductions: no-data-loss increment, int-vs-double typing + absent-field-from-zero, ServerTimestamp, ArrayUnion/ArrayRemove, transform inside WriteBatch and RunTransaction, `!=` absent-field exclusion, stale/matching updateTime precondition, Delete-guarded-by-exists on missing. The #639/#655 tests (value types, dotted updateMask, cursors, operators, paging) stay green.

### DIFF
--- a/server/gcp/firestore/firestore_transform_test.go
+++ b/server/gcp/firestore/firestore_transform_test.go
@@ -1,0 +1,329 @@
+// These tests drive the REAL cloud.google.com/go/firestore SDK against the
+// emulator's GCP HTTP server, asserting field-transform write semantics
+// (serverTimestamp / increment / arrayUnion / arrayRemove), the "!=" absent-
+// field exclusion, and currentDocument.updateTime / delete preconditions.
+//
+// The central regression they guard: a transform-only write (e.g. Update with
+// only firestore.Increment) must MERGE onto the stored document, never replace
+// it with an empty doc.
+package firestore_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	gcpfirestore "cloud.google.com/go/firestore"
+	"google.golang.org/grpc/codes"
+)
+
+// TestDatabaseIncrementNoDataLoss is the CRITICAL data-loss regression: a
+// transform-only Update must increment the target and preserve every other
+// field, not wipe the document.
+func TestDatabaseIncrementNoDataLoss(t *testing.T) {
+	ctx, client, _ := newDBClient(t, "counters")
+
+	doc := client.Collection("counters").Doc("c1")
+
+	if _, err := doc.Set(ctx, map[string]any{"count": 10, "keep": "me"}); err != nil {
+		t.Fatalf("Set: %v", err)
+	}
+
+	if _, err := doc.Update(ctx, []gcpfirestore.Update{
+		{Path: "count", Value: gcpfirestore.Increment(5)},
+	}); err != nil {
+		t.Fatalf("Update Increment: %v", err)
+	}
+
+	snap, err := doc.Get(ctx)
+	if err != nil {
+		t.Fatalf("Get: %v", err)
+	}
+
+	got := snap.Data()
+	if got["count"] != int64(15) {
+		t.Errorf("count=%v (%T), want int64(15)", got["count"], got["count"])
+	}
+
+	if got["keep"] != "me" {
+		t.Errorf("keep=%v, want %q (transform-only write must not wipe other fields)", got["keep"], "me")
+	}
+}
+
+// TestDatabaseIncrementTyping pins numeric typing: int increments stay int64,
+// double increments stay float64, and an increment on an absent field starts
+// from zero.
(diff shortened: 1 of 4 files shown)


NitinKumar004 (commented):
**Verdict: APPROVE** — the data-loss root cause is correctly fixed; verified adversarially against the real `cloud.google.com/go/firestore` SDK.

**Gates (pushed SHA de66660e):** build ✓, vet ✓, `go test` server+provider+services/database ✓, `-race` on both `providers/gcp/firestore` and `server/gcp/firestore` ✓, golangci-lint new-only ✓ 0 issues, gofmt ✓, `go generate` zero-diff ✓, compatgen green + `docs/compat/` zero-diff ✓.

**No data loss (the critical fix):** confirmed. `Set({count:10,keep:"me"})` → `Update(count, Increment(5))` yields `{count:15, keep:"me"}` — sibling not wiped. The design is robust to whether the SDK sends the transform-only write with a present-but-empty `updateMask` (→ `mergeUpdateBase` case 1) or no mask at all (→ case 2); both merge onto the stored doc. Plain `Set` still wholesale-replaces (old fields dropped), and a masked `Update` still preserves unmasked siblings — the `mergeUpdateBase` refactor did not regress either.

**Transform semantics verified against Firestore's documented behavior:**
- increment: int+int→int64, double→float64, int+double→float64 (promotion), absent field starts at 0, non-numeric field resets to the operand.
- arrayUnion dedups (including on an absent/non-array field, which creates the array); arrayRemove strips all matches.
- serverTimestamp resolves to the commit time; nested dotted transform (`m.cnt`) applies and preserves the sibling.
- transforms apply and merge correctly inside both `WriteBatch` and `RunTransaction`.
- `!=` excludes absent AND null-valued fields, includes present-but-different, excludes the matching value — mirrors the proven NOT_IN shape.
- updateTime precondition: stale→FAILED_PRECONDITION, exact match→succeeds; `Delete(Exists)` on a missing doc→NOT_FOUND.

**Non-blocking notes (no fix required to merge):**
1. A `Set` carrying ONLY a `ServerTimestamp` sentinel (empty `update.fields`, no mask) hits `mergeUpdateBase` case 2 and MERGES onto the stored doc rather than replacing it — real Firestore would replace, leaving only the stamped field. This is a rare shape and it errs toward data *preservation*, the opposite of the bug being fixed. Worth a follow-up but not a blocker.
2. `maximum`/`minimum` transforms are implemented on the wire but are not reachable through the Go SDK's public API (Increment/ArrayUnion/ArrayRemove/ServerTimestamp only), so they're covered by inspection, not an SDK e2e.
3. Batch/transaction commits remain non-atomic (a precondition failure mid-batch leaves earlier writes persisted) — pre-existing, documented in `beginTransaction`, out of scope here.

PR #704, opened 2026-08-26
fix(gcp-pubsub): Topic/Subscription PATCH update; evaluate subscription filter; enforce deadLetterPolicy; detach subs on topic delete

NitinKumar004:
Confirming re-audit residue on GCP Pub/Sub — 2 HIGH + 2 correctness gaps, all reconciled against the Pub/Sub v1 REST docs (projects.subscriptions.patch, projects.topics.patch, Filtering syntax, DeadLetterPolicy).

## Fixes

- **[HIGH] Topic/Subscription PATCH (update) returned 405.** `serveTopic`/`serveSubscription` only accepted PUT/GET/DELETE, so any in-place update — Terraform `google_pubsub_topic`/`google_pubsub_subscription` apply, or the Go SDK `Subscription.Update`/`Topic.Update` — failed. Added `http.MethodPatch` handlers that decode `UpdateTopicRequest`/`UpdateSubscriptionRequest`, honor `updateMask` field paths, and merge only the masked fields. Supported subscription masks: `ackDeadlineSeconds`, `retryPolicy`, `deadLetterPolicy`, `pushConfig`, `labels`, `messageRetentionDuration`, `expirationPolicy`, `retainAckedMessages`, `enableMessageOrdering`, `detached`. `filter`/`topic`/`name` are immutable and rejected. Topic patch supports `labels`.

- **[HIGH] Subscription filter stored but never evaluated.** `deliver()` ignored `sub.cfg.Filter`, so non-matching messages were delivered. Added a Pub/Sub filter evaluator (`attributes.key = "v"`, `!=`, `attributes:key`, `hasPrefix(...)`, `NOT`/`AND`/`OR`, parentheses). The filter is immutable, so it is parsed once at create and evaluated on delivery; a filtered-out message is auto-acked (not delivered, not backlogged), matching real Pub/Sub.

- **[MED] deadLetterPolicy not enforced.** Delivery counted attempts and surfaced `deliveryAttempt` but never routed. Now once attempts exceed `maxDeliveryAttempts` (default 5), the message is forwarded to the dead-letter topic's log and auto-acked on the source, so a subscription on the DLQ topic receives it and the source stops redelivering.

- **[LOW] Topic delete didn't detach subscriptions.** `deleteTopic` now sets `subscription.topic = "_deleted-topic_"` for attached subs and keeps their message log so already-published messages can still drain.

## Notes
- Server-local only; no driver/interface change. `go generate` and `internal/compatgen` produce zero diff.
- Round #638 behavior preserved (fan-out, ordering, ackDeadline/modifyAckDeadline, snapshots/seek, extended-field round-trip, IAM, pagination) — existing tests stay green, including `-race`.

## Tests
Real `google.golang.org/api/pubsub/v1` reproductions: subscription/topic PATCH reflected on Get; mask scoping; immutable-filter rejection; `attributes.color="red"` delivers only red; `hasPrefix(...) AND NOT attributes:skip`; DLQ routing after nacks; topic-delete detach.

### DIFF
--- a/server/gcp/pubsub/filter.go
+++ b/server/gcp/pubsub/filter.go
@@ -1,0 +1,476 @@
+package pubsub
+
+import (
+	"strconv"
+	"strings"
+
+	cerrors "github.com/stackshy/cloudemu/v2/errors"
+)
+
+// filterExpr is a parsed Pub/Sub subscription filter, evaluated against a
+// message's attributes. See
+// https://cloud.google.com/pubsub/docs/subscription-message-filter for the
+// grammar: attributes.KEY = "V" | attributes.KEY != "V" | attributes:KEY |
+// hasPrefix(attributes.KEY, "P"), combined with NOT / AND / OR and parentheses.
+type filterExpr interface {
+	eval(attrs map[string]string) bool
+}
+
+// filterAll matches every message (empty filter).
+type filterAll struct{}
+
+func (filterAll) eval(map[string]string) bool { return true }
+
+// hasAttr is `attributes:KEY` — true when the attribute is present.
+type hasAttr struct{ key string }
+
+func (h hasAttr) eval(a map[string]string) bool {
+	_, ok := a[h.key]
+	return ok
+}
+
+// equals is `attributes.KEY = "V"` (or != when negate) — exact value match.
+type equals struct {
+	key, val string
+	negate   bool
+}
+
+func (e equals) eval(a map[string]string) bool {
+	v, ok := a[e.key]
+	match := ok && v == e.val
+
+	if e.negate {
+		return !match
+	}
+
+	return match
+}
+
+// prefix is `hasPrefix(attributes.KEY, "P")`.
+type prefix struct{ key, pfx string }
+
+func (p prefix) eval(a map[string]string) bool {
+	v, ok := a[p.key]
+	return ok && strings.HasPrefix(v, p.pfx)
+}
+
+type notExpr struct{ x filterExpr }
+
+func (n notExpr) eval(a map[string]string) bool { return !n.x.eval(a) }
+
+type andExpr struct{ l, r filterExpr }
+
+func (n andExpr) eval(a map[string]string) bool { return n.l.eval(a) && n.r.eval(a) }
+
+type orExpr struct{ l, r filterExpr }
+
+func (n orExpr) eval(a map[string]string) bool { return n.l.eval(a) || n.r.eval(a) }
+
+func errFilter() error {
+	return cerrors.New(cerrors.InvalidArgument, "invalid filter expression")
+}
+
+// parseFilter compiles a Pub/Sub filter string. An empty filter matches all.
+func parseFilter(s string) (filterExpr, error) {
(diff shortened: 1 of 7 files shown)


NitinKumar004 (commented):
Adversarial review — verdict: **MERGE**. Reviewed the actual pushed tree at bdb328e (fetched pull/704/head, detached checkout, clean build — not the fixer's report).

**Gate matrix (GOMAXPROCS=3, -p=3):** build ✓ · vet ✓ · `go test ./server/gcp/pubsub ./providers/gcp/pubsub ./services/messagequeue` ✓ · `-race` on both pubsub packages ✓ · gofmt clean ✓ · `go generate ./...` zero-diff ✓ · compatgen exit 0 + `docs/compat/` zero-diff ✓ · golangci-lint on server/gcp/pubsub: 0 issues.

**Independent verification** (my own `google.golang.org/api/pubsub/v1` tests against a live httptest server — the correct REST client for this HTTP handler; `cloud.google.com/go/pubsub` is gRPC and would not exercise this handler):

- **PATCH / field-mask merge** — `Subscriptions.Patch(ackDeadlineSeconds=30, mask "ackDeadlineSeconds")` → Get shows 30, other fields untouched; a mask naming only `labels` leaves ackDeadline unchanged. `Topics.Patch(labels)` → labels reflected on both Get and List. camelCase masks match the lowerCamelCase JSON FieldMask representation the real GCP REST API accepts.
- **Immutability** — patching `filter`/`topic`/`name` in the mask → 400 INVALID_ARGUMENT.
- **Filter (correctness core)** — `attributes.color="red"`: only red delivered; blue is auto-acked, never delivered, never backlogged, and does not redeliver on a second pull. Grammar exercised: `!=` (incl. missing-attribute), `attributes:key` present/absent, `hasPrefix`, `NOT`, `AND`, `OR`, parenthesized precedence, quoted dotted key. No filter → all delivered.
- **Malformed filter → error at createSubscription, not a panic at deliver** — missing value, `==`, missing hasPrefix arg, dangling AND, unbalanced parens, unknown identifier, unterminated string all return 400 at create (parse-at-create is the safe design).
- **deadLetterPolicy** — maxDeliveryAttempts=5: delivered exactly 5× with deliveryAttempt = 1..5, the 6th pull dead-letters (source stops, no infinite redelivery) and the DLQ topic's subscription receives exactly 1. Without a DLQ policy: normal redelivery continues and deliveryAttempt stays 0.
- **Topic-delete detach** — deleted topic's subscription reports topic `_deleted-topic_`, and its already-published backlog is still drainable.

**Blast radius / #638** — fan-out, ordering, snapshots/seek, ackDeadline/modifyAckDeadline, IAM, pagination, extended-field round-trip all still green; filter eval is read-only and race-clean under concurrent delivery.

Non-blocking notes (no fix required): a lone `-` token is accepted as a NOT alias (not standard Pub/Sub grammar, but harmless — hyphens inside keys are consumed as ident parts); unsupported mask paths are silently ignored rather than rejected, which is the documented, over-permissive-but-safe choice.

PR #705, opened 2026-08-26
fix(azure-servicebus): SqlFilter OR (stop message loss) + relational/custom-property eval; subscription counts; maxDeliveryCount on update

NitinKumar004:
Confirming re-audit residue of Azure Service Bus subscription filters/counts. Fixes a HIGH message-loss bug plus routing, runtime-count and update-propagation gaps. First-round fixes (#602/#621/#629: DLQ wiring, system props, TTL-from-enqueue, scheduled, LockDuration peek-lock, basic SqlFilter/CorrelationFilter, namespace isolation) are preserved.

## Fixes

- **[HIGH — message loss] SqlFilter using OR silently dropped matching messages.** `sys.Label = 'a' OR sys.Label = 'b'` was parsed as a single malformed equality that always evaluated false, so a matching message got 204 No Content and was lost. The evaluator is now OR-of-AND-of-comparison-clauses, and the no-drop invariant is hardened: any clause with an unrecognized operator / unparseable syntax / unknown `sys.*` identifier falls back to match=true rather than dropping.
- **[MED — wrong routing] Relational and custom-property SqlFilters over-delivered.** Added the `>, <, >=, <=, !=, <>` operators and parsing of custom application properties (individual HTTP headers per the Service Bus REST protocol, reserved/`null` headers excluded) into the evaluated property set. `priority > 5` and `myprop = 'x'` now route only matching messages. (LIKE/IN/EXISTS/arithmetic remain deferred to the no-drop path.)
- **[LOW] CorrelationFilter user-defined `Properties` were ignored.** They are now matched against the message's custom application properties (case-insensitive key lookup, case-sensitive value compare).
- **[MED] Subscription CountDetails / messageCount always 0.** `toSubResource` now queries the backing store (active) and the paired `$DeadLetterQueue` (dead-letter), mirroring `toQueueResource`. ARM GET subscription reports `messageCount` / `activeMessageCount`.
- **[LOW-MED] PUT update didn't propagate maxDeliveryCount / deadLetteringOnMessageExpiration.** The create-or-update branch now propagates these to the backing store (`applyQueueAttributes` extended), so lowering `maxDeliveryCount` dead-letters at the new threshold.

Deferred (noted, out of scope for this PR): SQL `LIKE`/`IN`/`EXISTS`/`IS NULL`/arithmetic/parentheses; `forwardTo` / `forwardDeadLetteredMessagesTo` enforcement; AMQP-only session receive.

## Tests (real armservicebus SDK + raw-HTTP data plane)

OR SqlFilter delivers matches (not lost); numeric `priority > 5` routes only matching; custom-property `myprop='x'` routes only matching; CorrelationFilter `Properties{region:eu}` routes only matching; ARM GET subscription after 2 publishes reports counts >= 2; lowering maxDeliveryCount via PUT dead-letters at the new threshold. First-round tests kept green (DLQ, LockDuration peek-lock, scheduled/TTL, filter basics, namespace isolation), including -race.

### DIFF
--- a/providers/azure/servicebus/servicebus.go
+++ b/providers/azure/servicebus/servicebus.go
@@ -763,6 +763,19 @@ func applyQueueAttributes(qd *queueData, attrs map[string]int) {
 	if v, ok := attrs["MessageRetentionPeriod"]; ok {
 		qd.messageRetention = v
 	}
+
+	// MaxDeliveryCount reconfigures the dead-letter threshold (Service Bus
+	// maxDeliveryCount): lowering it dead-letters already-enqueued messages at the
+	// new threshold on their next delivery attempt.
+	if v, ok := attrs["MaxDeliveryCount"]; ok && qd.dlqConfig != nil {
+		qd.dlqConfig.MaxReceiveCount = v
+	}
+
+	// DeadLetterOnExpiration toggles Service Bus'
+	// deadLetteringOnMessageExpiration (encoded 0/1 over the int-only map).
+	if v, ok := attrs["DeadLetterOnExpiration"]; ok {
+		qd.deadLetterOnExpiration = v != 0
+	}
 }
 
 // PurgeQueue removes all messages from the specified queue.
--- a/server/azure/servicebus/dataplane.go
+++ b/server/azure/servicebus/dataplane.go
@@ -436,12 +436,9 @@ func (h *Handler) dataPlaneSend(w http.ResponseWriter, r *http.Request, url stri
 // real Service Bus). A topic with no subscriptions, or with none whose rules
 // match, still returns 201 (the message is accepted and dropped), matching
 // Azure. Each matching subscription persists the sender's system properties and
-// applies its own default message TTL.
-//
-// DEFER: only the well-known system properties are evaluated in filters;
-// arbitrary custom application-property headers are not parsed, so a SqlFilter
-// or CorrelationFilter predicate over a user-defined property cannot disqualify
-// a subscription yet.
+// applies its own default message TTL. Both the well-known system properties
+// and the sender's custom application-property headers are evaluated in
+// filters.
 func (h *Handler) dataPlaneTopicSend(w http.ResponseWriter, r *http.Request, targets []topicSubTarget) {
 	if r.Method != http.MethodPost {
 		azurearm.WriteError(w, http.StatusMethodNotAllowed, "MethodNotAllowed", "send requires POST")
@@ -449,7 +446,7 @@ func (h *Handler) dataPlaneTopicSend(w http.ResponseWriter, r *http.Request, tar
 	}
 
 	bp := parseWireBrokerProps(r)
-	filter := bp.filterProps()
+	filter := bp.filterProps(customHeaderProps(r))
 
 	body, ok := readSendBody(w, r)
 	if !ok {
@@ -499,9 +496,9 @@ func parseWireBrokerProps(r *http.Request) wireBrokerProps {
 	return bp
 }
 
-// filterProps projects the well-known system properties a subscription rule
-// filter evaluates against.
-func (b *wireBrokerProps) filterProps() messageProps {
+// filterProps projects the well-known system properties, plus the supplied
+// custom application properties, a subscription rule filter evaluates against.
+func (b *wireBrokerProps) filterProps(custom map[string]string) messageProps {
 	return messageProps{
 		MessageID:        b.MessageID,
 		CorrelationID:    b.CorrelationID,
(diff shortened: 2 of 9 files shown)


NitinKumar004 (commented):
**Verdict: CHANGES REQUESTED — parenthesized SqlFilter expressions still cause message loss.**

The core OR fix works for un-parenthesized expressions and the no-drop invariant holds for unrecognized operators / unknown `sys.*` / genuinely unparseable clauses. But the evaluator does not handle parentheses, and — contrary to the DEFER note — parenthesized clauses do **not** fall to the no-drop `match=true` path. A leading `(` is absorbed into the field name and a trailing `)` into the value, producing a *valid-but-false* comparison that drops the message instead of delivering it.

Reproduced directly against `evalSQLExpression` (white-box), message `{Label:"a", To:"x"}`:

- `(sys.Label = 'a' OR sys.Label = 'b') AND sys.To = 'x'` → **false** (message a real broker delivers is DROPPED)
- `(sys.Label = 'a')` → **false** (even a trivially-parenthesized single clause is DROPPED)

Mechanism: OR-split yields group `(sys.Label = 'a'` → `resolveSQLField("(sys.Label")` isn't a `sys.`/`user.` prefix, so it's looked up as an absent custom property → `present=false` → clause false; the other group carries the trailing `)` into the value (`'b')`), which unquotes to `'b')` and mismatches. Both groups false → not delivered.

This is a regression toward the very HIGH message-loss class this PR targets: the prior evaluator delivered both of these (via the no-drop path, over-delivering but never losing). Parentheses are the canonical way Azure docs and the Terraform `azurerm_servicebus_subscription_rule` sql_filter wrap OR groups, so this is a common real-user shape.

Suggested hardening: before parsing a clause, treat a token with unbalanced/leading `(` or trailing `)` (i.e. anything the grammar can't parse) as unparseable → no-drop `match=true`, or strip balanced surrounding parens on groups/clauses. Either restores the no-loss guarantee for the parenthesized form. A regression test over `(a OR b) AND c` should accompany it.

Everything else verified clean and correct:
- OR delivers matches / excludes non-matches; unknown-`sys.*` and `LIKE` fall to no-drop; quoted-literal edge (`sys.Label='cats or dogs'`) matches.
- Relational numeric (`priority > 5`), custom-string equality (`myprop='x'`), CorrelationFilter user `Properties` (case-insensitive key), reserved-header exclusion — all route correctly.
- Subscription CountDetails/messageCount from backing + `$DeadLetterQueue`; maxDeliveryCount lowered via PUT dead-letters at the new threshold; DeadLetterOnExpiration toggles.
- Prior-round behaviors (simple `sys.X='y'` equality, CorrelationFilter AND, DLQ, LockDuration, TTL/scheduled, namespace isolation) not regressed.

Gates on head b087b53: build, vet, `go test` (server+provider+messagequeue), `-race`, gofmt, golangci-lint (0 new), `go generate` zero-diff, compatgen + `docs/compat` zero-diff — all green.

NitinKumar004 (commented):
Re-reviewed the tokenizer + recursive-descent rewrite of `sqlMatches`/`evalSQLExpression` on head fe072ef3. The core fix is correct and gates are green.

**The parenthesized-OR fix is verified correct.** Direct unit tests over `evalSQLExpression` confirm:
- `(sys.Label='a' OR sys.Label='b') AND sys.To='x'` → {a,x}✓ {b,x}✓ {a,y}✗ {c,x}✗
- single-clause `(sys.Label='a')`, nested `((a OR b) AND c) OR d`, and AND-binds-tighter-than-OR precedence (`a OR b AND x` = `a OR (b AND x)`) all correct.
- Quoted atom containing parens/keywords (`sys.Label='a (OR) b'`) is tokenized as one atom and matches correctly.
- No-drop invariant holds for genuinely unparseable input: unbalanced paren, unterminated quote, empty group `()`, `LIKE`/`IN`/`EXISTS`, unknown operator, unknown `sys.*` identifier, leading/trailing `OR` → all match=true (no drop). CorrelationFilter Properties, subscription counts, and lowered-maxDeliveryCount-on-PUT all verified.

**One blocking finding — new message-loss path for relational operators over non-numeric operands.** `sqlRelational` (filter.go:504-508) returns `false` when `twoFloats` fails, so a relational clause whose operands aren't both numeric silently *excludes* the message:
- `sys.Label > 'm'` (Label='z') → drop
- `mydate > '2024-01-01'` (mydate='2024-06-01') → drop
- `sys.Label >= 'a'` → drop

These are valid Azure SQL filter shapes (relational operators apply to string/datetime, lexicographically), so real Service Bus delivers them. This is also a regression: before this PR every non-`=` clause fell through to match=true (over-deliver, safe); now relational-over-non-numeric drops — the exact message-loss class this PR set out to eliminate, and it contradicts the no-drop invariant documented on `evalSQLExpression`.

Note the asymmetry: `sqlEqual` (filter.go:496) *does* fall back to string comparison; `sqlRelational` does not. Suggested fix: when `twoFloats` fails, fall back to lexical string comparison (`got </<=/>/>= want`) to match Azure semantics — that both fixes correctness and restores no-drop.

Lower-risk edges (non-blocking): malformed clauses that still parse as a single comparison — `sys.Label='a' foo` (trailing token) and `sys.Label == 'a'` (double operator) — evaluate to a garbage-value compare and drop rather than falling to no-drop. These are invalid SQL that real Azure rejects at rule creation, so real-world risk is low.

Gates on fe072ef3 (GOMAXPROCS=3): build, vet, `-race` on server/azure/servicebus, gofmt, `go generate` (zero diff), compatgen (zero docs/compat diff), golangci-lint on both changed packages (0 issues) — all pass.

NitinKumar004 (commented):
Delta re-review of the new hunk since the last pass (fe072ef3..e668875d), on head e668875d.

**Diff scope confirmed clean**: only `server/azure/servicebus/filter.go` (compareSQL doc comment, sqlRelational fallback, new applyRelational helper) and the new `filter_relational_test.go`. No unrelated changes.

**Logic check**: `sqlRelational` still does `twoFloats` first (numeric path unchanged, byte-for-byte). On non-numeric operands it now falls back to `applyRelational(op, float64(strings.Compare(got, want)), 0)`. The operator switch in `applyRelational` is correct: `opLess`→`a<b`, `opGreater`→`a>b`, `opLessEqual`→`a<=b` (includes equality), `opGreaterEqual`→`a>=b` (includes equality) — no sign swap, no inversion. This mirrors `sqlEqual`'s existing numeric-then-string-fallback shape.

**No-drop invariant**: verified `evalSQLExpression` already returns `true` on tokenize/parse failure (lines 172/179, unchanged by this diff) — the "unparseable falls to match" test exercises pre-existing behavior, not new logic, and it passes. The prior blocking bug (relational-over-string dropping messages) is fixed: `sys.Label > 'm'` with `Label='z'` now matches per the new test.

**Test coverage**: `filter_relational_test.go` exercises string/date/numeric operands for `>`, `>=` through `evalSQLExpression`, including a boundary case (`>=` equal) and the unparseable-falls-to-match no-drop case. All pass.

**Gates (GOMAXPROCS=3, -p=3), on e668875d**:
- `go build ./...` — clean
- `go vet ./...` — clean
- `go test -p=3 ./server/azure/servicebus/... ./providers/azure/servicebus/... ./services/messagequeue/...` — all pass
- `go test -race -p=3 ./server/azure/servicebus/...` — clean
- `gofmt -l` on changed files — clean
- `golangci-lint run` on `server/azure/servicebus` — 0 issues
- `go run ./internal/compatgen` — exit 0, no red cells; `git diff docs/compat/` — zero diff

No blocking issues found in this hunk. Working tree clean.

PR #706, opened 2026-08-26
fix(gcp-cloudsql): round-trip instance settings (availabilityType/databaseFlags/backup/ip); databaseVersion patch; databases.update; serverCaCert

NitinKumar004:
Confirming re-audit residue on GCP Cloud SQL against the official Cloud SQL Admin REST docs. Fixes the HIGH Terraform perpetual-drift bug plus four lower-severity gaps.

## Fixes
- **[HIGH] Instance settings now round-trip.** `availabilityType`, `databaseFlags`, `backupConfiguration` and `ipConfiguration` were silently dropped on Insert/Patch, causing Terraform perpetual drift. `availabilityType` maps to the existing driver `Instance.MultiAZ` (REGIONAL<->true) purely in the wire layer. The other three are threaded through new **optional** opaque-JSON `Instance`/`InstanceConfig`/`ModifyInstanceInput` fields and round-trip on Insert->Get and Patch->Get.
- **[MED] Patched `databaseVersion` now visible on Get.** Patch mapped `databaseVersion`->`EngineVersion`, but Get renders from `Engine`. Cloud SQL's `ModifyInstance` now lands a patched version on `Engine`.
- **[LOW] `databases.update` (PUT/PATCH) no longer 405s.** Backed by a new optional `DatabaseUpdater` capability (charset/collation update); kept out of the shared `Databases` interface so Azure/AlloyDB are unaffected.
- **[LOW] `serverCaCert` populated.** Emitted as a synthetic `sslCert` so Terraform's computed `server_ca_cert` is never empty.

## Deferred
- **PUT `/instances/{i}` (Update) full-replace semantics.** Both PATCH and PUT still route to the merge path. True full-replace (absent settings fields reset-to-default) needs a clear/no-change distinction the shared `ModifyInstanceInput` does not carry; Terraform uses PATCH, so impact is low. Deferred rather than done unclean.

## Shared driver / no regression
The shared `services/relationaldb/driver` (RDS + Redshift + Cloud SQL) gains only **optional** fields and one optional interface that RDS/Redshift never set or read. RDS + Redshift server and provider tests pass unchanged.

## Tests
Real `sqladmin/v1` reproductions: settings round-trip (all four), patch availabilityType/databaseFlags, patch `POSTGRES_14`->`15`, `Databases.Update` charset/collation, `serverCaCert` populated. Plus provider-level `UpdateDatabase` and settings-persistence tests. #641 tests remain green.

Docs regenerated (`go generate`, `compatgen`), idempotent.

### DIFF
--- a/docs/coverage/aws/redshift.md
+++ b/docs/coverage/aws/redshift.md
@@ -138,6 +138,14 @@ DBProxies is an OPTIONAL capability for RDS Proxy, discovered by type
 | `ModifyDBProxy` |  |
 | `RegisterDBProxyTargets` |  |
 
+### DatabaseUpdater
+
+DatabaseUpdater is an OPTIONAL capability for updating a logical database's
+
+| Operation | Description |
+| --- | --- |
+| `UpdateDatabase` |  |
+
 ### Databases
 
 Databases is an OPTIONAL capability for managing the logical databases inside
--- a/docs/coverage/azure/mysqlflex.md
+++ b/docs/coverage/azure/mysqlflex.md
@@ -138,6 +138,14 @@ DBProxies is an OPTIONAL capability for RDS Proxy, discovered by type
 | `ModifyDBProxy` |  |
 | `RegisterDBProxyTargets` |  |
 
+### DatabaseUpdater
+
+DatabaseUpdater is an OPTIONAL capability for updating a logical database's
+
+| Operation | Description |
+| --- | --- |
+| `UpdateDatabase` |  |
+
 ### Databases
 
 Databases is an OPTIONAL capability for managing the logical databases inside
--- a/docs/coverage/coverage.json
+++ b/docs/coverage/coverage.json
@@ -8046,6 +8046,15 @@
         ]
       },
       {
+        "name": "DatabaseUpdater",
+        "doc": "DatabaseUpdater is an OPTIONAL capability for updating a logical database's",
+        "operations": [
+          {
+            "name": "UpdateDatabase"
+          }
+        ]
+      },
+      {
         "name": "Databases",
         "doc": "Databases is an OPTIONAL capability for managing the logical databases inside",
         "operations": [
--- a/docs/coverage/gcp/alloydb.md
+++ b/docs/coverage/gcp/alloydb.md
@@ -138,6 +138,14 @@ DBProxies is an OPTIONAL capability for RDS Proxy, discovered by type
 | `ModifyDBProxy` |  |
 | `RegisterDBProxyTargets` |  |
 
+### DatabaseUpdater
+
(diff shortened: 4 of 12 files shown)


NitinKumar004 (commented):
Reviewed at head 23478cd3 (clean-built the pushed tree, not the report). Verdict: LGTM / merge.

Gates (GOMAXPROCS=3, -p=3):
- go build ./... clean; go vet clean.
- go test ./server/gcp/cloudsql ./providers/gcp/cloudsql ./server/aws/rds ./providers/aws/rds ./server/aws/redshift ./providers/aws/redshift ./services/relationaldb/... all green.
- go test -race ./providers/gcp/cloudsql clean.
- go generate ./... => zero diff (coverage.json + redshift/mysqlflex/alloydb md already regenerated).
- go run ./internal/compatgen => suite green, git diff docs/compat zero.
- golangci-lint --new-from-rev on the three changed packages: 0 issues. gofmt clean.

Adversarial SDK verification (my own google.golang.org/api/sqladmin/v1 + aws-sdk-go-v2 rds tests, beyond the PR fixtures):
- Settings round-trip: Insert{availabilityType:REGIONAL, databaseFlags, backupConfiguration, ipConfiguration} => Get returns all four intact.
- availabilityType<->MultiAZ both directions: REGIONAL->true on Insert, Patch ZONAL->false on Get; default (no availabilityType) renders ZONAL.
- Full patch-merge: a label-only Patch (Settings present, only UserLabels set) preserves databaseVersion + all four settings sub-objects + tier. Empty-blob = no change is correct because ModifyInstance also guards InstanceClass!="" / AllocatedStorage>0 / Tags!=nil / EngineVersion!="".
- databaseVersion patch lands on inst.Engine (the field toSQLInstance renders) => Get shows the new version.
- databases.update AND databases.patch both apply charset/collation; partial patch merges (empty field preserved); missing db => NotFound.
- serverCaCert deterministic: two Gets of one instance return identical cert+fingerprint (sha1 over "serverCaCert/"+ID); distinct instances get distinct fingerprints.

Shared-driver blast radius (RDS/Redshift):
- The new Instance/InstanceConfig/ModifyInstanceInput opaque-JSON fields and the DatabaseUpdater capability are GCP-only; RDS/Redshift never set or read them.
- Verified with a real RDS create->ModifyDBInstance->Describe cycle: EngineVersion still lands on EngineVersion (NOT Engine, unlike the Cloud SQL-specific inst.Engine mapping), class/MultiAZ modify as before, no settings blob leaks into any RDS wire field. RDS/Redshift suites unchanged.

Notes (non-blocking): opaque-blob patch replaces the whole sub-object (e.g. databaseFlags list) rather than element-merging, which matches real Cloud SQL patch semantics for these fields. Instance-level PUT full-replace remains deferred as stated.

PR #707, opened 2026-08-26
fix(azure-keyvault): enforce nbf/exp on key crypto + secret GET; RotateKey; certificate mints backing secret+key

NitinKumar004:
confirming re-audit residue

### DIFF
--- a/docs/coverage/aws/secretsmanager.md
+++ b/docs/coverage/aws/secretsmanager.md
@@ -68,6 +68,7 @@ KeyVaultKeys is the Azure Key Vault keys data-plane surface: key lifecycle
 | `ListKeys` |  |
 | `PurgeDeletedKey` |  |
 | `RecoverDeletedKey` |  |
+| `RotateKey` |  |
 | `SignKey` |  |
 | `UnwrapKey` |  |
 | `UpdateKey` |  |
--- a/docs/coverage/azure/keyvault.md
+++ b/docs/coverage/azure/keyvault.md
@@ -68,6 +68,7 @@ KeyVaultKeys is the Azure Key Vault keys data-plane surface: key lifecycle
 | `ListKeys` |  |
 | `PurgeDeletedKey` |  |
 | `RecoverDeletedKey` |  |
+| `RotateKey` |  |
 | `SignKey` |  |
 | `UnwrapKey` |  |
 | `UpdateKey` |  |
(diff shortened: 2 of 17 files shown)


NitinKumar004 (commented):
Verified the crypto/security behavior against the real Azure SDK (azkeys/azsecrets) plus raw-REST cert probes, including edge cases beyond the added tests. All gates green on the pushed tree (build, vet, targeted + root tests, `-race`, gofmt, `go generate` zero-diff, compatgen zero-diff, golangci-lint 0 issues).

**Confirmed correct**
- Key window direction is right: `now < nbf` (not-yet-valid) and `now >= exp` (expired) block; a key valid right now permits encrypt/sign/wrap; the `>=` gives the correct exp-boundary semantics.
- Allow-list is complete in both directions: decrypt / unwrapKey / verify keep working on an expired/not-yet-valid key (I wrapped-then-expired-then-unwrapped successfully), while encrypt / sign / wrapKey are blocked 403.
- `Enabled=false` still blocks regardless of window; a key with no nbf/exp is unrestricted.
- Secret GET returns 403 on both the bare (`/secrets/{name}`) and versioned (`/secrets/{name}/{version}`) paths — they funnel through the same handler after the Enabled check — while LIST is not over-blocked (metadata for expired/nbf secrets still enumerated).
- RotateKey: new version becomes current with a distinct kid, old version stays gettable, kty/curve/keyOps preserved, rotation-policy ISO-8601 expiry honored (`P30D` → exp ~30d out), non-existent key → 404, concurrent rotate is race-clean.
- Certificate mint: SID/KID resolve to the managed secret (PEM bundle, `managed=true`) and managed key (`managed=true`); the `managed` field round-trips through snapshot/restore. AWS/GCP untouched.

**Follow-ups (non-blocking)**
1. `DeleteCertificate` does not clean up the managed secret+key it minted — real Key Vault soft-deletes the backing key and secret alongside the certificate; here they stay live and gettable (orphaned). Worth a follow-up.
2. No managed-object write protection: a caller can `SetSecret`/`DeleteSecret`/`UpdateKey` directly on the managed name; Azure returns Forbidden ("Managed") for direct writes to certificate-managed objects.
3. Enforcement asymmetry: the key window is enforced in the provider (`opKey`), but the secret window is enforced only in the HTTP handler (`getSecret`) using `time.Now()` rather than the configured Clock. So the typed/in-process library path does not enforce secret expiry, and the secret window is not FakeClock-deterministic. The key path is correct on both counts.
4. `RotateKey` on an imported `oct` key returns InvalidArgument (generateVersion only mints RSA/EC) — acceptable since Azure doesn't rotate imported material, just noting.

PR #708, opened 2026-08-26
fix(gcp): hard-delete Secret Manager secrets, 400 on illegal version transition, stable IAM list pagination

NitinKumar004:
## Summary

Fixes three real correctness bugs in GCP Secret Manager and GCP IAM.

### Bug 1 (HIGH) — DeleteSecret then re-create same name wrongly 409 ALREADY_EXISTS
GCP Secret Manager `secrets.delete` is a permanent hard delete with no recovery window; the same `secretId` must be creatable immediately afterwards. The GCP provider previously soft-deleted (stamped `deletedAt` and kept the entry), so `CreateSecret`'s existence guard still saw it and returned AlreadyExists. `DeleteSecret` now hard-deletes (`m.secrets.Delete`, dropping versions) and all GCP-only `deletedAt`/soft-delete machinery is removed from the GCP path. After delete: GetSecret 404s, ListSecrets omits it, and re-creating the same id succeeds. **AWS Secrets Manager soft-delete + recovery window + RestoreSecret is a separate provider and is untouched** (proven green).

### Bug 2 (LOW) — illegal version-lifecycle transition returned 409 instead of 400
Enable/Disable/Destroy on a DESTROYED (illegal-state) version returns `FailedPrecondition`, which the shared `gcprest.WriteCErr` maps to 409 conditionNotMet. Real GCP returns HTTP 400 FAILED_PRECONDITION. Fixed **locally** in the secretmanager version-lifecycle handler (hardcoded 400, mirroring the existing `:access` path) — the global `gcprest` mapping is left unchanged to avoid cross-service blast radius.

### Bug 3 (MED) — ListServiceAccounts / ListRoles pagination duplicated/skipped across pages
`listServiceAccounts`/`listRoles` build their result from `ListUsers`/`ListRoles`, which return `memstore.All()` in random map order, then an offset page token indexes into that unsorted slice — so page boundaries shifted between requests, duplicating or skipping entries. Both now sort `matched` by a stable key (resource name) before `paginate`.

## Tests
- Provider: delete then re-create same id succeeds; GetSecret 404 and ListSecrets omits after delete.
- Server (real GCP SDK): Enable/Disable/Destroy on a destroyed version each returns HTTP 400 FAILED_PRECONDITION; delete then recreate same secretId succeeds with Get 404 in between.
- Server (real GCP SDK): 150 service accounts and 150 roles paged fully with a small page size — union across pages equals the full set, each returned exactly once (no dups/skips).

## Deferrals (out of scope, tracked separately)
replication/ttl/rotation decode, `?filter=`, project-level cloudresourcemanager IAM, etag preconditions, dataCrc32c, role etag.

### DIFF
--- a/providers/gcp/secretmanager/iam.go
+++ b/providers/gcp/secretmanager/iam.go
@@ -19,10 +19,6 @@ func (m *Mock) GetSecretIAMPolicy(_ context.Context, name string) (*driver.GCPIA
 	sd.mu.RLock()
 	defer sd.mu.RUnlock()
 
-	if !sd.deletedAt.IsZero() {
-		return nil, errors.Newf(errors.NotFound, "secret %q is scheduled for deletion", name)
-	}
-
 	if sd.iam == nil {
 		return &driver.GCPIAMPolicy{Version: 1, Etag: newEtag()}, nil
 	}
@@ -40,10 +36,6 @@ func (m *Mock) SetSecretIAMPolicy(_ context.Context, name string, policy driver.
 
 	sd.mu.Lock()
 	defer sd.mu.Unlock()
-
-	if !sd.deletedAt.IsZero() {
-		return nil, errors.Newf(errors.NotFound, "secret %q is scheduled for deletion", name)
-	}
 
 	stored := clonePolicy(&policy)
 	if stored.Version == 0 {
@@ -66,10 +58,6 @@ func (m *Mock) TestSecretIAMPermissions(_ context.Context, name string, permissi
 
 	sd.mu.RLock()
 	defer sd.mu.RUnlock()
-
-	if !sd.deletedAt.IsZero() {
-		return nil, errors.Newf(errors.NotFound, "secret %q is scheduled for deletion", name)
-	}
 
 	granted := make([]string, len(permissions))
 	copy(granted, permissions)
--- a/providers/gcp/secretmanager/lifecycle.go
+++ b/providers/gcp/secretmanager/lifecycle.go
@@ -42,10 +42,6 @@ func (m *Mock) mutateVersion(name, versionID string,
 	sd.mu.Lock()
 	defer sd.mu.Unlock()
 
-	if !sd.deletedAt.IsZero() {
-		return nil, errors.Newf(errors.NotFound, "secret %q is scheduled for deletion", name)
-	}
-
 	v, ok := findVersion(sd, versionID)
 	if !ok {
 		return nil, errors.Newf(errors.NotFound, "version %q not found for secret %q", versionID, name)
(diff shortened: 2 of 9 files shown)


NitinKumar004 (commented):
Deep adversarial review — verdict: MERGE (clean-built on pushed head a45a0cfe).

Gates on the pushed SHA (GOMAXPROCS=3, -p=3): build, vet, gofmt -l, go generate (zero diff, 51 services), compatgen (exit 0, zero docs/compat diff), and golangci-lint --new-from-rev (0 new issues) all clean. The pre-existing gocyclo on iam/handler.go routeServiceAccounts is untouched by this PR (handler.go is not in the diff). Tests green including -race on providers/gcp/secretmanager and server/gcp/iam, plus AWS secretsmanager and root integration.

BUG1 (hard-delete rework) — verified adversarially with my own tests:
- Delete then recreate the same secretId: exactly one version numbered 1 (numbering restarts), AccessSecretVersion returns only the new payload, and the deleted secret's old version id ("2") no longer resolves — versions are genuinely dropped, not orphaned. secretData holds versions inline (no separate global version index/map), so memstore.Delete drops everything cleanly.
- Removal is complete: no residual deletedAt/DeletedAt/soft-delete/RestoreSecret code anywhere in the package (only two descriptive comments mention "recovery window"). The `time` import in secretmanager.go is still legitimately used (RFC3339 timestamps); snapshot.go's `time` import removal is coherent.
- Persist round-trip: snapshot a secret with multiple versions + IAM, restore into a fresh mock — value, version history, IAM, and verCounter all restore intact (next PutSecretValue correctly numbers 3, not 1). Restored secret is Gettable/listable/accessible.
- Backward-compat: an OLD snapshot JSON still carrying a `deletedAt` field decodes without error (unknown field ignored) and restores the secret as live. No crash, no data loss. Note (non-blocking): a pre-PR snapshot captured while a secret was soft-deleted now restores it as a live/accessible secret rather than a tombstone. This is a defensible semantic change (matches real GCP hard-delete), just worth being aware of.

BUG2 (400 on illegal version transition): the local override in server/gcp/secretmanager/operations.go mutateVersion returns 400 failedPrecondition only for IsFailedPrecondition; the shared gcprest.WriteCErr mapping is byte-for-byte unchanged (FailedPrecondition still -> 409 conditionNotMet globally), so other GCP services are unaffected. Real-SDK test confirms Destroy/Disable/Enable on a DESTROYED version each return HTTP 400.

BUG3 (stable IAM list pagination): sort by the stable resource Name before offset-paginate in both listServiceAccounts and listRoles. Real-SDK test walks 150 SAs and 150 roles across pages with no dups/skips; the Name sort is a total order over unique names, so page boundaries are deterministic across repeated requests.

AWS unaffected: AWS Secrets Manager soft-delete / recovery-window / RestoreSecret lives in the separate providers/aws/secretsmanager + server/aws/secretsmanager packages (not in this PR's 9-file diff); their tests (incl. RestoreSecret cycle) pass.

Convention: testify in providers/gcp/secretmanager/secretmanager_test.go was already present pre-PR (assert+require imported at the merge-base), so the new provider test matching it is consistent with the file's existing state — not introduced by this PR.

PR #710, opened 2026-08-26
fix(azure): Queue Storage message plane no longer shadowed by Service Bus routing

NitinKumar004:
## Summary

Azure Queue Storage's message plane was **entirely dead** in the full wire server. The dispatcher is first-match-wins, and the Service Bus handler (registered before Queue Storage) claimed **any** non-`/subscriptions/` path containing `/messages` via `isDataPlanePath`. So a real `azqueue` SDK call — `POST/GET/DELETE /{queue}/messages` — was swallowed by Service Bus, which found no SB entity named `{queue}` and returned `404 ResourceNotFound`. Queue create/delete/list/metadata have no `/messages` segment, so only the message plane (enqueue/dequeue/peek/delete/clear) was broken.

## Root cause

Both Queue Storage and Service Bus address a flat `/{entity}/messages` URL. Real Azure separates them by hostname (`*.queue.core.windows.net` vs `*.servicebus.windows.net`), which isn't available behind CloudEmu's shared emulator endpoint. Registration order can't disambiguate — both shapes genuinely collide.

## Fix (entity-resolution, least blast radius)

Narrowed the Service Bus data-plane `Matches`: a flat `/{entity}/messages` request is claimed **only when `{entity}` resolves to a Service Bus queue or topic the handler actually holds** (consulting the SB provider store it already owns). Otherwise it declines and the request falls through to the Queue Storage handler. Subscription paths (`/{topic}/subscriptions/{sub}/messages`) have no Queue Storage counterpart, so they remain unambiguously Service Bus (claimed when the parent topic resolves). Because Service Bus is registered first, a name collision still resolves in Service Bus's favor for entities it holds.

### Trade-off (documented)

A flat send to a **non-existent** Service Bus entity is no longer 404'd by Service Bus — it falls through to Queue Storage (which itself 404s an unknown queue). Real Service Bus also requires the entity to exist, so the outcome (a not-found error) is preserved; only the error's wire shape differs. A genuinely-addressed Service Bus miss on the subscription path still 404s correctly. Same-name collision on the namespace-less bare path is a known limitation, consistent with the existing Blob/Queue shared-endpoint collision docs.

## Verification (adversarial, full server)

New `server/azure/queue_servicebus_routing_test.go` boots the **full production server** (`NewFromProvider(cloudemu.NewAzure())` — the same wiring `cloudemu serve` uses) and drives the real `azqueue`/`azblob`/`aztables` SDKs plus the Service Bus REST data plane:

- **Queue plane alive**: azqueue CreateQueue -> EnqueueMessage -> PeekMessage -> DequeueMessage -> DeleteMessage -> EnqueueMessage -> ClearMessages, all succeed (no 404).
- **Service Bus unaffected**: create SB queue via ARM, REST send + receive.
- **SB topic+subscription unaffected**: create topic + subscription, publish to topic, receive on subscription.
- **Coexistence**: a Storage queue and an SB queue with different names both work simultaneously, no cross-talk.
- **Blob/Table unaffected**: blob PUT/GET and table insert/query still route correctly.

The pre-existing `TestDataPlaneSendToMissingQueue` (which encoded the old "SB owns every /messages path" contract) was updated to `TestDataPlaneUnknownFlatEntityDeclined`, asserting the new decline behavior.

## Gates

`go build ./...`, `go vet`, `go test ./server/azure/... ./providers/azure/...`, `go test -race ./server/azure/queue/... ./server/azure/servicebus/...`, root integration `go test .`, `gofmt -l` clean, `golangci-lint` 0 issues on changed packages, `go generate ./...` zero diff, compatgen + `git diff docs/compat/` zero — all green.

### DIFF
--- a/server/azure/queue_servicebus_routing_test.go
+++ b/server/azure/queue_servicebus_routing_test.go
@@ -1,0 +1,345 @@
+package azure_test
+
+// Routing regression tests for the Azure Queue Storage / Service Bus data-plane
+// collision (real-user BLOCKER). Both surfaces address a flat
+// "/{entity}/messages" URL; behind CloudEmu's shared endpoint the hostname that
+// separates them on real Azure is unavailable. The Service Bus handler is
+// registered before Queue Storage and used to claim EVERY "/messages" path,
+// which swallowed the entire Queue Storage message plane (create/list still
+// worked, but enqueue/dequeue/peek/delete/clear 404'd as "Service Bus entity
+// not found").
+//
+// The fix narrows Service Bus's Matches so it claims a flat "/{entity}/messages"
+// request only when {entity} resolves to a Service Bus queue or topic it holds.
+// These tests drive the FULL production server (NewFromProvider — the same
+// wiring `cloudemu serve` uses) with the real azqueue/azblob/aztables SDKs plus
+// the Service Bus REST data plane, proving both planes work simultaneously.
+
+import (
+	"context"
+	"encoding/json"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/data/aztables"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azqueue"
+
+	"github.com/stackshy/cloudemu/v2"
+	azureserver "github.com/stackshy/cloudemu/v2/server/azure"
+)
+
+const (
+	sbAPIVer = "?api-version=2022-10-01-preview"
+	sbSubID  = "sub-routing"
+	sbRG     = "rg-routing"
+	sbNS     = "ns-routing"
+)
+
+// newFullAzureServer boots the full production Azure server (all handlers
+// registered, same as `cloudemu serve`).
+func newFullAzureServer(t *testing.T) *httptest.Server {
+	t.Helper()
+
+	ts := httptest.NewServer(azureserver.NewFromProvider(cloudemu.NewAzure()))
+	t.Cleanup(ts.Close)
+
+	return ts
+}
+
+// anonOpts returns SDK client options that route through the test server with
+// retries disabled (anonymous access; no SharedKey signing needed).
+func anonOpts(ts *httptest.Server) policy.ClientOptions {
+	return policy.ClientOptions{
+		Transport: ts.Client(),
+		Retry:     policy.RetryOptions{MaxRetries: -1},
+	}
+}
+
+// sbDo issues a raw HTTP request against the full server (Service Bus REST),
+// asserts the status code, and returns the response body as a string.
+func sbDo(t *testing.T, ts *httptest.Server, method, path, body string, wantStatus int) string {
+	t.Helper()
+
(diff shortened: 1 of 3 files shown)


NitinKumar004 (commented):
Reviewed at head `9004292b`, full gate matrix + adversarial real-SDK e2e on the full production server (`NewFromProvider(cloudemu.NewAzure())`). Verdict: looks good to merge.

## Verification

**Gates (GOMAXPROCS=3):** `go build ./...` clean · `go vet` clean · `go test ./server/azure/... ./providers/azure/...` all ok · `go test -race ./server/azure/queue/... ./server/azure/servicebus/...` and the new full-server routing test under `-race` clean · root `go test .` ok · `gofmt -l` clean · `go generate ./...` zero diff · compatgen + `git diff docs/compat` zero.

**Lint:** handler.go + the two test files are clean. The 3 `canonicalheader` findings are in `dataplane.go` (the SB REST `BrokerProperties` header, intentional exact-casing) — an untouched file, so 0 NEW.

**No SB data-plane verb regressed** (the highest-risk failure mode). `claimsDataPlane` and `serveDataPlane` share the same `parseDataPlanePath` + resolve functions, so they agree on every shape. The one asymmetry is safe: subscription paths are claimed on *topic* existence (`resolveTopicSubURLs`) while served on *subscription* existence (`resolveSubscription`) — topic-existence is a strict superset, so no legitimate subscription receive is ever declined. Confirmed live on the full server: topic send 201, subscription receive 200, subscription `$DeadLetterQueue` receive 204, flat-queue `$DeadLetterQueue` receive 204, plus the existing peek-lock/renew-lock/complete/abandon suite passing through the dispatcher's `Matches` under `-race`. The #705 SqlFilter subscription path also still routes.

**Queue plane revived:** azqueue Create→Enqueue→Peek→Dequeue→Delete→Enqueue→Clear all succeed end-to-end. Blob PUT/GET and table insert/query unaffected.

**No shared-store cross-talk:** `ServiceBus` and `QueueStorage` are distinct `servicebus.New(o)` instances (providers/azure/azure.go:183-184), and SB's `resolveQueue` only searches SB's ARM-created namespace store — it can never see a Storage queue, so it never false-claims one.

**Non-existent SB flat send:** falls through to Queue Storage and 404s `QueueNotFound` — no 200 auto-create masking a typo. Confirmed against `resolveQueueURL` (returns NotFound, no create) and live.

**TOCTOU:** `Matches` and `ServeHTTP` each take their own RLock, so an entity deleted between them makes SB's `serveEntityData` 404 within SB rather than mis-route — the dispatcher never retries another handler after a match. Because the two planes are separate stores, no single request can ever land a message in the wrong store. `-race` clean.

## One documented footgun (non-blocking)

The name-collision case is sharper than "SB wins": with an SB queue `x` in *any* namespace, a real azqueue client's namespace-less `POST /x/messages` is claimed by SB (namespace-less `resolveQueue("", "x")` searches all namespaces), so the enqueue silently lands in the SB store and the client's `GET /x/messages` peek/dequeue then returns `405 send requires POST` from SB — the Storage queue `x` is effectively unusable while the SB entity exists. This is inherent to the shared endpoint (real Azure separates by host, and both real azqueue and real SB REST send namespace-less `/{entity}/messages`), SB-registered-first is the least-surprising default, and it's consistent with the existing blob/queue shared-endpoint collision docs. Worth a line in those collision docs so users hitting a same-name clash know SB takes precedence, but not a blocker.

PR #712, opened 2026-08-26
fix(azure): Queue Storage message-plane semantics (metadata, receipts, times, idempotency)

NitinKumar004:
## Summary

Fixes five Azure Queue Storage message-plane bugs, verified end-to-end against the full wire server (`azureserver.NewFromProvider(cloudemu.NewAzure())`) with the real `azqueue` SDK. These were previously e2e-blocked behind the Queue↔Service Bus routing fix (#710), now merged.

1. **CreateQueue metadata dropped** — `x-ms-meta-*` headers on Create Queue were ignored. Added `QueueConfig.Metadata`, persisted it in the provider, and populated it in the handler via the existing `metadataFromHeaders` helper. Get Properties / Get Metadata now returns it.
2. **DeleteMessage wrong error code** — a stale/invalid pop receipt returned `QueueNotFound`; now returns 404 `MessageNotFound`, mirroring the Update Message path.
3. **Enqueue pop receipt unusable until first dequeue** — Put Message returned the message ID as the pop receipt, but the stored receipt was empty until dequeue, so delete/update of a just-enqueued message failed. The provider now mints a real pop receipt at enqueue time and returns it (`SendMessageOutput.PopReceipt`); a later dequeue supersedes it, matching Azure.
4. **Dequeue reported times wrong** — response reported `InsertionTime = now` and `TimeNextVisible = now + rawVisibilityTimeout` (so `= now` when omitted). Added an insertion timestamp to the message model; the dequeue response now reports the actual enqueue time and `TimeNextVisible` pushed out by the effective visibility (30s default applied).
5. **Duplicate CreateQueue idempotency** — a duplicate create always returned 204. Now 204 when the requested metadata matches the existing queue's, and 409 `QueueAlreadyExists` when it differs.

Service Bus routing and behavior are untouched; the shared provider change (enqueue-time receipt) is verified not to affect the Service Bus wire package.

## Tests

New `server/azure/queue/message_semantics_test.go` (full server + real `azqueue`, one test per bug):
- create-with-metadata → GetProperties returns it (B1)
- delete with stale receipt → MessageNotFound (B2)
- enqueue then delete using the enqueue-returned receipt without dequeue → succeeds (B3)
- dequeue reports InsertionTime ≈ enqueue time and TimeNextVisible in the future by the effective visibility (B4)
- duplicate create same metadata → 204, different metadata → 409 (B5)

## Gates

`go build ./...`, `go vet`, `go test` (queue / servicebus provider / messagequeue / servicebus wire), `-race` on the queue package, root integration test, `gofmt -l`, `go generate` (zero diff), and `compatgen` (zero `docs/compat` diff) all green. golangci-lint on changed packages: 0 new issues.

### DIFF
--- a/providers/azure/servicebus/servicebus.go
+++ b/providers/azure/servicebus/servicebus.go
@@ -154,6 +154,11 @@ func (m *Mock) CreateQueue(_ context.Context, cfg driver.QueueConfig) (*driver.Q
 		tags[k] = v
 	}
 
+	metadata := make(map[string]string, len(cfg.Metadata))
+	for k, v := range cfg.Metadata {
+		metadata[k] = v
+	}
+
 	visibilityTimeout := cfg.VisibilityTimeout
 	if visibilityTimeout == 0 {
 		visibilityTimeout = defaultVisibilityTimeout
@@ -182,7 +187,7 @@ func (m *Mock) CreateQueue(_ context.Context, cfg driver.QueueConfig) (*driver.Q
 		deduplicationIndex:     make(map[string]time.Time),
 		dlqConfig:              cfg.DeadLetterQueue,
 		deadLetterOnExpiration: cfg.DeadLetterOnExpiration,
-		metadata:               make(map[string]string),
+		metadata:               metadata,
 	}
 
 	m.queues.Set(url, qd)
@@ -292,6 +297,12 @@ func (m *Mock) SendMessage(_ context.Context, input driver.SendMessageInput) (*d
 	// For a non-scheduled send visibleAt == now, so this is unchanged.
 	expiresAt := computeExpiry(visibleAt, input.MessageTTLSeconds)
 
+	// Mint a pop receipt at enqueue time so the message can be deleted or updated
+	// with the Put Message-returned receipt before its first dequeue, as Azure
+	// Queue Storage allows. A subsequent dequeue issues a fresh receipt that
+	// supersedes this one.
+	receiptHandle := idgen.GenerateID("sb-lock-")
+
 	msg := &sbMessage{
 		ID:              msgID,
 		Body:            input.Body,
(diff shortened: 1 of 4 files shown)


NitinKumar004 (commented):
Reviewed at head f99a9d7d against a full `azureserver.NewFromProvider` server with the real `azqueue` SDK, plus the Service Bus HTTP data-plane suite.

**Gates (GOMAXPROCS=3):** build, vet, `go test` (queue / servicebus provider / messagequeue / servicebus wire), `-race` (queue + servicebus wire + shared provider), root integration test, `gofmt -l`, `go generate` (zero diff), `compatgen` + `git diff docs/compat` (zero), and `golangci-lint --new-from-rev=origin/development` on the changed packages (0 new) — all green.

**Per-fix verification:**
- **B1 metadata** — Create with `x-ms-meta-*` round-trips through GetProperties; no-metadata create yields an empty (non-nil) map, no panic; SetMetadata updates. The provider defensively copies `cfg.Metadata` and the map is always allocated.
- **B2 delete code** — stale/garbage pop receipt on a resolved queue → 404 `MessageNotFound`; DeleteMessage on a missing queue is caught earlier by `resolveQueueURL` → `QueueNotFound` (not over-mapped). UpdateMessage's pre-existing `MessageNotFound` mapping is unregressed.
- **B3 receipt (shared provider)** — enqueue-returned receipt deletes/updates the message pre-dequeue. Supersede semantics are correct: `buildReceivedMessage` / `UpdateMessage` overwrite `msg.ReceiptHandle`, so the old enqueue receipt R1 fails after a dequeue mints R2 (the blocking case), and R2 succeeds. Two enqueues mint distinct receipts and IDs. Receipt minting is inside `qd.mu`.
- **B4 times** — dequeue reports `InsertionTime` = actual enqueue time (`SentAt`) and `TimeNextVisible` pushed out by the effective visibility (30s default applied when omitted). Peek leaves visibility/dequeue-count untouched.
- **B5 idempotency** — duplicate create with identical metadata → 204; differing metadata → 409 `QueueAlreadyExists`; no-metadata-either-side → 204 (`maps.Equal(nil, emptyMap)` is true).

**Service Bus unaffected** — the shared `SendMessage` now populates `SendMessageOutput.PopReceipt`, but the SB wire handler discards the output (`_, err := h.mq.SendMessage(...)`), and SB receive mints its own lock token, so send→receive→complete, peek-lock→abandon, dead-letter, and SqlFilter subscriptions are unchanged; the full SB HTTP data-plane + filter suites pass, including under `-race`.

No blocking issues.

PR #713, opened 2026-08-26
fix(dynamodb): preserve exact decimal strings in Number Sets (NS)

NitinKumar004:
## Summary

DynamoDB Number Sets (NS) silently corrupted any element beyond `2^53`. `decodeNumberSet` parsed each element via `strconv.ParseFloat(s, 64)` and `NumberSet` was `[]float64`, so a `PutItem` NS of `["123456789012345678901234567890","1"]` read back `["123456789012345680000000000000","1"]`. Large ids, high-precision decimals, and money values were rounded away. The scalar `N` type was already fixed to keep its exact decimal string (`expr.Number`); NS never got the same treatment.

## Fix

Mirror the scalar-N approach for the set type:

- `expr.NumberSet` is now `[]expr.Number` (exact decimal strings) instead of `[]float64`.
- Uniqueness, membership, `contains()`, `ADD` (union) and `DELETE` (difference) compare by **exact numeric value** using `big.Rat` — so `"1"`, `"1.0"` and `"+1"` are one element, while two 30-digit values that share a `float64` stay distinct (the exact bug being fixed). Survivors keep their original strings.
- Wire decode keeps each element's exact string and dedups numerically-equal forms (union with an empty set); wire encode emits the exact strings.

### Why the shared `expr` package changed

`NumberSet`'s representation lives in `services/database/driver/expr` (`sets.go`, `number.go`, `eval.go`), and the exact-string wire encoder lives in `services/database/driver/stream_event.go` (shared by control-plane responses, Streams, and Lambda stream delivery). All were use-sites of the old `[]float64` set and had to move to the string-backed form together so N and NS stay consistent.

## Tests

`server/aws/dynamodb/dynamodb_number_set_test.go` — real aws-sdk-go-v2 round-trips:

- NS with two 30-digit integers sharing one `float64` → both exact strings back, distinct
- high-precision decimal NS element (>15 sig digits) round-trips exactly
- `ADD` unions by numeric value, large element keeps exact string, `1.0` dedups against `1`
- `DELETE` removes by numeric value, survivors keep exact strings
- numerically-equal NS elements dedup to one (set semantics)
- scalar `N` still round-trips (regression guard)

### DIFF
--- a/server/aws/dynamodb/dynamodb_number_set_test.go
+++ b/server/aws/dynamodb/dynamodb_number_set_test.go
@@ -1,0 +1,160 @@
+// dynamodb_number_set_test.go — real aws-sdk-go-v2 round-trips proving the
+// DynamoDB Number Set (NS) preserves each element's exact decimal string.
+// Elements beyond float64's mantissa (large ids, high-precision decimals) must
+// not be corrupted by parsing through float64, and set operations (ADD/DELETE)
+// must act by numeric value while keeping the survivors' exact strings.
+package dynamodb_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
+	ddbtypes "github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func nsAttr(vals ...string) ddbtypes.AttributeValue {
+	return &ddbtypes.AttributeValueMemberNS{Value: vals}
+}
+
+func attrNS(t *testing.T, item map[string]ddbtypes.AttributeValue, field string) []string {
+	t.Helper()
+
+	v, ok := item[field].(*ddbtypes.AttributeValueMemberNS)
+	require.True(t, ok, "attribute %q should be NS, got %T", field, item[field])
+
+	return v.Value
+}
+
+func suiteDDBUpdate(t *testing.T, client *dynamodb.Client, table, expr string,
+	key, vals map[string]ddbtypes.AttributeValue) {
+	t.Helper()
+
+	_, err := client.UpdateItem(context.Background(), &dynamodb.UpdateItemInput{
+		TableName:                 aws.String(table),
+		Key:                       key,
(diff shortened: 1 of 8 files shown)


NitinKumar004 (commented):
Verdict: MERGE (clean-built the pushed head 6b5284cb in an isolated worktree).

Data-integrity edges probed with real aws-sdk-go-v2 round-trips plus targeted internal tests; the exact bug is fixed and does not regress:

- Exact round-trip beyond 2^53: NS `["123456789012345678901234567890","1"]` and a 30-significant-digit decimal `3.141592653589793238462643383279` return byte-identical; negative/`0`/`-0.5` exact. Scalar N unregressed.
- big.Rat equality is correct at the boundary: `100000000000000000000000000001` and `...002` (equal as float64) stay TWO distinct elements, while `1`/`1.0`/`+1`/`1.00`/`1e2==100`/`0==-0` collapse to one (first-seen string wins). This is precisely the collapse a float dedup caused.
- ADD unions by numeric value (a numerically-equal add is a no-op, large survivor keeps its exact string); DELETE removes by value keeping survivors' exact strings; `contains()` matches a >2^53 element by value and does not match its near neighbour.
- Encoder: control-plane GetItem (`MarshalItem`) and Lambda/Streams delivery share one path — `marshalBinaryOrSet`/`formatNumberSet` now emits the exact decimal strings, so the stream record is not re-corrupted through FormatFloat.
- No remaining NS->float coercion: `decodeNumberSet` keeps `expr.Number`; the only float sites are `valueSize` (len-based) and numeric ADD/comparison arithmetic (the documented scalar-N float limit, unchanged, out of scope for set membership).
- big.Rat robustness: malformed operands fall back to exact string compare (no panic); a huge-exponent input (`1e1000000000`) is rejected by the stdlib guard — no hang/DoS.

Blast radius: `NumberSet` is DynamoDB-only; the shared expr evaluator's NS branches are type-switched and inert for other drivers (cosmossql constructs no NS).

Gates on the pushed SHA (GOMAXPROCS=3): build, vet, `go test` (server/aws/dynamodb, providers/aws/dynamodb, services/database/...), `-race` (providers/aws/dynamodb), root `go test .` — all green; gofmt clean; `go generate` zero diff; compatgen exit 0 + `docs/compat` zero diff; golangci-lint `--new-from-rev=origin/development` 0 new issues.

PR #715, opened 2026-08-26
fix(azure): overlay never reflects write-only secrets back on reads

NitinKumar004:
## Summary

The generic Azure ARM property-echo overlay (`server/azure/echo_properties.go`) reflected **write-only secrets** back on reads — a credential leak.

The overlay's job is to preserve request `properties` keys a handler didn't model, echoing them on the create response and later GETs. But when a handler *deliberately* omits a write-only secret from its response (mirroring real Azure, which never returns these), the overlay treated the dropped key as merely "unmodeled" and reflected the caller's secret verbatim — on create **and every subsequent GET** (persisted in-memory keyed by resource id).

Concrete leak: `armsql` PUT `Microsoft.Sql/servers` with `administratorLoginPassword` → `toARMServer` correctly omits it → overlay re-added it → GET echoed the password.

## Fix

Added a conservative **write-only / sensitive key denylist** to the overlay. Denylisted keys are never captured, persisted, or echoed. Match is **case-insensitive** and applies at **every nesting depth** (the overlay descends into nested `properties.*`). Wholesale-captured objects (a whole subtree the handler dropped) are sanitized too, so a nested secret can't slip past the per-key skip.

### Denylist entries (each verified against the owning handler's to-ARM function)

| Key | Services | Justification |
|-----|----------|---------------|
| `administratorLoginPassword` | `Microsoft.Sql/servers`, `Microsoft.DBforMySQL/flexibleServers`, `Microsoft.DBforPostgreSQL/flexibleServers`, Cosmos DB for PostgreSQL (`serverGroupsv2`) | Each handler accepts it on write and its `toARM*` omits it; real Azure never returns it on read. |
| `adminPassword` | `Microsoft.Compute/virtualMachines` (osProfile) | The VM `osProfile` shape models no password field, so the handler drops it; real Azure returns null for it. |
| `initialCassandraAdminPassword` | `Microsoft.DocumentDB/cassandraClusters` | Handler accepts it on write; `toARMCluster` omits it; real Azure never returns it. |

The list is deliberately conservative: a key is included only when (a) real Azure genuinely omits it on read AND (b) a handler relies on that omission — so it never suppresses a field a handler legitimately returns. (The overlay only ever *adds* keys a handler dropped, so denylisting never affects a field a handler itself echoes.)

## Blast radius

This overlay is cross-cutting across all Azure ARM handlers. Verified:
- The overlay still reflects legitimate non-secret unmodeled properties (SQL database `maxSizeBytes`, VM nested leaves, flex `maintenanceWindow`) — regression guard test included.
- Broad Azure server + provider test pass green; no handler that legitimately echoed a denylisted key regresses (none do — all omit them).

## Tests (hand-rolled, matching `server/azure` style)

- SQL server password not echoed on create/GET/PATCH; non-secret fields still returned.
- MySQL + PostgreSQL flexible-server password not echoed.
- VM nested `osProfile.adminPassword` not echoed; sibling non-secret leaf still round-trips.
- Managed Cassandra `initialCassandraAdminPassword` not echoed.
- Regression guard: non-secret `maxSizeBytes` still reflected.

## Gates

build · vet · `go test ./server/azure/... ./providers/azure/...` · `-race ./server/azure/sql/...` · root test · gofmt · golangci-lint (0 issues) · `go generate` zero diff · compatgen + `docs/compat` zero diff — all green.

### DIFF
--- a/server/azure/echo_properties.go
+++ b/server/azure/echo_properties.go
@@ -327,10 +327,70 @@ func captureUnmodeled(
 	return fresh
 }
 
+// writeOnlyProperty reports whether an ARM request-body property key is a
+// write-only secret that real Azure accepts on write but never returns on a
+// read. Such a key must never be captured by the overlay or echoed back:
+// because the owning handler deliberately omits it from its response (mirroring
+// Azure), the generic overlay would otherwise treat it as an unmodeled property
+// and reflect the caller's secret on the create response and on every later GET
+// — a credential leak. Matched case-insensitively at any nesting depth, since
+// missingProperties descends into nested objects.
+//
+// Every entry is a value some Azure handler in this server accepts on write and
+// drops from its to-ARM output, verified against how that handler models the
+// key:
+//   - administratorLoginPassword: the admin password for Microsoft.Sql/servers
+//     (server/azure/sql), Microsoft.DBforMySQL/flexibleServers
+//     (server/azure/mysqlflex), Microsoft.DBforPostgreSQL/flexibleServers
+//     (server/azure/postgresflex) and Cosmos DB for PostgreSQL clusters
+//     (server/azure/cosmospostgresql). Each handler's toARM* omits it.
+//   - adminPassword: the VM local-admin password under osProfile for
+//     Microsoft.Compute/virtualMachines (server/azure/virtualmachines). The
+//     handler's osProfile shape carries no password field, so it is dropped.
+//   - initialCassandraAdminPassword: the seed admin password for managed
+//     Cassandra clusters (server/azure/managedcassandra). toARMCluster omits it.
+//
+// The list is deliberately conservative: a key belongs here only when real
+// Azure genuinely omits it on read AND a handler relies on that omission, so it
+// never suppresses a field a handler legitimately returns.
+func writeOnlyProperty(key string) bool {
+	switch strings.ToLower(key) {
+	case "administratorloginpassword", "adminpassword", "initialcassandraadminpassword":
+		return true
+	default:
+		return false
+	}
+}
+
+// sanitizeUnmodeled deep-copies a captured request value with every write-only
+// secret key (writeOnlyProperty) removed at every depth. It guards the wholesale
+// capture path in missingProperties: when the handler drops an entire object,
+// that object is preserved as-is, so a secret nested inside it would otherwise
+// slip past the per-key denylist. A non-map value passes through unchanged.
+func sanitizeUnmodeled(v any) any {
+	m, ok := v.(map[string]any)
+	if !ok {
+		return v
+	}
+
+	out := make(map[string]any, len(m))
+
+	for k, val := range m {
+		if writeOnlyProperty(k) {
+			continue
+		}
+
+		out[k] = sanitizeUnmodeled(val)
+	}
+
+	return out
+}
+
 // missingProperties returns the entries of req that resp does not already carry,
 // descending into nested objects so an unmodeled leaf under a modeled parent is
 // still captured. A key present in both as scalars is considered modeled (the
-// handler owns it) and is omitted.
+// handler owns it) and is omitted. Write-only secret keys (writeOnlyProperty)
+// are skipped at every level so they are never captured, persisted, or echoed.
 func missingProperties(req, resp map[string]any) map[string]any {
 	if len(req) == 0 {
 		return nil
@@ -339,9 +399,18 @@ func missingProperties(req, resp map[string]any) map[string]any {
 	out := map[string]any{}
 
 	for k, reqVal := range req {
+		if writeOnlyProperty(k) {
+			continue
+		}
+
 		respVal, present := resp[k]
 		if !present {
-			out[k] = reqVal
+			// A wholly-unmodeled object is captured verbatim, so strip any
+			// write-only secret nested inside it — the per-key skip above only
+			// covers keys the loop visits directly, not those buried in a
+			// subtree the handler dropped entirely (e.g. a VM osProfile the
+			// response omits, carrying adminPassword).
+			out[k] = sanitizeUnmodeled(reqVal)
 			continue
 		}
 
--- a/server/azure/echo_properties.go
+++ b/server/azure/echo_properties.go
@@ -336,49 +336,60 @@ func captureUnmodeled(
 // — a credential leak. Matched case-insensitively at any nesting depth, since
 // missingProperties descends into nested objects.
 //
-// Every entry is a value some Azure handler in this server accepts on write and
-// drops from its to-ARM output, verified against how that handler models the
-// key:
-//   - administratorLoginPassword: the admin password for Microsoft.Sql/servers
-//     (server/azure/sql), Microsoft.DBforMySQL/flexibleServers
-//     (server/azure/mysqlflex), Microsoft.DBforPostgreSQL/flexibleServers
-//     (server/azure/postgresflex) and Cosmos DB for PostgreSQL clusters
-//     (server/azure/cosmospostgresql). Each handler's toARM* omits it.
-//   - adminPassword: the VM local-admin password under osProfile for
-//     Microsoft.Compute/virtualMachines (server/azure/virtualmachines). The
-//     handler's osProfile shape carries no password field, so it is dropped.
-//   - initialCassandraAdminPassword: the seed admin password for managed
-//     Cassandra clusters (server/azure/managedcassandra). toARMCluster omits it.
-//   - password: a Cosmos DB for PostgreSQL role password (toARMRole,
-//     server/azure/cosmospostgresql, drops it) and a Container Instances
-//     imageRegistryCredentials[].password (server/azure/containerinstances does
-//     not model the array at all, so it is captured verbatim — inside an array,
-//     hence the []any recursion in sanitizeUnmodeled). Real Azure never returns
-//     either. The one place a `password` is legitimately returned is the
-//     Container Instances exec action response, but that is a bare
-//     {webSocketUri, password} object with no id/properties, so the overlay
-//     (which only rewrites ARM resources and only ever touches keys inside a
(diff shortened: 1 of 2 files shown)


NitinKumar004 (commented):
**Verdict: CHANGES REQUESTED** — the approach is right and the three denylisted secrets are correctly suppressed, but the same shared overlay still leaks several other write-only secrets that this PR does not cover. Since the overlay fronts every ARM handler, a per-key denylist has to be complete to be a security fix. All findings below were reproduced against a booted full Azure server (httptest) via raw ARM REST on the PR head SHA (5242cd9).

### Blocking: write-only secrets still echoed on create + every GET

1. **Cosmos DB for PostgreSQL role password** — `Microsoft.DBforPostgreSQL/serverGroupsv2/{c}/roles/{r}`. `toARMRole` (server/azure/cosmospostgresql/types.go:313) returns only `provisioningState`, dropping the caller-supplied `properties.password`. Key `password` is not denylisted → overlay captures and echoes it on the create response and on GET. Real Azure never returns a role password.

2. **Container Instances registry password** — `Microsoft.ContainerInstance/containerGroups`. `containerGroupProperties` (server/azure/containerinstances/types.go:27) does not model `imageRegistryCredentials` at all, so the whole array is captured verbatim, including `imageRegistryCredentials[].password`. Leaked on create + GET.

3. **Notification Hubs PNS credentials** — `Microsoft.NotificationHubs/namespaces/{ns}/notificationHubs/{h}`. `toHubJSON` (server/azure/notificationhubs/types.go:216) returns only `name`+`registrationTTL`; the credential blocks (`gcmCredential`/`apnsCredential`/`wnsCredential`/`admCredential`/`baiduCredential`/`mpnsCredential`, carrying `googleApiKey`, `secretKey`, `clientSecret`, APNS token/cert key, …) are dropped from the hub response and only meant to be served — separately — by `GetPnsCredentials`. The overlay captures the credential subtree and echoes it on the generic hub GET.

4. **AKS service-principal secret** — `Microsoft.ContainerService/managedClusters`. `armManagedClusterProperties` (server/azure/aks/types.go:44) omits `servicePrincipalProfile`, so `properties.servicePrincipalProfile.secret` is captured verbatim and echoed. (`windowsProfile.adminPassword` here is incidentally covered because `adminPassword` is on the denylist, but `secret` is not.)

### Blocking (correctness): `sanitizeUnmodeled` does not descend into arrays

`sanitizeUnmodeled` only recurses into `map[string]any`; a `[]any` is returned unchanged, so a denylisted secret inside an array element survives (reproduced: `adminPassword` nested in an unmodeled array round-tripped). The PR comment/tests claim "an array of objects each containing a secret → all stripped", but that path is not implemented. This also means simply adding `password`/`secret` to the denylist would still not fix finding #2, since `imageRegistryCredentials` is an array — the array recursion must be added too.

### What is correct and should be kept
- The three denylisted keys are fully suppressed: SQL/MySQL-flex/PostgreSQL-flex/Cosmos-PG-cluster `administratorLoginPassword`, VM `osProfile.adminPassword` (wholesale-subtree case verified — dropped osProfile, secret stripped, sibling leaf preserved), managed-Cassandra `initialCassandraAdminPassword`.
- Non-secret regression is clean: `maxSizeBytes` and sibling unmodeled leaves still round-trip; case-insensitive exact-key match (no substring false-positives); the deep-copy avoids cross-request map aliasing; `-race` on the package is clean.

### Suggested direction
A static per-key denylist is hard to keep complete for a cross-cutting overlay. Consider either (a) an allowlist/opt-in per handler for which unmodeled props may be echoed, or (b) at minimum extend the denylist to `password`, `secret`, and the PNS credential keys, AND make `sanitizeUnmodeled` recurse through arrays. Either way, add e2e leak tests for the four resources above.

NitinKumar004 (commented):
Delta re-review (head 4937cbe5, clean-built from the pushed tree).

## Gates — all green
build, vet, gofmt clean; `go test ./server/azure/... ./providers/azure/...` pass; `-race ./server/azure/...` clean; root `go test .` pass; `go generate ./...` zero diff; compatgen green + `docs/compat/` zero diff; golangci-lint `--new-from-rev=origin/development` 0 new issues.

## Confirmed correct
- **All seven targeted secrets suppressed** on create + GET + PATCH (SQL/MySQL-flex/PG-flex/Cosmos-PG admin password, VM osProfile.adminPassword, Cassandra initialCassandraAdminPassword, Cosmos-PG role password, CI imageRegistryCredentials[].password, Notification Hubs PNS credentials, AKS servicePrincipalProfile.secret). Verified independently, not just via the added tests.
- **Array recursion correct + non-destructive**: `sanitizeUnmodeled` strips the secret inside `imageRegistryCredentials[]` and a 2-level array->map->map secret while preserving non-secret siblings. It allocates a fresh map/slice at every level, so no request-owned map/slice is aliased into the overlay store.
- **Matcher is exact-key, case-insensitive** (`switch strings.ToLower(key)`): `passwordPolicy`, `secretName`, `clientSecretSetting`, `mySecretPassword` are all preserved — no substring false-positives.
- **No over-suppression on reads**: no response/`toARM*` struct emits `json:"secret"`, and the only `json:"password"` is the Container Instances exec-action `{webSocketUri,password}`, which has no `id`/`properties` and so bypasses the overlay entirely. The denylist's `password`/`secret` entries do not strip any legitimately read-returned field.

## Blocking: one write-only secret still leaks
`Microsoft.ContainerService/managedClusters` — `properties.aadProfile.serverAppSecret`.

The AKS handler's `armManagedClusterProperties` does not model `aadProfile`, so the whole block is captured wholesale via the missing-key path and deep-copied by `sanitizeUnmodeled`. `serverAppSecret` matches none of the denylist keys (`secret` is an exact match, `serverappsecret` is not), so it is reflected on the create response and on every later GET. `serverAppSecret` (legacy AAD integration) is write-only in real Azure — never returned on GET — so this is the same class of leak the PR fixes.

Reproduced live against the booted server: a PUT with `aadProfile.serverAppSecret` round-trips the value on GET. (`windowsProfile.adminPassword` in the same block is correctly stripped by the `adminpassword` entry.)

Fix: add `serverappsecret` to `writeOnlyProperty`. Recommend also adding `clientsecret` defensively and re-sweeping `*Secret`-suffixed write-only inputs on wholly-unmodeled blocks, since the denylist is a blocklist and this block is captured verbatim.

NitinKumar004 (commented):
**Final security re-review — verdict: MERGE** (reviewed pushed head `bf0cb33a`, clean-built).

**Gates (GOMAXPROCS=3, -p=3):** build ✓ · vet ✓ · `go test ./server/azure/... ./providers/azure/...` ✓ · `-race ./server/azure/...` ✓ · root `go test .` ✓ · gofmt clean ✓ · `go generate ./...` zero diff ✓ · `go run ./internal/compatgen` exit 0 + zero `docs/compat/` diff ✓ · golangci-lint `--new-from-rev=origin/development ./server/azure/...` → 0 issues.

**serverAppSecret + all prior secrets suppressed.** The suffix rule (`HasSuffix(lower,"password") || HasSuffix(lower,"secret")`) plus the 6 `*Credential` exact keys covers AKS `aadProfile.serverAppSecret`/`clientSecret` and `servicePrincipalProfile.secret`, SQL/MySQL-flex/PG-flex/Cosmos-PG `administratorLoginPassword`, VM `osProfile.adminPassword`, Cassandra `initialCassandraAdminPassword`, Cosmos-PG role `password`, Container Instances `imageRegistryCredentials[].password`, and the Notification Hubs PNS credential blocks. Tests exercise each on create+GET(+PATCH).

**Suffix is a true endsWith, not contains.** `secretName`/`passwordPolicy`/`secretUri`/`clientSecretSetting`/`serverAppID` and the plural `secrets`/`accessKeys` outputs all round-trip (verified by `TestEchoSuffixRuleIsEndsWithNotContains` and by grep — `"secrets"` does not end with `secret`).

**Additive-only invariant holds — over-suppression is structurally impossible.** `writeOnlyProperty` is only ever applied to captured *request* values (`missingProperties`/`sanitizeUnmodeled`), never to `respProps`. `missingProperties` only emits keys absent from the response (or nested deltas); `mergeProperties` copies the handler's response verbatim and only *adds* unmodeled keys where absent (never overwrites/removes). So a handler that legitimately *returns* a secret-suffixed field is untouched. Independent grep confirms no returned `server/azure` field ends in password/secret except the Container Instances exec-action `{webSocketUri, password}` — a bare object with no `id`/`properties`, so `rewrite` bails before merging, and it's a POST action carrying no `properties` to capture.

**Missed-leak hunt on the remaining surface (non-suffix write-only inputs in unmodeled subtrees): none found.** Swept every implemented Azure ARM handler. `connectionString`/`primaryKey`/`secondaryKey`/`accessKeys`/`sharedKey` are all action *outputs* (listKeys/listConnectionStrings/regenerateKeys — server-generated, never accepted in a create/update `properties` block, and those action responses carry no `id`/`properties` so the overlay skips them). Data-plane secrets (keyvault JWK, databricks PAT/token, search/AI keys) live on non-`/subscriptions/` paths the middleware never engages. ServiceBus/EventGrid keys are generated, not PUT inputs. No Data Factory / APIM / Logic Apps (the usual `protectedSettings`/`connectionString` write-only surfaces) are implemented, and no VM/VMSS extension `protectedSettings` handler exists. eventgrid event-subscriptions echo `properties` verbatim, but that is the handler's own doing (nothing is dropped, so the overlay never captures it) — outside this PR's scope.

**One non-blocking parity note (not a credential leak):** VM `osProfile` is omitted from the VM response entirely, so the overlay captures the whole subtree; `adminPassword` is correctly stripped, but `customData` (write-only in real Azure) round-trips back. It is the caller's own provisioning payload returned to the same caller — no cross-boundary exposure — and is pre-existing overlay behavior orthogonal to the secret-suffix fix. Worth a follow-up for strict read-parity, not a blocker here.

No still-leaking write-only credential and no over-suppressed returned field. The overlay secret-suppression fix is complete.

PR #716, opened 2026-08-26
fix(iam): embed entity Path in ARN on create; guard tag-map reads

NitinKumar004:
Fixes two AWS IAM bugs in `providers/aws/iam/iam.go`.

## Bug 1 (MED) — entity Path dropped from ARN on create
An IAM entity created with a non-default `Path` did not embed the path in its ARN. `CreateUser/CreateRole/CreatePolicy/CreateGroup/CreateInstanceProfile` built the ARN resource as `"<type>/"+Name`, so a path-set entity got `.../user/bob` instead of `.../user/div/sub/bob`. The `Path` field was returned correctly; only the ARN was wrong. This broke Terraform `aws_iam_{user,role,policy}.arn` and any trust/resource policy interpolating the ARN whenever a path was set.

Fix: build the resource as `"<type>/" + strings.TrimPrefix(path, "/") + Name` using the normalized (leading+trailing slash) path — matching `service_linked_role.go`. Default path `/` yields the exact old ARN (regression-safe).

## Bug 2 (LOW, concurrency) — tag map read without lock
User/Role `Tags` maps are mutated under `m.mu` (TagUser/TagRole) but were read via `toUserInfo`/`toRoleInfo` in GetUser/ListUsers/GetRole/ListRoles without holding `m.mu` — a data race on the inner map. Fixed by taking `m.mu.RLock()` on those read paths (lock ordering matches the rest of the file; `go test -race` now clean).

## Tests
- `providers/aws/iam/arn_path_test.go` — all five entities with `Path="/div/sub/"` embed the path; default path unchanged; a `-race` concurrent Tag+Get loop (confirmed to catch the race when the guard is removed).
- `server/aws/iam/arn_path_test.go` — real `aws-sdk-go-v2/service/iam` e2e over the booted wire server asserting the ARN for all five entities, plus a default-path regression guard.

## Gates (GOMAXPROCS=3)
build, vet, `go test` on server/providers/services iam, `go test -race` on providers/aws/iam, gofmt, golangci-lint (0 new), `go generate` zero-diff, compatgen + `docs/compat` zero-diff — all green.

Not included (tracked separately): UpdateUser/UpdateGroup rename, account alias CRUD, login profile CRUD, policy/instance-profile tagging.

### DIFF
--- a/providers/aws/iam/arn_path_test.go
+++ b/providers/aws/iam/arn_path_test.go
@@ -1,0 +1,131 @@
+package iam
+
+import (
+	"context"
+	"sync"
+	"testing"
+
+	"github.com/stackshy/cloudemu/v2/services/iam/driver"
+)
+
+const nestedPath = "/div/sub/"
+
+// TestCreateEntityARNEmbedsPath verifies that a non-default Path is folded into
+// the entity ARN (arn:aws:iam::123456789012:user/div/sub/bob), while the default
+// path "/" leaves the ARN unchanged (regression guard).
+func TestCreateEntityARNEmbedsPath(t *testing.T) {
+	ctx := context.Background()
+
+	makeUser := func(m *Mock, path string) (string, error) {
+		info, err := m.CreateUser(ctx, driver.UserConfig{Name: "bob", Path: path})
+		if err != nil {
+			return "", err
+		}
+		return info.ARN, nil
+	}
+	makeRole := func(m *Mock, path string) (string, error) {
+		info, err := m.CreateRole(ctx, driver.RoleConfig{Name: "r1", Path: path})
+		if err != nil {
+			return "", err
+		}
+		return info.ARN, nil
+	}
+	makePolicy := func(m *Mock, path string) (string, error) {
+		info, err := m.CreatePolicy(ctx, driver.PolicyConfig{Name: "p1", Path: path, PolicyDocument: makePolicyDoc(nil)})
+		if err != nil {
+			return "", err
+		}
+		return info.ARN, nil
+	}
+	makeGroup := func(m *Mock, path string) (string, error) {
+		info, err := m.CreateGroup(ctx, driver.GroupConfig{Name: "g1", Path: path})
+		if err != nil {
+			return "", err
+		}
+		return info.ARN, nil
+	}
+	makeProfile := func(m *Mock, path string) (string, error) {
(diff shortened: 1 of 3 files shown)


NitinKumar004 (commented):
**Verdict: MERGE** — verified against the pushed head `cdab32c5`.

**ARN path fold (all 5 sites).** CreateUser/CreateRole/CreatePolicy/CreateGroup/CreateInstanceProfile all build the resource as `"<type>/" + strings.TrimPrefix(path, "/") + Name` off the normalized `path` (empty → `/`). `Path="/div/sub/"` yields `.../{type}/div/sub/{name}`; default path yields the exact old flat ARN — regression-safe, no double/missing slash on well-formed input. The instance-profile reorder (path normalization moved above id/arn) changes nothing else. Real `aws-sdk-go-v2/iam` e2e over the wire server confirms all five ARNs plus the default-path regression guard.

**Concurrency (the risk) — no deadlock.**
- No reentrant RLock: GetUser/ListUsers/GetRole/ListRoles each take `m.mu.RLock()` once and, while holding it, call only `toUserInfo`/`toRoleInfo` → `copyTags`, all pure with no re-lock. No re-entrant acquisition, so no writer-contention hang.
- Lock ordering sound: the read paths do `memstore Get/All` (which acquires+releases the memstore lock internally) and only then take `m.mu.RLock()` — the memstore lock is never held across the `m.mu` acquisition. `TagRole` does `m.mu.Lock` then `memstore.Get` (the inverse nesting), but since no path ever holds memstore while blocking on `m.mu`, there is no lock cycle. That `TagRole` pattern is pre-existing and untouched here.
- The added `TestConcurrentTagAndGet` genuinely exercises the race: removing the four guards makes it fail under `-race` with DATA RACE; with the guards it passes clean.

**Gates (GOMAXPROCS=3, pushed SHA):** build, vet, `go test` (server/providers/services iam), `go test -race` (providers+server iam), root integration test — all pass. gofmt clean. `golangci-lint --new-from-rev=origin/development` on both changed packages: 0 new (the `hugeParam` on CreateRole:237 is pre-existing — present in a full run, absent from the new-only run). `go generate ./...` zero diff; `compatgen` + `git diff docs/compat/` zero diff. Diff limited to the 3 named files.

Minor, non-blocking: path normalization enforces the leading slash but not a trailing one, so a caller passing `Path="/div/sub"` (no trailing slash) would produce `.../user/div/subbob`. Real IAM requires the trailing slash and cloudemu does not validate paths anywhere (pre-existing), and this mirrors the `service_linked_role.go` convention — the happy path is correct.

PR #717, opened 2026-08-26
fix(vpc): round-trip CreateVpc InstanceTenancy instead of hardcoding default

NitinKumar004:
## Summary

`CreateVpc` with `InstanceTenancy=dedicated` was accepted but `DescribeVpcs` always reported `default`, causing perpetual Terraform drift on `aws_vpc.instance_tenancy`. Root cause: the wire `createVpc` never read `InstanceTenancy` from the request, `toVpcXML` hardcoded `"default"`, and the shared networking driver's `VPCConfig`/`VPCInfo` lacked the field.

## Changes

- **Driver** (`services/networking/driver`): add optional `InstanceTenancy` to `VPCConfig` and `VPCInfo`. Azure/GCP/OCI use keyed struct literals and leave it unset — no behavior change.
- **Provider** (`providers/aws/vpc`): `CreateVPC` reads and stores tenancy, defaulting to `default` when empty and validating it is `default` or `dedicated`. Real EC2 `CreateVpc` rejects `host` (a placement-only tenancy) and any other value with `InvalidParameterValue`, so those are rejected too. Returned on `VPCInfo`.
- **Server** (`server/aws/ec2/vpc.go`): `createVpc` reads `InstanceTenancy` from the form; `toVpcXML` emits the stored value.

## Tests

Real aws-sdk-go-v2 EC2 client against a booted in-process server, plus provider-level:
- `dedicated` create round-trips to `dedicated` on describe
- unset defaults to `default`
- `host` and a bogus value are rejected with `InvalidParameterValue`

## Deferral

`ModifyVpcTenancy` (dedicated→default) is not implemented and is out of scope here; this PR covers the create+describe round-trip only.

### DIFF
--- a/providers/aws/vpc/vpc.go
+++ b/providers/aws/vpc/vpc.go
@@ -43,6 +43,14 @@ const (
 	defaultSGDescription = "default VPC security group"
 )
 
+// VPC instance tenancy. Real EC2 CreateVpc accepts only "default" and
+// "dedicated"; "host" is a valid instance placement tenancy but is rejected by
+// CreateVpc, and any other value is an InvalidParameterValue.
+const (
+	tenancyDefault   = "default"
+	tenancyDedicated = "dedicated"
+)
+
 // Compile-time checks. The optional capabilities are asserted too: without
 // this a signature drifting out of shape would silently stop satisfying the
 // interface and every call would answer InvalidAction at runtime instead of
@@ -176,6 +184,7 @@ type vpcData struct {
 	EnableDNSSupport   bool
 	EnableDNSHostnames bool
 	DhcpOptionsID      string
+	InstanceTenancy    string
 }
 
 type subnetData struct {
@@ -279,6 +288,11 @@ func (m *Mock) CreateVPC(_ context.Context, cfg driver.VPCConfig) (*driver.VPCIn
 		return nil, err
 	}
 
+	tenancy, err := validateInstanceTenancy(cfg.InstanceTenancy)
+	if err != nil {
+		return nil, err
+	}
+
 	id := idgen.GenerateID("vpc-")
 	tags := copyTags(cfg.Tags)
 
@@ -289,6 +303,7 @@ func (m *Mock) CreateVPC(_ context.Context, cfg driver.VPCConfig) (*driver.VPCIn
 		Tags:      tags,
 		// EC2 defaults DNS support on and DNS hostnames off for a new VPC.
 		EnableDNSSupport: true,
+		InstanceTenancy:  tenancy,
 	}
 	m.vpcs.Set(id, v)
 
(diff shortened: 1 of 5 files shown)


NitinKumar004 (commented):
Reviewed at head SHA `2b798b00be89cc73469f874bbe149cc0c6473477` in an isolated worktree.

**Gates (GOMAXPROCS=3):**
- `go build ./...` clean
- `go vet ./...` clean
- `go test -p=3 ./server/aws/ec2/... ./providers/aws/vpc/... ./services/networking/... ./providers/azure/vnet/... ./providers/gcp/vpc/... ./providers/oci/vcn/...` — all pass
- `go test -race -p=3 ./providers/aws/vpc/...` — pass
- `go test -p=3 ./...` (full suite) — 294 packages ok, zero FAIL
- `gofmt -l` — clean on the 5 changed files (4 pre-existing unrelated unformatted files in the tree, untouched by this PR)
- `golangci-lint --new-from-rev=origin/development` on the 3 changed packages — 0 issues
- `go run ./internal/compatgen` — exit 0, `git diff --exit-code docs/compat/` — zero diff

**Functional verification (real aws-sdk-go-v2 EC2 client against a booted server, via the PR's own `vpc_tenancy_test.go` + independent re-run):**
- `CreateVpc(InstanceTenancy=dedicated)` → `DescribeVpcs` reports `dedicated`
- Unset `InstanceTenancy` → defaults to `default`
- `InstanceTenancy=host` and a bogus value → both rejected with `InvalidParameterValue`. Confirmed against the AWS SDK's `Tenancy` enum (`default`/`dedicated`/`host`) and AWS docs: `CreateVpc` only accepts `default`/`dedicated`; `host` is a legitimate tenancy value elsewhere (instance placement) but is correctly rejected here, not over-rejected.

**Shared driver inertness (Azure/GCP/OCI):** `grep -rln "InstanceTenancy" --include='*.go' .` returns only the 5 files in the PR diff — no reference in `providers/azure/vnet`, `providers/gcp/vpc`, or `providers/oci/vcn`. All three build `driver.VPCInfo{...}`/`driver.VPCConfig{...}` with keyed struct literals, so the new field is simply zero-valued there; their package tests pass unchanged. Additive-optional, zero behavior change confirmed.

**Error mapping:** `validateInstanceTenancy` returns `cerrors.InvalidArgument`, which `writeErrWithNotFound` (via the generic `cerrors.IsInvalidArgument` branch in `handler.go`) already maps to `InvalidParameterValue` — same mechanism used elsewhere in this file (e.g. `IamInstanceProfile`, filter validation). No new wire error-mapping code was needed, and none was added.

**Scope:** Confirmed no `ModifyVpcTenancy` handler exists anywhere in the tree — the PR's stated deferral (create+describe only) doesn't leave a half-wired action.

No blocking issues found. Nit (non-blocking): `tenancyDefaultXML` in `server/aws/ec2/vpc.go` duplicates the literal `"default"` also defined as `tenancyDefault` in `providers/aws/vpc/vpc.go` — intentional per the comment (covers `VPCInfo` records with an empty tenancy from outside the AWS wire path, e.g. old persisted snapshots), so this is fine as-is.

PR #718, opened 2026-08-26
fix(ec2): apply request Filters to 5 VPC Describe handlers (NetworkAcls/NatGateways/Addresses/VpcPeeringConnections/DhcpOptions)

NitinKumar004:
## Summary

Five EC2 `Describe*` handlers read only their resource-id list and never applied request `Filters`, so `--filter` queries and Terraform data-source lookups returned **every** resource across all VPCs instead of the matching subset. This mirrors the existing working pattern in `describeSubnets` for each:

| Handler | Filters added |
|---|---|
| `DescribeNetworkAcls` | `vpc-id`, `default`, `association.subnet-id`, `network-acl-id`, `tag:*`, `tag-key` |
| `DescribeNatGateways` | `vpc-id`, `subnet-id`, `state`, `nat-gateway-id`, `tag:*` — **plus MaxResults/NextToken pagination** |
| `DescribeAddresses` | `domain`, `instance-id`, `allocation-id`, `public-ip`, `association-id`, `tag:*` |
| `DescribeVpcPeeringConnections` | `status-code`, `requester-vpc-info.vpc-id`, `accepter-vpc-info.vpc-id`, `vpc-peering-connection-id`, `tag:*` |
| `DescribeDhcpOptions` | `key`, `value`, `dhcp-options-id`, `tag:*` |

## Approach

- Reuses the shared `awsquery.Filters` parser and the `tagFilterMatch` tag helper — no divergent hand-rolled matcher.
- Two small generic helpers added to `networking_common.go` (`validateNetworkingFilters`, `matchNetworkingFilters`) so the five handlers do not each duplicate the validate/match loop. Matchers report `(matched, known)`; an unrecognized filter name is rejected with `InvalidParameterValue`, matching real EC2 (and the existing subnet/route-table handlers).
- Multiple filters AND together; multiple values within one filter OR; the explicit id list still works and combines with filters.
- Driver and provider layers are untouched — pure server-layer filter additions. Only DescribeNatGateways gains pagination (mirroring the sibling paginating describe handlers).

## Tests

Real `aws-sdk-go-v2/ec2` tests against the booted server, one per handler: resources created in 2+ VPCs / with distinct attributes, then a filtered Describe returns only the matching subset, a non-matching filter returns empty, and an id-list still resolves. DescribeNatGateways adds a pagination test asserting a NextToken and a clean multi-page walk.

## Gates

`go build`, `go vet`, `go test ./server/aws/ec2/... ./providers/aws/...`, `go test -race ./server/aws/ec2/...`, `gofmt -l`, `golangci-lint` (0 new), `go generate` (zero diff), `compatgen` + `docs/compat` (zero diff) — all green.

### DIFF
--- a/server/aws/ec2/address.go
+++ b/server/aws/ec2/address.go
@@ -112,9 +112,20 @@ func (h *Handler) describeAddresses(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	filters := awsquery.Filters(r.Form)
+	if err := validateNetworkingFilters(filters, addressFilterMatch); err != nil {
+		writeVPCErr(w, err)
+
+		return
+	}
+
 	out := make([]addressXML, 0, len(eips))
 
 	for i := range eips {
+		if !matchNetworkingFilters(&eips[i], filters, addressFilterMatch) {
+			continue
+		}
+
 		addr := addressXML{
 			PublicIP:           eips[i].PublicIP,
 			AllocationID:       eips[i].AllocationID,
@@ -138,6 +149,26 @@ func (h *Handler) describeAddresses(w http.ResponseWriter, r *http.Request) {
 	awsquery.WriteXMLResponse(w, describeAddressesResponseXML{
 		Xmlns: awsquery.Namespace, RequestID: awsquery.RequestID, AddressSet: out,
 	})
+}
+
+// addressFilterMatch reports whether e satisfies filter f and whether f is a
+// filter DescribeAddresses recognizes. Every allocation is a VPC domain address
+// (EC2-Classic is retired), so a domain filter only matches "vpc".
+func addressFilterMatch(e *netdriver.ElasticIP, f awsquery.Filter) (matched, known bool) {
+	switch f.Name {
+	case "allocation-id":
+		return containsString(f.Values, e.AllocationID), true
+	case "public-ip":
+		return containsString(f.Values, e.PublicIP), true
+	case "association-id":
+		return containsString(f.Values, e.AssociationID), true
+	case "instance-id":
+		return containsString(f.Values, e.InstanceID), true
+	case "domain":
+		return containsString(f.Values, domainVPC), true
+	default:
+		return tagFilterMatch(f.Name, f.Values, e.Tags)
+	}
 }
 
 // allocationIDForPublicIP resolves a public address back to its allocation id.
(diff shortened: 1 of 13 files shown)


NitinKumar004 (commented):
Reviewed at head ef9bf159 (clean checkout of the pushed tree). The filter work itself is correct and well-factored; one lint gate blocks.

## Blocking: 1 new goconst finding from the collateral refactor
`golangci-lint run --new-from-rev=origin/development ./server/aws/ec2/...` reports:

```
server/aws/ec2/route_table.go:338:24: string `association.main` has 2 occurrences, make it a constant (goconst)
```

On `development`, line 338 read `"association.subnet-id", "association.main":`. The goconst refactor here extracted `"association.subnet-id"` -> `filterAssocSubnetID` but left its line-mate `"association.main"` as a literal (it also appears at line 375). Because the PR edited that exact line, `--new-from-rev` now attributes the remaining `association.main` (2 occurrences) as a NEW finding, so the "0 new lint" gate fails with 1 issue. Fix is to finish the same refactor: extract `"association.main"` to a constant (e.g. `filterAssocMain`) used at both 338 and 375, mirroring the `filterAssocSubnetID` extraction already done on those lines. Pure literal->constant, no behavior change.

## Everything else verified green
- Build / vet / `go test -p=3 ./server/aws/ec2/... ./providers/aws/...` / `go test -race ./server/aws/ec2/...` / gofmt -l: all clean.
- `go generate ./...` and `compatgen` + `docs/compat/`: zero diff.
- Real aws-sdk-go-v2 verification (added throwaway tests, then removed): unknown filter name -> InvalidParameterValue confirmed for all five handlers (NAT/NACL/Addresses/Peering/DHCP); multi-value OR within one filter; multi-filter AND; explicit id-list combines with a filter (id in vpc B AND vpc-id=A -> empty); NAT pagination combined with vpc-id filter walks the matching subset with no dup/skip and no filter leak across pages.
- Shared helper design is sound: `validateNetworkingFilters` probes a zero-value `T` and consults only the matcher's `known` flag, so the accepted-filter set cannot drift from what each `xFilterMatch` actually honors, and an unknown name is rejected before any match runs. Each matcher returns `known=true` for a recognized name independent of item state (tag prefix detection doesn't depend on map contents), so zero-value validation is correct.
- Driver/provider layers untouched; server-layer only. No regression in the existing NACL/NAT/Addresses/Peering/DHCP describe tests or the broader ec2 suite.

Once `association.main` is extracted (or the goconst finding is otherwise resolved), this is a clean merge.

PR #719, opened 2026-08-26
fix(aws-vpc): materialize instance primary ENI so subnet/SG delete guards fire

NitinKumar004:
## Summary

Fixes two AWS VPC delete-guard bugs: `DeleteSubnet` and `DeleteSecurityGroup` wrongly succeeded while a running instance used the subnet / security group.

Root cause: `RunInstances` never materialized an instance's primary (eth0) network interface in the VPC provider's ENI store, so the existing `eniInSubnet` and `securityGroupInUse` guards — which scan that store — saw nothing to block on.

## What changed (approach A — real-EC2-faithful)

- `RunInstances` now materializes the primary eth0 ENI real EC2 creates for every instance: in the instance's subnet, carrying its security groups, at device index 0, attached to the instance, delete-on-termination.
- `TerminateInstances` (and mid-batch launch rollback) releases that ENI.
- Wired compute -> VPC via a new EC2 `Networking` hook (`SetNetworking(p.VPC)`) in the provider factory, mirroring the existing `SetSubnetResolver` pattern. No import cycle — the hook interface is defined on the consumer (EC2) side.

With the primary ENI present, the existing guards work unchanged:
- `DeleteSubnet` with a running instance -> `DependencyViolation`; terminate -> succeeds.
- `DeleteSecurityGroup` referenced by a running instance -> `DependencyViolation`; terminate -> succeeds.

Side effect (correct AWS behavior): `DescribeNetworkInterfaces` now lists an instance's eth0 ENI.

## Tests

New real-SDK e2e tests in `server/aws/ec2/instance_delete_guards_test.go`:
- DeleteSubnet blocked by a running instance, then succeeds after terminate.
- DeleteSecurityGroup blocked by a running instance, then succeeds after terminate.
- RunInstances materializes the eth0 ENI (in-use, device index 0, attached to the instance); released on terminate.

Existing guards verified unregressed: SG-in-use-by-ENI, cannot-delete-default-SG, SG-referenced-by-another-SG rule, delete-VPC-with-subnet, delete-attached-ENI.

### DIFF
--- a/providers/aws/aws.go
+++ b/providers/aws/aws.go
@@ -251,6 +251,10 @@ func New(opts ...config.Option) *Provider {
 	p.RDS.SetSubnetResolver(p.VPC)
 	p.ElastiCache.SetSubnetResolver(p.VPC)
 	p.EC2.SetSubnetResolver(p.VPC)
+	// RunInstances materializes the instance's primary (eth0) ENI in the VPC, and
+	// TerminateInstances releases it — so a running instance's interface blocks
+	// DeleteSubnet / DeleteSecurityGroup the way real EC2 does.
+	p.EC2.SetNetworking(p.VPC)
 	// A load balancer's VpcId is derived from its subnets, matching ELBv2.
 	p.ELB.SetSubnetResolver(p.VPC)
 	// EFS mount targets derive their VpcId and AZ from the subnet, so all mount
--- a/providers/aws/ec2/ec2.go
+++ b/providers/aws/ec2/ec2.go
@@ -288,6 +288,10 @@ type Mock struct {
 	// role->profile->instance chain reads back on DescribeInstances. nil until
 	// wired by the provider.
 	instanceProfileResolver InstanceProfileResolver
+	// networking materializes an instance's primary (eth0) ENI at launch and
+	// releases it on terminate, so the VPC subnet / security-group delete guards
+	// see a running instance's interface. nil until wired by the provider.
+	networking Networking
 	// mu guards managedResourceVisibility, clientTokens and subnetIPCounters,
 	// which are scalar/map shared state that (unlike the memstores) has no
 	// internal locking of their own.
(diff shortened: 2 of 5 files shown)


NitinKumar004 (commented):
Deep review on the pushed head (9814ab2) — clean checkout, not the fixer's report. Verdict: approve to merge on green CI.

**Gates (GOMAXPROCS=3):** build, vet, `go test ./server/aws/ec2/... ./providers/aws/... ./services/... ./features/topology/...`, `-race` on vpc+ec2, root `go test .`, gofmt, `golangci-lint --new-from-rev=origin/development` (0 new), `go generate` (zero diff), compatgen suite + `docs/compat` (zero diff) — all green.

**Guards / lifecycle (PR tests + adversarial runtime checks):**
- DeleteSubnet / DeleteSecurityGroup return DependencyViolation while an instance runs; both succeed after terminate.
- Stopped (not terminated) instance still blocks — StopInstances doesn't release the ENI, so it correctly keeps residing in the subnet.
- COUNT=N materializes N ENIs; terminating one leaves the others blocking; terminating instance A never releases B's ENI.
- rollbackInstances releases any ENI it created — no orphan.

**Release precision (the blast-radius risk) — verified at runtime under -race:** `ReleaseInstanceNetworkInterfaces` keys on `InstanceID == id && deleteOnTermination`. A standalone `CreateNetworkInterface` ENI attached to the instance as a secondary interface (deleteOnTermination=false) survives terminate; only the launch-created eth0 primary is dropped. No over-broad delete.

**Blast radius:**
- Lock ordering: cross-provider call is one-directional EC2→VPC, and EC2 does not hold its own `mu` when calling in (mirrors the existing `resolveSubnet` path); VPC guards read only VPC-local stores and never call back into EC2 — no cycle. A concurrent RunInstances/Terminate vs DeleteSubnet/DescribeNetworkInterfaces stress ran clean under -race with no deadlock. `memstore.All()` returns a copy, so delete-during-iterate in the release loop is safe.
- IP accounting stays consistent: `AvailableIpAddressCount` is CIDR-derived/static and independent of the `eniIPCounters` private-IP allocator, so materializing an ENI doesn't half-wire the count.
- No import cycle: the `Networking` hook is defined on the EC2 consumer side; both sides depend only on `services/networking/driver`.
- Diff is purely additive (293/0) — no existing test was modified to mask the new ENI, and the topology suite is unaffected.

**Non-blocking follow-up:** `DescribeNetworkInterfaces` now correctly lists the eth0 ENI, but EC2 `DescribeInstances` still doesn't populate the instance's `networkInterfaceSet` (it reads only the EC2 store, not the VPC ENI store). That's a pre-existing fidelity gap, unchanged here — worth a later ticket, not a blocker.

PR #722, opened 2026-08-26
fix(cloudwatch): paginate CloudWatch reads so no history/metrics are lost or unreachable

NitinKumar004:
## Summary

Fixes five CloudWatch pagination bugs across both wire paths (rpc-v2-cbor and the legacy query/XML path). CloudWatch reads either lost data or returned unbounded, tokenless responses that don't match real AWS.

### Bugs fixed

1. **DescribeAlarmHistory (data loss)** — the cbor handler truncated history to `MaxRecords` but the response carried **no `NextToken`**, so every entry past the first page was silently unreachable. Added `NextToken` to the request and response and switched to offset-based paging (default cap 100), so an SDK paginator retrieves **all** entries.
2. **GetMetricData** — had no paging fields. Added `MaxDatapoints` + `NextToken` to the request and `NextToken` to the response, paging over result rows within the datapoint budget.
3. **DescribeAlarms (cbor)** — the `MaxRecords==0 && NextToken==""` early-return bypassed pagination and returned every alarm tokenless. It now always sorts and paginates (cap 100 + token), matching AWS.
4. **ListMetrics (legacy XML)** — no pagination; now sorts and pages (500/page) with `NextToken`, mirroring the already-correct cbor path.
5. **DescribeAlarms (legacy XML)** — no pagination; now honors `MaxRecords`/`NextToken` with a stable sort.

All paths use the repo's existing offset-token helpers (`pageWindow`/`encode`/`decodeOffsetToken`) and stable sorting before slicing.

### Tests (real aws-sdk-go-v2/cloudwatch)

- **DescribeAlarmHistory**: builds a multi-page history and asserts the SDK paginator retrieves every entry exactly once across pages — proving the loss is fixed.
- **GetMetricData / DescribeAlarms (cbor)**: paginators walk more than one page, each result once, and terminate.
- **ListMetrics / DescribeAlarms (XML)**: legacy query path walked across pages via `NextToken` to completion; single page yields no token.

### DIFF
--- a/server/aws/cloudwatch/metric_data_ops.go
+++ b/server/aws/cloudwatch/metric_data_ops.go
@@ -57,7 +57,13 @@ type getMetricDataInput struct {
 	MetricDataQueries []metricDataQueryCBR `cbor:"MetricDataQueries"`
 	StartTime         *time.Time           `cbor:"StartTime,omitempty"`
 	EndTime           *time.Time           `cbor:"EndTime,omitempty"`
+	MaxDatapoints     int                  `cbor:"MaxDatapoints,omitempty"`
+	NextToken         string               `cbor:"NextToken,omitempty"`
 }
+
+// defaultMaxDatapoints is the AWS GetMetricData datapoint budget per page when a
+// caller omits MaxDatapoints; results past it spill onto a NextToken page.
+const defaultMaxDatapoints = 100800
 
 type metricDataResultCBR struct {
 	ID         string      `cbor:"Id"`
@@ -69,6 +75,7 @@ type metricDataResultCBR struct {
 
 type getMetricDataOutput struct {
 	MetricDataResults []metricDataResultCBR `cbor:"MetricDataResults"`
+	NextToken         string                `cbor:"NextToken,omitempty"`
 }
 
 // getMetricData implements the modern GetMetricData read API used by SDK v2,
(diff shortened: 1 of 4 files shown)


NitinKumar004 (commented):
**Verdict: MERGE** (pending green CI)

Reviewed on the actual pushed head `25f11df`. Clean detached-HEAD build of the pushed tree.

**Data-loss regression (DescribeAlarmHistory) — FIXED, verified adversarially.** The old handler did `kept[:MaxRecords]` with no NextToken, silently dropping page-2+ entries. `filterAlarmHistory` no longer truncates; `pageAlarmHistory` applies offset paging via `pageWindow`. Beyond the PR's own SDK paginator test, I ran an adversarial test (12 history entries, `MaxRecords=3`, real `aws-sdk-go-v2` `DescribeAlarmHistoryPaginator`): every distinct entry retrieved exactly once across ≥4 pages, none dropped/duplicated, terminates. Order is timestamp-descending (non-increasing), and the token/page boundary is stable across repeated identical requests (same first-page NextToken and same top entry). Only one AlarmHistory wire path exists (cbor) — no XML legacy path to mirror.

**GetMetricData (cbor) — no stall.** `pageMetricData` always emits ≥1 row (the `end > start` guard skips the budget check on the first row), so a row larger than `MaxDatapoints` still makes progress; the NextToken (`end`) is emitted only when `end < len(rows)`, so it strictly advances and terminates. Empty result set and offset==len both return empty page + no token (no empty non-terminating page). `MaxDatapoints` budget honored (defaults to 100800).

**DescribeAlarms (cbor) — early-return removed correctly.** No-arg call still returns results: `pageWindow(total,0,100)` yields the first ≤100 alarms plus a token; common no-arg path unbroken (PR test seeds >100 and walks all).

**ListMetrics XML + DescribeAlarms XML legacy paths** paginate, mirror the cbor path, stable-sort (`sort.Strings` / `sort.SliceStable` by name), token emitted only when truncated.

**Stable sort / invalid token / scope.** All paged paths sort before windowing (no dup/skip). `decodeOffsetToken` maps empty/malformed/negative tokens to offset 0 gracefully, consistent across all paths. No provider changes; the already-correct cbor ListMetrics path unregressed.

**Gates (GOMAXPROCS=3, -p=3):** go build ./... OK · go vet ./... OK · go test ./server/aws/cloudwatch/... ./providers/aws/cloudwatch/... OK · go test -race ./server/aws/cloudwatch/... OK · gofmt: 4 changed files clean (a pre-existing gofmt flag on metric_statistics_test.go exists on development, unrelated) · golangci-lint --new-from-rev=origin/development: 0 new issues · go generate: zero diff · compatgen suite green + docs/compat zero diff.

Changes limited to the 4 stated files. No remaining data loss.

PR #724, opened 2026-08-26
fix(aws-route53): paginate ListHostedZones/ListHealthChecks and continue ListResourceRecordSets by SetIdentifier

NitinKumar004:
## Summary

Fixes three AWS Route 53 pagination bugs so the real `aws-sdk-go-v2/route53` client walks every page correctly.

### 1. ListHostedZones
Previously hardcoded `IsTruncated: false`, ignored `marker`/`maxitems`, and never emitted a `NextMarker` — a large account came back as one untruncated page. Now sorts zones by id (stable), resumes at the request `marker` (inclusive), bounds the page by `maxitems`, and emits `IsTruncated` + `NextMarker` in Route 53 Marker style.

### 2. ListHealthChecks
Same bug, same fix: stable id sort, `marker`/`maxitems` honored, `IsTruncated` + `NextMarker` emitted. The shared logic lives in one generic `markerPage` helper.

### 3. ListResourceRecordSets — SetIdentifier continuation
The `StartRecordName`/`StartRecordType` -> `NextRecordName`/`NextRecordType` continuation was already correct, but `StartRecordIdentifier`/`NextRecordIdentifier` were not modeled. A weighted/latency/geo record **set** sharing one name+type whose siblings straddled a page boundary re-emitted the whole group, producing duplicate rows on the next page. Now the records sort by (name, type, SetIdentifier), the continuation resumes from the exact `SetIdentifier`, and `NextRecordIdentifier` is emitted — no record is duplicated or skipped.

## Tests (real route53 SDK)
- `TestSDKListHostedZonesPaginates` / `TestSDKListHealthChecksPaginates` — small `MaxItems` over 5 items, walk every page via Marker, assert each id appears exactly once and the walk terminates.
- `TestSDKListResourceRecordSetsIdentifierNoDupSkip` — a 3-record weighted set (distinct SetIdentifiers) straddling a `MaxItems=2` boundary, paginated with the SDK paginator; asserts each sibling appears exactly once. Verified to fail without the SetIdentifier continuation.

## Gates
build, vet, `-count=1` + `-race` tests, gofmt, golangci-lint (0 new), `go generate`, and `compatgen` + `git diff docs/compat` all clean.

### DIFF
--- a/server/aws/route53/health_check.go
+++ b/server/aws/route53/health_check.go
@@ -63,8 +63,12 @@ type listHealthChecksResponse struct {
 	XMLName      xml.Name         `xml:"ListHealthChecksResponse"`
 	Xmlns        string           `xml:"xmlns,attr"`
 	HealthChecks []healthCheckXML `xml:"HealthChecks>HealthCheck"`
-	IsTruncated  bool             `xml:"IsTruncated"`
-	MaxItems     int              `xml:"MaxItems"`
+	// Marker echoes the request marker; NextMarker is present only on a truncated
+	// page and carries the health-check id the caller passes back as Marker.
+	Marker      string `xml:"Marker,omitempty"`
+	IsTruncated bool   `xml:"IsTruncated"`
+	NextMarker  string `xml:"NextMarker,omitempty"`
+	MaxItems    int    `xml:"MaxItems"`
 }
 
 type updateHealthCheckRequest struct {
@@ -153,6 +157,9 @@ func (h *Handler) getHealthCheck(w http.ResponseWriter, r *http.Request, id stri
 	wire.WriteXML(w, http.StatusOK, getHealthCheckResponse{Xmlns: xmlns, HealthCheck: toHealthCheckXML(info)})
 }
 
+// listHealthChecks answers ListHealthChecks. Health checks are returned in a
+// stable id order and paginated Route 53 Marker-style: maxitems bounds the page
+// and NextMarker (echoed back as the next marker) resumes at the following id.
 func (h *Handler) listHealthChecks(w http.ResponseWriter, r *http.Request) {
 	infos, err := h.dns.ListHealthChecks(r.Context())
 	if err != nil {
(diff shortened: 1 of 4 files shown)


NitinKumar004 (commented):
**Verdict: approve (MERGE).** Verified against the pushed head `d5287a9` with the real `aws-sdk-go-v2/route53` client booted over `httptest`.

**ListResourceRecordSets SetIdentifier continuation (the priority) — correct, no dup and no skip.**
- The sort is a well-defined total order `(compareDNSName(name), Type, SetID)`, and `recordAtOrAfter` is the matching "at-or-after" predicate in that same order (name via `compareDNSName`, then plain string `>=` on Type, then on SetID) — monotonic over the sorted slice, so `recordsFrom`'s linear lower-bound scan resumes at exactly the right sibling.
- The truncation token is `records[maxItems]` (first record *not* emitted); the next page resumes inclusively at that `(name,type,identifier)`. Previous page excluded it via `records[:maxItems]`, next page re-includes it — one appearance, no gap.
- Beyond the PR's own `TestSDKListResourceRecordSetsIdentifierNoDupSkip`, I paginated a zone (SOA+NS + a 4-member weighted A group + a trailing simple record = 7 records) at **every MaxItems from 1..8** via the SDK paginator, forcing the page boundary at every position — including inside the weighted group at the first, middle, and last sibling. Every record (keyed by name|type|setid) appeared **exactly once** in all cases; walks terminated. `identifier` query param ↔ `NextRecordIdentifier` response element round-trips on the wire (MaxItems=1 resumes inside the group at every step).
- Simple name/type continuation unregressed — the trailing no-identifier record paginates once.

**ListHostedZones / ListHealthChecks** — shared `markerPage[T]` sorts by id, resumes inclusively at `marker` (`id >= marker`), bounds by maxItems, emits `NextMarker` = first id not included. Self-consistent inclusive resume = real Route 53 Marker semantics; SDK Marker walk returns each id once and terminates, last page carries no NextMarker.

**Other gates:** `//nolint:gosec G115` on `int32(maxItems)` is justified — `parseMaxItems` always returns `[1,100]`, matching the existing codebase clamp pattern. build / vet / `-count=1` / `-race` (server + provider route53) green; gofmt clean; golangci-lint `--new-from-rev=origin/development` = 0 new; `go generate` zero diff; `compatgen` + `git diff docs/compat/` zero. Changes limited to the 4 route53 files; no provider changes; existing route53 tests pass.

PR #725, opened 2026-08-26
fix(ec2): paginate Describe handlers that ignored MaxResults/NextToken

NitinKumar004:
## Summary

Eight AWS EC2 `Describe*` handlers applied request `Filters` but silently ignored `MaxResults`/`NextToken` and never emitted a token, so any client paginating them (real AWS paginates every one) got the whole set on page one and looped forever waiting for a token. This wires them through the same pagination path `DescribeInstances`/`DescribeSubnets`/`DescribeVpcs` already use.

Handlers fixed:
- DescribeVolumes
- DescribeSnapshots
- DescribeNetworkInterfaces
- DescribeImages
- DescribeTags
- DescribeVpcEndpoints
- DescribeVpcPeeringConnections
- DescribeDhcpOptions

## What changed

Each handler now, after applying its existing filters and explicit id-list, routes the projected result set through the shared `pageNetworkingXML(...)` helper (stable-sort by resource id, then apply request `MaxResults`/`NextToken`) and adds `nextToken` to its response struct. No new pagination logic is introduced — the shared helper is reused, so behaviour (MaxResults cap, base64 cursor, unparsable-token restart) is identical to the already-paginated Describes.

`DescribeTags` returns `(resource-id, key)` tuples gathered from map iteration, so it pages on a composite `resourceId\x00key` cursor to keep the token deterministic; a resource holds each key at most once, so the pair is unique.

`DescribeNatGateways` (already paginated) and `DescribeAddresses` (does not paginate in real AWS) are intentionally left untouched.

## Tests

New `describe_pagination_test.go` drives a booted in-process AWS server with the real `aws-sdk-go-v2/ec2` client. For every one of the eight ops it asserts:
- `MaxResults=2` over 5 items walks the whole set exactly once and terminates (no duplicates, gaps, or skips across the cursor)
- pagination combines with an explicit id-list / resource-id filter (set is filtered first, then paged)
- an unparsable `NextToken` restarts from the beginning (matching the shared `pageNetworkingXML` helper) rather than erroring or skipping

## Verification

`go build ./...`, `go vet ./...`, `go test ./server/aws/ec2/... ./providers/aws/...`, `go test -race ./server/aws/ec2/...`, `gofmt -l`, `go generate` (zero diff), and `compatgen` + `git diff docs/compat` (zero drift) all green. golangci-lint on the changed package introduces zero new issues over baseline.

### DIFF
--- a/server/aws/ec2/describe_pagination_test.go
+++ b/server/aws/ec2/describe_pagination_test.go
@@ -1,0 +1,399 @@
+package ec2_test
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
+)
+
+// These tests pin that the EC2 Describe handlers that gained MaxResults/NextToken
+// paging honour it the way DescribeSubnets/DescribeInstances already do: a small
+// MaxResults walks the whole set exactly once via the echoed NextToken, paging
+// combines with an explicit id-list (the set is filtered first, then paged), and
+// an unparsable NextToken restarts from the beginning (shared pageNetworkingXML
+// behaviour) rather than erroring.
+const (
+	pageSize      = 2
+	itemCount     = 5
+	filterCount   = 3
+	maxAll        = 1000
+	volumeSizeGiB = 10
+	bogusToken    = "!!!not-a-valid-token!!!"
+	nameTagKey    = "Name"
+)
+
+// pageFunc issues one Describe request: max is the MaxResults, ids restricts the
+// result set (explicit id-list or resource-id filter), tok is the NextToken to
+// resume from. It returns the page's stable ids plus the echoed NextToken.
+type pageFunc func(max int32, ids []string, tok *string) ([]string, *string, error)
+
+// mapIDs projects a slice of SDK shapes to their stable ids.
+func mapIDs[T any](xs []T, id func(T) string) []string {
+	out := make([]string, 0, len(xs))
+	for i := range xs {
+		out = append(out, id(xs[i]))
+	}
+
+	return out
+}
+
+// collectAll fetches the whole (optionally id-filtered) set in one request and
+// asserts a within-cap MaxResults leaves no dangling NextToken.
+func collectAll(t *testing.T, f pageFunc, ids []string) []string {
+	t.Helper()
+
+	got, next, err := f(maxAll, ids, nil)
+	if err != nil {
+		t.Fatalf("list: %v", err)
+	}
+
+	if aws.ToString(next) != "" {
+		t.Fatalf("MaxResults=%d returned a NextToken, want the full set in one page", maxAll)
+	}
+
+	return got
+}
+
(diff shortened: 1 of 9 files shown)


NitinKumar004 (commented):
Reviewed at head 20f5298 (isolated worktree, clean build of the pushed tree).

Gates -- all green (GOMAXPROCS=3, -p=3):
- go build ./... -- clean
- go vet ./... -- clean
- go test -p=3 ./server/aws/ec2/... ./providers/aws/... -- 45 packages, all ok, exit 0
- go test -race -p=3 ./server/aws/ec2/... -- ok, exit 0
- gofmt -l on the 9 changed files -- clean
- golangci-lint --new-from-rev=origin/development ./server/aws/ec2/... -- 0 new issues
- go generate ./... -- zero diff (coveragegen: 51 services, unchanged)
- go run ./internal/compatgen + git diff --exit-code docs/compat/ -- exit 0, zero drift
- Diff scope: exactly the 9 files in the PR description, nothing else touched

Verification against a booted server with the real aws-sdk-go-v2/ec2 client:
- All 8 ops (Volumes/Snapshots/NetworkInterfaces/Images/Tags/VpcEndpoints/VpcPeeringConnections/DhcpOptions) route through the same pageNetworkingXML helper as DescribeSubnets/DescribeVpcs/DescribeInstances, invoked as the last step after the existing filter/id-list projection in every handler -- confirmed filter-then-page ordering in all 9 files.
- The PR's own describe_pagination_test.go (399 lines, real SDK client) passes under -race: full walk with MaxResults=2, id-list-filtered walk, and invalid-token restart, for all 8 ops.
- Independently re-verified two of the strongest cases using the actual generated SDK Paginator types (ec2.NewDescribeVolumesPaginator, ec2.NewDescribeTagsPaginator) rather than the PR's hand-rolled loop, including a DescribeTags stress case the PR doesn't cover: 6 resources x 4 tags each (24 tuples, vs. the PR's 1-tag-per-resource case), MaxResults=2 -- every (resourceId,key) tuple observed exactly once, no dup/skip. Scratch test deleted after use; tree left clean.
- tagItemCursor (resourceId + "\x00" + key) is deterministic under pageNetworkingXML's pre-sort despite tags being gathered via map iteration; NUL is not producible by any AWS resource-id or tag-key charset, so no separator collision. The "each resource holds a key at most once" uniqueness premise holds structurally -- appendTagRecords emits one record per (id, key) pair from a single tags map per resource, and compute/network id prefixes (i-/vol-/snap-/ami-/vpc-/subnet-/sg-/sgr-/...) are disjoint so no cross-source collision on resourceId.
- Restart-at-zero on an unparsable NextToken: confirmed via decodePageToken (returns index 0 on any base64/lookup failure) -- this is the same paginateXML/pageNetworkingXML path the other already-paginated EC2 describes use, pinned by TestPaginateXMLUnknownTokenStartsAtZero. The 8 new handlers inherit this by construction; not a new behavior and not a bug.
- No regression: DescribeSubnets/DescribeVpcs/DescribeInstances/DescribeSecurityGroups/DescribeRouteTables/DescribeNetworkAcls/DescribeInternetGateways are untouched by the diff and pass in the full server/aws/ec2 suite.
- DescribeNatGateways and DescribeAddresses -- confirmed untouched in the diff.

No blocking issues found.

PR #727, opened 2026-08-26
fix(pagination): pace DynamoDB ListTagsOfResource and EFS Describe* in the handler

NitinKumar004:
## Summary

Four AWS list endpoints ignored their page-size/token inputs: the provider returned the full slice and the **server handler** wrote it all back, never emitting a continuation cursor. This pages each one in the handler via the repo `internal/pagination` helper over a **stable-sorted** slice, so a real SDK paginator walks every record exactly once and terminates.

Fixed endpoints:

- **DynamoDB `ListTagsOfResource`** — decode `NextToken`, sort tags by key, emit `NextToken` when truncated. DynamoDB caps a resource at 50 tags so a real response is a single page (matching AWS); the plumbing is exercised with a small page size. Invalid token -> `ValidationException`.
- **EFS `DescribeFileSystems`** — `Marker`/`MaxItems` -> `NextMarker`, sorted by `FileSystemId`.
- **EFS `DescribeMountTargets`** — `Marker`/`MaxItems` -> `NextMarker`, sorted by `MountTargetId`, existing FileSystemId/AccessPointId filter preserved.
- **EFS `DescribeAccessPoints`** — `MaxResults`/`NextToken` -> `NextToken`, sorted by `AccessPointId`.

A malformed EFS cursor returns the real `BadRequest`. Stable sort precedes offset paging so a page boundary never duplicates or skips a record. Providers were not changed.

## Why

Offset-token pagination over an unsorted, map-iteration-ordered slice would duplicate or skip records across page boundaries; the handlers also never returned a cursor, so paginators saw everything as one page.

## Tests (real aws-sdk-go-v2)

- EFS: three SDK paginator journeys (page size 2, 5 items) assert every record is visited exactly once and the walk terminates; single page emits no cursor; invalid Marker/NextToken -> BadRequest.
- DynamoDB: SDK single-page-no-token + invalid-NextToken -> ValidationException, plus a unit test driving the pagination plumbing with a small page size to prove the stable-sorted cursor walks all pages once.

## Gates

build, vet, `go test` (server+provider+pagination), `-race` on EFS, gofmt, golangci-lint (0 new), go generate (no diff), compatgen (no docs/compat diff) — all green.

### DIFF
--- a/server/aws/dynamodb/list_tags_pagination_sdk_test.go
+++ b/server/aws/dynamodb/list_tags_pagination_sdk_test.go
@@ -1,0 +1,70 @@
+// list_tags_pagination_sdk_test.go — real aws-sdk-go-v2 journeys for
+// ListTagsOfResource pagination: a resource's tags (well under the 50-tag cap)
+// return in a single page with no NextToken, and a malformed NextToken is
+// rejected with a ValidationException.
+package dynamodb_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
+	ddbtypes "github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
+	"github.com/aws/smithy-go"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDDBListTagsSinglePageNoToken(t *testing.T) {
+	client, _ := newSuiteDDBEnv(t)
+	ctx := context.Background()
+
+	suiteDDBCreateTable(t, client, "tagged", "pk", "")
+
+	desc, err := client.DescribeTable(ctx, &dynamodb.DescribeTableInput{TableName: aws.String("tagged")})
+	require.NoError(t, err)
+	arn := aws.ToString(desc.Table.TableArn)
+
+	_, err = client.TagResource(ctx, &dynamodb.TagResourceInput{
+		ResourceArn: aws.String(arn),
+		Tags: []ddbtypes.Tag{
+			{Key: aws.String("env"), Value: aws.String("prod")},
+			{Key: aws.String("team"), Value: aws.String("core")},
+			{Key: aws.String("cost"), Value: aws.String("42")},
+		},
+	})
+	require.NoError(t, err)
+
+	out, err := client.ListTagsOfResource(ctx, &dynamodb.ListTagsOfResourceInput{ResourceArn: aws.String(arn)})
+	require.NoError(t, err)
+	assert.Len(t, out.Tags, 3)
+	assert.Nil(t, out.NextToken, "a sub-cap tag set must not emit a NextToken")
+
+	got := map[string]string{}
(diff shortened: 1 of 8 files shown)


NitinKumar004 (commented):
Reviewed pushed head `5e04e69` in an isolated worktree (GOMAXPROCS=3, `-p=3`).

**Gates — all green**
- `go test -p=3 ./server/aws/dynamodb/... ./server/aws/efs/... ./providers/aws/dynamodb/... ./providers/aws/efs/... ./internal/pagination/... -count=1` — all pass
- `go test -race -p=3 ./server/aws/efs/... -count=1` — all pass, no races
- `gofmt -l` — clean
- `golangci-lint run --new-from-rev=origin/development ./server/aws/dynamodb/... ./server/aws/efs/...` — 0 new issues
- Diff scope matches: exactly the 8 listed files; `providers/` untouched

**Correctness**
- DynamoDB `ListTagsOfResource`: `pageTags` sorts by key via `pagination.PaginateSorted`, page size 50 (matches the real 50-tag cap), emits `NextToken` only when truncated, omits it entirely (not empty string) when not. Empty `Tags` never serializes as `null` (explicit `[]tagJSON{}` fallback). Invalid token → 400 `ValidationException`. `TestPageTagsWalksAllPagesStable` drives page size 2 over 5 shuffled-input tags and proves each key visited exactly once in sorted order, confirming the plumbing beyond the single-page SDK-visible case.
- EFS `respondPaged` generic helper: sorts via `sort.SliceStable` over element pointers (no per-comparison copy of larger structs), then delegates to `pagination.Paginate` (not `PaginateSorted` — correct, since it already sorted the same slice reference in place, avoiding a redundant second sort). A driver error short-circuits before any success write. A malformed token maps to `writeError(...BadRequest...)`. All 3 handlers (`DescribeFileSystems`, `DescribeMountTargets`, `DescribeAccessPoints`) wire distinct `less` (by `FileSystemID`/`MountTargetID`/`AccessPointID`) and `build` closures through the shared helper — no duplicated pagination logic, no drift risk between the three.
- `DescribeMountTargets` filter signature unchanged (`FileSystemId`/`MountTargetId`/`AccessPointId` all still passed to the driver before pagination is applied) — confirmed by diff, filtering still happens pre-page.
- Driver-returned slices for all three Describe* calls are freshly built copies (map iteration into a new slice each call), never aliases into internal store state, so in-place `sort.SliceStable` in the handler is safe and race-free — matches the `-race` result.
- `writeError` retype (`errType` → `genericErrorType` const, `"BadRequest"`): checked all 4 call sites pre- and post-diff (`notFound`, `methodNotAllowed`, `decodeJSON`, and the new pagination path) — every one already passed the literal `"BadRequest"` before this change. The unparam-driven retype is behavior-preserving; no error response shape or code changed for any existing EFS caller.
- Wire response structs (`describeFileSystemsResponse.NextMarker`, `describeMountTargetsResponse.NextMarker`, `describeAccessPointsResponse.NextToken`) all carry `omitempty`, so an un-truncated page serializes without the cursor field, matching real EFS behavior (confirmed by the SDK single-page test asserting a nil pointer).

**Verdict: MERGE** — no blocking issues found.

PR #728, opened 2026-08-26
fix(pagination): reject negative-offset page tokens instead of panicking

NitinKumar004:
## Summary

A well-formed continuation token (valid base64 + valid JSON) carrying a **negative** offset made the shared `internal/pagination.Paginate` helper panic with "slice bounds out of range". `net/http` recovers per-connection, so callers saw a connection reset instead of a proper error. Real AWS returns an invalid-token error.

This affects ~15 callers that route through this helper (RDS, Redshift, ACM, ECS, EKS, DynamoDB ListTagsOfResource, EFS, plus several GCP/Azure handlers).

## Root cause

`DecodeToken` accepted any integer offset. `Paginate` guarded `offset >= len(items)` but not `offset < 0`, so a negative offset slid past the guard straight into `items[offset:end]` and panicked.

## Fix

- Added an exported sentinel `pagination.ErrInvalidToken`.
- `DecodeToken` now rejects a decoded negative offset with `ErrInvalidToken` (a negative offset is never a token we produced).
- `Paginate` adds a defensive `offset < 0` guard at the slice site so no token input can panic there regardless of decode path.

The behavior is **consistent with the existing contract**: this helper already returns an error for malformed (non-base64 / non-JSON) tokens, and an empty page (no error) for offset-beyond-len. Negative offset now returns an error, matching the malformed-token path. Malformed and offset-beyond-len behavior are unchanged. No caller changes were needed — RDS and Redshift already map any pagination error to `InvalidParameterValue`.

## Tests

- Helper unit tests: negative offset (via `EncodeToken` and a hand-crafted `{"offset":-1}` token) rejected with `ErrInvalidToken`; offset-beyond-len still returns empty page; malformed token behavior unchanged; valid mid-list token returns the correct page; explicit `NotPanics` assertions.
- Caller-level SDK round-trip tests: RDS `DescribeDBInstances` and Redshift `DescribeClusters` with a well-formed negative-offset Marker return an error (mapped to `InvalidParameterValue`), not a connection reset. Existing normal-pagination round-trips continue to pass.

### DIFF
--- a/internal/pagination/pagination.go
+++ b/internal/pagination/pagination.go
@@ -26,6 +26,12 @@ func Paginate[T any](items []T, pageToken string, maxResults int) (Page[T], erro
 	}
 
 	offset := token.Offset
+	// Defensive: DecodeToken already rejects negative offsets, but guard the
+	// slice bound directly so no token input can panic here.
+	if offset < 0 {
+		return Page[T]{}, ErrInvalidToken
+	}
+
 	if offset >= len(items) {
 		return Page[T]{Items: nil, HasMore: false}, nil
 	}
--- a/internal/pagination/pagination_test.go
+++ b/internal/pagination/pagination_test.go
@@ -1,6 +1,7 @@
 package pagination
 
 import (
+	"encoding/base64"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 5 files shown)


NitinKumar004 (commented):
Reviewed at head 23f8857 (clean-built the pushed tree, GOMAXPROCS=3).

Verdict: MERGE.

Verified:
- No panic on ANY token input. Table-tested the helper directly (-race) with EncodeToken(-5), crafted {"offset":-1} and {"offset":-999999}, offset==len, offset>len, malformed base64, valid-b64-non-JSON, valid-JSON-non-int-offset ({"offset":"x"}), empty JSON object {}, and empty token. None panic. Negative -> ErrInvalidToken; offset>=len -> empty page, no error; empty/{} -> offset 0 from start; mid-list -> correct page.
- Malformed contract unchanged. DecodeToken still returns the pre-existing base64/JSON decode error for non-base64 and valid-b64-non-JSON tokens; it is NOT remapped to ErrInvalidToken (confirmed errors.Is negative). TestPaginate_InvalidToken semantics preserved.
- Error path never returns items (no swallow into a partial/valid page).
- Broad caller regression sweep. Enumerated every internal/pagination.Paginate / PaginateSorted / DecodeToken / EncodeToken caller (grep, non-test): AWS s3, rds, redshift, ecs, eks, acm, ecr, wafv2; GCP gcs, cloudsql, vpc, pubsub, secretmanager, artifactregistry, eventarc, compute, firestore; Azure blobstorage; portable services/database. EVERY call site checks err and maps it to that API's invalid-token response (InvalidParameterValue / InvalidParameterException / InvalidArgument / OutOfRangeInput / INVALID_ARGUMENT / cerrors.InvalidArgument). ACM/Redshift/CloudSQL return the error upward through helpers that all propagate to a proper error response. No caller swallows ErrInvalidToken into a success.
- The awsquery wire codec (offset-0-restart, TestPaginateXMLUnknownTokenStartsAtZero) does not route through this helper and is untouched. MSK/EFS/DynamoDB ListTagsOfResource do not use internal/pagination (own paginators), so they are unaffected.

Gates (head 23f8857): go build clean; go vet clean; go test -race ./internal/pagination pass; all caller pkgs pass; full suite go test -p=3 ./... -count=1 = 0 failures; gofmt clean; golangci-lint --new-from-rev=origin/development = 0 new; go generate = 0 diff; compatgen exit 0 + docs/compat/ 0 diff. Changes limited to the 5 declared files. Tree left clean.

No blocking issues.

PR #729, opened 2026-08-26
fix(aks): round-trip networkProfile and inline agentPool advanced fields

NitinKumar004:
## Summary

Fixes two HIGH Terraform-drift bugs in Azure AKS (`Microsoft.ContainerService/managedClusters`) where a submitted cluster PUT did not round-trip on GET.

Root cause (both bugs): the AKS handler is wrapped by `echoUnmodeledProperties`, which echoes any request `properties.*` subtree the handler does not model. But when the handler emitted a **partial hardcoded model** of a field, the overlay treated those keys as "modeled" and dropped the caller's real values. The fix fully models each field: parse the caller's value, echo it, synthesize a default only when omitted.

### 1. networkProfile fabricated / did not round-trip
`toARMCluster` emitted `defaultNetworkProfile()` unconditionally (hardcoded kubenet / 10.0.0.0/16 / 10.0.0.10), and `buildClusterInput` never read `body.Properties.NetworkProfile`. Now `networkProfile` (networkPlugin, networkPolicy, serviceCidr, dnsServiceIP, podCidr, loadBalancerSku, outboundType) is modeled on `ManagedCluster` + `ClusterInput`, parsed from the request, and echoed verbatim. The default block is synthesized only when the caller omits `networkProfile` entirely (real AKS defaults: kubenet / 10.0.0.0/16 / 10.0.0.10). Sub-keys the caller sets but we do not model are not fabricated, so they still round-trip through the overlay.

### 2. Inline agentPoolProfiles advanced fields dropped
Terraform submits `default_node_pool` inline in the cluster PUT, but `availabilityZones`, `enableAutoScaling`, `minCount`, `maxCount`, `vnetSubnetID`, `osSKU`, and `enableNodePublicIP` were dropped on GET. These are now modeled on a shared embedded `armAgentPoolAdvanced` struct (used by both the inline profile and the standalone-pool shapes), on `AgentPool`/`AgentPoolInput`/`buildAgentPool`, and parsed on both the inline and standalone wire paths through a shared `applyTo` mapping — so both paths behave identically.

Out of scope (deferred, untouched): upgradeProfile, UserAssigned identity map, systemData/etag, sku.name, data-plane kubeconfig. `echo_properties.go` is unchanged — the #715 write-only secret stripping stays intact.

## Tests

New real-SDK e2e tests (`armcontainerservice/v6`) in `sdk_terraform_drift_test.go`:
- networkProfile submitted -> GET echoes exactly (B1)
- networkProfile omitted -> GET returns real defaults (B1-default)
- inline agentPoolProfiles advanced fields -> GET echoes all (B2)
- standalone agentPools PUT->GET still round-trips (regression)
- servicePrincipalProfile.secret absent on GET (#715 regression)

## Gates
build, vet, `-race`, gofmt, golangci-lint (0 new), `go generate` (zero diff), compatgen + `docs/compat` (zero diff) all green.

### DIFF
--- a/providers/azure/aks/aks.go
+++ b/providers/azure/aks/aks.go
@@ -37,10 +37,18 @@ const (
 	defaultPoolType      = "VirtualMachineScaleSets"
 	defaultNodeImage     = "AKSUbuntu-2204gen2containerd-202401.09.0"
 	poolPowerRunning     = "Running"
-	emulatorTenantID     = "11111111-1111-1111-1111-111111111111"
-	cpuMetricRunning     = 0.35
-	memMetricRunning     = 0.50
-	podMetricRunning     = 12.0
+	// Network-profile defaults the real AKS service synthesizes when a create
+	// omits properties.networkProfile entirely.
+	defaultNetworkPlugin   = "kubenet"
+	defaultServiceCidr     = "10.0.0.0/16"
+	defaultDNSServiceIP    = "10.0.0.10"
+	defaultPodCidr         = "10.244.0.0/16"
+	defaultLoadBalancerSKU = "standard"
+	defaultOutboundType    = "loadBalancer"
+	emulatorTenantID       = "11111111-1111-1111-1111-111111111111"
+	cpuMetricRunning       = 0.35
+	memMetricRunning       = 0.50
+	podMetricRunning       = 12.0
 )
 
 // ManagedCluster is the in-memory representation of an AKS cluster.
@@ -60,6 +68,11 @@ type ManagedCluster struct {
 	Tags           map[string]string
 	AgentPoolNames []string
 
+	// NetworkProfile echoes properties.networkProfile. When a create omits it,
+	// the mock stores the standard AKS defaults (see defaultNetworkProfile);
+	// otherwise it stores exactly the submitted values so a read round-trips.
+	NetworkProfile NetworkProfile
+
 	// EnableRBAC mirrors properties.enableRBAC; defaults to true, the AKS
 	// default when a create omits it.
 	EnableRBAC bool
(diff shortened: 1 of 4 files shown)


NitinKumar004 (commented):
## Review: fix(aks): round-trip networkProfile and inline agentPool advanced fields

Verified on pushed SHA `dc6e8921` in an isolated checkout, GOMAXPROCS=3.

**Gates (all green):**
- `go build ./...`, `go vet ./...`, `go test ./...`, `go test -race` (Bedrock set) — PASS
- `gofmt`, `go mod tidy` — clean
- `go run ./internal/compatgen` + `git diff docs/compat/` — zero diff
- `golangci-lint` scoped to `providers/azure/aks/...` + `server/azure/aks/...` — **0 issues** (repo-wide `golangci-lint run ./...` reports 273 pre-existing issues elsewhere, none in the touched files)
- `go test -race ./providers/azure/aks/...` — clean

**Root cause and fix confirmed correct.** The overlay (`echoUnmodeledProperties` / `missingProperties`) treats any key present in both the request and the handler's response as "modeled" regardless of value equality, so a hardcoded/partial model silently swallowed the caller's real value without ever looking wrong to the overlay. Fully modeling `networkProfile` (with `omitempty` throughout) and the inline/standalone agent-pool advanced fields via the shared `armAgentPoolAdvanced` + `applyTo` closes that gap correctly.

**Independently re-verified (beyond the PR's own tests):**
- `networkProfile` submitted → echoed verbatim; omitted → real AKS defaults (kubenet/10.0.0.0/16/10.0.0.10).
- The critical overlay-interaction case: PUT `networkProfile{networkPlugin, serviceCidr, dnsServiceIP}` **plus an unmodeled sub-key** (`natGatewayProfile`, which this PR does not model) → GET echoes the modeled keys from the model *and* `natGatewayProfile` from the overlay, with no fabricated zero-value keys for the sub-keys the caller never sent (`podCidr`, `networkPolicy`, `loadBalancerSku` all absent). This is the exact mechanism the two HIGH bugs came from, and it now round-trips correctly.
- Inline `agentPoolProfiles[0]` advanced fields (availabilityZones, enableAutoScaling, min/maxCount, vnetSubnetID, osSKU) round-trip; the standalone `agentPools/{name}` path shares the same `applyTo` mapping and round-trips identically (regression-guarded by the PR's own new test).
- `#715` regression check: `servicePrincipalProfile.secret` still absent on GET, `clientId` still round-trips; `echo_properties.go`/`echo_properties_test.go` (including `TestEchoDoesNotLeakAKSAADServerAppSecret`) are untouched by this diff and still pass.
- Concurrent `CreateOrUpdateAgentPool` + `GetAgentPool` under `-race` on the new pointer fields (`MinCount`/`MaxCount`/`EnableAutoScaling`/`EnableNodePublicIP`) is clean — `copyBoolPtr`/`copyInt32Ptr` always allocate fresh pointers rather than mutating in place, so `Get`'s shallow struct copy never aliases a mutable pointee.
- The 2 new pointer-param helpers (`toNetworkProfile`, `toAgentPoolAdvanced`) are read-only renderers; passing by pointer vs. value doesn't change behavior — consistent with the 0 new lint issues.

**Advisory note (non-blocking):** `mirror-check.sh` flags this as a single-provider change (Azure only). That's expected and correctly scoped here — the bug is specific to Azure's ARM `echoUnmodeledProperties` overlay mechanism, which has no AWS/GCP equivalent; there's no portable-API behavior being silently left un-mirrored.

No blocking issues found.

**Verdict: MERGE** (pending green CI).

PR #730, opened 2026-08-26
fix(lambda): Publish, Architectures/EphemeralStorage/Layers round-trip, GetFunctionConcurrency 200 when unset

NitinKumar004:
## Summary

Fixes 5 real AWS Lambda parity bugs surfaced by driving a real `aws-sdk-go-v2/lambda` client at the booted wire server.

1. **CreateFunction `Publish=true` was ignored** — Version stayed `$LATEST` and no version was cut. Now `Publish=true` publishes version 1 and the response is that version's configuration (`Version:"1"`, `:1`-qualified ARN). Also honored on `UpdateFunctionCode` (both APIs really carry a `Publish` field). Fixes Terraform `aws_lambda_function{publish=true}`.
2. **`Architectures` never round-tripped** — now echoes `[arm64]` when set and defaults to `[x86_64]` when omitted.
3. **`EphemeralStorage` never round-tripped** — now echoes `{Size:2048}` when set and defaults to `{Size:512}`; a size outside 512–10240 is rejected with `InvalidParameterValueException`.
4. **`GetFunctionConcurrency` returned 404 when reserved concurrency was unset** — real AWS returns HTTP 200 with an empty body; only a missing function is 404. The wire GET now checks function existence separately from concurrency being set.
5. **`Layers` were not echoed** — imported layers are now persisted and returned as `[{Arn, CodeSize}]` on Create/Get and after an UpdateFunctionConfiguration that changes them.

## Approach

The AWS-only fields (`Architectures`, `EphemeralStorage`, `Layers`) are carried on `AWSFunctionConfig` and applied/read through the existing AWS-only optional interface, consistent with how `VpcConfig`/`DeadLetterConfig`/`TracingConfig` are already kept off the provider-agnostic `Serverless` surface. `toConfiguration` applies the `[x86_64]` / `512` defaults.

`Publish` is honored only on the two APIs that really expose it — `CreateFunction` and `UpdateFunctionCode`; `UpdateFunctionConfiguration` has no `Publish` parameter in the AWS API and is left unchanged.

## Tests

Added hand-rolled real-SDK e2e tests: Publish round-trip on Create + `GetFunction Qualifier=1`, Publish on UpdateFunctionCode, Architectures round-trip + default, EphemeralStorage round-trip + default + range rejection, GetFunctionConcurrency 200-when-unset / 404-when-missing, and Layers echoed. Updated the existing concurrency roundtrip test whose post-delete assertion encoded the old (buggy) 404 behavior.

## Gates

build, vet, `go test` (lambda + serverless + server/persist/seed/root), `-race`, gofmt, golangci-lint (0 new), `go generate` (zero diff), compatgen (exit 0, `docs/compat` clean).

### DIFF
--- a/providers/aws/lambda/awsconfig.go
+++ b/providers/aws/lambda/awsconfig.go
@@ -14,6 +14,8 @@ import (
 // It implements the AWS-only optional interface the Lambda server type-asserts
 // for, keeping these AWS-specific settings off the provider-agnostic Serverless
 // surface.
+//
+//nolint:gocritic // hugeParam: interface method signature cannot be changed.
 func (m *Mock) SetFunctionAWSConfig(_ context.Context, name string, cfg driver.AWSFunctionConfig, create bool) error {
 	m.mu.Lock()
 	defer m.mu.Unlock()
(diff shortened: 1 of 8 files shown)


NitinKumar004 (commented):
Delta re-review of aa6de917 (fix(lambda): drop Publish from UpdateFunctionConfiguration).

**Delta verified correct and complete:**
- `updateFunctionConfigurationRequest` (types.go) no longer has a `Publish` field; `serveConfiguration`'s PUT branch (handler.go) no longer special-cases `req.Publish` — it always falls through to `writeJSON(w, http.StatusOK, toConfiguration(info, awsCfg))`. Matches the real API: `UpdateFunctionConfiguration` has no `Publish` parameter.
- `Publish` is intact and still exercised by real `aws-sdk-go-v2/lambda`:
  - `CreateFunction{Publish:true}` → version 1, `:1`-qualified ARN (via `publishConfiguration` at handler.go:863, `TestSDKCreateFunctionPublish`).
  - `UpdateFunctionCode{Publish:true}` → new version, qualified ARN (via `writePublished`/`publishConfiguration` at handler.go:595-596, `TestSDKUpdateFunctionCodePublish`).
- No dead code: `writePublished` (handler.go:537) now has exactly one call site (serveCode, line 596) and is still used; `publishConfiguration` (handler.go:929) is called from both `writePublished` and the CreateFunction handler (line 863). Nothing orphaned.
- Other 4 fixes from the base commit (e690dd3) are intact and gate-covered: Architectures round-trip + `[x86_64]` default, EphemeralStorage round-trip + 512MB default + range validation, `GetFunctionConcurrency` 200-empty-when-unset vs 404-when-function-missing, Layers echoed with Arn/CodeSize.

**Non-blocking nit:** two doc comments are now stale after this delta and reference the removed path — worth a follow-up cleanup, not blocking:
- handler.go:534-536, `writePublished`'s comment: "...the shared tail of the Publish=true UpdateFunctionCode/**UpdateFunctionConfiguration** paths." — no longer shared, only UpdateFunctionCode calls it.
- handler.go:926-928, `publishConfiguration`'s comment: "...the response body AWS returns for a Publish=true Create/UpdateFunctionCode/**Configuration**." — same stale reference.

**Gates on aa6de9175354084f2071409a0d1a06fe30ae1ab2 (GOMAXPROCS=3):**
- `go build ./...` clean
- `go vet ./...` clean
- `go test -p=3 ./server/aws/lambda/... ./providers/aws/lambda/... ./services/serverless/...` — all pass
- `go test -race -p=3 ./server/aws/lambda/...` — pass
- `gofmt -l` — 4 pre-existing unrelated files flagged (acr, cloudwatch, s3 x2), none touched by this PR
- `golangci-lint --new-from-rev=origin/development` on changed packages — 0 new issues

Verdict: MERGE.

PR #731, opened 2026-08-26
fix(sqs): FIFO in-flight per-group blocking + honor WaitTimeSeconds long polling

NitinKumar004:
## Summary

Fixes two AWS SQS `ReceiveMessage` correctness bugs.

### 1. [HIGH] FIFO per-message-group in-flight blocking
For a FIFO queue, once a message from a `MessageGroupId` is received and still in-flight (received, visibility not expired, not deleted), no further messages of that group may be returned until the in-flight one is deleted or its visibility expires. Previously `collectVisibleMessages` iterated purely by `VisibleAt`, so receiving `a` of group `G1` then receiving again returned `b` of `G1`, breaking FIFO ordering under a poll-then-process consumer.

Now, for FIFO queues only, `collectVisibleMessages` tracks the head message per group: a group with an earlier in-flight/undelivered message is skipped so only the single next-in-line message per group is deliverable at a time. Different groups still deliver in parallel. Non-FIFO queues are unchanged; dedup and in-group send-order are preserved.

### 2. [MED] WaitTimeSeconds long polling was a no-op
An empty-queue receive with `WaitTimeSeconds=N` returned immediately (busy-spin) and missed a message a concurrent producer sent during the window. `ReceiveMessages` now runs a bounded poll loop honoring `WaitTimeSeconds` (falling back to the queue's `ReceiveMessageWaitTimeSeconds` attribute when unset, capped at 20s): it retries on a short interval until a message appears, the deadline elapses, or the request context is cancelled. The queue lock is never held across the wait (each attempt re-acquires it). `WaitTimeSeconds=0` returns immediately as before.

## Tests
Added real `aws-sdk-go-v2/sqs` e2e tests (httptest-booted server), run clean under `-race`:
- FIFO: send `a,b` to `G1` + `x` to `G2`; receive returns `{a,x}` not `b`; delete `a` unblocks `b`; two groups deliver in parallel.
- Long poll: mid-window concurrent send is picked up within the window; empty queue with no producer returns empty after ~the full window; a cancelled context aborts the wait promptly.

## Gates
build, vet, `go test`, `go test -race`, gofmt, golangci-lint (0 new), `go generate` (no diff), compatgen (exit 0, no `docs/compat` diff) all green.

### DIFF
--- a/providers/aws/sqs/sqs.go
+++ b/providers/aws/sqs/sqs.go
@@ -857,23 +857,23 @@ func (m *Mock) collectVisibleMessages(
 
 	var toRemove []int
 
-	isFIFO := qd.info.FIFO
-	blocked := make(map[string]bool)
+	// FIFO: any group that already has an in-flight message (received in a prior
+	// call, visibility not yet expired) is skipped entirely this call, so a
+	// group's messages are delivered strictly in order across consumers. Within
+	// a single call a non-blocked group still yields its consecutive messages
+	// (up to MaxNumberOfMessages), matching real SQS FIFO batch receives.
+	blocked := fifoInFlightGroups(qd, now)
 
 	for i, msg := range qd.messages {
 		if len(results) >= maxMessages {
 			break
 		}
 
-		// FIFO: a group with an earlier in-flight/undelivered message is blocked
-		// until that message is deleted or its visibility timeout expires.
-		if fifoGroupBlocked(isFIFO, msg.GroupID, blocked) {
+		if blocked[msg.GroupID] {
 			continue
 		}
 
 		if msg.VisibleAt.After(now) {
-			markFIFOGroup(isFIFO, msg.GroupID, blocked)
-
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


NitinKumar004 (commented):
**Verdict: MERGE** (pending green CI)

Reviewed at pushed head `70be4ee7`. Clean-built the actual pushed tree.

**Correctness — FIFO per-group in-flight blocking**
- `collectVisibleMessages` iterates `qd.messages` in send order and blocks a group once its head is either delivered or found in-flight (`VisibleAt.After(now)`), so ordering within a group is preserved and dedup is untouched.
- Verified against real `aws-sdk-go-v2/sqs` and at the provider level:
  - two groups deliver in parallel (a + x), b never returned while a is in flight; deleting a unblocks b.
  - visibility EXPIRY (not delete): a is redelivered as the head and b stays blocked — only deleting a advances the group. Confirmed with FakeClock.
  - 3+ messages in one group: exactly one in flight; each delete advances exactly one.
  - `MaxNumberOfMessages=10` on a FIFO queue returns at most one message per group in a single batch (never a+b together).
  - non-FIFO queue unchanged (multiple messages returned in one batch).

**Long-poll / concurrency**
- Lock discipline is correct: `receiveOnce` takes/releases `qd.mu` per attempt (deferred unlock); the wait (`time.After(50ms)`) happens in `pollForMessages` OUTSIDE the lock. A concurrent send is not blocked while a caller long-polls — verified by the mid-window pickup test.
- ctx cancellation returns promptly via `select { case <-ctx.Done() }`; the wire handler passes `r.Context()` (signature changed `_ ctx` → real `ctx`, all callers pass a real context).
- Empty-queue full-window and mid-window-pickup behaviors verified; 50ms interval bounds CPU; loop terminates on deadline and ctx, no leaked goroutine.
- `-race -count=2` on `server/aws/sqs` + `providers/aws/sqs` is clean. `emitReceiveMetrics` reads `qd.info.Name` lock-free, which is safe: Name is immutable after CreateQueue.

**WaitTimeSeconds=0 nuance (flagged): not a SDK-reachable bug.**
The driver `WaitTimeSeconds` is a plain int, so the server cannot distinguish an explicit 0 from omitted. However `aws-sdk-go-v2` serializes `WaitTimeSeconds int32` with `if v.WaitTimeSeconds != 0`, so an explicit 0 is omitted on the wire — any server sees absence. The real-emulator surface therefore behaves correctly; only a typed-Go-API caller who explicitly passes `WaitTimeSeconds=0` against a queue with a non-zero `ReceiveMessageWaitTimeSeconds` would wait. Acceptable documented edge (a `*int` on the driver would fully fix it) — non-blocking.

**Gates (GOMAXPROCS=3)**: build ✓ · vet ✓ · sqs tests -count=1 ✓ · -race -count=2 ✓ · root `go test .` ✓ · gofmt clean ✓ · golangci-lint `--new-from-rev=origin/development` 0 new ✓ · `go generate` zero diff ✓ · compatgen exit 0 + `docs/compat` zero diff ✓.

NitinKumar004 (commented):
Delta re-review — verdict: MERGE. Verified against pushed head c4fea3b2.

FIFO semantics now correct:
- Multiple messages per group within one call: a FIFO queue with 5 msgs of one group + ReceiveMessage(Max=10) returns all 5 in send order (TestSDKFIFOBatchReturnsConsecutiveSameGroup). This was the regression; it is fixed.
- Cross-package SNS->FIFO-SQS fanout (TestSDKSNSFifoTopicToFifoQueueDelivers) passes.
- Across-call in-flight blocking preserved: receive(Max=1)->a; receive->[] while a in-flight; delete a -> receive->b.
- Two groups deliver in parallel (G1 in-flight does not block G2).

The blocked-group set is computed correctly: fifoInFlightGroups(qd, now) is evaluated once at the top of collectVisibleMessages, before the receive marks anything, and the collection loop only reads it — it is never mutated mid-iteration. So a group is blocked only by messages already in-flight (VisibleAt.After(now)) at the start of the call, and a fresh group still yields its consecutive messages within the batch. Non-FIFO queues get an empty (never nil) blocked set, so behavior is unchanged. In-group ordering is preserved by in-order iteration over qd.messages.

Long poll: pollForMessages re-acquires the lock inside receiveOnce each attempt and releases it before the time.After wait — the queue lock is never held across the sleep; resolveWaitDuration takes/releases the lock separately. ctx cancellation aborts promptly; empty window returns [] near full window; mid-window send is picked up.

Gates (GOMAXPROCS=3): build, vet clean; go test ./server/aws/sqs ./providers/aws/sqs ./server/aws/sns pass; -race -count=2 clean; gofmt clean; golangci-lint --new-from-rev=origin/development 0 new; go generate no diff; compatgen exit 0, docs/compat no diff. Diff limited to providers/aws/sqs/sqs.go + the new test file.

PR #732, opened 2026-08-26
fix(gcp-cloudrun): honor updateMask on service PATCH; model resources.limits + template labels/annotations

NitinKumar004:
## Summary

Fixes four GCP Cloud Run (`run.googleapis.com` v2) Services bugs that caused data loss and Terraform/gcloud round-trip drift.

### 1. [HIGH data-loss] Field-masked PATCH ignored `updateMask` and wiped unspecified fields
A `PATCH .../services/{id}?updateMask=traffic` carrying only `traffic` (what `gcloud run services update-traffic` and field-masked SDK clients send) previously overwrote every mutable field from the partial body, wiping `template.containers` etc. Now `updateService` parses `updateMask` and the driver merges only the masked paths onto the existing service, preserving everything unmasked. A maskless PUT (Terraform's `google_cloud_run_v2_service`) still full-replaces. Supported masks: `traffic`, `template` (and dotted `template.<sub>`), `labels`, `annotations`, `ingress`, `description`.

### 2. [LOW] Spurious revision on template-unchanged update
Every update materialized a new revision and advanced `latestCreated`/`latestReady`, even for a traffic-only change. Now a new revision is cut only when the revision template actually changes; a traffic-only update updates traffic without a new revision.

### 3. [HIGH drift] `template.containers[].resources.limits` not modeled
`resources` (limits map + `cpuIdle`/`startupCpuBoost`) is now modeled end to end (wire, driver, conv, clone) and round-trips on create->get and update.

### 4. `template.labels` / `template.annotations` dropped
`revisionTemplate` labels/annotations (where `autoscaling.knative.dev/maxScale` etc. live) are now mapped through the driver and emitted on Get.

## Tests
New real-SDK (`google.golang.org/api/run/v2`) + raw field-masked PATCH e2e tests against a booted `gcpserver.New`:
- masked `updateMask=traffic` PATCH preserves the container template and updates traffic (B1)
- traffic-only update cuts no new revision, `latestCreatedRevision` unchanged (B2)
- `resources.limits{cpu,memory}` round-trips on create+get (B3)
- `template.labels`/`annotations` round-trip (B4)
- maskless full PUT still full-replaces, clearing omitted fields (regression)

### DIFF
--- a/providers/gcp/cloudrun/cloudrun.go
+++ b/providers/gcp/cloudrun/cloudrun.go
@@ -291,13 +291,35 @@ func cloneContainers(in []driver.Container) []driver.Container {
 	out := make([]driver.Container, len(in))
 	for i := range in {
 		out[i] = driver.Container{
-			Name:    in[i].Name,
-			Image:   in[i].Image,
-			Command: append([]string(nil), in[i].Command...),
-			Args:    append([]string(nil), in[i].Args...),
-			Env:     cloneMap(in[i].Env),
-			Ports:   append([]int(nil), in[i].Ports...),
+			Name:      in[i].Name,
+			Image:     in[i].Image,
+			Command:   append([]string(nil), in[i].Command...),
+			Args:      append([]string(nil), in[i].Args...),
+			Env:       cloneMap(in[i].Env),
+			Ports:     append([]int(nil), in[i].Ports...),
+			Resources: cloneResources(in[i].Resources),
 		}
+	}
+
+	return out
+}
+
+// cloneResources deep-copies a container's resource requirements, or nil.
+func cloneResources(in *driver.ResourceRequirements) *driver.ResourceRequirements {
+	if in == nil {
+		return nil
+	}
+
+	out := &driver.ResourceRequirements{Limits: cloneMap(in.Limits)}
+
+	if in.CPUIdle != nil {
+		v := *in.CPUIdle
+		out.CPUIdle = &v
+	}
+
+	if in.StartupCPUBoost != nil {
+		v := *in.StartupCPUBoost
+		out.StartupCPUBoost = &v
 	}
 
 	return out
--- a/providers/gcp/cloudrun/services.go
+++ b/providers/gcp/cloudrun/services.go
@@ -3,6 +3,7 @@ package cloudrun
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"time"
 
 	cerrors "github.com/stackshy/cloudemu/v2/errors"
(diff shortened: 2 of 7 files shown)


NitinKumar004 (commented):
**Deep review — verdict: approve (MERGE on green CI).**

Reviewed the pushed head `27ac389`, clean-built and driven with the real `run/v2` SDK + raw field-masked PATCH against a booted `gcpserver.New`. The data-loss fix holds up and the four bugs are addressed correctly.

**Data-loss (updateMask) — confirmed fixed**
- `PATCH ?updateMask=traffic` carrying only traffic preserves the full container template (image/env/serviceAccount) and updates traffic — no wipe.
- `updateMask=labels` updates service labels while template + description + serviceAccount + ingress all survive.
- `updateMask=template.containers` swaps the image while traffic + service labels survive.
- Multi-mask `updateMask=traffic,labels` applies both, template preserved.
- Set-to-value semantics verified: `updateMask=description` with an empty value clears description (not merge-if-present).
- Maskless full PUT still full-replaces (omitted serviceAccount cleared, generation advances) — the Terraform `google_cloud_run_v2_service` path is unregressed. The empty-mask vs non-empty-mask split in `applyUpdate` is the right seam.

**No-spurious-revision (B2)** — `snapshotTemplate` + `reflect.DeepEqual` compare is sound. Traffic-only and labels-only updates cut no revision and leave `latestCreatedRevision` unchanged while still updating traffic status via `reconcileStatus`; a template-changing update advances the pointers and adds a revision. No false-positive: a no-op masked update applied twice still yields exactly one revision (both sides normalize empty→nil through `cloneContainers`, so nil-vs-empty does not spuriously differ). No false-negative on the paths tested.

**resources.limits (B3)** — `{cpu,memory}` limits plus `cpuIdle`/`startupCpuBoost` round-trip on create→get, survive a subsequent masked update, and land on the materialized revision. Wire JSON tags (`cpuIdle`/`startupCpuBoost`) match `run/v2`.

**template labels/annotations (B4)** — `revisionTemplate.labels`/`annotations` round-trip and stay distinct from service-level labels/annotations (`hasTemplate` vs `has` keep the two planes separate).

**Blast radius** — the additive `driver.Container.Resources` pointer and `ServiceConfig.UpdateMask`/`TemplateLabels`/`TemplateAnnotations` fields are backward-compatible; jobs share the same conv/clone paths and are unaffected (nil default). `cloneResources` deep-copies the limits map and both bool pointers.

**Gates (GOMAXPROCS=3):** build, vet, `-race` on providers/gcp/cloudrun, server+provider+services cloudrun tests, root integration tests — all pass. gofmt clean; `golangci-lint --new-from-rev=origin/development` 0 new issues; `go generate ./...` zero diff; compatgen + `docs/compat/` zero diff.

No blocking issues.

PR #734, opened 2026-08-26
fix(dynamodb): parallel-scan segments, Select mode, numeric key identity, transact idempotency+limit

NitinKumar004:
## Summary

Fixes five AWS DynamoDB correctness bugs, verified end-to-end against the real `aws-sdk-go-v2/dynamodb` client driving the emulator's HTTP server.

### 1. Parallel Scan `Segment`/`TotalSegments` (HIGH, data-duplication)
`Scan` ignored the parallel-scan parameters, so every segment returned the whole table (4 segments × 10 items = 40, each item 4×), breaking parallel-scan ETL/worker patterns. Now `Segment`/`TotalSegments` thread through `ScanInput` into the provider, which keeps an item only when a stable FNV-1a hash of its (canonical) primary key falls in the segment. The union of all segments covers every item **exactly once** — no duplicates, no skips. Validated (`Segment` required with `TotalSegments`; `0 <= Segment < TotalSegments`; `TotalSegments` 1..1,000,000 → `ValidationException`).

### 2. `Select` mode on Scan + Query (MED)
`Select` was ignored. Now `Select=COUNT` returns `Count`/`ScannedCount` with an empty `Items` array; `SPECIFIC_ATTRIBUTES` requires a `ProjectionExpression`; `ALL_PROJECTED_ATTRIBUTES` is only valid on an index; and `Select`/`ProjectionExpression` conflicts (e.g. `ALL_ATTRIBUTES` + a projection) are `ValidationException`s.

### 3. Numeric key identity normalization (MED, data-correctness)
`PutItem` "100" then "100.0" created two distinct items because the key string used the raw value text. DynamoDB compares Number keys numerically, so both (and "1e2") must map to one item. `itemKey` and `keyIdentity` now canonicalize an `expr.Number` via its exact rational form (reusing the same big.Rat comparison the Number-set dedup relies on). String/Binary keys are unaffected. Applies to hash and composite sort keys, and to Get/Delete/Batch/Transact resolution.

### 4. `TransactWriteItems` `ClientRequestToken` idempotency (MED)
A retried transaction re-applied its writes (e.g. `ADD cnt :one` applied 3× → 3 instead of 1). The provider now remembers recent tokens (10-minute window, mirroring the SQS FIFO dedup index) and short-circuits a replay, returning the cached success without re-applying. The same token with a **different** request body is an `IdempotentParameterMismatchException`.

### 5. Transaction 100-op limit (LOW)
`TransactWriteItems` and `TransactGetItems` now reject more than 100 operations with a `ValidationException` before applying anything (mirroring the existing BatchWrite=25 / BatchGet=100 checks).

## Tests
New real-SDK e2e tests: parallel-scan segment union is disjoint and complete; `Select=COUNT` returns empty Items; numeric-key put "100"/"100.0" leaves one item that GetItem "100"/"1e2" resolves; token-idempotent transact replays increment once (and mismatch is rejected); 101-op transaction is a `ValidationException` that writes nothing.

## Gates
`go build`, `go vet`, targeted `go test` (+ `-race` on the provider), `gofmt`, `go generate` (zero diff), `compatgen` (zero diff) all green; no new golangci-lint findings in the changed packages.

### DIFF
--- a/providers/aws/dynamodb/conditional.go
+++ b/providers/aws/dynamodb/conditional.go
@@ -2,13 +2,30 @@ package dynamodb
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
 	"maps"
 	"strings"
+	"time"
 
 	cerrors "github.com/stackshy/cloudemu/v2/errors"
 	"github.com/stackshy/cloudemu/v2/services/database/driver"
 	"github.com/stackshy/cloudemu/v2/services/database/driver/expr"
 )
+
+// transactIdempotencyWindow is how long a TransactWriteItems ClientRequestToken
+// is remembered so a retry with the same token is treated as a replay rather
+// than a second application (real DynamoDB uses ~10 minutes).
+const transactIdempotencyWindow = 10 * time.Minute
+
+// txIdempotencyRecord is a remembered ClientRequestToken: the fingerprint of the
+// request it first committed and when, so a replay with a matching body is a
+// no-op while a different body with the same token is a parameter mismatch.
+type txIdempotencyRecord struct {
+	fingerprint string
+	storedAt    time.Time
+}
 
 // This file holds the ATOMIC conditional-write and transaction primitives. The
 // wire handler must NOT evaluate a ConditionExpression with a standalone GetItem
@@ -244,12 +261,16 @@ func updateBaseImage(item map[string]any, present bool, key map[string]any) (bas
 // a single hold of m.mu, so the transaction is all-or-nothing and isolated from
 // concurrent single-item writes. On any failed condition it returns a
 // *driver.TransactionCanceled naming the failed operations and writes nothing.
-func (m *Mock) TransactWrite(ctx context.Context, ops []driver.TransactOp) error {
+func (m *Mock) TransactWrite(ctx context.Context, ops []driver.TransactOp, clientRequestToken string) error {
 	m.mu.Lock()
 	// flush registers before the unlock defer so it runs after the lock is
 	// released (defers are LIFO), delivering stream records outside m.mu.
 	defer func() { m.flushStreamDeliveries(ctx) }()
 	defer m.mu.Unlock()
+
+	if done, err := m.checkTransactToken(clientRequestToken, ops); done {
+		return err
+	}
 
 	tds, err := m.resolveTransactTables(ops)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


NitinKumar004 (commented):
**Verdict: MERGE** — all five DynamoDB fixes are correct, additive, and AWS-scoped; full gate matrix green on the pushed head (bcb3997).

**Diff scope**: exactly the 9 expected dynamodb/database/expr files, pure data/expr logic — no network/exec/exfil (the security-classifier warning is a confirmed false positive).

**B1 parallel scan** — `itemInSegment` hashes `itemKey` (FNV-1a of the canonical primary key) mod TotalSegments, so each item maps to exactly one shard; the union covers the table exactly once with no dup/skip. Verified with uneven counts (10/4 in-PR, plus 13/3) and segment assignment proven stable across repeated calls. Segment/TotalSegments validation (both-or-neither, 1..1e6, 0<=Segment<Total) correct; ordinary scan + pagination unaffected. `int64(h.Sum32())%int64(*total)` has no narrowing.

**B3 numeric key identity** — `expr.CanonicalKey` collapses N spellings via `big.Rat.RatString()` and is used consistently in `itemKey` (provider) and `keyIdentity` (wire batch/transact dedup). "100"/"100.0"/"1e2" resolve to one item; composite keys canonicalize on both parts; S/B keys unaffected (`fmt.Sprintf %v` path — string "100" != "100.0", stay 2 items). A 38-digit N key round-trips exactly and a differing 38-digit value does not collide — no precision loss.

**B4 transact idempotency** — token cache guarded by `m.mu` (held for the whole TransactWrite), so replays serialize: same token+body short-circuits (ADD applied once), same token+different body -> IdempotentParameterMismatchException, no token applies each time. `pruneTransactTokens` evicts >10min entries (no unbounded growth). 30 concurrent same-token replays under -race increment the counter exactly once with zero data races.

**B2 Select** — COUNT empties Items while Count/ScannedCount stay correct (Scan+Query); SPECIFIC_ATTRIBUTES requires ProjectionExpression; Select+ProjectionExpression conflict and ALL_PROJECTED-on-base-table rejected.

**B5** — TransactWrite and TransactGet both reject >100 ops with ValidationException at the wire layer before any apply.

**Prior fixes unregressed** — conditional-write TOCTOU, TransactWrite atomicity/CancellationReasons, NS/scalar-N precision all still pass. Blast radius clean: `TransactWrite` signature change is AWS-only (no Azure/GCP/chaos/persist callers); ScanInput Segment/TotalSegments and the new IdempotentParameterMismatch type are additive — all consumers still build.

**Gates (GOMAXPROCS=3)**: build, vet, targeted dynamodb/database tests, `-race -count=2` providers, root test, gofmt, `golangci-lint --new-from-rev` (0 new), `go generate` zero diff, compatgen exit 0 + docs/compat zero diff — all green.

PR #735, opened 2026-08-26
fix(s3): system object metadata, storage class, GetObjectAttributes, HeadBucket region

NitinKumar004:
## Summary

Fixes four AWS S3 fidelity gaps around object system metadata, storage class, and bucket/object attribute reads.

1. **System-defined object metadata** — `Cache-Control`, `Content-Encoding`, `Content-Disposition`, `Content-Language`, and `Expires` are now recorded on `PutObject` and returned on `GetObject`/`HeadObject`. `CopyObject`/`CopyObjectV2` inherit them from the source (unless `REPLACE` overrides). Carried through a new optional `driver.SystemPropsBucket` capability so the shared `Bucket.PutObject` signature is unchanged and GCS/Azure Blob are unaffected.
2. **Storage class** — persisted from the `x-amz-storage-class` request header, echoed in `ListObjectsV2`/`ListObjectVersions`, and emitted on read as `x-amz-storage-class` for non-STANDARD classes only (S3 omits it for STANDARD).
3. **GetObjectAttributes** — new `GET /{bucket}/{key}?attributes` branch returning `ETag`, `ObjectSize`, and `StorageClass` per the `x-amz-object-attributes` header, plus `Last-Modified`/version-id headers. Missing key → `NoSuchKey`, missing bucket → `NoSuchBucket`.
4. **HeadBucket** — returns `x-amz-bucket-region` on 200 (SDK `HeadBucketOutput.BucketRegion`).

## Shared-driver safety

`driver.PutObject`/`ObjectInfo` stay backward-compatible: system props travel through the optional `SystemPropsBucket` capability and additive struct fields. GCS and Azure Blob providers compile unchanged and their tests pass — verified in the gate below.

## Tests

Real `aws-sdk-go-v2/s3` against a booted `httptest` server (`server/aws/s3/system_props_sdk_test.go`):
- system metadata round-trips PUT→HEAD and CopyObject inherits Cache-Control
- StorageClass=STANDARD_IA reported on HeadObject + ListObjectsV2; STANDARD omits the header
- GetObjectAttributes populates ETag/ObjectSize/StorageClass; missing key → NoSuchKey
- HeadBucket returns BucketRegion

Gates: build, vet, `-p=3` tests (AWS S3 + storage + GCS + Azure Blob), `-race`, gofmt, golangci-lint (0 new), `go generate` (coverage docs regenerated), compatgen (no compat diff).

### DIFF
--- a/docs/coverage/aws/s3.md
+++ b/docs/coverage/aws/s3.md
@@ -163,6 +163,14 @@ StorageAccountKeys is an OPTIONAL Azure-specific capability, discovered by
 | `ListStorageAccountKeys` | ListStorageAccountKeys returns the account's access keys, generating a |
 | `RegenerateStorageAccountKey` | RegenerateStorageAccountKey rotates the value of the named key (key1/key2) |
 
+### SystemPropsBucket
+
+SystemPropsBucket is an OPTIONAL capability (discovered by type assertion like
+
+| Operation | Description |
+| --- | --- |
+| `PutObjectWithSystemProps` |  |
+
 ### VersionedBucket
 
 VersionedBucket is an optional extension a storage provider implements when
--- a/docs/coverage/azure/blobstorage.md
+++ b/docs/coverage/azure/blobstorage.md
@@ -163,6 +163,14 @@ StorageAccountKeys is an OPTIONAL Azure-specific capability, discovered by
 | `ListStorageAccountKeys` | ListStorageAccountKeys returns the account's access keys, generating a |
 | `RegenerateStorageAccountKey` | RegenerateStorageAccountKey rotates the value of the named key (key1/key2) |
 
+### SystemPropsBucket
+
+SystemPropsBucket is an OPTIONAL capability (discovered by type assertion like
+
+| Operation | Description |
+| --- | --- |
+| `PutObjectWithSystemProps` |  |
+
 ### VersionedBucket
 
 VersionedBucket is an optional extension a storage provider implements when
--- a/docs/coverage/coverage.json
+++ b/docs/coverage/coverage.json
@@ -10851,6 +10851,15 @@
         ]
       },
       {
+        "name": "SystemPropsBucket",
+        "doc": "SystemPropsBucket is an OPTIONAL capability (discovered by type assertion like",
+        "operations": [
+          {
+            "name": "PutObjectWithSystemProps"
+          }
+        ]
+      },
+      {
         "name": "VersionedBucket",
         "doc": "VersionedBucket is an optional extension a storage provider implements when",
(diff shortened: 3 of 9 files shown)


NitinKumar004 (commented):
**Verdict: MERGE**

Clean-built and gate-verified against the pushed head `d567cc48`.

**What was checked**
- `go build ./...`, `go vet` (target pkgs): clean.
- `go test -p=3` on `server/aws/s3`, `providers/aws/s3`, `services/storage/...`, `providers/gcp/gcs`, `providers/azure/blobstorage`: all pass. `-race` on `providers/aws/s3`: clean.
- `gofmt -l`: clean. `go generate ./...`: zero diff. `go run ./internal/compatgen` + `git diff docs/compat/`: exit 0, zero diff.
- `golangci-lint --new-from-rev=origin/development` on changed pkgs: 0 new issues.

**Behavior verified (real aws-sdk-go-v2/s3)**
- System metadata (Cache-Control/Content-Encoding/Content-Disposition/Content-Language/Expires) round-trips PUT→HEAD/GET; default CopyObject inherits, REPLACE overrides via `copyDstSystemProps`.
- StorageClass persisted and reported on HeadObject + ListObjectsV2; STANDARD correctly OMITS the `x-amz-storage-class` header on read (`writeSystemPropHeaders`) while listings still show `STANDARD` (`storageClassOrDefault`) — matches real S3.
- GetObjectAttributes returns populated ETag/ObjectSize/StorageClass (ObjectSize is `*int64` so a 0-byte object still emits `<ObjectSize>0`); missing key → NoSuchKey, missing bucket → NoSuchBucket (via `bucketMissing`); `?attributes` branch is isolated so a plain GET still returns the body.
- HeadBucket sets `x-amz-bucket-region` (us-east-1 default).

**Shared-driver safety (critical): PASS**
- `Bucket.PutObject`/`ObjectInfo` signatures unchanged. New behavior travels through the additive optional `SystemPropsBucket` capability + additive struct fields (`ObjectInfo.Expires`, `ObjectVersion.StorageClass`, `CopyObjectRequest.SystemProps`). GCS + Azure Blob build unchanged and their provider tests pass; those providers do not implement `SystemPropsBucket`, so their object-metadata behavior is inert.

**Non-blocking observation**
- A PutObject carrying BOTH a conditional header (If-None-Match/If-Match) AND system props routes to `storePutConditional` → `PutObjectConditional`, which does not carry `ObjectSystemProps`, so system metadata/storage-class is dropped for that specific combination. Uncommon intersection and not a regression (conditional writes never modeled system props). Worth a follow-up but does not block this PR.

PR #736, opened 2026-08-26
fix(azure-functions): model site kind and managed identity on Microsoft.Web/sites

NitinKumar004:
## Summary

Two Azure Functions (`Microsoft.Web/sites`) bugs surfaced by driving the real `armappservice/v3` SDK against a booted `azureserver`:

1. **`kind` always overwritten to `functionapp`.** A site PUT with `kind="app,linux"` (or any non-default) came back as `functionapp` on both create and GET because `toSiteResource` hardcoded the kind. The request kind is now stored on `SiteMeta.Kind` and echoed, falling back to `functionapp` only when the request omits kind.

2. **Managed identity not modeled (hard failure for Terraform).** A PUT with `identity:{type:SystemAssigned}` dropped the `identity` envelope entirely, so azurerm's `identity[0].principal_id` hit index-out-of-range. The top-level `identity` field is now modeled:
   - **SystemAssigned** synthesizes a deterministic `principalId` + emulator `tenantId` (via `idgen.SyntheticGUID`, matching the VM/AKS/ACR pattern), stable across GET/List.
   - **UserAssigned** echoes each submitted identity key back with a synthesized `principalId`/`clientId`.
   - **SystemAssigned,UserAssigned** does both.
   - `None`/omitted yields no identity block.

The `echoUnmodeledProperties` overlay only rescues unmodeled `properties.*`, not top-level envelope fields, so both required real modeling. `echo_properties.go` is untouched (the #715 secret-stripping stays intact).

Deferred (unchanged): PATCH update, connection strings, publish profile, slots, source control.

## Tests

New real-SDK e2e (`sites_identity_sdk_test.go`): kind round-trip; kind default when omitted; SystemAssigned deterministic principal/tenant GUIDs across GETs; UserAssigned key echo; no-identity nil; and a #715 regression asserting an app-setting secret is still not exposed on a plain GET.

### DIFF
--- a/providers/azure/functions/site_meta.go
+++ b/providers/azure/functions/site_meta.go
@@ -6,8 +6,10 @@ import (
 	"encoding/base64"
 	"maps"
 	"sort"
+	"strings"
 
 	cerrors "github.com/stackshy/cloudemu/v2/errors"
+	"github.com/stackshy/cloudemu/v2/internal/idgen"
 	"github.com/stackshy/cloudemu/v2/services/serverless/driver"
 )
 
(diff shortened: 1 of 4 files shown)


NitinKumar004 (commented):
Gates on the pushed SHA (`abd6fca9`) are all green: `go build`/`go vet` clean, `go test` + `go test -race` on `server/azure/functions` and `providers/azure/functions` pass, `gofmt` clean, `golangci-lint --new-from-rev` 0 new issues, `go generate ./...` zero diff, `compatgen` + `docs/compat/` zero diff, `go mod tidy` zero diff, `contrib/testcontainers` builds+vets. Diff is exactly the 4 files claimed.

Verified against a real `armappservice/v3` client on a booted server: kind round-trips on create+GET, defaults to `functionapp` when omitted, and is reflected in List; SystemAssigned identity synthesizes deterministic GUIDs stable across repeated GETs and List; UserAssigned echoes the submitted key with synthesized ids; `SystemAssigned,UserAssigned` combined works; identity is nil when type is omitted or explicitly `"None"`; `SiteMeta.clone()`/`SiteIdentity.clone()` deep-copy with no aliasing across GETs; the #715 regression (app-setting secret not on a plain GET) still holds, `echo_properties.go` untouched.

One real gap: `kind` is not preserved on an update PUT that omits it, unlike `Identity`.

In `UpsertSiteMeta`'s update branch (`providers/azure/functions/site_meta.go:171`):

```go
existing.Kind = in.Kind
```

this overwrites unconditionally, while two lines below `Identity` is deliberately guarded:

```go
// A nil in.Identity means the request omitted the identity block, so the
// already-attached identity is preserved (real ARM PUT semantics).
if in.Identity != nil {
    existing.Identity = resolveSiteIdentity(in.Identity, identitySeed(&in))
}
```

Reproduced: create a site with `kind="app,linux"`, then send an update PUT that only changes app settings (kind omitted, so it decodes to `""`) — both the update response and a subsequent GET report `kind="functionapp"` again, silently losing the value just set.

Since `Kind`'s zero value (`""`) is unambiguous with "omitted" the same way a nil `*SiteIdentity` is, the fix is the same shape already used for identity: guard the assignment with `if in.Kind != "" { existing.Kind = in.Kind }` (or preserve-by-default and only overwrite on a non-empty value). Given the PR's own rationale for Identity — a `Get → mutate → Put` client shouldn't lose fields it didn't touch — `kind` should get the same treatment; right now it doesn't.

NitinKumar004 (commented):
Delta re-review (head 020c0f41).

Verified only the new commit (`020c0f41`) vs the prior-reviewed state (`abd6fca9`, kind+identity feature). Diff is exactly:

- `providers/azure/functions/site_meta.go:177-181` — update path now guards `if in.Kind != "" { existing.Kind = in.Kind }`, mirroring the existing `if in.Identity != nil` guard right below it. Confirmed:
  - Update PUT omitting kind (only touching app settings) preserves the existing kind — the prior bug is fixed.
  - Create path (`meta := in.clone()` at line 194) still sets `meta.Kind = in.Kind` unconditionally, so an explicit kind on create still lands.
  - Create with no kind stores `Kind: ""`; the "functionapp" default is applied at the response layer (`toSiteResource` in `server/azure/functions/handler.go`), unchanged by this delta — still correct.
- `server/azure/functions/sites_identity_sdk_test.go` — new `TestSDKSiteKindPreservedOnUpdateOmit`: create with kind="app,linux" → update PUT omitting kind → GET still returns "app,linux". Ran with `-race`, passes.

No other files touched. Previously-reviewed behavior (identity synthesis, kind round-trip on create, #715 secret-leak regression test) is untouched by this commit.

Gates on pushed SHA (GOMAXPROCS=3):
- `go build ./...` clean
- `go vet ./...` clean
- `go test -p=3 ./server/azure/functions/... ./providers/azure/functions/...` — pass
- `go test -race -p=3 ./providers/azure/functions/...` and targeted `-race` run of the new test — pass
- `gofmt -l` — no diffs
- `golangci-lint --new-from-rev=origin/development` on the two changed packages — 0 new issues
- `go run ./internal/compatgen` + `git diff docs/compat/` — zero diff

Verdict: MERGE.

PR #737, opened 2026-08-26
fix(gcp-clouddns): validate zone/record semantics and round-trip dnssec + private visibility

NitinKumar004:
## Summary

Fixes seven real-cloud-parity bugs in the GCP Cloud DNS (dns.googleapis.com v1) wire handler, verified end-to-end against the real `google.golang.org/api/dns/v1` SDK pointed at a booted `gcpserver`.

### Validation / semantics
- **Non-empty zone delete (data-shaped):** `managedZones.delete` on a zone that still holds user record sets now returns `400 containerNotEmpty` instead of silently cascading every record and returning 200. Only the auto-created apex SOA/NS may remain for a delete to succeed.
- **CNAME coexistence:** a change that would leave any name with a CNAME beside another record type (in either add order) is rejected `400`.
- **Apex NS/SOA protection:** a plain deletion of the apex NS or SOA rrset is rejected `400`. A delete paired with a re-add of the same rrset (the canonical apex update) is still allowed.
- **Fresh-zone change log:** zone creation now logs the initial apex-provisioning change (id `"0"`), so `changes.list` on a fresh zone returns it as real Cloud DNS does.
- **Change-id sequencing:** change ids are now numbered per-zone from `"0"` via the change log, independent of the managed-zone Operation-id counter, so change ids no longer skip after a zone patch.

### Round-trip (Terraform drift)
- **dnssecConfig** (state / nonExistence / defaultKeySpecs) now round-trips on create, get, and patch.
- **privateVisibilityConfig.networks[].networkUrl** now round-trips.

Both are modeled on `driver.ZoneInfo`/`ZoneConfig` (GCP-only, other providers leave them nil), mirroring the existing `VPCAssociation`/`AliasTarget` pattern.

### Tests
New real-SDK e2e coverage in `validation_sdk_test.go` for each of the seven behaviors, plus an update to `TestSDKChangesListAndGet` for the seeded initial change.

Out of scope (not implemented): forwardingConfig/peeringConfig/reverseLookup/cloudLogging config, rrsets direct create/patch/delete API, DNS Policies, Response Policies.

### DIFF
--- a/providers/gcp/clouddns/dns.go
+++ b/providers/gcp/clouddns/dns.go
@@ -46,6 +46,51 @@ func recordKey(zoneID, name, recordType, setID string) string {
 	return key
 }
 
+// cloneDNSSEC deep-copies a DNSSEC config so the stored zone does not alias the
+// caller's value. Returns nil for a nil input.
+func cloneDNSSEC(in *driver.DNSSECConfig) *driver.DNSSECConfig {
+	if in == nil {
+		return nil
+	}
+
+	out := &driver.DNSSECConfig{State: in.State, NonExistence: in.NonExistence}
+	if in.DefaultKeySpecs != nil {
+		out.DefaultKeySpecs = append([]driver.DNSKeySpec(nil), in.DefaultKeySpecs...)
+	}
+
+	return out
+}
+
+// cloneNetworks copies a visibility-network slice so the stored zone does not
+// alias the caller's slice. Returns nil for a nil input.
+func cloneNetworks(in []driver.VisibilityNetwork) []driver.VisibilityNetwork {
+	if in == nil {
+		return nil
+	}
+
+	return append([]driver.VisibilityNetwork(nil), in...)
+}
+
+// mergeDNSSEC returns the incoming DNSSEC config (cloned) when a patch carries
+// one, else preserves the existing value — so a patch that omits it is a no-op.
+func mergeDNSSEC(existing, incoming *driver.DNSSECConfig) *driver.DNSSECConfig {
+	if incoming == nil {
+		return existing
+	}
+
+	return cloneDNSSEC(incoming)
+}
+
+// mergeNetworks returns the incoming visibility networks (cloned) when a patch
+// carries them, else preserves the existing slice.
+func mergeNetworks(existing, incoming []driver.VisibilityNetwork) []driver.VisibilityNetwork {
+	if incoming == nil {
+		return existing
+	}
+
+	return cloneNetworks(incoming)
+}
+
 // CreateZone creates a new Cloud DNS managed zone.
 func (m *Mock) CreateZone(_ context.Context, cfg driver.ZoneConfig) (*driver.ZoneInfo, error) {
 	if cfg.Name == "" {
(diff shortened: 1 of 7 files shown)


NitinKumar004 (commented):
Reviewed at head `5f8096ab`, clean-built and driven against the real `google.golang.org/api/dns/v1` SDK on a booted `gcpserver`. Verdict: LGTM.

**Delete guard (data-shaped)** — `deleteZone` calls `rejectIfNotEmpty` *before* `dns.DeleteZone`, so the provider's cascading delete is unreachable while the zone holds any non-apex rrset (400 `containerNotEmpty`, zone + records intact). `isApexRRSet` treats only `Name==dnsName && Type∈{NS,SOA}` as non-blocking; a delegation NS/SOA at a non-apex name still blocks (verified with an added `sub.<dns>` NS → 400). Apex-only and apex+records-then-cleared zones delete cleanly. No path deletes a non-empty zone.

**CNAME coexistence** — `checkCNAME` evaluates `postApplyTypes` (current − deletions + additions), so it's order-independent and does NOT false-reject a change that deletes the conflicting record in the same batch. Verified: A-then-CNAME → 400, CNAME-then-A → 400, standalone CNAME OK, and A→CNAME swap (delete A + add CNAME, same change) → allowed.

**Apex NS/SOA protection** — plain delete of apex NS or SOA → 400 `invalid`; a delete paired with a re-add of the same apex rrset (canonical apex update) is exempted via the `adding` set and passes (verified NS TTL update in one change → 200). Not over-blocked.

**Change log / id sequencing** — `seedApexRecords` logs the initial apex change id `"0"`; `recordChange` assigns per-zone log-index ids under `mu`. `opSeq` (managed-zone Operations) is split from the change log, so a zone patch no longer consumes a change id — verified ids stay `0,1,2` across a patch. `changeSeq` is fully removed. The `TestSDKChangesListAndGet` bump to 2 (seed + created) is the correct intended behavior, matching real Cloud DNS's initial-provisioning change, not a masked regression.

**Round-trips** — `dnssecConfig{state,nonExistence,defaultKeySpecs}` echoes on create/get and is preserved across a patch that omits it (`mergeDNSSEC`); `privateVisibilityConfig.networks[].networkUrl` echoes. Modeled GCP-only on `driver.ZoneInfo/ZoneConfig` (nil for other providers), mirroring the `VPCAssociation`/`AliasTarget` pattern; deep-cloned so stored zones don't alias caller input.

**Regression & scope** — full clouddns package tests pass (prior #259/#321/dup-zone-409/numeric-id/rrsets.list filters+pagination intact). Changes limited to the 7 expected files.

Gates on the pushed SHA (GOMAXPROCS=3): `go build ./...`, `go vet ./...`, `go test ./server/gcp/clouddns/... ./providers/gcp/clouddns/... ./services/dns/...`, `go test -race ./providers/gcp/clouddns/...`, `gofmt -l`, `golangci-lint --new-from-rev=origin/development` (0 new — the removed gocyclo was pre-existing), `go generate ./...` (0 diff), `compatgen` + `docs/compat` (0 diff) — all green.

PR #738, opened 2026-08-26
fix(gcp-artifactregistry): round-trip dockerConfig.immutableTags + cleanupPolicies, honor list filter/orderBy

NitinKumar004:
## Summary

Fixes four Artifact Registry (`artifactregistry.googleapis.com` v1) round-trip / list gaps that cause perpetual Terraform drift on `google_artifact_registry_repository`. No logic bugs existed — these were wire-layer fields being dropped or query params being ignored.

## Changes

- **dockerConfig.immutableTags** — was dropped on create → `get` returned `dockerConfig=nil`. Now folded into a reserved tag and re-emitted, so it round-trips on create + get + patch (`updateMask=dockerConfig`).
- **cleanupPolicies + cleanupPolicyDryRun** — both were dropped on create. `cleanupPolicies` now round-trips as opaque JSON in a reserved tag; `cleanupPolicyDryRun` as a bool reserved tag. Honored by the existing updateMask PATCH path.
- **list filter** — `repositories.list?filter=` was ignored. Now honors `name=` (exact, or prefix when the value ends with `*`); unsupported/malformed expressions return the full set unfiltered (lenient, no error) to match API behavior.
- **list orderBy** — `repositories.list?orderBy=` was ignored (always name asc). Now honors `name`, `createTime`, `updateTime`, each with an optional ` desc`.

Implementation follows the package's existing reserved-tag pattern (already used for format/description/mode/kmsKeyName), so create/get/patch stay in lockstep and the driver interface is untouched.

## Tests

Added real `google.golang.org/api/artifactregistry/v1` SDK e2e tests against a booted `gcpserver.New`:
- immutableTags round-trip (create+get+patch)
- cleanupPolicies + cleanupPolicyDryRun round-trip (create+get+patch)
- list filter `name=` narrows results
- list `orderBy=name desc` reorders
- regression: format/mode round-trip + updateMask PATCH + label preservation

## Gates

`go build ./...`, `go vet`, scoped + `-race` tests, `gofmt -l`, golangci-lint (0 issues), `go generate` (zero diff), `compatgen` + `docs/compat` (zero diff) — all green.

### DIFF
--- a/server/gcp/artifactregistry/operations.go
+++ b/server/gcp/artifactregistry/operations.go
@@ -34,11 +34,11 @@ func (h *Handler) createRepository(w http.ResponseWriter, r *http.Request, rt *r
 }
 
 // reservedTagsFrom folds the GCP-only Repository fields (format, description,
-// mode, kmsKeyName) the driver does not model into reserved tags alongside the
-// user labels, so they round-trip on read.
+// mode, kmsKeyName, dockerConfig, cleanupPolicies) the driver does not model
+// into reserved tags alongside the user labels, so they round-trip on read.
 // reservedFieldCount is the number of GCP-only Repository fields folded into
-// reserved tags (format, description, mode, kmsKeyName).
-const reservedFieldCount = 4
+// reserved tags.
+const reservedFieldCount = 7
 
 func reservedTagsFrom(body *repositoryJSON) map[string]string {
 	tags := make(map[string]string, len(body.Labels)+reservedFieldCount)
@@ -62,7 +62,41 @@ func reservedTagsFrom(body *repositoryJSON) map[string]string {
 		tags[kmsKeyTag] = body.KmsKeyName
 	}
 
+	addExtraTags(tags, body)
+
 	return tags
+}
+
+// addExtraTags folds the dockerConfig / cleanupPolicies / cleanupPolicyDryRun
+// fields into reserved tags. Shared with the patch path via the same setters so
+// create and update stay in lockstep.
+func addExtraTags(tags map[string]string, body *repositoryJSON) {
+	setBoolTag(tags, immutableTagsTag, body.DockerConfig != nil && body.DockerConfig.ImmutableTags)
+	setCleanupPolicies(tags, body.CleanupPolicies)
+	setBoolTag(tags, cleanupDryRunTag, body.CleanupPolicyDryRun)
+}
+
+// setBoolTag stores key="true" when val, else drops it.
+func setBoolTag(tags map[string]string, key string, val bool) {
+	if val {
+		tags[key] = trueTag
+		return
+	}
+
+	delete(tags, key)
+}
+
+// setCleanupPolicies stores the cleanupPolicies map as opaque JSON, or drops the
+// reserved tag when empty.
+func setCleanupPolicies(tags map[string]string, policies map[string]json.RawMessage) {
+	if len(policies) == 0 {
+		delete(tags, cleanupPolicyTag)
+		return
+	}
+
+	if b, err := json.Marshal(policies); err == nil {
+		tags[cleanupPolicyTag] = string(b)
+	}
 }
 
 // repositoryTypeURL is the protobuf Any type URL a GAPIC client expects in a
@@ -106,8 +140,9 @@ func (h *Handler) listRepositories(w http.ResponseWriter, r *http.Request, rt *r
 		return
 	}
 
-	page, err := pagination.PaginateSorted(repos,
-		func(a, b crdriver.Repository) bool { return repoName(a.Name) < repoName(b.Name) },
+	repos = filterRepositories(rt, repos, r.URL.Query().Get("filter"))
+
+	page, err := pagination.PaginateSorted(repos, repoLess(rt, r.URL.Query().Get("orderBy")),
 		r.URL.Query().Get("pageToken"), pageSize(r))
 	if err != nil {
 		gcprest.WriteError(w, http.StatusBadRequest, "invalid", "invalid page token")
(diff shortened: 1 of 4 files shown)


NitinKumar004 (commented):
## Review: PR #738 (fix/gcp-artifactregistry-dockerconfig-cleanup-filter)

Reviewed pushed commit `4167690` on the actual pushed tree (build/vet/test/race/lint/generate/compatgen all re-run on this SHA).

### Gates (all green)
- `go test ./server/gcp/artifactregistry/... ./providers/gcp/artifactregistry/...` — all pass, including the 5 new SDK e2e tests (immutableTags round-trip, cleanupPolicies round-trip, list filter, list orderBy, no-regression)
- `go test -race ./providers/gcp/artifactregistry/...` and `... ./server/gcp/artifactregistry/...` — clean
- `gofmt -l` — clean
- `golangci-lint run --new-from-rev=origin/development` on the package — 0 new issues

### Verified working
- `dockerConfig.immutableTags` round-trips create→get, patch(`updateMask=dockerConfig`) toggles it, and a repo that never sets it never gets a spurious `dockerConfig` key on GET (checked the raw JSON body, not just the SDK struct).
- `cleanupPolicies`/`cleanupPolicyDryRun` round-trip create→get, including opaque fields not modeled by any SDK struct (e.g. an unrecognized future field inside a policy survives byte-for-byte) — confirms the "opaque JSON preserved exactly" design.
- `list filter`: `name="<full>"` exact match, trailing-`*` prefix match, and unsupported/malformed filters (e.g. `format="DOCKER"`, `garbage(((`) all correctly fall back to returning the full set unfiltered.
- Reserved `cloudemu:*` tags are stripped from returned `labels` — never leaked.
- No regression: format/mode/description/labels/kmsKeyName round-trip and `updateMask` PATCH still work exactly as before (existing + new regression test both pass).

### Blocking issue found

**[High] `orderBy=createTime` / `orderBy=updateTime` are non-deterministic and cause duplicate/skipped items across pagination** — `server/gcp/artifactregistry/operations.go:219-259` (`repoLess`/`repoSortKey`).

Root cause: `CreatedAt`/`UpdatedAt` are formatted with second granularity (`time.RFC3339`, `providers/gcp/artifactregistry/artifactregistry.go:102`), so any two repositories created within the same second (the normal case for any script/Terraform run) tie on the sort key. `ListRepositories` (`providers/gcp/artifactregistry/artifactregistry.go:181`) builds its slice from `m.repos.All()`, which returns a Go `map[string]V` — iteration order is randomized per call. `pagination.PaginateSorted` stable-sorts that already-randomized input, so ties are preserved in whatever (random) order the map happened to yield *this* call — which differs from the *next* call.

Reproduced directly against the pushed SHA (raw HTTP against a booted `gcpserver`, no test-harness quirk):
- `orderBy=createTime` on 3 repos created back-to-back returns a different first element across repeated identical requests (flips between the actual creation order and an arbitrary one).
- Paginating `orderBy=createTime` 2-at-a-time over 8 repos created back-to-back produces real duplicates and skips: repeated runs returned as few as 4-6 of the 8 repositories, with some names appearing 2-3 times and others never appearing.

This only affects `createTime`/`updateTime` — `orderBy=name` (the default, and the only variant the PR's own `TestSDKArtifactRegistryListOrderBy` exercises) is unaffected because names are unique, so no ties occur and the underlying map-order randomness never surfaces.

Cascade: a client that creates several repositories in one run (any realistic IaC/script usage) and then paginates `repositories.list?orderBy=createTime` — a normal reconciliation/cleanup-ordering pattern — silently gets duplicate and missing repositories across pages.

Suggested fix: give `repoLess`/`repoSortKey` a deterministic secondary key so ties always resolve the same way regardless of `ListRepositories`' input order — e.g. when the primary field compares equal, fall back to comparing by resource name. This is contained entirely within the PR's new code in `operations.go` and doesn't require touching the pre-existing `memstore`-backed `ListRepositories`.

### Verdict: CHANGES

Everything else in the PR (dockerConfig, cleanupPolicies, list filter, regression coverage, gates) is solid and ready. The orderBy tie-breaking gap is real, reproducible, and hits the PR's own headline claim ("Now honors name, createTime, updateTime") for 2 of the 3 documented fields, so it should be fixed before merge.

NitinKumar004 (commented):
Delta re-review at 40c0a9a4 (fast pass, previously-reviewed dockerConfig/cleanupPolicies/filter code untouched).

**Delta verified in isolation** — `git show 40c0a9a4` is exactly 7 lines in `operations.go` + the new `repository_list_paging_test.go` (137 lines total, 0 deletions). Nothing else in the PR changed since the prior review.

`repoLess` now falls back to `repoName(a.Name)`/`repoName(b.Name)` on a primary-key tie:
```go
ka, kb := repoSortKey(rt, field, &a), repoSortKey(rt, field, &b)
if ka == kb {
    ka, kb = repoName(a.Name), repoName(b.Name)
}
```
This gives `sort.SliceStable` (in `pagination.PaginateSorted`) a true total order independent of the input slice's order, which is what actually fixes the bug — `PaginateSorted` re-sorts the full slice on every call from `m.repos.All()` (randomized map iteration), so any comparator that ties on the primary key was letting `SliceStable` fall back to relying on that random input order, producing a different sort result (and therefore dup/skip under offset paging) each call. With the name fallback the comparator is now a strict total order, so the result is deterministic regardless of map iteration.

Confirmed the new test genuinely exercises the tie: `CreateRepository` stamps `CreatedAt`/`UpdatedAt` via `m.opts.Clock.Now().UTC().Format(time.RFC3339)` (second granularity), and the test freezes a `FakeClock` before creating 5 repos, so all 5 share an identical `createTime`/`updateTime`. `TestSDKArtifactRegistryOrderByTiesDeterministic` paginates one-at-a-time under both `orderBy=createTime` and `orderBy=updateTime`, asserts every repo appears exactly once, and re-sweeps to check order stability across calls. Ran it 20x back-to-back (fresh randomized map iteration each `go test` invocation) — all pass. `orderBy=name` is a no-op here since names are already unique, so the default path is unaffected.

Gates on the pushed SHA (GOMAXPROCS=3):
- `go build ./...` — clean
- `go vet ./...` — clean
- `go test -p=3 ./server/gcp/artifactregistry/... ./providers/gcp/artifactregistry/...` — all pass, including the 6 new tests (`TestSDKArtifactRegistryImmutableTagsRoundTrip`, `CleanupPoliciesRoundTrip`, `ListFilter`, `ListOrderBy`, `ConfigNoRegression`, `OrderByTiesDeterministic`)
- `go test -race -p=3 ./providers/gcp/artifactregistry/...` and server pkg — clean
- `gofmt -l` — 0 hits in this PR's files (4 pre-existing unrelated hits elsewhere in the repo)
- `golangci-lint --new-from-rev=origin/development ./server/gcp/artifactregistry/...` — 0 issues

No blocking issues. This closes out the HIGH pagination bug from the prior review.

PR #739, opened 2026-08-26
fix(gcp-memorystore): redisConfigs round-trip, updateMask-honoring PATCH, clearable labels, readReplicasMode-gated readEndpoint, list pagination

NitinKumar004:
## Summary

Fixes several TF-drift and semantic bugs in the GCP Memorystore for Redis (redis.googleapis.com v1) server layer, verified end-to-end with the real `google.golang.org/api/redis/v1` SDK against a booted server.

All GCP-specific fields round-trip through the existing reserved-tag pattern (JSON-encoded for maps); the provider `ListCaches` already returns a stable sort, so no provider change was needed.

## Changes

- **redisConfigs round-trip** — `redisConfigs` (e.g. `{maxmemory-policy: allkeys-lru}`) now round-trips on create, get, and patch via a JSON-encoded reserved tag.
- **PATCH honors updateMask** — the update mask is parsed and only masked fields are applied; a body field outside the mask (e.g. `memorySizeGb` under `updateMask=displayName`) is left untouched. Absent updateMask keeps the prior apply-all behavior.
- **Labels/redisConfigs are whole-replaced when masked** — a removed label (or config) disappears instead of merge-lingering.
- **New modeled fields** — `authEnabled`, `transitEncryptionMode`, `replicaCount`, `readReplicasMode` round-trip on create/get/patch.
- **readEndpoint gating fixed** — `readEndpoint`/`readEndpointPort` are now emitted only when `readReplicasMode == READ_REPLICAS_ENABLED` (matching real Memorystore), not on `tier == STANDARD_HA`.
- **List pagination + filter** — `listInstances` now honors `pageSize`/`pageToken` (stable-sorted, emits `nextPageToken` when truncated) and a single AIP-160 equality `filter` term.

Custom verbs (`:upgrade`/`:failover`/`:getAuthString`/`:import`/`:export`) and a CREATING→READY transition are intentionally out of scope.

## Tests

Added real-SDK e2e tests: redisConfigs round-trip; updateMask honored (size unchanged outside mask); labels clearable via whole-replace; auth/transit/replica round-trip; readEndpoint gated on readReplicasMode; list pagination visits every instance once and terminates. Updated the existing read-endpoint test to the new gating.

### DIFF
--- a/server/gcp/memorystore/configs_sdk_test.go
+++ b/server/gcp/memorystore/configs_sdk_test.go
@@ -104,6 +104,79 @@ func TestSDKMemorystorePatchLabelsClearable(t *testing.T) {
 	}
 }
 
+// TestSDKMemorystorePatchNoMaskPreservesMapFields guards the regression where a
+// PATCH omitting updateMask entirely silently wiped labels/redisConfigs: with no
+// mask the map fields must MERGE FORWARD (unspecified entries survive).
+func TestSDKMemorystorePatchNoMaskPreservesMapFields(t *testing.T) {
+	svc := newRedisService(t)
+	ctx := context.Background()
+
+	if _, err := svc.Projects.Locations.Instances.Create(parent(), &redis.Instance{
+		Tier:         "BASIC",
+		MemorySizeGb: 1,
+		Labels:       map[string]string{"env": "prod"},
+		RedisConfigs: map[string]string{"maxmemory-policy": "allkeys-lru"},
+	}).InstanceId("nomask").Context(ctx).Do(); err != nil {
+		t.Fatalf("Create: %v", err)
+	}
+
+	// PATCH with NO updateMask, changing only displayName. The body carries no
+	// labels/redisConfigs, so they must be preserved (not whole-replaced away).
+	if _, err := svc.Projects.Locations.Instances.Patch(instanceName("nomask"), &redis.Instance{
+		DisplayName: "renamed",
+	}).Context(ctx).Do(); err != nil {
+		t.Fatalf("Patch (no updateMask): %v", err)
+	}
+
+	got, err := svc.Projects.Locations.Instances.Get(instanceName("nomask")).Context(ctx).Do()
+	if err != nil {
+		t.Fatalf("Get: %v", err)
+	}
+
+	if got.DisplayName != "renamed" {
+		t.Errorf("displayName = %q, want renamed", got.DisplayName)
+	}
+
+	if got.Labels["env"] != "prod" {
+		t.Errorf("labels = %v, want env=prod preserved after no-mask patch", got.Labels)
+	}
+
+	if got.RedisConfigs["maxmemory-policy"] != "allkeys-lru" {
+		t.Errorf("redisConfigs = %v, want maxmemory-policy preserved after no-mask patch", got.RedisConfigs)
+	}
+}
+
+// TestSDKMemorystorePatchReplicaCountZero guards that replicaCount=0 under an
+// explicit updateMask=replicaCount is applied (0 is a valid "no replicas"
+// setting), not swallowed as if unset.
+func TestSDKMemorystorePatchReplicaCountZero(t *testing.T) {
+	svc := newRedisService(t)
+	ctx := context.Background()
+
+	if _, err := svc.Projects.Locations.Instances.Create(parent(), &redis.Instance{
+		Tier:         "STANDARD_HA",
+		MemorySizeGb: 5,
+		ReplicaCount: 2,
+	}).InstanceId("rc").Context(ctx).Do(); err != nil {
+		t.Fatalf("Create: %v", err)
+	}
+
+	if _, err := svc.Projects.Locations.Instances.Patch(instanceName("rc"), &redis.Instance{
+		ReplicaCount: 0,
+	}).UpdateMask("replicaCount").Context(ctx).Do(); err != nil {
+		t.Fatalf("Patch: %v", err)
+	}
+
+	got, err := svc.Projects.Locations.Instances.Get(instanceName("rc")).Context(ctx).Do()
+	if err != nil {
+		t.Fatalf("Get: %v", err)
+	}
+
+	if got.ReplicaCount != 0 {
+		t.Errorf("replicaCount = %d, want 0 applied under explicit updateMask", got.ReplicaCount)
+	}
+}
+
 // TestSDKMemorystoreSecurityFieldsRoundTrip (B4) guards that authEnabled,
 // transitEncryptionMode and replicaCount round-trip on create and get.
 func TestSDKMemorystoreSecurityFieldsRoundTrip(t *testing.T) {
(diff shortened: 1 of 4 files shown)


NitinKumar004 (commented):
Verified on the pushed head (a385034) with GOMAXPROCS=3: `go build ./...`, `go vet ./...`, full `go test -p=3 ./...` (0 failures), `go test -race ./providers/gcp/memorystore/...`, `gofmt -l` on the 4 changed files, `golangci-lint --new-from-rev=origin/development` on `server/gcp/memorystore/...` (0 issues), `go generate ./...` (zero diff), `internal/compatgen` (exit 0, zero diff in `docs/compat/`) — all green. Diff is exactly the 4 files listed.

Confirmed correct via the real `google.golang.org/api/redis/v1` SDK against a booted `gcpserver.New(...)`:
- redisConfigs create→get round-trip, and masked-PATCH whole-replace.
- `updateMask=displayName` with a body that also sets `memorySizeGb` leaves `memorySizeGb` untouched (B2 test).
- `updateMask=labels` whole-replaces the label set (dropped key actually disappears, B3).
- authEnabled/transitEncryptionMode/replicaCount/readReplicasMode round-trip on create+get (B4).
- readEndpoint/readEndpointPort now gated on `readReplicasMode==READ_REPLICAS_ENABLED`, not `tier==STANDARD_HA` — `TestSDKMemorystoreReadEndpointStandardHA` was correctly updated to set `ReadReplicasMode` rather than masking the gating change (B5).
- List pagination visits every instance exactly once and terminates (B6); no regression in memorySizeGb/redisVersion/displayName/reservedIpRange/authorizedNetwork/connectMode/tier round-trip, or LRO/404/409 handling.

Two real bugs found while probing edge cases the added tests don't cover (both reproduced against the pushed SHA with throwaway SDK tests, then reverted — tree is clean):

**1. Regression: PATCH without `updateMask` now silently wipes labels/redisConfigs when the body omits them.** `carryForwardTags` (types.go) computes `dropLabels := mask.has("labels")`, and `fieldMask.has` treats a nil mask (no `updateMask` query param at all) as "select everything" — so a mask-less PATCH drops every existing user label before `applyLabels` re-adds only what the body supplies. If the body doesn't set `Labels` (e.g. a client PATCHing only `displayName` without ever calling `.UpdateMask(...)`), labels are wiped rather than preserved. `applyRedisConfigs` has the identical failure mode (unconditional `delete(out, redisConfigsTag)` when `mask.has("redisConfigs")`, which is always true under a nil mask). Pre-PR, this path always merged forward (never dropped), so this is new behavior, not a pre-existing gap. Repro:
```go
// Create with Labels: {"env":"prod"}, then:
svc.Projects.Locations.Instances.Patch(name, &redis.Instance{DisplayName: "renamed"}).Do() // no .UpdateMask(...)
// GET afterwards: Labels == map[] (env dropped)
```
Suggest: when mask is nil, only whole-replace labels/redisConfigs if the body actually sets them (e.g. treat a nil/absent `body.Labels`/`body.RedisConfigs` under a nil mask as "leave unchanged", matching how the string/scalar fields already treat a zero value as "not provided").

**2. `replicaCount` can't be explicitly set to 0 via a masked PATCH.** `applyScalarTags` guards with `mask.has("replicaCount") && body.ReplicaCount > 0` (types.go). `replicaCount: 0` is a valid, meaningful Memorystore value (read replicas disabled), but the `>0` guard — copied from the `memorySizeGb` guard, where 0 is actually invalid — means `PATCH ...?updateMask=replicaCount` with body `{"replicaCount": 0}` is a silent no-op; the prior non-zero value persists. This is the same TF-drift class of bug the PR fixes elsewhere. Repro:
```go
// Create with ReplicaCount: 2, then:
svc.Projects.Locations.Instances.Patch(name, &redis.Instance{ReplicaCount: 0}).UpdateMask("replicaCount").Do()
// GET afterwards: ReplicaCount == 2, not 0
```
Suggest mirroring `setBoolTag`'s pattern: when masked, always write or delete the tag rather than gating on `> 0`.

Everything else in the diff — updateMask parsing, the redisConfigs JSON-encoded round-trip, the filter/pagination logic, and the readEndpoint gating fix — is solid and well-tested.

NitinKumar004 (commented):
Delta re-review at 45838d6f (types.go + configs_sdk_test.go only, confirmed via `git show --stat`).

**Both prior bugs confirmed fixed, verified against the real `google.golang.org/api/redis/v1` SDK:**

1. **Nil-mask map wipe (labels/redisConfigs)** — fixed via the new `mapFieldMode` (Preserve/Merge/Replace). `carryForwardTags` no longer drops user labels unless the mask *explicitly* names `labels` (`dropLabels := mask.mapMode("labels") == mapModeReplace`), and `applyRedisConfigs`/`applyLabels` merge onto the existing map when the mask is nil. Verified:
   - No-mask PATCH (`displayName` only, no updateMask) preserves both `labels` and `redisConfigs` — existing `configs_sdk_test.go::TestSDKMemorystorePatchNoMaskPreservesMapFields`.
   - No-mask PATCH whose body *does* set new labels merges them onto existing (`env=prod` survives, `team=core` added) — verified manually against the pushed tree.
   - Explicit `updateMask=labels` still whole-replaces (`team` dropped) — `TestSDKMemorystorePatchLabelsClearable`.
   - Non-nil mask omitting `labels`/`redisConfigs` (e.g. `updateMask=displayName`) leaves both untouched — verified manually.

2. **replicaCount:0 no-op** — fixed via `fieldMask.explicit()` distinguishing "field named in a non-nil mask" from "nil mask." `applyReplicaCount` applies 0 when explicitly masked (`TestSDKMemorystorePatchReplicaCountZero`, passes), and preserves the existing count on a no-mask PATCH that leaves `ReplicaCount` at its Go zero value — verified manually.

3. **authEnabled — same nil-mask-wipe hazard, proactively fixed** — same `explicit()` pattern applied to `applyAuthEnabled`. Verified manually: an existing `authEnabled=true` survives an unrelated no-mask PATCH (previously this would have been wiped, since the old code called `setBoolTag` whenever `mask.has("authEnabled")`, which is unconditionally true for a nil mask); an explicit `updateMask=authEnabled` with `false` still clears it correctly.

**No regressions** — full `server/gcp/memorystore` suite green (redisConfigs round-trip, extra-field masking, readEndpoint gating on `readReplicasMode`, list pagination, `memorySizeGb>0` guard all still pass unchanged).

**Gates on 45838d6f (GOMAXPROCS=3):**
- `go build ./...` clean
- `go vet ./...` clean
- `go test -p=3 ./server/gcp/memorystore/... ./providers/gcp/memorystore/...` — PASS
- `go test -race -p=3 ./providers/gcp/memorystore/...` — PASS
- root `go test -p=3 .` — PASS
- `gofmt -l` — clean on touched files (4 pre-existing unrelated dirty files elsewhere in the repo, not touched by this PR)
- `golangci-lint --new-from-rev=origin/development` on `server/gcp/memorystore/...` + `providers/gcp/memorystore/...` — 0 new issues

Delta is scoped exactly to `types.go` (+134/-17) and `configs_sdk_test.go` (+73), matching the stated intent.

PR #740, opened 2026-08-26
fix(azure-eventgrid): round-trip topic/domain publicNetworkAccess & domain inputSchema; add Topics PATCH + RegenerateKey

NitinKumar004:
## Summary

Fixes 4 Azure Event Grid (`Microsoft.EventGrid`) real bugs where topic/domain properties were echoed from hardcoded defaults instead of stored request state, and two topic operations were missing.

### Bugs fixed
1. **Topic `publicNetworkAccess` not round-tripped** — a topic created with `PublicNetworkAccess=Disabled` returned `Enabled` on both the create response and GET. Added `PublicNetworkAccess` to the eventbus driver (mirroring `InputSchema`), stored it in the provider (default `Enabled` only when omitted), and echoed the stored value from the wire handler.
2. **`Topics.Update` (PATCH) → 405** — the topic router only handled PUT/GET/DELETE. Added `PATCH`, which merges the supplied tags onto the topic's existing tags and applies the mutable `publicNetworkAccess` via `UpdateEventBus`, returning the updated topic (200).
3. **`Topics.RegenerateKey` → 400** — topics only handled `listKeys`. Added a per-topic key-generation counter (handler-owned state); `regenerateKey(key1|key2)` bumps only that key's generation, so key1 rotates while key2 is untouched and `listKeys` reflects the new value. Mirrors the existing domain `regenerateKey` behavior.
4. **Domain `inputSchema`/`publicNetworkAccess` hardcoded** — a domain created with `CloudEventSchemaV1_0` GETs back `EventGridSchema`. The domain create path now reads both from the request body, stores them, and echoes them. `inputSchema` is immutable on update (like topics); `publicNetworkAccess` is mutable.

### Unregressed
Topic `inputSchema` round-trip + immutability, endpoint synthesis, event-subscription destination/filter round-trip, the filter engine, WebHook cross-service delivery, and `listKeys` returning real `key1`/`key2` (#715) are all preserved and still covered by their tests.

### Tests (real `armeventgrid/v2` SDK against a booted `httptest.NewTLSServer`)
- Topic `PublicNetworkAccess=Disabled` round-trips + defaults to `Enabled` when omitted.
- `Topics.BeginUpdate` PATCH succeeds and merges tags + applies `publicNetworkAccess`.
- `Topics.BeginRegenerateKey` key1 → key1 changes, key2 unchanged, `listKeys` reflects it.
- Domain `CloudEventSchemaV1_0` + `PublicNetworkAccess` round-trip, and domain `inputSchema` immutability.

### Gates
`go build`, `go vet`, targeted `go test` (server/provider/eventbus), `-race` on the provider, `gofmt -l`, `go generate` (zero diff), and `compatgen` (zero `docs/compat` diff) all clean. No new lint issues on authored lines.

### DIFF
--- a/providers/azure/eventgrid/eventgrid.go
+++ b/providers/azure/eventgrid/eventgrid.go
@@ -29,6 +29,9 @@ const (
 	// defaultInputSchema is the event schema a topic accepts when the caller
 	// doesn't specify one, matching real Event Grid's default.
 	defaultInputSchema = "EventGridSchema"
+	// defaultPublicNetworkAccess is the public-network-access setting a topic
+	// gets when the caller doesn't specify one, matching real Event Grid.
+	defaultPublicNetworkAccess = "Enabled"
 )
 
 // Compile-time check that Mock implements driver.EventBus.
@@ -128,15 +131,21 @@ func (m *Mock) CreateEventBus(_ context.Context, cfg driver.EventBusConfig) (*dr
 		inputSchema = defaultInputSchema
 	}
 
+	publicNetworkAccess := cfg.PublicNetworkAccess
+	if publicNetworkAccess == "" {
+		publicNetworkAccess = defaultPublicNetworkAccess
+	}
+
 	info := driver.EventBusInfo{
-		Name:        cfg.Name,
-		Scope:       cfg.Scope,
-		ARN:         topicID,
-		State:       activeTopicState,
-		CreatedAt:   m.opts.Clock.Now().UTC().Format(time.RFC3339),
-		Tags:        tags,
-		Region:      cfg.Region,
-		InputSchema: inputSchema,
+		Name:                cfg.Name,
+		Scope:               cfg.Scope,
+		ARN:                 topicID,
+		State:               activeTopicState,
+		CreatedAt:           m.opts.Clock.Now().UTC().Format(time.RFC3339),
+		Tags:                tags,
+		Region:              cfg.Region,
+		InputSchema:         inputSchema,
+		PublicNetworkAccess: publicNetworkAccess,
 	}
 
 	bd := &busData{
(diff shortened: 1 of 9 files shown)


NitinKumar004 (commented):
**Review — verdict: MERGE**

Verified against the pushed head (`30196d0f`), clean-built and tested on a detached checkout.

**Gates (GOMAXPROCS=3)**
- `go test ./server/azure/eventgrid/... ./providers/azure/eventgrid/... ./services/eventbus/... ./server/aws/eventbridge/... ./providers/aws/eventbridge/...` — pass; also ran GCP eventarc + OCI eventbus consumers, all pass
- `go test -race` on both azure eventgrid packages — pass
- `gofmt -l` clean; `go generate ./...` zero diff; `compatgen` green + `docs/compat` zero diff
- `golangci-lint --new-from-rev=origin/development` on the three changed non-test packages — 0 new issues

**Behavior confirmed**
- Topic `publicNetworkAccess` round-trips (Disabled create+GET; defaults to Enabled when omitted).
- `Topics.Update` PATCH merges tags onto existing + applies mutable `publicNetworkAccess`, 200. `inputSchema` immutable: PATCH body has no schema field, and the PUT-update path routes through `UpdateEventBus`, which never writes `InputSchema` (only Region/PublicNetworkAccess).
- `RegenerateKey(key1)` rotates key1, leaves key2 untouched; `listKeys` reflects it; key2 regen independent.
- Domain `inputSchema` + `publicNetworkAccess` round-trip; inputSchema immutable, PNA mutable.

**RegenerateKey concurrency** — no race. `topicKeyGens` is only touched inside `regenerateTopicKey` (h.mu.Lock) and `topicKeysFor` (called under h.mu.RLock in listTopicKeys / under Lock in regenerate). The `bus.GetEventBus` call sits before the lock, so no nested-lock/deadlock. `-race` clean.

**Shared driver inert** — `EventBusInfo/EventBusConfig.PublicNetworkAccess` is purely additive. Grep shows the only consumers outside Azure eventgrid are the driver definitions themselves; AWS EventBridge, GCP Eventarc, and OCI construct these structs with named fields and neither set nor read the field. EventBridge/Eventarc/OCI tests unchanged and green.

**#715** — `listKeys` (topic + domain) still return real derived key1/key2; the added round-trip fields carry no secret.

Change set limited to the 9 stated files.

PR #741, opened 2026-08-26
fix(azure-acr): round-trip UserAssigned identity map; stop webhook callback config leaking on GET; add getCallbackConfig

NitinKumar004:
## Summary

Fixes three real-user Azure Container Registry (`Microsoft.ContainerRegistry`) bugs, verified end-to-end against a booted `azureserver` and the real `armcontainerregistry` SDK.

### 1. UserAssigned identity map was dropped on GET (Terraform drift)
`PUT identity{type:UserAssigned, userAssignedIdentities:{<id>:{}}}` then GET returned the identity type but an empty `userAssignedIdentities` map, causing perpetual drift on `azurerm_container_registry.identity`. The map is a top-level field, so the property overlay could not rescue it.

- Added `UserAssignedIdentities` to the driver config/update/stored types plus `AzureUserAssignedIdentity{PrincipalID, ClientID}`.
- The provider synthesizes a deterministic principal/client pair per identity; the wire layer decodes the request map and echoes it on GET.
- SystemAssigned (principalId/tenantId) and combined `SystemAssigned,UserAssigned` still work; `None`/omitted yields no identity.

### 2. Webhook `serviceUri` + `customHeaders` leaked on plain GET
Real ACR `Webhooks_Get` returns only `{status, scope, actions, provisioningState}`; `serviceUri`/`customHeaders` are write-only and readable only via `getCallbackConfig`. Because `toARMWebhook` omitted them, the generic property overlay captured them from the create request and replayed them on every read — leaking a webhook `Authorization: Bearer` header.

- Model `serviceUri`/`customHeaders` on the create/update response (`toARMWebhookWithCallback`) so the overlay treats them as modeled and does not capture them.
- The plain read shape (`toARMWebhook`) still omits them, so GET/List no longer leak.

### 3. Webhook `getCallbackConfig` returned 405
`POST .../webhooks/{w}/getCallbackConfig` now returns `{serviceUri, customHeaders}` — the only supported read path for those two fields (Terraform `azurerm_container_registry_webhook` read calls it).

## Tests (real armcontainerregistry SDK)
- UserAssigned round-trip: GET echoes the map with deterministic principalId/clientId.
- SystemAssigned still works + combined type.
- Webhook create with `customHeaders{Authorization}`: raw plain GET contains no `serviceUri`/`customHeaders`/`Authorization`; `getCallbackConfig` returns them. (Confirmed the test fails without the fix.)
- #715 regression guard: registries `listCredentials` still returns admin passwords (not stripped).

## Gates
build, vet, scoped tests, `-race`, gofmt, `go generate` (zero diff), compatgen (zero `docs/compat` diff), golangci-lint (0 new) all green.

### DIFF
--- a/providers/azure/acr/registry.go
+++ b/providers/azure/acr/registry.go
@@ -45,6 +45,8 @@ func subResourceKey(rg, registry, name string) string {
 // CreateOrUpdateRegistry creates or replaces an ACR registry (ARM PUT). It
 // reports whether the registry was newly created (vs. replaced) so the wire
 // layer can distinguish 201 Created from 200 OK.
+//
+//nolint:gocritic // cfg mirrors the driver interface's value-type config; pointer would invite caller mutation.
 func (m *Mock) CreateOrUpdateRegistry(
 	_ context.Context, rg, name string, cfg driver.AzureRegistryConfig,
 ) (*driver.AzureRegistry, bool, error) {
@@ -82,7 +84,7 @@ func (m *Mock) CreateOrUpdateRegistry(
 	rd.reg.AdminUserEnabled = cfg.AdminUserEnabled
 	rd.reg.ProvisioningState = registryProvisioned
 	rd.reg.Tags = copyTags(cfg.Tags)
-	applyIdentity(&rd.reg, cfg.IdentityType, rg, name)
+	applyIdentity(&rd.reg, cfg.IdentityType, cfg.UserAssignedIdentities, rg, name)
 
 	m.registries.Set(key, rd)
 
@@ -122,7 +124,7 @@ func (m *Mock) UpdateRegistry(
 	}
 
 	if upd.IdentityType != nil {
-		applyIdentity(&rd.reg, *upd.IdentityType, rg, name)
+		applyIdentity(&rd.reg, *upd.IdentityType, upd.UserAssignedIdentities, rg, name)
 	}
 
 	m.registries.Set(key, rd)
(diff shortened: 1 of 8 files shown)


NitinKumar004 (commented):
Reviewed at pushed head 0134e69 (clean checkout, GOMAXPROCS=3). Verdict: LGTM / approve.

**UserAssigned identity round-trip** — `identityKeys` sorts the map keys before storing, and `synthUserAssigned` derives the principal/client pair from `SyntheticGUID("uai-{principal,client}/"+id)`, so the synthesized pair is deterministic per identity and stable across repeated GETs. Verified via real `armcontainerregistry` SDK against a booted TLS httptest server: single- and multi-identity maps survive create→GET, each GET returns identical pairs, distinct identities get distinct principals. SystemAssigned still synthesizes principalId/tenantId; combined `SystemAssigned, UserAssigned` yields both; `None`/omitted returns no identity block.

**Webhook GET no-leak (the security fix)** — Root cause understood and correctly addressed: the property overlay (`server/azure/echo_properties.go`) captures request `properties` the create/update *response* omits and replays them on later reads. `serviceUri`/`customHeaders` don't match the `writeOnlyProperty` password/secret denylist, so pre-fix they were captured and replayed onto plain `Webhooks_Get` — leaking an `Authorization: Bearer …` header. Modeling them on `toARMWebhookWithCallback` (PUT/PATCH response) makes `missingProperties` see them as modeled, so nothing is captured and plain GET (`toARMWebhook`) has nothing to replay. Confirmed on RAW JSON for both `Webhooks_Get` and `Webhooks_List`: no `serviceUri`, `customHeaders`, `Authorization`, host, or secret substring present.

**getCallbackConfig** — POST `.../webhooks/{w}/getCallbackConfig` returns 200 `{serviceUri, customHeaders}` (the only read path); 405 gone.

**#715 intact** — `registries listCredentials` and `regenerateCredential` both still return admin username + non-empty passwords (verified via SDK); `echo_properties.go` untouched.

**Shared driver scope** — `services/containerregistry/driver/azure.go` additions (`AzureUserAssignedIdentity`, config `UserAssignedIdentities`) are consumed only by `providers/azure/acr` and `server/azure/acr`; AWS ECR / GCP Artifact Registry use the cross-cloud `ContainerRegistry` interface, so no cross-provider impact.

**Lint** — removed `//nolint:dupl` on `serveReplication` is genuinely unused now (`serveWebhook` diverged with the getCallbackConfig branch); full lint on the package is 0 issues.

Gates (pushed SHA): build, vet, `go test` (server/azure/acr, providers/azure/acr, services/containerregistry), `-race` provider — all green; gofmt clean on the 8 changed files (incl. under go1.26 — no reformatted untouched file committed); golangci-lint `--new-from-rev` 0 new; `go generate` and compatgen produce zero diff.

PR #742, opened 2026-08-26
fix(azure-dns): RecordSets.ListByType, CNAME coexistence, apex SOA/NS delete protection, maxRecordSets=10000

NitinKumar004:
## Summary

Fixes four Azure DNS (Microsoft.Network/dnsZones) fidelity gaps against the real armdns SDK, verified end-to-end through a booted azureserver + real `armdns` client.

- **RecordSets.ListByType (MED, functional):** A type-only GET `.../dnsZones/{zone}/{type}` (no record name) parsed with an empty record name and was routed to `getRecordSet` → `GetRecord(zone,"",type)` → 404. It now routes to a type-filtered list of the zone's record sets, matching `RecordSets.ListByType`. `ListByDnsZone`/`ListAllByDnsZone` share the same list machinery via a new `writeRecordSetList(filterType)` helper (no new duplication).
- **CNAME coexistence (LOW):** The Azure Mock's `CreateRecord` now rejects (400) a CNAME at a name that already holds another record type, and any other type at a name that already holds a CNAME (both orders). Same-type collisions remain `AlreadyExists`.
- **Apex SOA/NS delete protection (LOW):** `deleteRecordSet` now rejects deleting the auto-provisioned apex SOA and apex NS record sets (name `@`) with a 400, matching Azure.
- **maxNumberOfRecordSets (LOW, cosmetic):** Corrected the reported public-zone cap from 5000 to 10000.

## Tests

New real-`armdns` SDK e2e tests in `server/azure/dns/constraints_sdk_test.go`:
- `ListByType(A)` returns the zone's A record sets (not 404)
- CNAME-over-A and A-over-CNAME both 400
- Delete apex SOA and apex NS both 400
- `maxNumberOfRecordSets == 10000` on zone GET

Existing round-trip / list-by-zone / apex auto-provision / idempotent-delete tests remain green.

## Not in scope (deferred)

PATCH Update, ETag If-Match/If-None-Match, privateDnsZones, list `$top`/nextLink pagination, and 201-vs-200 status are intentionally untouched.

### DIFF
--- a/providers/azure/dns/dns.go
+++ b/providers/azure/dns/dns.go
@@ -35,6 +35,10 @@ func New(opts *config.Options) *Mock {
 	}
 }
 
+// cnameType is the DNS CNAME record type. Azure enforces CNAME coexistence
+// rules against it (a CNAME cannot share a name with any other record set).
+const cnameType = "CNAME"
+
 // recordKey builds the key used to store a record in the memstore.
 // For weighted records (non-empty SetID), the SetID is appended.
 func recordKey(zoneID, name, recordType, setID string) string {
@@ -172,6 +176,11 @@ func (m *Mock) CreateRecord(_ context.Context, cfg driver.RecordConfig) (*driver
 		return nil, cerrors.New(cerrors.InvalidArgument, "record type is required")
 	}
 
+	if m.hasCNAMEConflict(cfg.ZoneID, cfg.Name, cfg.Type) {
+		return nil, cerrors.Newf(cerrors.InvalidArgument,
+			"a CNAME record set cannot coexist with another record set of a different type at name %q", cfg.Name)
+	}
+
 	key := recordKey(cfg.ZoneID, cfg.Name, cfg.Type, cfg.SetID)
 
 	if m.records.Has(key) {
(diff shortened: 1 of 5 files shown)


NitinKumar004 (commented):
Verified against pushed HEAD `cecb3b5` in an isolated worktree (real `armdns` SDK client against a booted `azureserver`).

**Functional checks (all green):**
- `RecordSets.ListByType(zone, A)` now returns the zone's A record sets (previously 404'd via the type-only-path → `getRecordSet` fallthrough). Verified another type (TXT with zero matches) returns an empty list, not an error.
- Single-record `RecordSets.Get(zone, type, name)` is unaffected — the new type-only routing check (`rp.SubResourceName == "" && GET`) only intercepts paths with no record name, confirmed via `TestSDKAzureDNSRecordSets` (existing, still passing) plus manual probe.
- `ListByDnsZone`/`ListAllByDnsZone` unchanged in behavior, now share `writeRecordSetList` — no observed duplication regression.
- `maxNumberOfRecordSets` = 10000 on zone GET.
- CNAME coexistence: CNAME-over-A and A-over-CNAME both 400, verified both orders. Probed that an unrelated name in the same zone is unaffected, and that a same-type upsert at an already-occupied name is *not* rejected by the new check (falls through to `AlreadyExists`/update as before).
- Apex SOA/NS delete protection: apex `@` SOA and NS deletes both 400. Probed that a non-apex delegation NS (e.g. `sub`) still creates and deletes cleanly — the `isApexProtectedRecord` gate correctly short-circuits on `name != "@"`.
- Full existing DNS suite (round-trip, apex auto-provision, nameServers synthesis, list-by-RG/subscription, idempotent delete, unmodeled-type list passthrough) stays green.

**Scope:** diff limited to `providers/azure/dns/dns.go`, `server/azure/dns/{handler,operations,types}.go`, and the new `server/azure/dns/constraints_sdk_test.go`. `server/azure/echo_properties.go` untouched.

**Gates (GOMAXPROCS=3):** `go build ./...`, `go vet ./...` clean. `go test -p=3 ./server/azure/dns/... ./providers/azure/dns/... ./services/dns/...` and `go test -race -p=3 ./providers/azure/dns/...` green. `gofmt -l` clean on changed files. `golangci-lint --new-from-rev=origin/development` on the changed packages: 0 new issues. `go generate ./...` zero diff. `go run ./internal/compatgen` exits 0; `docs/compat/` zero diff.

**Minor, non-blocking:**
- `hasCNAMEConflict` in `providers/azure/dns/dns.go` reads via `m.records.SortedValues()` and only later `Set()`s, with no lock held across the read-then-write — a concurrent CreateRecord race could theoretically let a CNAME and a same-name A both pass the check before either is stored. This mirrors the pre-existing `Has()`-then-`Set()` TOCTOU pattern already in `CreateRecord` for the `AlreadyExists` check, so it's not a new class of issue, just worth knowing it isn't atomic under concurrency.
- `hasCNAMEConflict` compares record names case-insensitively (`strings.EqualFold`), while `recordKey`/storage elsewhere is case-sensitive on name. Slightly stricter than the rest of the store, but in the safe direction (Azure DNS names are case-insensitive in practice) and didn't surface as a real inconsistency in testing.

No blocking issues found.

PR #743, opened 2026-08-26
fix(gcp-loadbalancer): round-trip backends/target/cdn, scope-aware keying, list filter+pagination, patch/update verbs

NitinKumar004:
## Summary

Fixes 8 GCP Cloud Load Balancing bugs surfaced by driving the booted wire server (`gcpserver.New`) with the real `google.golang.org/api/compute/v1` SDK. All fixes are server-layer (reserved-tag serialization + list/verb plumbing); the portable `TargetGroup`/`LBInfo` drivers are unchanged except for one opaque-store update method.

### Round-trip drops (Terraform drift)
- **backendServices** `backends[]` (group/balancingMode/capacityScaler), `connectionDraining`, `cdnPolicy`, `enableCDN` now round-trip via reserved tags (patch preserves them).
- **forwardingRules** `target` is now stored and re-emitted, so the LB has an entrypoint on read.

### List defects
- BackendServices / ForwardingRules / HealthChecks / TargetPools / UrlMaps list endpoints now honor `filter` (`name eq/=/ne/!=`) and `maxResults`/`pageToken`, emitting `nextPageToken`. The filter/pagination helpers (`NameMatches`, `MaxResults`) are centralized in `server/wire/gcprest` and reused by the vpc addresses handler (removing its private copies).

### Verbs + scope
- healthChecks / urlMaps / targetPools now accept **PATCH** (merge into stored body) and **UPDATE** (replace body), returning a DONE Operation instead of 405 — so `terraform apply` on change works.
- Regional vs global resources of the same name no longer collide: the driver store key is scope-prefixed and lists are scope-filtered, and a regional Operation now carries a `region` self-link (new `Operation.Region` field mirroring `Zone`).

Not in scope (separate deferral): the L7 proxy tier (targetHttp(s)Proxies / sslCertificates / targetTcpProxies).

### Deferral
A fully scope-keyed opaque store would need a deeper driver change; this PR does the collision-avoidance at the server layer (scope-prefixed driver key + scope-filtered lists + scope-aware Operation metadata), which is sufficient for the global/regional same-name case.

## Tests

Real-SDK e2e added against a booted server:
- backends[] + connectionDraining + cdnPolicy + enableCDN round-trip
- forwardingRule target round-trip
- List `Filter("name eq ...")` returns only matches
- List `MaxResults(1)` paginates via nextPageToken, all items once
- HealthChecks.Patch merge + UrlMaps.Update replace reflect on Get
- global + regional same-name backendService coexist; regional insert Operation carries the region

## Gates
build, vet, `go test` (loadbalancer server/provider/service + vpc + gcprest), `-race` (provider), gofmt, golangci-lint (0 new on changed pkgs), `go generate` idempotent (coverage regenerated), compatgen + `docs/compat` clean.

### DIFF
--- a/docs/coverage/aws/elb.md
+++ b/docs/coverage/aws/elb.md
@@ -64,6 +64,7 @@ GCPComputeResourceStore is an OPTIONAL, type-asserted capability implemented
 | `GetGCPResource` | GetGCPResource returns the stored resource, or NotFound. |
 | `ListGCPResources` | ListGCPResources returns every resource in a (collection, scope) bucket. |
 | `PutGCPResource` | PutGCPResource stores res, returning AlreadyExists when a resource with |
+| `UpdateGCPResource` | UpdateGCPResource applies mutate to the stored resource in place under the |
 
 ### LBAttributeUpdater
 
--- a/docs/coverage/azure/lb.md
+++ b/docs/coverage/azure/lb.md
@@ -64,6 +64,7 @@ GCPComputeResourceStore is an OPTIONAL, type-asserted capability implemented
 | `GetGCPResource` | GetGCPResource returns the stored resource, or NotFound. |
 | `ListGCPResources` | ListGCPResources returns every resource in a (collection, scope) bucket. |
 | `PutGCPResource` | PutGCPResource stores res, returning AlreadyExists when a resource with |
+| `UpdateGCPResource` | UpdateGCPResource applies mutate to the stored resource in place under the |
 
 ### LBAttributeUpdater
 
--- a/docs/coverage/coverage.json
+++ b/docs/coverage/coverage.json
@@ -5783,6 +5783,10 @@
           {
             "name": "PutGCPResource",
             "doc": "PutGCPResource stores res, returning AlreadyExists when a resource with"
+          },
+          {
+            "name": "UpdateGCPResource",
+            "doc": "UpdateGCPResource applies mutate to the stored resource in place under the"
           }
         ]
       },
--- a/docs/coverage/gcp/lb.md
+++ b/docs/coverage/gcp/lb.md
@@ -64,6 +64,7 @@ GCPComputeResourceStore is an OPTIONAL, type-asserted capability implemented
 | `GetGCPResource` | GetGCPResource returns the stored resource, or NotFound. |
 | `ListGCPResources` | ListGCPResources returns every resource in a (collection, scope) bucket. |
 | `PutGCPResource` | PutGCPResource stores res, returning AlreadyExists when a resource with |
+| `UpdateGCPResource` | UpdateGCPResource applies mutate to the stored resource in place under the |
 
 ### LBAttributeUpdater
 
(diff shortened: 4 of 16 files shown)


NitinKumar004 (commented):
Reviewed at head 3ba5ab40 (clean-built the pushed tree, not the report). Verdict: MERGE.

- go build ./... and go vet on touched packages: clean.
- go test ./server/gcp/loadbalancer/... ./providers/gcp/loadbalancer/... ./services/loadbalancer/... ./server/gcp/vpc/... ./server/wire/gcprest/... ./server/aws/elbv2/... ./server/azure/loadbalancer/... ./providers/azure/loadbalancer/... and root: pass.
- go test -race ./providers/gcp/loadbalancer/...: clean.
- Full ./server/gcp/... tree: no failures.
- gofmt -l: clean; go generate ./... zero diff (coverage regen idempotent); compatgen + git diff docs/compat/: zero diff.
- golangci-lint --new-from-rev=origin/development on changed packages: 0 new issues.

Fix verification (real google.golang.org/api compute SDK against a booted gcpserver):
- B1 backends[] / B3-B4 connectionDraining + cdnPolicy + enableCDN: round-trip verbatim through JSON reserved tags (shipped BackendsRoundTrip test asserts group/balancingMode/capacityScaler/drainingTimeoutSec/enableCDN/cacheMode/defaultTtl).
- B2 forwardingRule target: echoed via frTargetTag; L4/L7 covered.
- B5 list filter and B6 pagination: filter selects only matches; MaxResults(1) walks all pages via nextPageToken, stable sort keys on name, each item exactly once (verified independently — no dup/skip).
- B7 patch/update verbs: HealthChecks.Patch merge, UrlMaps.Update replace, targetPools reflect on GET and return a DONE Operation; applyBody force-pins name so a body can't orphan the store key.
- B8 scope collision: global + regional backend service of the same name coexist in distinct driver slots via scope-prefixed key; each GET resolves to its own scope self-link; regional insert Operation carries the /regions/<r> link. Independently confirmed cross-scope delete isolation — deleting the global one leaves the regional record intact with its own distinct body (no cross-leak).

Blast radius:
- gcprest.NameMatches/MaxResults promotion: extracted bodies are byte-identical to the former vpc/handler.go copies; the vpc wrappers now delegate. Independently drove the real SDK against VPC global addresses list — filter and MaxResults(1) pagination unregressed, no dup/skip.
- UpdateGCPResource added to the shared GCPComputeResourceStore interface: the only implementer is the GCP loadbalancer Mock (compile-time _ = (*Mock)(nil) assertion holds); it got a correct read-modify-write under memstore.Update. AWS ELB / Azure LB do not implement the interface — they only surface it in generated coverage docs, which regen correctly and idempotently. Their tests pass.
- NUL scopeSep ("\x00"): safe — GCP RFC1035 resource names cannot contain NUL, so no collision with a real name segment; global names stay plain, preserving existing self-link stability.

No blocking issues.

PR #744, opened 2026-08-26
fix(aws-secretsmanager): BatchGetSecretValue rejects empty SecretIdList with no Filters

NitinKumar004:
## Summary

The Secrets Manager resource-policy ops (Put/Get/Delete/ValidateResourcePolicy) and `BatchGetSecretValue` were already implemented and merged in #672. This PR closes the one remaining real-cloud-parity gap in `BatchGetSecretValue`.

Real AWS constrains `SecretIdList` to a minimum of 1 item and requires the caller to supply either `SecretIdList` or `Filters`. The previous implementation returned an empty result for an empty `SecretIdList` with no `Filters`; real AWS rejects it with `InvalidParameterException`.

## Changes
- `BatchGetSecretValue` now returns `InvalidParameterException` when neither `SecretIdList` nor `Filters` is provided (matching real AWS). When both are provided, the list is preferred (unchanged).
- Updated the real-SDK e2e test to assert `InvalidParameterException` for an empty `SecretIdList`.

## Verification
- `go build ./...`, `go vet`, `gofmt -l` clean
- `go test` + `go test -race` on server/provider secretsmanager packages pass (real aws-sdk-go-v2 client against a booted server)
- `golangci-lint` 0 issues on changed package
- `go generate ./...` and `compatgen` zero diff (ops surface unchanged)

### DIFF
--- a/server/aws/secretsmanager/batch.go
+++ b/server/aws/secretsmanager/batch.go
@@ -11,10 +11,18 @@ import (
 // batchGetSecretValue retrieves the AWSCURRENT value of up to a page of secrets
 // named by SecretIdList (or matched by Filters). A missing (or otherwise
 // unreadable) secret is reported per-item in Errors rather than failing the
-// whole batch, matching the real service.
+// whole batch, matching the real service. A request with neither SecretIdList
+// nor Filters is rejected as InvalidParameterException, as real AWS does.
 func (h *Handler) batchGetSecretValue(w http.ResponseWriter, r *http.Request) {
 	var req batchGetSecretValueRequest
 	if !wire.DecodeJSON(w, r, &req) {
+		return
+	}
+
+	if len(req.SecretIDList) == 0 && len(req.Filters) == 0 {
+		wire.WriteJSONError(w, http.StatusBadRequest,
+			"InvalidParameterException", "either SecretIdList or Filters must be provided")
+
 		return
 	}
 
--- a/server/aws/secretsmanager/batch_get_secret_value_test.go
+++ b/server/aws/secretsmanager/batch_get_secret_value_test.go
@@ -2,10 +2,12 @@ package secretsmanager_test
 
 import (
 	"context"
+	"errors"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


NitinKumar004 (commented):
Verified against pushed SHA 31c28f0 (GOMAXPROCS=3):

- `go build ./...`, `go vet ./...` — clean.
- `go test -p=3 ./server/aws/secretsmanager/... ./providers/aws/secretsmanager/...` — all pass, including the updated `TestSDKBatchGetSecretValueEmpty`.
- `go test -race -p=3 ./providers/aws/secretsmanager/...` — clean.
- `gofmt -l` — no issues in the changed files (4 pre-existing unrelated files elsewhere in the repo need gofmt, untouched by this PR).
- `golangci-lint run --new-from-rev=origin/development ./server/aws/secretsmanager/...` — 0 new issues.
- `go generate ./...` — zero diff.
- `go run ./internal/compatgen` — exits 0, `git diff docs/compat/` — zero diff.

Real-user e2e against the booted server with `aws-sdk-go-v2/service/secretsmanager`:
- `SecretIdList=[]`, no `Filters` → `InvalidParameterException` ("either SecretIdList or Filters must be provided"). Matches real AWS.
- No `SecretIdList` field at all (nil) and no `Filters` → also correctly rejected.
- Valid `SecretIdList` (2 secrets) → still returns both values, no errors. Unregressed.
- `Filters`-only request → still resolves and returns matching secret value. Unregressed.
- Both `SecretIdList` + `Filters` set → list wins, filters ignored (pre-existing behavior from #672, untouched by this PR — note real AWS actually documents "must include Filters or SecretIdList, but not both" as mutually exclusive rather than list-precedence, but that's an existing behavior gap outside this PR's scope).

Diff is exactly as advertised: `server/aws/secretsmanager/batch.go` (+9/-1) and its test (+9/-8), nothing else touched. The resource-policy ops and BatchGetSecretValue happy path from #672 are untouched and still pass.

PR #746, opened 2026-08-26
fix(glue): GetPartitions Expression filter + crawler/job TF-drift round-trip

NitinKumar004:
## Summary

Fixes three AWS Glue gaps that caused wrong query results and Terraform drift. All verified end-to-end against `aws-sdk-go-v2/service/glue`.

### 1. GetPartitions honors the Expression filter
Previously `GetPartitions` ignored `Expression` and returned **all** partitions. Implemented the common Glue partition-filter grammar (a small tokenizer + recursive-descent parser):
- `key = 'value'`, comparison operators `=`, `<>`, `>`, `<`, `>=`, `<=`
- `AND` / `OR`, parentheses, string and number literals
- keys matched positionally against a partition's `Values` via the table's partition-key order; numeric literals compare numerically when the value parses as a number, else lexically
- a malformed expression is a typed `InvalidInputException`; no `Expression` still returns all partitions (unchanged)

### 2. Crawler fields round-trip (aws_glue_crawler drift)
`SchemaChangePolicy`, `RecrawlPolicy`, `LineageConfiguration` and `CrawlerSecurityConfiguration` now round-trip `CreateCrawler`/`UpdateCrawler` -> `GetCrawler` (stored opaquely as decoded JSON). Create-time `Tags` are stored under the crawler ARN and read back via `GetTags`, matching real Glue (GetCrawler omits tags).

### 3. Job fields round-trip (aws_glue_job drift)
`ExecutionProperty`, `Connections`, `NotificationProperty` and `SecurityConfiguration` now round-trip `CreateJob`/`UpdateJob` -> `GetJob`. Create-time `Tags` are stored under the job ARN and read back via `GetTags`.

## Tests (real glue SDK, booted httptest server)
- `TestSDKGetPartitionsExpressionFilter` — single equality and AND of two keys return only matching partitions
- `TestSDKGetPartitionsNoExpressionReturnsAll` — no filter returns all
- `TestSDKGetPartitionsBadExpression` — malformed filter -> InvalidInputException
- `TestSDKCrawlerFieldsRoundTrip` — SchemaChangePolicy/RecrawlPolicy via GetCrawler, Tags via GetTags
- `TestSDKJobFieldsRoundTrip` — ExecutionProperty/Connections/NotificationProperty via GetJob, Tags via GetTags

## Gates
build, vet, `go test` (server+provider glue), `-race` (provider), gofmt, `go generate` (zero diff), compatgen (`docs/compat` zero diff), golangci-lint (no new issues) — all green.

### DIFF
--- a/providers/aws/glue/copy.go
+++ b/providers/aws/glue/copy.go
@@ -97,6 +97,10 @@ func copyCrawler(in driver.Crawler) driver.Crawler {
 	out := in
 	out.Classifiers = copyStrings(in.Classifiers)
 	out.Targets = copyAnyMap(in.Targets)
+	out.SchemaChangePolicy = copyAnyMap(in.SchemaChangePolicy)
+	out.RecrawlPolicy = copyAnyMap(in.RecrawlPolicy)
+	out.LineageConfiguration = copyAnyMap(in.LineageConfiguration)
+	out.Tags = copyTags(in.Tags)
 
 	return out
 }
@@ -114,6 +118,10 @@ func copyJob(in driver.Job) driver.Job {
 	out := in
 	out.Command = copyAnyMap(in.Command)
 	out.DefaultArguments = copyTags(in.DefaultArguments)
+	out.ExecutionProperty = copyAnyMap(in.ExecutionProperty)
+	out.Connections = copyAnyMap(in.Connections)
+	out.NotificationProperty = copyAnyMap(in.NotificationProperty)
+	out.Tags = copyTags(in.Tags)
 
 	return out
 }
--- a/providers/aws/glue/crawlers.go
+++ b/providers/aws/glue/crawlers.go
@@ -16,7 +16,7 @@ type crawlerData struct {
 // CreateCrawler creates a crawler in the READY state, atomically.
 //
 //nolint:gocritic // hugeParam: taken by value to match the driver interface / copy semantics
-func (m *Mock) CreateCrawler(_ context.Context, c driver.Crawler) error {
+func (m *Mock) CreateCrawler(ctx context.Context, c driver.Crawler) error {
 	if !validName(c.Name) {
 		return invalidInput("crawler name %q is invalid", c.Name)
 	}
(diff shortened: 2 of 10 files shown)


NitinKumar004 (commented):
Reviewed the Glue partition-filter parser and the crawler/job round-trip fields at 2a75ab4.

Gates: build, vet, `go test` (server+provider), and `-race` all pass on the pushed SHA. Exactly the 10 expected files changed; `go generate ./...` and `go run ./internal/compatgen` leave zero diff; gofmt clean; 0 new golangci-lint issues.

Parser (partition_filter.go) — attacked the recursive-descent parser with a white-box harness:
- Precedence is correct: `a=1 AND b=2 OR c=3` binds as `(a AND b) OR c` (AND tighter than OR). Verified with a distinguishing case.
- Operator tokenizing correct: `<>`, `>=`, `<=` are scanned as two-char ops before falling back to single `<`/`>`; strings containing operator chars/spaces parse fine.
- Numeric vs lexical: an unquoted number literal compares numerically (`part > 9` -> 10,100 true, 2 false); a quoted literal compares lexically (`part > '9'` -> '10','100' false). This is a guess-from-literal, which matches how Athena/Terraform emit filters (numeric columns unquoted, string columns quoted) and is documented in the file header. Worth noting it deviates from real Glue, which keys on the partition-key column *type*: `strcol > 9` (unquoted literal against a string column) would compare numerically here. Not blocking for realistic TF/Athena input.
- No panic: 19 malformed inputs (unbalanced parens, trailing/leading operators, empty, `= =`, unterminated string, unknown char, adjacent comparisons) all return a typed InvalidInput error, no nil-deref/OOB. `next()` can reach the EOF sentinel but the parse never reads past it.
- Positional matching correct: filtering `month` reads Values[1], not Values[0].
- No-Expression path returns all partitions; filter is applied before pagination, over sorted keys, so NextToken stays stable.

Minor (non-blocking): a malformed numeric literal like `1.2.3` has its ParseFloat error dropped (comparePred) and silently becomes 0; escaped/doubled quotes in strings aren't supported but are rejected as trailing tokens rather than mis-parsed.

Round-trip fields: crawler SchemaChangePolicy/RecrawlPolicy/LineageConfiguration/CrawlerSecurityConfiguration and job ExecutionProperty/Connections/NotificationProperty/SecurityConfiguration round-trip Create->Get as opaque decoded JSON. Create-time Tags are stripped from the stored object and pushed to the tag store under the resource ARN; the wire response structs carry no Tags field, so GetCrawler/GetJob correctly omit tags (readable only via GetTags) — matches real Glue. UpdateCrawler/UpdateJob are full-replace, which is the correct Glue update semantic (the API takes the complete object), not a partial-update wipe.

Verdict: looks correct and well-tested. No blocking issues.

PR #747, opened 2026-08-26
fix(cloudwatch): support dashboards and composite alarms

NitinKumar004:
## Summary

Adds two AWS CloudWatch capabilities that unblock common Terraform workflows against the emulator:

- **Dashboards** — `PutDashboard`, `GetDashboard`, `ListDashboards`, `DeleteDashboards` (backs the `aws_cloudwatch_dashboard` resource).
- **Composite alarms** — `PutCompositeAlarm`, with `DescribeAlarms` surfacing them in the `CompositeAlarms` list (separate from `MetricAlarms`) and `DeleteAlarms` removing them alongside metric alarms (backs the `aws_cloudwatch_composite_alarm` resource).

## Behavior (matched to real AWS)

- `PutDashboard` validates the body is valid JSON; a non-JSON body is rejected (CloudWatch's dashboard-body validation). Existing dashboards are overwritten.
- `GetDashboard` returns `DashboardArn` + `DashboardName` + `DashboardBody`; an unknown name is a not-found error.
- `ListDashboards` honors `DashboardNamePrefix` and paginates with `NextToken`; entries carry `LastModified` and `Size`.
- `DeleteDashboards` is plural and all-or-nothing: an unknown name returns not-found and deletes nothing.
- `PutCompositeAlarm` stores `AlarmRule` + actions verbatim and round-trips faithfully. `DescribeAlarms` supports the `AlarmTypes` filter (`CompositeAlarm` / `MetricAlarm`, both when omitted). Composite state is left at `INSUFFICIENT_DATA` — the definition is round-tripped rather than shipping a half-working boolean rule evaluator (there is no trigger that re-evaluates a composite when a referenced alarm transitions).

## Implementation notes

Dashboards and composite alarms are AWS-local optional capabilities reached by type assertion in the wire handler, so the shared `driver.Monitoring` interface — and the Azure/GCP providers — stay unchanged. Operations are dispatched over the existing rpc-v2-cbor path (the protocol modern aws-sdk-go-v2 / Terraform use).

## Tests

Provider-level (in-memory) and wire-level (real aws-sdk-go-v2 client against an in-process server) tests cover: dashboard put/get/list(prefix)/delete, bad-JSON rejection, not-found; composite alarm put + describe (`AlarmTypes` filter round-trip) + delete-via-DeleteAlarms.

### DIFF
--- a/providers/aws/cloudwatch/cloudwatch.go
+++ b/providers/aws/cloudwatch/cloudwatch.go
@@ -52,13 +52,15 @@ type metricKey struct {
 
 // Mock is an in-memory mock implementation of the AWS CloudWatch service.
 type Mock struct {
-	mu       sync.RWMutex
-	metrics  map[metricKey][]driver.MetricDatum
-	alarms   *memstore.Store[*alarmData]
-	channels *memstore.Store[*driver.NotificationChannelInfo]
-	history  []driver.AlarmHistoryEntry
-	opts     *config.Options
-	sns      ActionPublisher
+	mu              sync.RWMutex
+	metrics         map[metricKey][]driver.MetricDatum
+	alarms          *memstore.Store[*alarmData]
+	compositeAlarms *memstore.Store[*compositeAlarmData]
+	dashboards      *memstore.Store[*storedDashboard]
+	channels        *memstore.Store[*driver.NotificationChannelInfo]
+	history         []driver.AlarmHistoryEntry
+	opts            *config.Options
+	sns             ActionPublisher
 }
 
 // SetSNSPublisher wires the SNS backend so an alarm state transition delivers
@@ -97,10 +99,12 @@ type alarmData struct {
 // New creates a new CloudWatch mock with the given configuration options.
 func New(opts *config.Options) *Mock {
 	return &Mock{
-		metrics:  make(map[metricKey][]driver.MetricDatum),
-		alarms:   memstore.New[*alarmData](),
-		channels: memstore.New[*driver.NotificationChannelInfo](),
-		opts:     opts,
+		metrics:         make(map[metricKey][]driver.MetricDatum),
+		alarms:          memstore.New[*alarmData](),
+		compositeAlarms: memstore.New[*compositeAlarmData](),
+		dashboards:      memstore.New[*storedDashboard](),
+		channels:        memstore.New[*driver.NotificationChannelInfo](),
+		opts:            opts,
 	}
 }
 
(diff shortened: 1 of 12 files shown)


NitinKumar004 (commented):
Verdict: MERGE.

Gates (pushed SHA e041c61): build, vet, gofmt, and the scoped tests
(`server/aws/cloudwatch`, `providers/aws/cloudwatch`, `services/monitoring/...`)
all pass; `-race` on `providers/aws/cloudwatch` is clean;
`golangci-lint --new-from-rev=origin/development` reports 0 new issues. Exactly
the 12 expected files changed.

Docs exemption is legitimate, not a missed regen. `go generate ./...` and
`go run ./internal/compatgen` (suite green, exit 0) both leave the tree clean.
The new ops are AWS-local optional capabilities reached by type assertion in the
wire handler (`dashboardStore` / `compositeAlarmStore` defined in the server
package); they are NOT added to the shared `services/monitoring/driver`
`Monitoring` interface (which still exposes only `DeleteAlarm` /
`DescribeAlarms`). The interface-driven coverage generator therefore correctly
does not surface them, and `docs/coverage/aws/cloudwatch.md` lists the
driver-backed ops (e.g. `DescribeAlarms`) but not dashboards/composite — exactly
consistent with zero-diff.

DescribeAlarms partitioning is correct: `AlarmTypes` gates the metric-alarm and
composite-alarm branches independently; empty `AlarmTypes` returns both; neither
list leaks into the other. Composites are attached only on the first page
(offset 0) to avoid duplication across metric-alarm pages — a small documented
deviation from AWS interleaving, acceptable for the TF flow.

DeleteAlarms is faithful to AWS: metric names tolerate not-found
(`!IsNotFound`), and `DeleteCompositeAlarms` silently skips absent names, so a
mixed / already-gone batch (terraform destroy) never errors. Deleting a
composite name works via the type-assertion tail.

Dashboards round-trip correctly: PutDashboard rejects invalid JSON
(InvalidArgument -> DashboardInvalidInputError shape), GetDashboard/DeleteDashboards
return NotFound as expected, ListDashboards prefix filter + sorted offset-token
pagination is stable. DeleteDashboards is all-or-nothing (validates all names
first) matching real AWS.

Composite-alarm state remains INSUFFICIENT_DATA (no rule engine) — an accepted,
documented deferral; the definition round-trips faithfully, satisfying the TF
lifecycle.

Minor non-blocking note: metric and composite alarms live in separate stores, so
cloudemu does not enforce AWS's global alarm-name uniqueness across the two
types. Edge case, no impact on the TF resources this targets.

PR #748, opened 2026-08-26
fix(ssm): SecureString KeyId round-trip and AllowedPattern enforcement

NitinKumar004:
## Summary

Closes two AWS SSM Parameter Store fidelity gaps around `aws_ssm_parameter` `key_id` and `allowed_pattern`.

### 1. SecureString KeyId round-trip + validation
- `PutParameter` with `Type=SecureString` accepts `KeyId`, defaulting to `alias/aws/ssm` (the AWS-managed key) when omitted.
- Supplying `KeyId` for a `String`/`StringList` parameter is rejected with `ValidationException` (real Parameter Store only honors `KeyId` for SecureString).
- `KeyId` is reflected back **only** on `DescribeParameters` `ParameterMetadata.KeyId` (and `GetParameterHistory`), matching real AWS — the `GetParameter`/`GetParameters` `Parameter` shape has no `KeyId` field and does not echo it.

### 2. AllowedPattern enforcement
- `PutParameter` accepts `AllowedPattern` (a regex); the `Value` must match or the call is rejected with `ValidationException`.
- An `AllowedPattern` that is not a valid regular expression is rejected with `ValidationException`.
- Enforced on overwrite as well, and reflected back on `DescribeParameters` `ParameterMetadata.AllowedPattern`.

## Why
Without these, IaC/SDK callers that set `key_id` or `allowed_pattern` on a parameter saw the fields silently dropped and invalid values silently accepted — drift from real Parameter Store behavior and its validation errors.

## Notes
- SecureString values remain stored in the clear (no real KMS integration); this change models the `KeyId` **association** and round-trip, not encryption.

## Tests
- Provider-level: KeyId default/explicit/rejection on non-SecureString; AllowedPattern match/mismatch/invalid-regex/overwrite; DescribeParameters reflects both; GetParameter omits both.
- Server-level (real aws-sdk-go-v2): KeyId + AllowedPattern round-trip through DescribeParameters, ValidationException on String+KeyId and on pattern mismatch.

### DIFF
--- a/providers/aws/ssm/keyid_allowedpattern_test.go
+++ b/providers/aws/ssm/keyid_allowedpattern_test.go
@@ -1,0 +1,213 @@
+package ssm_test
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/stackshy/cloudemu/v2/services/parameterstore/driver"
+)
+
+// describeByName returns the metadata for a single parameter name.
+func describeByName(t *testing.T, m interface {
+	DescribeParameters(context.Context) ([]driver.ParameterMetadata, error)
+}, name string) driver.ParameterMetadata {
+	t.Helper()
+
+	metas, err := m.DescribeParameters(context.Background())
+	if err != nil {
+		t.Fatalf("DescribeParameters: %v", err)
+	}
+
+	for _, md := range metas {
+		if md.Name == name {
+			return md
+		}
+	}
+
+	t.Fatalf("parameter %q not found in DescribeParameters", name)
+
+	return driver.ParameterMetadata{}
+}
+
+func TestPutSecureStringDefaultsKeyID(t *testing.T) {
+	m := newMock()
+	ctx := context.Background()
+
+	if _, _, err := m.PutParameter(ctx, driver.PutConfig{
+		Name: "/sec", Value: "v", Type: driver.TypeSecureString,
+	}); err != nil {
+		t.Fatalf("PutParameter: %v", err)
+	}
+
+	md := describeByName(t, m, "/sec")
+	if md.KeyID != driver.DefaultSecureStringKeyID {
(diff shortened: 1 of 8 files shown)


NitinKumar004 (commented):
MERGE. Verified on pushed SHA a4b4285.

Gates (GOMAXPROCS=3, -p=3): go build, go vet clean. Tests pass for ./server/aws/ssm/..., ./providers/aws/ssm/..., ./services/parameterstore/...; -race clean on providers/aws/ssm. gofmt clean on all 8 changed files; 0 new golangci-lint issues. go generate + compatgen produce zero diff (compatgen exit 0).

Diff hygiene: commit touches exactly the 8 expected SSM files and nothing else — no glue or other leakage.

Correctness:
- KeyId surfaces only on DescribeParameters metadata and GetParameterHistory. toParameter leaves KeyID/AllowedPattern empty and parameterJSON wire struct has neither field, so GetParameter/GetParameters do not echo them. TestGetParameterOmitsKeyID asserts both are empty; the SDK-level test corroborates.
- KeyId validation: String/StringList + KeyId -> ErrKeyIDOnNonSecure (ValidationException); SecureString with omitted KeyId defaults to alias/aws/ssm. Covered by tests.
- AllowedPattern: validated up front in PutParameter so it applies to create and overwrite; value mismatch and invalid regex both rejected as ValidationException; reflected on DescribeParameters. TestPutAllowedPatternEnforcedOnOverwrite confirms the overwrite path.
- Naming: fields renamed KeyID (revive) while wire json tags preserve "KeyId" (request json:"KeyId", metadata/history json:"KeyId,omitempty") so the SDK sees KeyId.
- KeyId filter key wired in server filter.go and provider path_filter.go for GetParametersByPath.

Ciphertext stored in clear (no real KMS) is an accepted deferral. No blocking issues.

PR #749, opened 2026-08-26
fix(eventbridge): PutPermission/RemovePermission resource policy, ListRuleNamesByTarget, reject empty {} event pattern

NitinKumar004:
## Summary

Closes three genuinely-open AWS EventBridge gaps.

### 1. Event-bus resource policy — `PutPermission` / `RemovePermission`
Adds the resource-based policy surface behind `aws_cloudwatch_event_permission`:
- **PutPermission** supports both forms — the legacy `Action` (must be `events:PutEvents`) / `Principal` (12-digit account id or `*`) / `StatementId` trio, optionally with a `Condition` (Type/Key/Value, e.g. `StringEquals aws:PrincipalOrgID`), and a full `Policy` JSON document (mutually exclusive with the trio). Legacy statements are stored keyed by `StatementId`; a `*` principal is stored as-is while an account id expands to its root ARN.
- **RemovePermission** removes a statement by `StatementId` (unknown id → `ResourceNotFoundException`) or clears the whole policy with `RemoveAllPermissions`.
- The resulting policy document is reflected on **DescribeEventBus** (`Policy` field).
- Validation: wrong `Action`, and the both-forms / neither-form cases, are rejected as `ValidationException`.

Resource policies are AWS-EventBridge-specific, so — following the existing `resourceTagger` precedent — this is exposed through an AWS-only optional interface (`permissionManager`) asserted against the provider, not the portable `EventBus` driver. The shared `PermissionInput`/`PermissionCondition` types and an `EventBusInfo.Policy` data field live in the driver package.

### 2. `ListRuleNamesByTarget`
Given a `TargetArn`, returns the names of rules routing to it on the given bus (default bus when omitted), with `NextToken`/`Limit` pagination like the other list ops. Derived purely from the existing `ListRules`/`ListTargets` driver methods, so no new provider hook is required.

### 3. Reject empty `{}` event pattern
`eventmatch.ValidatePattern` now rejects a top-level empty pattern `{}` (and `null`) with `InvalidEventPatternException` ("Empty patterns are not allowed"), matching real EventBridge on `PutRule`/`TestEventPattern`. Legitimately-empty **nested** objects and all existing valid patterns still pass (gate is exactly the top-level object).

## Tests
- Provider (hand-rolled helpers): PutPermission legacy / wildcard principal / Policy-doc / Condition / upsert-by-Sid; RemovePermission by-id, RemoveAll, unknown-id error; both-forms & neither-form & wrong-action & unknown-bus validation; DescribeEventBus reflects the policy; empty `{}` pattern rejected while a valid pattern passes.
- Wire (real aws-sdk-go-v2): PutPermission+Condition reflected on DescribeEventBus; RemovePermission by-id / RemoveAll / unknown-id → ResourceNotFoundException; ListRuleNamesByTarget correct names + pagination.
- `eventmatch` unit test for the empty-pattern gate.

## Gates
`go build ./...`, `go vet`, scoped `go test`, `gofmt -l`, and `golangci-lint --new-from-rev=origin/development` all clean. `go generate ./...` (coverage) and `go run ./internal/compatgen` (compat) produce zero diff — consistent with the tags precedent, these AWS-specific optional-interface ops are not on the portable driver, so they don't add coverage/compat rows.

### DIFF
--- a/internal/eventmatch/eventmatch.go
+++ b/internal/eventmatch/eventmatch.go
@@ -472,6 +472,13 @@ func ValidatePattern(raw string) error {
 		return &PatternError{Reason: "event pattern is not valid JSON"}
 	}
 
+	// A top-level empty object {} (or null) is rejected: real EventBridge refuses
+	// an empty pattern on PutRule. Empty nested objects stay legal — they are
+	// validated recursively by validatePatternObject, not here.
+	if len(p) == 0 {
+		return &PatternError{Reason: "Empty patterns are not allowed"}
+	}
+
 	return validatePatternObject(p)
 }
 
--- a/internal/eventmatch/eventmatch_test.go
+++ b/internal/eventmatch/eventmatch_test.go
@@ -280,3 +280,26 @@ func TestParsePatternRejectsGarbage(t *testing.T) {
 		t.Fatal("expected empty pattern to fail parse")
 	}
 }
+
+// TestValidatePatternEmptyObject asserts that a top-level empty pattern {} (and
+// null) is rejected the way real EventBridge rejects it, while a legitimately
+// empty NESTED object and ordinary valid patterns are still accepted.
+func TestValidatePatternEmptyObject(t *testing.T) {
+	reject := []string{"{}", " {} ", "null"}
+	for _, raw := range reject {
+		if err := eventmatch.ValidatePattern(raw); err == nil {
+			t.Fatalf("ValidatePattern(%q) = nil, want rejection", raw)
+		}
+	}
+
+	accept := []string{
+		`{"source":["aws.ec2"]}`,
+		`{"detail":{}}`,
+		`{"detail":{"state":["running"]}}`,
+	}
+	for _, raw := range accept {
+		if err := eventmatch.ValidatePattern(raw); err != nil {
+			t.Fatalf("ValidatePattern(%q) = %v, want nil", raw, err)
+		}
+	}
+}
--- a/providers/aws/eventbridge/eventbridge.go
+++ b/providers/aws/eventbridge/eventbridge.go
@@ -42,6 +42,9 @@ type busData struct {
 	rules  *memstore.Store[*ruleData]
 	mu     sync.RWMutex
 	events []driver.Event
+	// policyStmts holds the event bus's resource-policy statements in insertion
+	// order, keyed for removal by each statement's "Sid". Guarded by mu.
+	policyStmts []map[string]any
 }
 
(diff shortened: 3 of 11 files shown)


NitinKumar004 (commented):
Verdict: MERGE

Verified against the pushed head SHA 27c0c02. Clean build + vet pass; exactly the 11 expected files changed; gofmt clean; 0 new lint issues; `go generate ./...` and `internal/compatgen` both produce zero diff (compatgen exit 0).

## Shared `internal/eventmatch` change (blast radius)
The only behavioral change to the shared codec is a top-level empty `{}`/`null` rejection inside `ValidatePattern`. Full importer sweep of `internal/eventmatch`:
- providers/aws/eventbridge/eventbridge.go — calls `ValidatePattern` (PutRule) plus `ParsePattern`/`MatchEvent` (matching)
- server/aws/eventbridge/operations.go — calls `ValidatePattern` (PutRule, TestEventPattern) plus `ParsePattern`/`MatchEvent`
- providers/aws/sns/filter_policy.go — uses only `ParsePattern`/`MatchEvent`/`MatchLeafAttr`/`IsOrClause`, never `ValidatePattern`

`ValidatePattern` has no callers outside EventBridge, so the `{}` rejection cannot affect SNS filter policies (or any other service). Confirmed the SNS filter path routes through `ParsePattern`, so an empty `{}` filter policy there is untouched by this change.

Gate scope confirmed exactly top-level empty: a throwaway table test (fields pattern, `{"detail":{}}`, `{"detail":{"inner":{}}}`, top-level `{}`, top-level `null`, a complex numeric/nested pattern) shows only the top-level empties are rejected; nested empty objects and all previously-valid patterns still pass. Reverse-dependent tests all green: internal/eventmatch, server+providers eventbridge, server+providers sns.

## EventBridge additive ops
- PutPermission: legacy trio (Action must be events:PutEvents, Principal, StatementId) and full Policy-doc form both handled; both-given or neither-given -> InvalidArgument (ValidationException); Condition rendered into the statement; statements stored in insertion order keyed by Sid with upsert-by-Sid.
- RemovePermission: RemoveAllPermissions clears; unknown StatementId -> NotFound (ResourceNotFoundException); missing statementId without removeAll -> validation error.
- DescribeEventBus reflects the rendered policy (GetEventBus renders under RLock; response field is `Policy,omitempty` so it is omitted when the bus has no policy).
- ListRuleNamesByTarget: derived from ListRules/ListTargets, requires TargetArn, reuses the existing stable `paginateByCursor` and `busNameOrDefault` helpers.

No blocking issues.

PR #750, opened 2026-08-26
fix(config): recordingMode round-trip + EvaluationResultIdentifier in GetComplianceDetailsByConfigRule

NitinKumar004:
## Summary

Two AWS Config fidelity gaps that caused Terraform/SDK drift.

### 1. recordingMode round-trip on the configuration recorder
Real `PutConfigurationRecorder` accepts a `RecordingMode { RecordingFrequency, RecordingModeOverrides[] }` alongside `recordingGroup`. It was not modeled anywhere, so `DescribeConfigurationRecorders` never echoed it and Terraform saw perpetual drift.

- Added `RecordingMode` / `RecordingModeOverride` to the recorder driver model.
- Stored what `PutConfigurationRecorder` sends (create + idempotent upsert paths) and reflected it faithfully on `DescribeConfigurationRecorders`, including per-resource-type overrides.
- Validate `RecordingFrequency` is `CONTINUOUS` or `DAILY` when present (top-level and per override).
- Round-trip fidelity only; actual recording behavior is unchanged.

### 2. EvaluationResultIdentifier in GetComplianceDetailsByConfigRule
The response only carried `ComplianceType` / `Annotation` / `ResultRecordedTime`. Consumers that key results by resource need the full identifier.

- Populated `EvaluationResultIdentifier.EvaluationResultQualifier` (`ConfigRuleName`, `ResourceType`, `ResourceId`) plus `OrderingTimestamp`, `ConfigRuleInvokedTime`, and `ResultRecordedTime`.
- `ResourceType` / `ResourceId` come from the stored evaluation (already carried on the record — nothing fabricated). `ConfigRuleName` comes from the request. The wire `toDriver` now threads the client-supplied `OrderingTimestamp` (previously dropped); `PutEvaluations` / `PutExternalEvaluation` default it to now when omitted so timestamps are always populated.
- The by-resource and conformance-pack paths span multiple rules and do not surface a single rule name, so their qualifier's `ConfigRuleName` is left empty (ResourceType/ResourceId still populated).

## Tests
- `TestSDKRecorderRecordingModeRoundTrip` — CONTINUOUS, then DAILY + override, round-trip Put -> Describe.
- `TestSDKRecorderInvalidRecordingFrequency` — invalid frequency rejected.
- `TestSDKComplianceDetailsEvaluationResultIdentifier` — identifier qualifier fields + timestamps match the recorded evaluation.

All via the real AWS SDK against the wire server; no testify.

### DIFF
--- a/providers/aws/configservice/evaluations.go
+++ b/providers/aws/configservice/evaluations.go
@@ -76,10 +76,18 @@ func (m *Mock) PutEvaluations(
 		return nil, invalidResultToken(resultToken)
 	}
 
+	now := m.now()
+
+	for i := range evals {
+		if evals[i].OrderingTimestamp.IsZero() {
+			evals[i].OrderingTimestamp = now
+		}
+	}
+
 	rd.mu.Lock()
 	rd.evals = append(rd.evals, evals...)
 	rd.rule.Compliance = rollUpCompliance(rd.evals)
-	rd.rule.LastSuccessfulEval = m.now()
+	rd.rule.LastSuccessfulEval = now
 	rd.mu.Unlock()
 
 	return nil, nil
@@ -143,6 +151,10 @@ func (m *Mock) PutExternalEvaluation(_ context.Context, ruleName string, eval dr
 	rd, ok := m.rules.Get(ruleName)
 	if !ok {
 		return noSuchConfigRule(ruleName)
+	}
+
+	if eval.OrderingTimestamp.IsZero() {
+		eval.OrderingTimestamp = m.now()
 	}
 
 	rd.mu.Lock()
--- a/providers/aws/configservice/recorders.go
+++ b/providers/aws/configservice/recorders.go
@@ -23,6 +23,10 @@ func (m *Mock) PutConfigurationRecorder(_ context.Context, rec driver.Configurat
 		return invalidParameter("RoleARN is required")
 	}
 
+	if err := validateRecordingMode(rec.RecordingMode); err != nil {
+		return err
+	}
+
 	now := m.now()
 
 	// Hold createMu across the scan+insert: the single-recorder cap is not atomic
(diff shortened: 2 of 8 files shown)


NitinKumar004 (commented):
**Verdict: approve.** AWS Config recordingMode round-trip + EvaluationResultIdentifier land cleanly. Gates green on the pushed SHA (build/vet, scoped tests incl. `-race`, gofmt, 0 new lint, `go generate` + compatgen zero-diff). 8 files, all expected.

**Shared `evalResults` helper blast radius — correct.** All three callers compile and are semantically right:
- `GetComplianceDetailsByConfigRule` passes `req.ConfigRuleName` (has it) — full qualifier.
- `getAggregateComplianceDetailsByConfigRule` passes `req.ConfigRuleName`, which is a *required* input parameter of that API, so the qualifier's rule name is real, not empty.
- `getComplianceDetailsByResource` and `getConformancePackComplianceDetails` pass `""`. This is honest: the driver signatures (`GetComplianceDetailsByResource(resourceType, resourceID, page)`, `GetConformancePackComplianceDetails(name, page)`) carry no per-evaluation rule name, so there is nothing to fabricate. `ConfigRuleName,omitempty` drops the key rather than emitting a wrong rule, and ResourceType/ResourceId are still populated from the stored evaluation. Acceptable known limitation.

The added `EvaluationResultIdentifier` / `ConfigRuleInvokedTime` fields match the SDK shapes for `EvaluationResult`, `ConformancePackEvaluationResult`, and `AggregateEvaluationResult`; no pre-existing test asserted the old flat shape, and the by-resource/pack paths were not previously result-asserted, so no regression.

**recordingMode round-trip — solid.** Nested wire keys (`recordingMode`/`recordingModeOverrides`, lowerCamel) correct; stored on Put (both insert and upsert), reflected on Describe. Frequency validation rejects anything outside `""`/CONTINUOUS/DAILY on both the top-level frequency and each override. No slice-aliasing bug: `copyRecordingMode` value-copies each override and deep-copies `ResourceTypes` via `copyStrings` on every read, and each Put supplies a fresh per-request struct, so two recorders never share an overrides slice.

**OrderingTimestamp threading — sensible.** Now threaded through wire `toDriver` (`timeFromEpoch`), and `PutEvaluations`/`PutExternalEvaluation` default a zero timestamp to `now()`. Round-trips at second granularity (test asserts `.Equal(ordering)`). `ConfigRuleInvokedTime`/`ResultRecordedTime` both derive from ordering — a reasonable approximation. Minor: `PutEvaluations` mutates the caller's slice elements in place when defaulting; harmless here (wire supplies a fresh slice) but worth a note for direct driver callers.

No secret/plaintext exposure.

PR #751, opened 2026-08-26
fix(rds): ModifyDBInstance honors ApplyImmediately and records PendingModifiedValues

NitinKumar004:
## What

AWS RDS `ModifyDBInstance` now honors the `ApplyImmediately` flag and exposes `PendingModifiedValues`, matching real RDS semantics. Previously the flag was ignored — every modification applied immediately and no pending state was ever surfaced, so a user's deferred change (e.g. an instance-class resize meant to wait for the maintenance window) was silently applied and `PendingModifiedValues` was always absent. This drove `aws_db_instance` Terraform drift.

## Behavior (matches real AWS)

- **`ApplyImmediately=true`**: the deferrable target fields (DBInstanceClass, AllocatedStorage, EngineVersion, MasterUserPassword, MultiAZ, BackupRetentionPeriod, Iops, StorageType) update on the instance now; `PendingModifiedValues` is cleared. The master password is rotated on the backing engine.
- **`ApplyImmediately=false` (default)**: the instance keeps its current values; the requested changes are recorded in `PendingModifiedValues` (nested `DBInstance` element on `DescribeDBInstances`) and the engine password is left untouched.
- Only fields that actually **differ** from the current value are recorded (real RDS omits no-op changes).
- A pending master-password change is echoed masked as `****` — the plaintext is never stored or returned.
- Parameter/option groups, tags, backup/maintenance windows and deletion protection continue to apply immediately regardless of the flag.

## Layers touched

- `services/relationaldb/driver` — `ModifyInstanceInput.ApplyImmediately`, `Instance.PendingModifiedValues`, `PendingModifiedValues` type + `MaskedPassword`.
- `providers/aws/rds` — deferred-vs-immediate branch in `ModifyInstance`; pending-diff computation.
- `server/aws/rds` — parse the `ApplyImmediately` form field; nested `PendingModifiedValues` wire element on the `DBInstance` response.

## Tests

Provider-level (hand-rolled helpers) and SDK wire round-trip tests cover: apply-immediately updates now with empty pending; deferred records pending + leaves current unchanged + reflected on describe; password masked in pending / never rotated when deferred; no-op field omitted from pending. Existing tests that relied on the old always-immediate default were updated to set `ApplyImmediately=true`.

### DIFF
--- a/providers/aws/rds/engine_test.go
+++ b/providers/aws/rds/engine_test.go
@@ -357,7 +357,9 @@ func TestModifyInstanceRotatesPassword(t *testing.T) {
 		t.Fatalf("CreateInstance: %v", err)
 	}
 
-	if _, err := m.ModifyInstance(ctx, "pg", rdsdriver.ModifyInstanceInput{MasterUserPassword: "new"}); err != nil {
+	if _, err := m.ModifyInstance(ctx, "pg", rdsdriver.ModifyInstanceInput{
+		MasterUserPassword: "new", ApplyImmediately: true,
+	}); err != nil {
 		t.Fatalf("ModifyInstance: %v", err)
 	}
 
--- a/providers/aws/rds/instance_config_validation_test.go
+++ b/providers/aws/rds/instance_config_validation_test.go
@@ -83,7 +83,8 @@ func TestModifyInstanceAcceptsValidAndEmptyClass(t *testing.T) {
 	}
 
 	out, err := m.ModifyInstance(context.Background(), "db1", rdsdriver.ModifyInstanceInput{
-		InstanceClass: "db.r5.large",
+		InstanceClass:    "db.r5.large",
+		ApplyImmediately: true,
 	})
 	if err != nil {
 		t.Fatalf("valid class modify should succeed: %v", err)
--- a/providers/aws/rds/modify_apply_immediately_test.go
+++ b/providers/aws/rds/modify_apply_immediately_test.go
@@ -1,0 +1,168 @@
+package rds
+
+import (
+	"context"
+	"testing"
+
+	rdsdriver "github.com/stackshy/cloudemu/v2/services/relationaldb/driver"
+)
+
+// seedModifyInstance creates a standalone db.t3.micro instance the
+// ApplyImmediately tests modify.
+func seedModifyInstance(t *testing.T, m *Mock) {
+	t.Helper()
+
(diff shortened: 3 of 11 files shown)


NitinKumar004 (commented):
Verified the behavioral change end to end on the pushed SHA (55df80f). Clean build / vet / scoped tests / -race all green; gofmt clean; 0 new lint; go generate and compatgen both zero-diff; exactly the expected 11 files touched.

Correctness:

- Deferred path (ApplyImmediately unset, the default): ModifyInstance takes the else branch, calls neither applyDeferrableMods nor rotateEnginePassword, so current InstanceClass/AllocatedStorage/EngineVersion/etc. stay put and the engine credential is never rotated. Only PendingModifiedValues is populated. Confirmed by TestModifyInstanceApplyImmediatelyFalse, TestModifyInstancePendingPasswordMasked (asserts rootPasswords not rotated), and the SDK round-trip (describe still reports db.t3.micro/20).
- Immediate path (ApplyImmediately=true): applyDeferrableMods applies the fields now, rotateEnginePassword runs, and PendingModifiedValues is cleared to nil. Confirmed by TestModifyInstanceApplyImmediatelyTrue and TestSDKRDSModifyApplyImmediatelyClearsPending.
- Password masking: pendingModifiedValues writes rdsdriver.MaskedPassword (****), never the plaintext; the XML mapper echoes only that field. No path serializes the real password into pending. TestModifyInstancePendingPasswordMasked + SDK test assert the masking and non-leak.
- Diff-only pending: pendingStr/pendingInt/pendingBool omit a requested value equal to the current one; IsEmpty()->nil when nothing differs. TestModifyInstanceNoOpNotPending and TestModifyInstanceNoDeferrableChangeClearsPending cover both.
- Wire shape: PendingModifiedValues is a nested complex element under DBInstance, omitempty when nil, verified through the real aws-sdk-go-v2 round-trip (pmv.DBInstanceClass/AllocatedStorage/MasterUserPassword).
- The updated existing tests each assert the modification actually took effect on the instance (password rotated, class/storage changed), so adding ApplyImmediately=true is exactly what a real caller sends for that assertion under the new default-deferred semantics — legitimate, not masking a regression.
- Blast radius confined: only AWS RDS ModifyInstance reads the flag; Azure flex/SQL, GCP Cloud SQL/AlloyDB and Redshift implement their own ModifyInstance and ignore it. applyImmediateMods (param/option groups, windows, deletion protection, tags) still applies regardless of the flag, matching real RDS.

No blocking issues. LGTM.

PR #752, opened 2026-08-26
fix(eks): DescribeCluster returns logging, kubernetesNetworkConfig, and accessConfig

NitinKumar004:
## What

`DescribeCluster` now returns three fields it previously dropped, so `aws_eks_cluster` Terraform no longer drifts after create:

- **logging** — `{ "clusterLogging": [ { "types": [...], "enabled": bool } ] }`
- **kubernetesNetworkConfig** — `{ "serviceIpv4Cidr", "serviceIpv6Cidr", "ipFamily" }`
- **accessConfig** — `{ "authenticationMode", "bootstrapClusterCreatorAdminPermissions" }`

`CreateCluster` accepts all three and `DescribeCluster` faithfully echoes them (round-trip). When omitted on create, the real-EKS defaults are applied so the post-create read is populated and drift-free.

## Why

The wire `clusterJSON` had no `logging`, `kubernetesNetworkConfig`, or `accessConfig`, so an `aws_eks_cluster` resource read back empty values for these attributes right after apply, producing perpetual Terraform diffs.

## Defaults (AWS-accurate)

- **logging**: omitted -> one `clusterLogging` entry with all five control-plane types (`api, audit, authenticator, controllerManager, scheduler`) and `enabled: false` (the AWS default state).
- **kubernetesNetworkConfig**: `ipFamily` defaults to `ipv4`; `serviceIpv4Cidr` auto-assigned `10.100.0.0/16` when omitted. `ipFamily: ipv6` assigns a `serviceIpv6Cidr` (`fd00::/108`) instead.
- **accessConfig**: `authenticationMode` defaults to `CONFIG_MAP` — the documented default on the EKS API / SDK / CloudFormation path (only the AWS Console defaults to `API_AND_CONFIG_MAP`), which is the path Terraform uses. `bootstrapClusterCreatorAdminPermissions` defaults to `true`.

## Layering

Threaded through all three layers, following the existing VPCConfig pattern: driver model (`providers/aws/eks/driver`) -> provider store defaults/echo (`providers/aws/eks`) -> wire `toWire`/`toDriver` (`server/aws/eks`).

## Tests

- Provider-level (hand-rolled helpers): explicit echo, omitted-defaults, and the ipv6 path.
- Wire-level (real aws-sdk-go-v2 EKS client): CreateCluster -> DescribeCluster echo and defaults over the wire.

### DIFF
--- a/providers/aws/eks/cluster_config_test.go
+++ b/providers/aws/eks/cluster_config_test.go
@@ -1,0 +1,114 @@
+package eks
+
+import (
+	"context"
+	"testing"
+
+	eksdriver "github.com/stackshy/cloudemu/v2/providers/aws/eks/driver"
+)
+
+// TestCreateClusterConfigEcho verifies CreateCluster faithfully round-trips the
+// caller's logging, kubernetesNetworkConfig, and accessConfig into the stored
+// cluster read back by DescribeCluster.
+func TestCreateClusterConfigEcho(t *testing.T) {
+	m := newTestMock()
+	ctx := context.Background()
+
+	bootstrap := false
+	cfg := eksdriver.ClusterConfig{
+		Name:    "echo-cluster",
+		RoleArn: "arn:aws:iam::123456789012:role/eks-cluster",
+		Logging: []eksdriver.ClusterLogging{
+			{Types: []string{"api", "audit"}, Enabled: true},
+			{Types: []string{"authenticator", "controllerManager", "scheduler"}, Enabled: false},
+		},
+		NetworkConfig: eksdriver.NetworkConfig{
+			ServiceIPv4CIDR: "172.20.0.0/16",
+			IPFamily:        "ipv4",
+		},
+		AccessConfig: eksdriver.AccessConfigRequest{
+			AuthenticationMode:                      "API",
+			BootstrapClusterCreatorAdminPermissions: &bootstrap,
+		},
+	}
+
+	if _, err := m.CreateCluster(ctx, cfg); err != nil {
+		t.Fatalf("CreateCluster: %v", err)
+	}
+
+	got, err := m.DescribeCluster(ctx, "echo-cluster")
+	requireNoError(t, err)
+
+	if len(got.Logging) != 2 {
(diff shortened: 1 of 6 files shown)


NitinKumar004 (commented):
Reviewed at head e146c0e. MERGE.

Gates (clean checkout of the pushed SHA): build, vet, go test (server/aws/eks + providers/aws/eks) and -race on the provider all green. gofmt clean. golangci-lint --new-from-rev=origin/development: 0 new issues. go generate + compatgen produce zero diff. Exactly the 6 expected files changed.

Correctness:
- Round-trip: CreateCluster -> DescribeCluster faithfully echoes logging, kubernetesNetworkConfig, and accessConfig over the real aws-sdk-go-v2 wire. Nested JSON keys match AWS: clusterLogging:[{types,enabled}], {serviceIpv4Cidr,serviceIpv6Cidr,ipFamily}, {authenticationMode,bootstrapClusterCreatorAdminPermissions}.
- authenticationMode default CONFIG_MAP is correct. Per the AWS EKS API reference, CreateCluster via API/SDK/CloudFormation defaults to CONFIG_MAP; only the console defaults to API_AND_CONFIG_MAP. Terraform's aws_eks_cluster uses the SDK path, so CONFIG_MAP is the no-drift default.
- bootstrapClusterCreatorAdminPermissions *bool handling is correct: omitted -> true, explicit false -> false (nil-check before deref, no coercion). accessConfigResponse omits omitempty on the bool so an explicit false serializes.
- Omitted-value defaults are populated on describe with no drift: logging = all 5 control-plane types, enabled:false; ipFamily ipv4 + serviceIpv4Cidr 10.100.0.0/16; ipv6 path assigns serviceIpv6Cidr.
- JSON hygiene: response structs use pointer + omitempty so unset blocks are omitted (not null); casing is correct camelCase throughout; resolveAccessConfig always sets a mode so accessConfig is never dropped.

SDK tests cover both the explicit-echo (including bootstrap=false distinguishable from omitted) and the defaults path. No blocking issues.

PR #754, opened 2026-08-26
fix(vpc): Subnet AvailableIpAddressCount, public NAT requires EIP, Interface endpoints back ENIs

NitinKumar004:
Fixes three AWS VPC/EC2 gaps that drove Terraform drift and missing validation on core networking resources. All three are independent and land together.

## GAP 1 — Subnet AvailableIpAddressCount now moves
Real EC2 DescribeSubnets returns `availableIpAddressCount` = CIDR host space minus 5 AWS-reserved addresses minus IPs consumed by resident interfaces. Previously the wire layer computed a static CIDR-5 value that never decremented.

- The provider (`providers/aws/vpc`) is now the source of truth: base = 2^(32-prefix) - 5, minus the ENIs resident in the subnet. Populated on CreateSubnet and DescribeSubnets.
- Because instance launches materialize a primary ENI in the subnet (and NAT gateways / interface endpoints occupy ENIs too), the count decrements on launch and re-increments on terminate automatically.
- The wire layer reflects the provider value; the old static CIDR-only computation was removed.

## GAP 2 — Public NAT gateway requires an EIP AllocationId
Real AWS requires a `public` NAT gateway to reference an allocated EIP, and forbids one on a `private` gateway.

- connectivityType defaults to `public`.
- public + no AllocationId -> `MissingParameter`.
- public + unknown AllocationId -> `InvalidAllocationID.NotFound`.
- private + AllocationId -> `InvalidParameterValue`.
- A public NAT now reflects the looked-up EIP's public IP instead of fabricating one.

## GAP 3 — Interface VPC endpoints create backing ENIs
Real AWS creates one ENI per specified subnet for an `Interface`-type endpoint and returns `networkInterfaceIds`; Gateway endpoints create none.

- CreateVpcEndpoint (Interface) provisions one ENI per subnet through the existing ENI store, so DescribeNetworkInterfaces sees them and they block premature subnet/VPC deletes.
- DescribeVpcEndpoints returns `networkInterfaceIdSet`.
- DeleteVpcEndpoint releases the backing ENIs. Gateway endpoints create none.

## Notes / simplifications
- Subnet IP accounting counts resident ENIs rather than modeling every AWS edge (secondary private IPs, prefix delegation); the count is correct for real launches, standalone ENIs, NAT gateways and interface endpoints.
- ModifyVpcEndpoint does not add/remove ENIs when an Interface endpoint's subnet set changes (out of scope here).

## Tests
Provider (hand-rolled helpers) and wire (real SDK) tests: empty-subnet CIDR-5, decrement on instance launch + standalone ENI, re-increment on terminate; public NAT without/with-unknown allocation rejected, public NAT reflects the EIP's IP, private NAT with allocation rejected; Interface endpoint returns one networkInterfaceId per subnet (visible via DescribeNetworkInterfaces) and Gateway endpoint returns none. Existing NAT tests updated to allocate real EIPs / use private connectivity.

Gates (scoped): build, vet, gofmt, `go generate` + compatgen zero-diff, scoped tests, and `golangci-lint --new-from-rev=origin/development` all clean.

### DIFF
--- a/cloudemu_test.go
+++ b/cloudemu_test.go
@@ -4707,9 +4707,12 @@ func TestNATGateway(t *testing.T) {
 				t.Fatalf("CreateSubnet: %v", err)
 			}
 
+			// A private NAT gateway needs no Elastic IP, so this cross-provider CRUD
+			// check stays uniform: AWS requires an AllocationId only for a public one.
 			nat, err := p.d.CreateNATGateway(ctx, netdriver.NATGatewayConfig{
-				SubnetID: subnet.ID,
-				Tags:     map[string]string{"env": "test"},
+				SubnetID:         subnet.ID,
+				ConnectivityType: "private",
+				Tags:             map[string]string{"env": "test"},
 			})
 			if err != nil {
 				t.Fatalf("CreateNATGateway: %v", err)
--- a/providers/aws/vpc/endpoint.go
+++ b/providers/aws/vpc/endpoint.go
@@ -8,6 +8,16 @@ import (
 	"github.com/stackshy/cloudemu/v2/services/networking/driver"
 )
 
+// vpcEndpointTypeInterface is the endpoint type that provisions a backing ENI in
+// each specified subnet. Gateway-type endpoints hold no interfaces.
+const vpcEndpointTypeInterface = "Interface"
+
+// endpointENIDescription is the description stamped on the ENIs an Interface
+// endpoint occupies, so DeleteVpcEndpoint can release exactly this endpoint's set.
+func endpointENIDescription(endpointID string) string {
+	return "VPC Endpoint Interface " + endpointID
+}
+
 // copyStringSlice creates a shallow copy of a string slice.
 func copyStringSlice(src []string) []string {
 	if src == nil {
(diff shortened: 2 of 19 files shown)


NitinKumar004 (commented):
Verified this change end-to-end against real AWS semantics. Build, vet, gofmt, and golangci-lint (new-from-development) are clean; `go generate` and compatgen produce zero diff. Full reverse-dependency test sweep passes: `providers/aws/vpc`, `server/aws/ec2`, `services/networking`, `services/resourcediscovery`, `compat/aws`, `features/topology`, and the root integration suite — plus `-race` on `providers/aws/vpc`.

NAT-gateway Elastic IP validation — error codes match real EC2:
- public + no AllocationId -> `MissingParameter`
- public + unknown allocation -> `InvalidAllocationID.NotFound`
- private + AllocationId -> `InvalidParameterValue` (falls through the create mapper to the shared InvalidArgument path, which emits InvalidParameterValue)
- public NAT reflects the EIP's real public address rather than a fabricated one
Validation runs before the ENI is attached, so a rejected create leaves no trace.

The modified existing tests are legitimate real-user reflections, not workarounds: tests that previously passed fabricated allocation IDs (`eipalloc-12345678`) — which real AWS rejects — now either allocate a real EIP via AllocateAddress (the public-path tests) or switch to `ConnectivityType: private` where the test isn't exercising the EIP path at all (ENI-holding, lifecycle, filter, walk tests). The public/EIP behavior gets dedicated new coverage.

AvailableIpAddressCount is computed dynamically by counting resident ENIs (base = 2^(32-prefix) - 5), so there is no stale accounting to drift. Verified adversarially: overfilling a /28 past capacity clamps to 0 and never goes negative; /30 and /32 base to 0; /29 = 3; a launched instance's primary ENI counts exactly once and terminate restores the count exactly. Recompute uses the thread-safe memstore only, no lock reentrancy — race clean.

Interface-endpoint ENIs: one per subnet, visible to DescribeNetworkInterfaces, blocking premature subnet/VPC delete via the existing eniInSubnet guard, and released on DeleteVpcEndpoint by an endpoint-id description marker (independent of the endpoint's current SubnetIDs) — so no ENI leak even after a subnet modify. Gateway endpoints provision none.

Deferrals (secondary-IP / prefix-delegation not modeled; ModifyVpcEndpoint subnet-change doesn't re-ENI) are genuine edges and documented; the common create/delete flows are correct and delete never leaks.

LGTM.

PR #755, opened 2026-08-26
fix(redshift): preserve snapshot KmsKeyId across CreateClusterSnapshot/RestoreFromClusterSnapshot

NitinKumar004:
## What
Redshift cluster snapshots now carry the encryption key. `ClusterSnapshot` gained a `KmsKeyID` field that:

- `CreateClusterSnapshot` copies from the source cluster (alongside the existing `Encrypted` flag).
- `DescribeClusterSnapshots` returns via the wire `snapshotXML` (`KmsKeyId`, omitempty).
- `RestoreFromClusterSnapshot` reads back onto the restored cluster — inheriting the snapshot's key, or honoring a `KmsKeyId` override supplied on the restore request.

An encrypted cluster created without an explicit key now gets the account-default `alias/aws/redshift` (mirroring the existing RDS `resolveKMSKeyID` behavior), so there is always a key to preserve.

## Why
Previously a `ClusterSnapshot` had only `Encrypted bool` and no `KmsKeyId`, so the KMS key was lost across CreateClusterSnapshot -> RestoreFromClusterSnapshot and omitted from DescribeClusterSnapshots. Real Redshift preserves the encryption key; the gap drove `aws_redshift_cluster`/snapshot Terraform drift.

## Scope / safety
`ClusterSnapshot` is shared with RDS. RDS never sets the new field, so its cluster-snapshot output is unchanged (field is `omitempty`); RDS provider + server tests pass.

## Tests
- Provider (hand-rolled helpers): encrypted cluster -> snapshot captures KmsKeyId -> Describe shows it -> restore preserves it on the new cluster (verified via DescribeClusters); restore with explicit KmsKeyId override honored; encrypted-without-key gets the default alias; unencrypted -> empty KmsKeyId + Encrypted=false through snapshot and restore.
- Server SDK round-trip: KmsKeyId survives Create -> Describe -> Restore -> DescribeClusters over the real wire.

### DIFF
--- a/providers/aws/redshift/redshift.go
+++ b/providers/aws/redshift/redshift.go
@@ -25,9 +25,12 @@ import (
 )
 
 const (
-	defaultEngine            = "redshift"
-	defaultPort              = 5439
-	singleNodeCount          = 1
+	defaultEngine   = "redshift"
+	defaultPort     = 5439
+	singleNodeCount = 1
+	// defaultKMSKeyAlias is the account-default Redshift KMS key real AWS fills in
+	// when a cluster is created encrypted without an explicit KmsKeyId.
+	defaultKMSKeyAlias       = "alias/aws/redshift"
 	snapshotBackupSizeMB     = 100.0
 	cpuUtilizationRunning    = 25.0
 	databaseConnectionsRun   = 5.0
@@ -500,7 +503,7 @@ func (m *Mock) reserveCluster(cfg rdbdriver.ClusterConfig) (rdbdriver.Cluster, e
 		NodeType:                    cfg.NodeType,
 		NumberOfNodes:               numberOfNodes,
 		Encrypted:                   cfg.Encrypted,
-		KmsKeyID:                    cfg.KmsKeyID,
+		KmsKeyID:                    resolveKMSKeyID(cfg.Encrypted, cfg.KmsKeyID),
 		PubliclyAccessible:          cfg.PubliclyAccessible,
 		AvailabilityZone:            cfg.AvailabilityZone,
 		CreatedAt:                   m.opts.Clock.Now().UTC(),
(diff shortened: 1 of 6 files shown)


NitinKumar004 (commented):
Verdict: MERGE.

Scope: exactly the 6 expected files. Build + vet clean on the pushed SHA; gofmt clean; 0 new golangci-lint issues on the changed packages; `go generate ./...` + compatgen produce zero diff.

Shared-type blast radius (KmsKeyID added to relationaldb/driver.ClusterSnapshot and RestoreClusterInput):
- RDS is unaffected. server/aws/rds/xml.go has its own dbClusterSnapshotXML (no KmsKeyId element) and toClusterSnapshotXML never reads the new snap.KmsKeyID field, so RDS DescribeDBClusterSnapshots XML is byte-identical. RDS never populates the new field.
- Azure (mysqlflex/postgresflex/sql) and GCP (alloydb/cloudsql) consumers never set the field; omitempty keeps their output unchanged.
- Full consumer test run passed: services/relationaldb/..., providers/aws/{redshift,rds}, server/aws/{redshift,rds}, providers/azure/{mysqlflex,postgresflex,sql}, providers/gcp/{alloydb,cloudsql}, server/gcp/alloydb, services/cost.

Redshift KmsKeyId logic:
- Capture: CreateClusterSnapshot copies cluster.KmsKeyID onto the snapshot; DescribeClusterSnapshots returns it via KmsKeyId (omitempty when unencrypted). Verified.
- Restore precedence: restoredKMSKeyID = override > snapshot key > default, gated on Encrypted. All three branches covered by unit + SDK-roundtrip tests (override honored, inheritance, unencrypted stays empty).
- Default resolution: resolveKMSKeyID fills alias/aws/redshift only when Encrypted=true and no explicit key (mirrors the RDS precedent). Unencrypted clusters correctly carry empty KmsKeyId with Encrypted=false.
- Sensible safety net: restore of an encrypted snapshot that predates key capture (empty snap key) falls back to the account default rather than emitting an encrypted cluster with no key.

No correctness, wire-parity, or blast-radius concerns. Good to merge.

PR #756, opened 2026-08-26
fix(elbv2): DescribeTargetHealth Target.NotRegistered; LB provisioning->active settle

NitinKumar004:
## What

Two AWS ELBv2 fidelity fixes to match real cloud behavior.

**GAP 1 — DescribeTargetHealth on an unregistered target.** Real ELBv2 does not silently drop a target that an explicit `Targets` query names but that isn't registered to the target group — it returns a `TargetHealthDescription` with `State=unused` and `Reason=Target.NotRegistered`. The wire handler's `filterHealth` previously dropped such targets. It now iterates the requested targets, returns real health for registered ones (matching on `(ID, Port)`, or all registrations of an ID when no port is given), and emits a synthetic `unused` / `Target.NotRegistered` description for the rest. With no explicit `Targets`, only registered targets are returned (unchanged).

**GAP 2 — Load balancer provisioning -> active.** Real ELBv2 returns a new load balancer in `provisioning` and settles it to `active` shortly after. `CreateLoadBalancer` now registers a `settle.Window` (intermediate `provisioning`, final `active`); `CreateLoadBalancer` and `DescribeLoadBalancers` overlay the observed state at read time. This reuses the existing `internal/settle` read-time overlay (same pattern as EC2 instance/volume and ACM cert settling), is clock-driven with no wall-clock sleeps, and is gated behind the opt-in `WithAsyncSettle()` option — default behavior reports the terminal state immediately.

## Why

Matches real AWS semantics so SDK waiters and health-polling flows behave as they do against the real service.

## Tests

- Provider: `CreateLoadBalancer` with `WithAsyncSettle` reports `provisioning`, then `active` after the clock advances past the settle window (FakeClock, deterministic).
- Server (SDK roundtrip): `DescribeTargetHealth` with a mixed registered + unregistered explicit target returns `initial` for the registered one and `unused` / `Target.NotRegistered` for the unregistered one; with no explicit targets, only the registered target is returned.

Gates: build, vet, scoped tests, gofmt, golangci-lint (0 issues), `go generate` + compatgen zero-diff.

### DIFF
--- a/internal/settle/settle.go
+++ b/internal/settle/settle.go
@@ -27,6 +27,7 @@ const (
 	DefaultDBRebootSettle    = 1 * time.Second // RDS available->rebooting->available
 	DefaultCertificateSettle = 2 * time.Second // ACM PENDING_VALIDATION->ISSUED
 	DefaultExecutionSettle   = 1 * time.Second // SFN RUNNING->SUCCEEDED
+	DefaultLBSettle          = 2 * time.Second // ELBv2 provisioning->active
 )
 
 // Window is a read-time overlay describing a resource still settling into its
--- a/providers/aws/elbv2/elb.go
+++ b/providers/aws/elbv2/elb.go
@@ -11,6 +11,7 @@ import (
 	"github.com/stackshy/cloudemu/v2/errors"
 	"github.com/stackshy/cloudemu/v2/internal/idgen"
 	"github.com/stackshy/cloudemu/v2/internal/memstore"
+	"github.com/stackshy/cloudemu/v2/internal/settle"
 	"github.com/stackshy/cloudemu/v2/services/loadbalancer/driver"
 )
 
@@ -27,6 +28,13 @@ var (
 
 // defaultIdleTimeoutSec is the default idle timeout for load balancers in seconds.
 const defaultIdleTimeoutSec = 60
+
+// Load balancer provisioning states. A new load balancer starts in
+// provisioning and settles to active after a short window.
+const (
+	lbStateProvisioning = "provisioning"
+	lbStateActive       = "active"
+)
 
 // targetKey is the identity of a registered target within a target group.
 // AWS lets the same instance ID or IP be registered multiple times with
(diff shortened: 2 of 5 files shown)


NitinKumar004 (commented):
Reviewed at head a466bd5. Verdict: MERGE.

Gates (pushed SHA): build, vet clean. Scoped tests pass (server/aws/elbv2, providers/aws/elbv2, internal/settle, services/loadbalancer); providers/aws/elbv2 also passes -race. EC2/ACM settle users still pass. gofmt clean; golangci-lint --new-from-rev=origin/development reports 0 new issues. go generate + compatgen produce zero diff. Exactly the 5 expected files changed.

filterHealth (GAP 1) — correct against real ELBv2:
- Exact (ID, Port) request returns that registration's real health; ID-only (Port 0) request returns all registrations of the ID via byID.
- Unregistered requested target yields exactly one synthetic entry with State="unused" / Reason="Target.NotRegistered" — the SDK-typed assertions (TargetHealthStateEnumUnused, TargetHealthReasonEnumNotRegistered) confirm the exact AWS strings match.
- No-explicit-Targets path is unchanged: filterHealth is only invoked when len(wanted) > 0, so all registered health is returned with no synthetic entries.
- Registered targets take the real-health branch with `continue`, so a registered target never also emits a synthetic entry. Empty/duplicate wanted lists cannot panic.

settle overlay (GAP 2) — correct and additive:
- internal/settle change is purely the DefaultLBSettle constant; Window/Observe/Pending unchanged, so EC2/ACM/RDS/SFN users are untouched (verified by their tests).
- Deterministic and clock-driven: Observe is a pure read-time comparison; the transition is driven by the clock (FakeClock advance in tests, wall-clock under serve). No time.Sleep / time.After introduced.
- Window is keyed by LB ARN and deleted in DeleteLoadBalancer alongside the LB — no leak.
- Opt-in: SettleDuration returns 0 unless WithAsyncSettle, so Pending yields an inactive window and default Create/Describe report terminal "active" immediately (TestCreateLoadBalancer still asserts active). With async on, Create reports "provisioning" and Describe transitions to "active" after the clock advances past the window. Stored lb.State stays "active"; the overlay only affects the read, so internal consumers are unaffected.

Minor (non-blocking, pre-existing modeling gap): a target registered by ID only (Port 0) that is then queried with an explicit port now falls through to the synthetic NotRegistered branch, because the mock stores the literal registered port and does not resolve an omitted port to the target-group default. Real ELBv2 resolves it. This requires an inconsistent client (register by ID-only, describe by explicit port); the normal register/describe/waiter flow uses matching Targets and is unaffected. Worth a future port-resolution follow-up, not a blocker for this PR.

PR #757, opened 2026-08-26
fix(route53): paginate ListHostedZonesByVPC (MaxItems/NextToken)

NitinKumar004:
## Summary

`ListHostedZonesByVPC` read `maxitems` but never truncated the result or emitted a `NextToken`, so it always returned every matching private hosted zone in one unbounded, unordered page. Real Route 53 returns at most `MaxItems` `HostedZoneSummary` entries and, when more remain, a `NextToken` that a follow-up call resumes from.

## What changed
- `server/aws/route53/vpc.go`: `listHostedZonesByVPC` now paginates via the package's shared `markerPage` helper (the same one backing `ListHostedZones`/`ListHealthChecks`), keyed by the unique `HostedZoneId` for a deterministic, stable order. Reads the `nexttoken` query param, echoes the clamped `MaxItems`, and returns `NextToken` only when the page is truncated.
- `maxitems` is now parsed through the shared `parseMaxItems` (clamped to [1,100]) for consistency with the other Route 53 list ops.

## Tests
- `TestSDKListHostedZonesByVPCPaginates`: walks 5 zones on one VPC with `MaxItems=2`; asserts each page is bounded, `NextToken` continues, and every zone is returned exactly once (no dup/skip).
- `TestSDKListHostedZonesByVPCStableOrder`: `MaxItems=1` walk asserts strictly increasing `HostedZoneId` order across pages.

## Note on the apex SOA/NS ordering (deferred)
This branch was also scoped to flip the apex record tiebreak to SOA-before-NS. Verified against the AWS Route 53 API reference: `ListResourceRecordSets` is documented to return records "in ASCII order" sorted by name then type, and in ASCII "NS" < "SOA", so NS sorts before SOA — which is exactly what the current code does. The docs do not support the SOA-first claim, so the apex ordering is left unchanged and this gap is deferred pending a real-cloud capture.

## Gates
build / vet / scoped tests / gofmt / golangci-lint (new-from-development) all green; `go generate` and `compatgen` produce zero doc diff.

### DIFF
--- a/server/aws/route53/vpc.go
+++ b/server/aws/route53/vpc.go
@@ -172,17 +172,17 @@ func (h *Handler) listHostedZonesByVPC(w http.ResponseWriter, r *http.Request) {
 		})
 	}
 
-	maxItems := listMaxItems
-
-	if v := q.Get("maxitems"); v != "" {
-		if n, cerr := strconv.Atoi(v); cerr == nil && n > 0 {
-			maxItems = n
-		}
-	}
+	// ListHostedZonesByVPC paginates NextToken-style: maxitems bounds the page and
+	// NextToken (echoed back as nexttoken) resumes at the following zone id. Reuse
+	// the shared marker paginator keyed by the unique HostedZoneId so tokens never
+	// duplicate or skip a summary.
+	maxItems := parseMaxItems(q.Get("maxitems"))
+	page, next := markerPage(summaries, q.Get("nexttoken"), maxItems, func(s hostedZoneSummaryXML) string { return s.HostedZoneId })
 
 	wire.WriteXML(w, http.StatusOK, listHostedZonesByVPCResponse{
 		Xmlns:               xmlns,
-		HostedZoneSummaries: summaries,
+		HostedZoneSummaries: page,
 		MaxItems:            strconv.Itoa(maxItems),
+		NextToken:           next,
 	})
 }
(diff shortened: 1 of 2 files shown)


NitinKumar004 (commented):
Verdict: MERGE.

ListHostedZonesByVPC pagination is correct and lands cleanly.

Correctness:
- Stable, unique ordering: the shared `markerPage` helper sorts by the unique `HostedZoneId` (stable sort) before paging, so the map/`ListZones` iteration order is not the ordering source — no dup/skip trap. This is the same helper, used the same way, as `ListHostedZones`/`ListHealthChecks`.
- No dup/skip across pages: resume is inclusive (next marker = id of the first unreturned summary; next call drops `id < marker`), so a MaxItems=1/2 walk over N zones returns each exactly once. The new tests prove this end-to-end with the real SDK: `assertEachOnce` (exactly-1, no unexpected id) and a strictly-increasing-id assertion.
- NextToken semantics: `NextToken` is `xml:"...,omitempty"` and set only when truncated — absent on the last page; a client echoing it continues correctly. Query keys match what the SDK serializes (`nexttoken`, `maxitems`, `vpcid`, `vpcregion`).
- MaxItems: parsed/clamped to [1,100] via the shared `parseMaxItems` and echoed back.
- Legitimately wire-layer-only: backend `ListZones` still returns all associated zones; pagination lives entirely in `server/aws/route53/vpc.go`. No provider change needed.

Gates on the pushed SHA: build, vet, scoped tests (server + provider route53) all pass; gofmt clean; 0 new golangci-lint issues; `go generate` + compatgen produce zero diff and compat suite is green. Only the 2 expected files changed.

Apex SOA-vs-NS ordering was correctly left out of scope.

PR #758, opened 2026-08-26
fix(kinesis): shard iterators expire after 5 minutes

NitinKumar004:
## What

AWS Kinesis shard iterators returned by `GetShardIterator` are valid for only **5 minutes**. Using an expired iterator in `GetRecords` must return `ExpiredIteratorException`. The emulator previously never expired iterators — the `iteratorToken` carried no timestamp, and `ExpiredIteratorException` was only raised when the shard itself was missing.

## Changes

- Extend the opaque `iteratorToken` with a `CreatedAtMillis` stamp (encoded into the existing base64/JSON token, so expiry needs no extra server state).
- `GetShardIterator` stamps the iterator with the current clock time.
- `GetRecords` rejects an iterator older than 5 minutes with `ExpiredIteratorException` ("Iterator expired. The iterator was created at time ... which is now expired.").
- The `NextShardIterator` returned by `GetRecords` gets a **fresh** 5-minute window, so a client polling in a loop keeps working — each poll refreshes.

## Why

Matches real Kinesis semantics: iterators are short-lived and refreshed by polling. Long-lived reuse of a stale iterator now fails as it does on AWS.

## Notes

- Uses the package's `config.Clock` (`m.now()`), so `FakeClock` tests are deterministic — no wall-clock `time.Now()`.
- The existing missing-shard `ExpiredIteratorException` path is unchanged.

## Tests

New `FakeClock`-driven tests (provider hand-rolled helpers, no testify): iterator valid within 5 min; expires past 5 min; `NextShardIterator` refreshes across many polls; missing-shard path still fires.

### DIFF
--- a/providers/aws/kinesis/kinesis_test.go
+++ b/providers/aws/kinesis/kinesis_test.go
@@ -3,6 +3,7 @@ package kinesis_test
 import (
 	"context"
 	"testing"
+	"time"
 
 	"github.com/stackshy/cloudemu/v2/config"
 	"github.com/stackshy/cloudemu/v2/providers/aws/kinesis"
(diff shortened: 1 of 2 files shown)


NitinKumar004 (commented):
Reviewed at head d265bab. Build, vet, scoped tests (server/providers/services kinesis) and `-race` on providers/aws/kinesis all pass. gofmt clean, 0 new lint issues, `go generate` + compatgen produce zero diff. Only the 2 expected files changed.

TTL check (records.go): `m.now().Sub(createdAt) > shardIteratorTTL` — strict `>`, 5m window, driven by `m.now()` = `opts.Clock.Now().UTC()`, not wall-clock, so FakeClock tests are deterministic. Emits `ExpiredIteratorException` via the `ExExpiredIterator` tag (= "ExpiredIteratorException"), which `writeErr` passes straight through as HTTP 400 — server layer correctly needs no change (opaque iterator passthrough + tagged-exception mapping already in place).

NextShardIterator refresh: `encodeFreshIterator` stamps `CreatedAtMillis = m.now().UnixMilli()` on every GetRecords return, so a client polling under 5m keeps a live iterator indefinitely. The 5-poll test genuinely proves this — it advances FakeClock by ~5m between polls and reuses `out.NextShardIterator`; a non-refreshing impl would fail by the 3rd poll (accumulated drift > 5m).

Token round-trip: `CreatedAtMillis` is field "c" in the base64/JSON token, encoded and decoded symmetrically; both GetShardIterator and GetRecords mint via the same helper. Malformed base64/JSON returns InvalidArgumentException (no panic). A legacy token lacking the field decodes to CreatedAtMillis=0 → treated as created at epoch → expired rather than panicking, which is graceful and consistent with iterator-expiry semantics. No overflow risk (epoch-to-now is well within int64 ns range).

Missing-shard path unchanged: the pre-existing `findShardByID == nil → expiredIterator("shard ... no longer exists")` branch still fires after the TTL gate; the new test recreates the stream so the shard id is gone and asserts ExpiredIteratorException.

Verdict: MERGE. Correct, deterministic, and the polling-loop refresh is real.

PR #759, opened 2026-08-26
fix(chaos): bind scenario windows to the engine clock at Apply so FakeClock tests activate

NitinKumar004:
## Summary

Fixes a HIGH-severity determinism bug: time-bounded chaos scenarios never activated under a `config.FakeClock`, silently breaking the documented deterministic-chaos-testing path.

## The bug

Scenario constructors (`ServiceOutage`, `LatencySpike`, `Throttle`, `ProbabilisticFailure`, and `Composite` children) stamped their active-window `start` from raw wall-clock `time.Now()` at **construction** time. But the engine evaluates `window.active(now)` using its **injected clock** (`e.clock.Now()`).

When a caller passes a `FakeClock` pinned to, e.g., 2024-01-01 (as every FakeClock test does), `fakeNow(2024)` is `Before` the wall-clock `start(2026)`, so `window.active()` is always false and no time-bounded scenario ever fires. The `chaos.New` doc comment explicitly advertises "Pass a config.FakeClock for deterministic tests" — a path that was broken and uncaught (no chaos test used a FakeClock).

## The fix

The window `start`/`end` are now bound to the **engine's clock at Apply time** rather than baked from wall-clock at construction:

- `window` now carries a `duration` and binds its absolute `start`/`end` via a new idempotent `bind(now)` method.
- `Engine.Apply` calls `bind(e.clock.Now())` on any scenario implementing the unexported `clockBinder` interface, so the window starts "now" on the engine's clock.
- `Composite` propagates `bind` to its children.

Real-clock behavior is unchanged (constructing then immediately applying binds to a near-identical time). Auto-recovery/expiry, the first-error-wins/latency-sum/composite merge logic, and event recording are all preserved.

## Tests

- Added `TestServiceOutageFakeClockActivatesAndRecovers`: a 5-minute outage under a FakeClock activates at fake now, stays active while the clock advances inside the window, and auto-recovers after it advances past the window. This test **fails against the old code** and passes with the fix.
- Added `TestCompositeFakeClockBindsChildren` covering composite child binding under a FakeClock.
- All existing real-clock tests remain green (behavior unchanged).

### DIFF
--- a/features/chaos/chaos.go
+++ b/features/chaos/chaos.go
@@ -93,9 +93,21 @@ func New(clock config.Clock) *Engine {
 	}
 }
 
+// clockBinder is implemented by scenarios whose active window is anchored to
+// the engine clock. Apply binds them at registration time so that time-bounded
+// scenarios start "now" on the engine's clock — including a FakeClock — rather
+// than on wall-clock at construction.
+type clockBinder interface {
+	bind(now time.Time)
+}
+
 // Apply registers a scenario. The returned Active can be used to stop it
 // before it expires.
 func (e *Engine) Apply(s Scenario) *Active {
+	if b, ok := s.(clockBinder); ok {
+		b.bind(e.clock.Now())
+	}
+
 	e.mu.Lock()
 	defer e.mu.Unlock()
 
--- a/features/chaos/scenarios.go
+++ b/features/chaos/scenarios.go
@@ -8,17 +8,40 @@ import (
 )
 
 // window holds the common "active from start to start+duration" timing logic.
-// All scenarios with a duration embed it.
+// All scenarios with a duration embed it. The start (and thus end) is bound to
+// the engine's clock when the scenario is applied — see bind — so the window is
+// relative to the engine clock, not wall-clock at construction. This keeps
+// FakeClock-driven deterministic tests working. Until bound, the window is
+// inactive.
 type window struct {
-	start time.Time
-	end   time.Time
+	duration time.Duration
+	start    time.Time
+	end      time.Time
+	bound    bool
 }
 
-func newWindow(start time.Time, duration time.Duration) window {
-	return window{start: start, end: start.Add(duration)}
+func newWindow(duration time.Duration) window {
+	return window{duration: duration}
+}
+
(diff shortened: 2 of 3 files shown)


NitinKumar004 (commented):
Reviewed at head f6da59f. Gates green on the pushed SHA: build, vet, `go test -p=3 ./features/chaos/... ./features/inject/...`, and `go test -race ./features/chaos/...` all pass; gofmt clean; golangci-lint `--new-from-rev` reports 0 new issues; `go generate ./...` + compatgen produce zero diff. Only the 3 expected files changed.

Verified the crux — the "inactive until bound" risk cannot silently disable chaos:

- Apply is the ONLY path a scenario enters `e.active` (`e.active[id] = s` at chaos.go:116 is the sole write besides map re-init in New/Stop), and Apply binds via the `clockBinder` assertion immediately before insertion. There is no direct append, constructor pre-registration, or deserialization path that adds a scenario unbound. Check only iterates `e.active`, so no windowed scenario is ever evaluated unbound.
- Every built-in windowed scenario (`outage`, `latencySpike`, `probabilisticFailure`, `throttle`) embeds `window` and is returned as a pointer, so each satisfies `clockBinder` and gets bound at Apply. `composite` implements its own `bind` that recurses into children (nested composites included). A hypothetical external Scenario without a window degrades gracefully — the type assertion fails, no window gating applies, and its own Active() drives it.
- bind is idempotent (`if w.bound { return }`), so re-Apply does not shift the window.
- Real-clock behavior is unchanged in practice: start moves from construction-time to Apply-time, but for RealClock both are `time.Now()` and Apply follows construction immediately; the shift is the more correct semantics and no existing real-clock test/contract depended on construction-time start.
- Auto-recovery/expiry preserved: `window.active` still returns false past end and Check drops expired scenarios lazily; first-error-wins and latency-sum in Check are untouched.

Confirmed the new tests are non-vacuous: reverting just the source (keeping the new tests) makes both `TestServiceOutageFakeClockActivatesAndRecovers` and `TestCompositeFakeClockBindsChildren` fail ("expected outage to be active at fake now" / composite child latency 0), and they pass with the fix.

No blocking issues. LGTM.

PR #760, opened 2026-08-26
fix(azure-vm): case-insensitive resource-group and resource-name lookup (VM + scale sets)

NitinKumar004:
## What

ARM resource-group names and resource names are **case-insensitive**, but Azure Virtual Machines GET/LIST and scale-set GET compared them with an exact `!=`/`==`. A user who created a VM under `rg-1` and then `GET .../resourceGroups/RG-1/...` got a spurious 404. The cascade-delete path already used `strings.EqualFold`, so the codebase already intended case-insensitivity — GET/LIST just missed it.

## Changes

- `server/azure/virtualmachines/instances.go` — `findByName` and `list` now compare resource group **and** VM name with `strings.EqualFold`.
- `server/azure/virtualmachines/scalesets.go` — `getScaleSet` compares the scale-set name with `strings.EqualFold`.
- `server/azure/echo_properties.go` — the shared property overlay (which preserves unmodeled request properties like `additionalCapabilities`) is keyed by the ARM resource id, which reflects the **request-path** RG casing. Without normalization, a differently-cased GET would resolve the VM but silently drop its unmodeled props. `capture`/`lookup`/`evict` now lowercase the **resource-group segment** of the key only — the response body id is never altered, and the resource-name segment is untouched, so no cross-service collision. This is safe for every service that shares the overlay.

## Why it's correct

Mirrors ARM semantics and the existing `purgeInstances` `EqualFold` intent. The overlay normalization touches only the internal map key, not any response.

## Tests

New `resourcegroup_case_insensitive_test.go` (real `azure-sdk-for-go` armcompute client over TLS httptest):
- GET a VM by an upper-cased RG returns 200 + the VM, **including** the unmodeled `additionalCapabilities` overlay prop.
- GET a VM by an upper-cased name resolves.
- LIST scoped by an upper-cased RG returns the VM.
- Scale-set GET by differently-cased name and RG resolves.

Verified **fail-before / pass-after** (both the findByName and the overlay-normalization halves are independently load-bearing).

## Scope

Focused on Virtual Machines only. The same case-sensitive RG/name anti-pattern exists in several other Azure services (notification hubs, snapshots, images, public IPs, NAT gateways, disks, SSH public keys, load balancer, managed Cassandra, databricks, AKS, ACR, search, AI/ML, functions, cosmos-PG) — reported separately for a batched systemic follow-up, not fixed here.

### DIFF
--- a/server/azure/echo_properties.go
+++ b/server/azure/echo_properties.go
@@ -33,6 +33,8 @@ func newPropertyOverlay() *propertyOverlay {
 // entry. An empty set clears the entry so a resource that no longer carries
 // unmodeled properties (e.g. re-created without them) does not keep stale ones.
 func (o *propertyOverlay) capture(id string, unmodeled map[string]any) {
+	id = normalizeOverlayKey(id)
+
 	o.mu.Lock()
 	defer o.mu.Unlock()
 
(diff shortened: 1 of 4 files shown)


NitinKumar004 (commented):
**Verdict: MERGE** — clean on pushed SHA `c8bd137`.

**Overlay key normalization (`echo_properties.go`) — the server-wide change**
- `normalizeOverlayKey` lowercases only the value between the `/resourceGroups/` marker and the next `/` (or end-of-string when the RG is the final segment). The resource-name segment and everything else are left byte-for-byte, and it returns the id unchanged when the marker is absent. So distinct resources never collide and no case-sensitive segment (blob name, sub-resource) is touched.
- It is applied identically on all three paths — `capture`, `lookup`, and `evict` — so the internal map key is consistent regardless of request casing. Symmetric by construction.
- The marker keyword casing is consistent across paths: capture/lookup key off the handler-generated response-body `id` (built via idgen, capital-G `/resourceGroups/`), and evict keys off `resourceIDFromPath` → `BuildResourceID` (also capital-G). Both hit the marker, both normalize the RG value the same way — no capture-vs-evict mismatch, so no overlay leak on delete/recreate.
- Response body `id` is never mutated — only the internal map KEY is normalized. Correct.

**VM/scale-set EqualFold changes**
- `findByName` (RG + VM name), `list` (RG filter), and `getScaleSet` (name) now use `strings.EqualFold`. Correct for ARM: resource-group and resource names are case-insensitively unique, so there is no risk of matching two distinct resources. No logic downstream assumes case-sensitive distinctness.

**Test**
- `resourcegroup_case_insensitive_test.go` drives the real armcompute v5 SDK: create under `rg-1`, GET with `RG-1` (asserts 200 + the unmodeled `additionalCapabilities.hibernationEnabled` overlay prop survives the differently-cased read), GET with `CASE-VM`, LIST scoped by `RG-1`, and scale-set GET by `VMSS-1`. Non-vacuous — pre-fix the `!=`/`==` comparisons 404 before any assertion is reached.

**Gates:** build + vet clean; `go test` VM (server+provider) + `-race` (provider) green; broad `./server/azure/...` sweep green (queue, resourcegroups, servicebus, storageaccount, vnet, sql, cosmosdb, aks, keyvault, functions, eventgrid, tablestorage, subscriptions, tenants, monitor, dns, iam, etc. — all pass; the earlier "no space left on device" build-fails were a local disk issue, re-run green after clearing cache); gofmt clean; `golangci-lint --new-from-rev` 0 new issues; `go generate` + `compatgen` zero diff; only the 4 expected files changed.

PR #761, opened 2026-08-26
fix(azure-blob): Range/206, content-property round-trip, committed-block refs, delimiter pagination, time conditionals

NitinKumar004:
Fixes five real Azure Blob Storage residue bugs found by a deep confirming audit. All in `server/azure/blobstorage/` + `providers/azure/blobstorage/` (+ driver types in `services/storage/driver/`), batched into one PR since they touch the same packages.

## Bugs fixed

1. **[HIGH data-corruption] Get Blob ignored Range / x-ms-range.** `getBlob` always wrote the full body with 200, so the azblob SDK's parallel ranged `DownloadFile`/`DownloadBuffer` (each chunk a ranged GET) received the whole blob per chunk and assembled a corrupt file. Now parses `x-ms-range` (preferred) then `Range` (`bytes=start-end`, `bytes=start-`, `bytes=-N`), returns `206 Partial Content` with `Content-Range`/partial `Content-Length` and only the requested slice, sets `Accept-Ranges: bytes`, and returns `416` with `Content-Range: bytes */total` for an unsatisfiable range. No-range GET stays a full-body 200.

2. **[MED] Content properties never round-tripped.** Put Blob dropped `Content-Encoding`/`Cache-Control`/`Content-Language`/`Content-Disposition`; `objectInfo()` omitted them and `writeBlobHeaders` never emitted them. Put Blob now persists all four (via a new `PutBlockBlob` extension), `objectInfo()` carries them, and GET/HEAD/Get Blob Properties emit them. Terraform `azurerm_storage_blob` cache_control/content_encoding no longer drifts.

3. **[MED] Commit Block List couldn't reference committed blocks.** Committed block bytes were discarded and assembly looked only in staging, and the parser collapsed `<Committed>`/`<Uncommitted>`/`<Latest>` into one list. Now committed block bytes are retained, the per-element source list is preserved through parsing, and each id resolves against the correct source — so the "append by re-committing existing blocks plus a new one" pattern works.

4. **[MED pagination] List Blobs delimiter returned all BlobPrefixes on every page.** Prefixes were computed over all keys and never counted toward `maxresults`. Blobs and prefixes now fold into one sorted, marker-paginated stream (both count toward `maxresults`, stable-sorted, no dup/skip across pages) with `NextMarker` when truncated.

5. **[LOW-MED] If-Modified-Since / If-Unmodified-Since ignored.** Only ETag conditionals were handled. Reads now honor `If-Modified-Since` (304) and `If-Unmodified-Since` (412); writes honor `If-Unmodified-Since`/`If-Modified-Since` (412) against the blob's LastModified. ETag conditions keep precedence.

## Tests
Provider-level (committed-block re-commit, content-property round-trip) and wire-level with the real azure-sdk-for-go azblob client: ranged GET 206 + correct bytes + Content-Range, open-ended range, 416, a multi-chunk `DownloadBuffer` that asserts the assembled buffer equals the original (the data-corruption guard), content-property round-trip via GetProperties + read headers, commit-block-list re-commit referencing a committed block, delimiter+maxresults pagination without dup/skip, and If-Modified-Since 304 / If-Unmodified-Since 412.

## Gates
`go build ./...`, `go vet`, scoped `go test` (server+providers+services/storage), gofmt, `golangci-lint --new-from-rev=origin/development` (0 issues), `go generate` + `compatgen` regenerated and idempotent.

### DIFF
--- a/docs/coverage/aws/s3.md
+++ b/docs/coverage/aws/s3.md
@@ -75,6 +75,7 @@ AzureBlobExtensions is an OPTIONAL Azure-specific blob data-plane capability,
 | `DeleteBlobSnapshots` | DeleteBlobSnapshots applies the Azure delete-snapshots directive |
 | `GetBlobSnapshot` | GetBlobSnapshot reads a previously captured snapshot (GET ?snapshot=…). |
 | `GetBlockList` | GetBlockList returns the blob's committed and uncommitted blocks (Get |
+| `PutBlockBlob` | PutBlockBlob writes a block blob's content together with its system content |
 | `ReleaseLease` | ReleaseLease releases the blob's current lease. |
 | `RenewLease` | RenewLease renews the blob's current lease. |
 | `SetBlobMetadata` | SetBlobMetadata replaces only a blob's metadata (Set Blob Metadata, |
--- a/docs/coverage/azure/blobstorage.md
+++ b/docs/coverage/azure/blobstorage.md
@@ -75,6 +75,7 @@ AzureBlobExtensions is an OPTIONAL Azure-specific blob data-plane capability,
 | `DeleteBlobSnapshots` | DeleteBlobSnapshots applies the Azure delete-snapshots directive |
 | `GetBlobSnapshot` | GetBlobSnapshot reads a previously captured snapshot (GET ?snapshot=…). |
 | `GetBlockList` | GetBlockList returns the blob's committed and uncommitted blocks (Get |
+| `PutBlockBlob` | PutBlockBlob writes a block blob's content together with its system content |
 | `ReleaseLease` | ReleaseLease releases the blob's current lease. |
 | `RenewLease` | RenewLease renews the blob's current lease. |
 | `SetBlobMetadata` | SetBlobMetadata replaces only a blob's metadata (Set Blob Metadata, |
--- a/docs/coverage/coverage.json
+++ b/docs/coverage/coverage.json
@@ -10688,6 +10688,10 @@
             "doc": "GetBlockList returns the blob's committed and uncommitted blocks (Get"
           },
           {
+            "name": "PutBlockBlob",
+            "doc": "PutBlockBlob writes a block blob's content together with its system content"
+          },
+          {
             "name": "ReleaseLease",
             "doc": "ReleaseLease releases the blob's current lease."
           },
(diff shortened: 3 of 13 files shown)


NitinKumar004 (commented):
Reviewed PR #761 (Azure Blob: Range/206, content-property round-trip, committed-block refs in Put Block List, delimiter pagination, time conditionals) at head 33570fb2.

Gates (all green): build, vet, scoped tests, `-race` on providers/azure/blobstorage, gofmt, golangci-lint (0 new). `go generate ./...` and `go run ./internal/compatgen` both idempotent — zero further diff.

Driver-interface change: `AzureBlobExtensions.CommitBlockList` signature changed (now takes `[]BlockListEntry` + `*BlobProperties`) and `PutBlockBlob` added. Sole implementer is `providers/azure/blobstorage.Mock` (compile-time-asserted); only caller is the Azure blob wire handler. Both updated; no other implementer/mock. Build confirms.

BUG 1 Range — verified airtight. `DownloadBuffer` 1024-byte parallel 64-byte ranged assembly passes. Adversarially probed parseByteRange/serveBlobContent across 12 cases, all correct:
- bytes=0-0 -> 206, Content-Range bytes 0-0/1024, CL=1, exact byte
- bytes=0- -> 206 full, bytes 0-1023/1024
- bytes=-100 suffix -> 206 last 100, bytes 924-1023/1024; suffix > total clamps to whole blob
- bytes=1023-2000 -> 206 clamped to 1023-1023/1024
- bytes=2000-3000 fully OOR -> 416 + Content-Range bytes */1024
- bytes=abc / bytes=5-3 / bytes=-0 / bytes=1024- / multi-range -> 416 + bytes */1024
- no range -> 200 full body
Content-Length/Content-Range/Accept-Ranges exact in every case. x-ms-range suffix is accepted as a lenient superset of Azure (Azure rejects suffix on x-ms-range) — harmless, not corruption.

BUG 3 block-list — verified. committedBlockData retained on the object; resolveBlock honors Committed/Uncommitted/Latest per element. Adversarial: commit [A,B] then explicit <Committed>A + <Uncommitted>C -> AAACCC (not AAABBB, not AAABBBCCC); referencing a committed block as Uncommitted correctly 4xx-fails. SDK Latest append (re-commit committed A + staged B) also passes.

BUG 4 pagination — verified. Blobs+prefixes fold into one stream sorted from a deterministic sorted-keys source (no map-iteration ordering); PaginateSorted with unique name + blob-before-prefix tiebreak. Walking maxresults=1 over a mixed set returns each prefix and blob exactly once, no dup/skip; NextMarker only when truncated. Rewritten test asserts real Azure behavior.

BUG 2 content props — verified. Content-Encoding/Cache-Control/Content-Language/Content-Disposition persist on Put Blob and Commit Block List and emit on GET/HEAD/GetProperties; objectInfo() carries all four.

BUG 5 time conditionals — verified. If-Modified-Since -> 304, If-Unmodified-Since -> 412 on read/write. On writes ETag conditions evaluated before time conditions (correct precedence); existing If-Match/If-None-Match unregressed.

Refactor safety — PutObject and PutBlockBlob share putBlockBlobInternal; plain PutObject passes props=nil (unchanged). services/storage/driver.go is shared but the interface change touches only the Azure-specific optional capability; server/aws/s3 and server/gcp/gcs tests pass.

Verdict: MERGE. No blocking issues.

PR #762, opened 2026-08-26
fix(azure): case-insensitive resource-group/name lookup across services

NitinKumar004:
## What

Fixes a systemic bug where Azure GET/LIST paths returned a spurious 404 when a
user supplied a resource-group (or a case-insensitive resource name) with
different casing than what was stored. Azure ARM resource-group names — and
almost all resource names — are **case-insensitive**, but ~20 get/list sites
used exact `==`/`!=` string compares.

This replaces those exact compares with `strings.EqualFold` across the affected
services. Only the **comparison** changes — stored casing and response bodies
are untouched. The shared overlay-key half of this bug was already fixed
globally in #760 (`echo_properties.go`); this PR closes the remaining
per-service exact-compare sites.

## Why

Real ARM treats `rg-1` and `RG-1` as the same resource group; a CLI/SDK/IaC
tool that lists or gets with different casing than a create used must still
resolve the resource. The exact compares caused false NotFound results.

## Services touched

Resource-group compares (server handlers): notificationhubs, disks, images,
snapshots, sshpublickeys, vnet (public IP + NAT gateway).

Resource-group + case-insensitive name compares (provider backends): aks
(cluster/pool/maintenance list + cluster & node-pool names), acr, search, ai
(+ machine learning), functions (site metadata + app service plan), databricks
(workspaces + access connectors), loadbalancer (list + frontend IP config
name), managedcassandra, cosmospostgresql.

## Not touched (intentionally)

Azure **Blob names are case-sensitive** — the blob-name compares in
`storageaccount/handler.go` are left as-is.

## Tests

Added representative fail-before/pass-after tests across structurally-distinct
layers:
- `aks`: `ListClustersByResourceGroup` with differently-cased RG
- `functions`: `GetSiteMeta` with differently-cased RG
- `disks` (server handler, tag-based): SDK list-by-resource-group with
  differently-cased RG
- `acr` (provider backend): `ListRegistries` with differently-cased RG

Remaining sites rely on the mechanical, identical transformation plus existing
tests passing.

### DIFF
--- a/providers/azure/acr/registry.go
+++ b/providers/azure/acr/registry.go
@@ -241,7 +241,7 @@ func (m *Mock) ListRegistries(_ context.Context, rg string) ([]driver.AzureRegis
 	out := make([]driver.AzureRegistry, 0, len(all))
 
 	for _, rd := range all {
-		if rg != "" && rd.reg.ResourceGroup != rg {
+		if rg != "" && !strings.EqualFold(rd.reg.ResourceGroup, rg) {
 			continue
 		}
 
--- a/providers/azure/acr/registry_test.go
+++ b/providers/azure/acr/registry_test.go
@@ -36,6 +36,21 @@ func TestCreateOrUpdateRegistry(t *testing.T) {
 	assert.Equal(t, "Standard", updated.SKUName) // default when SKU omitted
 }
 
+func TestListRegistriesResourceGroupCaseInsensitive(t *testing.T) {
+	m, _ := newTestMock()
+	ctx := context.Background()
+
+	_, _, err := m.CreateOrUpdateRegistry(ctx, "rg-1", "MyReg", driver.AzureRegistryConfig{Location: "eastus"})
+	require.NoError(t, err)
+
+	// ARM resource-group names are case-insensitive: a list against "RG-1"
+	// must still return the registry created in "rg-1".
+	got, err := m.ListRegistries(ctx, "RG-1")
+	require.NoError(t, err)
+	require.Len(t, got, 1)
+	assert.Equal(t, "rg-1", got[0].ResourceGroup)
+}
+
 func TestUpdateRegistryPartialMerge(t *testing.T) {
 	m, _ := newTestMock()
 	ctx := context.Background()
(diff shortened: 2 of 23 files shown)


NitinKumar004 (commented):
Verdict: approve.

Systemic case-insensitive RG/name comparison change reviewed against the pushed head SHA. Clean build + vet, gofmt clean on all changed files, `go generate ./...` and compatgen produce zero diff, and tests for every touched package pass.

Semantic safety (the real risk of a broad `EqualFold` sweep):
- All ~23 changed sites are READ-path lookups/filters, not create-time uniqueness guards or map-key equality. Verified each:
  - LIST-by-resource-group filters: acr, ai (accounts + ML workspaces), aks (clusters/agentpools/maintenance), cosmospostgresql, managedcassandra, search, databricks (workspaces + access connectors), functions (app service plans + site meta), loadbalancer, and the server-side tag-based collection filters in disks/images/snapshots/sshpublickeys/vnet publicIPs + NAT gateways.
  - GET/DELETE/UPDATE guards keyed by name where the RG is a validation predicate (not the map key): functions site_meta.go (Get/Delete/UpdateAppSettings/GetFunctionScoped/DeleteFunctionScoped) and notificationhubs wrongResourceGroup. The backing map key (`name`) is untouched, so relaxing the RG guard to EqualFold is correct under ARM's case-insensitive-unique guarantee.
  - loadbalancer hasFrontend: a reference-resolution lookup (a NAT rule resolving its frontend IP config by name), not a uniqueness check — correct to EqualFold. The LB map key `azureLBKey(rg, name)` is left exact (unchanged, out of scope).
- Comparison-only: no stored casing, response body, or map key is altered. Response id/name casing is byte-identical.
- Blob-name compares in storageaccount/handler.go (case-SENSITIVE in Azure) are correctly NOT in the diff.
- No site relied on case to distinguish two legitimately-different entries (impossible under ARM).

Added tests (acr, aks, functions, disks) are non-vacuous: each creates in `rg-1`/`rgA` and lists/gets against `RG-1`/`RGA`, asserting the resource is returned AND that the stored RG casing is unchanged — real fail-before/pass-after coverage.

`strings` import added where needed (databricks.go, azure_load_balancer.go); build confirms no unused/missing imports. golangci-lint over the touched packages reports 0 new issues (the 3 gocritic hits are in the unchanged lb.go).

PR #763, opened 2026-08-26
fix(azure-cosmos): functional ETag/If-Match concurrency, partition-scoped queries, FakeClock TTL, write-time _ts, delete-missing 404

NitinKumar004:
## Summary

Fixes five real Azure Cosmos DB bugs in `server/azure/cosmosdb/`, all confirmed against the real `azcosmos` SDK.

### Bugs fixed

1. **Optimistic concurrency was completely inert (HIGH, lost-update).** No `ETag` response header was emitted, the body `_etag` was a static function of the id, `If-Match` was never parsed, and `writeErr` had no 412 mapping — so `ItemOptions{IfMatchEtag}` never triggered a 412 and `ItemResponse.ETag` was always empty. Now every write stamps a **rotating** `_etag` at write time (persisted with the item), create/read/replace emit the `ETag` header, the replace (PUT) and delete paths parse `If-Match` (`*` matches any existing), and a stale/mismatched precondition returns `FailedPrecondition` → **412 PreconditionFailed**.

2. **Partition-scoped queries leaked across partitions (HIGH, wrong-result).** `queryDocuments` scanned the whole container and relied only on the WHERE clause, so a scoped `SELECT * FROM c` over a multi-partition container returned every partition's rows. Now a query pinned to a partition (via `x-ms-documentdb-partitionkey`) is filtered to that partition value before evaluation; a query with no partition key still fans out. Note: the azcosmos SDK sets `x-ms-documentdb-query-enablecrosspartition: true` on **every** query by default, so the *presence of the partition-key header* is the correct discriminator (a specific partition key always wins over the cross-partition permission), not that flag.

3. **Container TTL used the wall clock (MED, determinism).** `attrsStore` computed and evaluated expiry with raw `time.Now()`, so a `FakeClock` couldn't drive it. The clock is now threaded from the driver (new optional `Clock()` capability on the Azure cosmos `Mock`, mirroring how `AccountTables`/`PurgeAccount` are exposed) into `attrsStore`; both call sites use `clock.Now()`, defaulting to a `RealClock` when the driver doesn't expose one.

4. **`_ts` reflected read-time, not write-time (MINOR).** `addSystemProps` stamped `_ts = time.Now()` at response time. `_ts` is now persisted at write time from the injected clock (deterministic) and returned as-stored on read.

5. **DELETE of a missing document returned 204 (MINOR, error-code).** `deleteDocument` now existence-checks the `(id, partition-key)` under the write lock and returns **404 NotFound** for a missing document, matching real Cosmos.

### Existing tests corrected from wrong → correct behavior

These encoded the bugs above and were updated to assert real-Cosmos semantics:

- `cosmos_lifecycle_test.go` `TestCosmosWriteDivergences` — the "stale `IfMatchEtag` succeeds" block now asserts **412** and that the doc is unchanged; the "delete-missing → 204" block now asserts **404**. File-header and test doc comments updated to drop those two divergences (only the PATCH-unrouted divergence remains).
- `cosmos_lifecycle_test.go` `TestCosmosQueryAndPagination` and `TestCosmosQueryFidelity` — both stored documents in per-id partitions but issued a **partition-scoped** query while expecting cross-partition WHERE results (explicitly commented "advisory in the emulator"). With scoping now enforced these must be **cross-partition** queries (`azcosmos.NewPartitionKey()`); switched and re-commented. Their purpose (WHERE / ORDER BY / projection / aggregate fidelity) is unchanged.
- `ttl_test.go` — the three container-`DefaultTimeToLive` tests replaced `time.Sleep` with a `FakeClock` `Advance`, now that the container-TTL path is clock-driven.

### New tests

- `optimistic_concurrency_test.go` — ETag rotates across writes and is returned in the response header/`ItemResponse.ETag`; a matching `If-Match` succeeds, a stale one is 412 on both replace and delete, and `*` matches any existing.
- `partition_query_test.go` — a partition-scoped `SELECT * FROM c` over a multi-partition container returns only that partition (including correct continuation paging), while a cross-partition query returns all.

### Gates

`go build ./...`, scoped `go vet`, `go test -p=3 ./server/azure/cosmosdb/... ./services/database/... ./providers/azure/cosmosdb/...` (plus `-race` on the cosmos package and the adjacent cosmosaccount/compat tests), gofmt, `golangci-lint --new-from-rev=origin/development` (0 issues), `go generate ./...` and `go run ./internal/compatgen` — all green, zero unexpected diff.

### DIFF
--- a/providers/azure/cosmosdb/cosmosdb.go
+++ b/providers/azure/cosmosdb/cosmosdb.go
@@ -131,6 +131,13 @@ func (m *Mock) SetMonitoring(mon mondriver.Monitoring) {
 	m.monitoring = mon
 }
 
+// Clock exposes the mock's injected clock so the Cosmos wire handler can drive
+// its container-TTL expiry and write-time timestamps from the same clock,
+// letting a FakeClock set on the provider make those deterministic in tests.
+func (m *Mock) Clock() config.Clock {
+	return m.opts.Clock
+}
+
 func (m *Mock) emitMetric(container string, metrics map[string]float64) {
 	if m.monitoring == nil {
 		return
--- a/server/azure/cosmosdb/container_attrs.go
+++ b/server/azure/cosmosdb/container_attrs.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/stackshy/cloudemu/v2/config"
 	dbdriver "github.com/stackshy/cloudemu/v2/services/database/driver"
 )
 
@@ -37,10 +38,14 @@ type attrsStore struct {
 	// expiry maps an item identity (table|partitionValue|id) to its computed
 	// expiry time. An item absent here never expires.
 	expiry map[string]time.Time
+	// clock computes expiry at write time and evaluates it at read time, so a
+	// FakeClock injected into the provider drives container-TTL deterministically
+	// rather than the wall clock.
+	clock config.Clock
 }
 
-func newAttrsStore() *attrsStore {
-	return &attrsStore{attrs: make(map[string]containerAttrs), expiry: make(map[string]time.Time)}
+func newAttrsStore(clock config.Clock) *attrsStore {
+	return &attrsStore{attrs: make(map[string]containerAttrs), expiry: make(map[string]time.Time), clock: clock}
 }
 
 func (s *attrsStore) set(table string, ttl *int32, uk *uniqueKeyPolicy, indexing map[string]any) {
@@ -120,7 +125,7 @@ func (s *attrsStore) recordWrite(table string, cfg *dbdriver.TableConfig, item m
 		return
 	}
 
-	s.expiry[key] = time.Now().Add(time.Duration(effective) * time.Second)
+	s.expiry[key] = s.clock.Now().Add(time.Duration(effective) * time.Second)
 }
 
 // expired reports whether the item at (table, id, partition value) has
@@ -133,7 +138,7 @@ func (s *attrsStore) expired(table string, cfg *dbdriver.TableConfig, item map[s
 	exp, ok := s.expiry[key]
 	s.mu.RUnlock()
 
-	return ok && time.Now().After(exp)
+	return ok && s.clock.Now().After(exp)
 }
 
 // forget removes an item's TTL bookkeeping, called on delete.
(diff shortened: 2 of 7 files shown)


NitinKumar004 (commented):
Verdict: LGTM — merge.

Gates (clean-built the pushed SHA c2e508a): build, vet, `go test` on server/azure/cosmosdb + providers/azure/cosmosdb + services/database all pass; `-race` on the cosmos server package passes; gofmt clean; golangci-lint `--new-from-rev` reports 0 new issues; `go generate ./...` and compatgen both produce zero diff. Only the 7 expected files changed.

BUG 1 — optimistic concurrency (data-integrity): correct.
- `_etag` rotates on every write via an atomic counter, so it's distinct per mutation and does NOT collide on identical content (a re-write of the same body still yields a new etag). Emitted as the `ETag` response header on create, read, and replace, which is what azcosmos surfaces as ItemResponse.ETag.
- If-Match is parsed on both PUT and DELETE; empty = no precondition, `*` = match any, otherwise must equal the stored `_etag` else 412 (FailedPrecondition -> PreconditionFailed added to writeErr).
- The GetItem(existing) -> checkIfMatch -> stampWrite -> PutItem sequence runs entirely under the per-collection write lock, so there is no TOCTOU window. Verified adversarially: read E1, an intervening write makes E2, then a stale replace carrying E1 is rejected 412 with the doc left at the newer revision; and 12 goroutines racing a replace with the same base etag yield exactly one winner under `-race`.

BUG 2 — partition scoping: correct and real-Cosmos-faithful. Scoping keys off the PRESENCE of `x-ms-documentdb-partitionkey`, not the cross-partition flag — right call, since supplying a partition key scopes the query to that partition regardless of EnableCrossPartitionQuery, and the SDK sets the cross-partition flag on every query so it's useless as a discriminator. PK header present -> filter to that PK value before WHERE; absent -> whole container. Verified: 5-doc/2-partition container, scoped SELECT returns only the 3 team-a docs (no team-b leak), unscoped returns all 5, and paged scoping (page size 2) returns the same set with no dup/skip. The two changed query tests legitimately switch to a cross-partition key: their docs live in per-id partitions and the intent was always cross-partition WHERE — they only passed before because scoping was ignored.

BUG 3/4/5: clock is threaded through an optional `clockProvider` interface (Mock.Clock()), defaulting to RealClock so the non-injected path keeps real-time TTL; the FakeClock now drives container-TTL expiry and write-time `_ts`. TTL tests' Sleep->Advance and the divergence-test corrections (stale If-Match now 412, delete-missing now 404) are legitimate bug->correct changes, not regressions hidden. Point-read PK enforcement, create/upsert 409, unique-key, offers and isolation still pass.

No blocking issues.

PR #764, opened 2026-08-26
fix(azure-functions): PATCH site update, derive App Service Plan SKU tier, stable function keys, start/stop state

NitinKumar004:
## Summary

Fixes four Azure Functions / App Service parity bugs found by a confirming real-SDK audit, all in `server/azure/functions/` + `providers/azure/functions/`.

### Bugs fixed

1. **PATCH on `Microsoft.Web/sites` returned 405.** `serveResource` only switched PUT/GET/DELETE, so `WebApps_Update` (PATCH — `az functionapp update`, `az functionapp identity assign`, SDK `.Update()`) fell through to 405. Added a PATCH arm with true partial-update semantics: only the fields the body carries are applied, everything else is left as stored (a new `PatchSiteMeta` provider method + partial `FunctionConfig` that rides `applyConfigUpdates`' zero-field preservation). Identity assign/remove via PATCH works; unmodeled (overlay) properties are preserved via the existing PATCH-union path; a PATCH against a missing site (or wrong resource group) is 404, not a silent create.

2. **App Service Plan SKU tier hardcoded `"Dynamic"`.** When the tier was omitted, every plan was tagged `Dynamic` regardless of SKU name. Now derived server-side from the SKU-name prefix the way real Azure does: `Y`→Dynamic, `EP`→ElasticPremium, `WS`→WorkflowStandard, `B`→Basic, `S`→Standard, `P#v2`→PremiumV2, `P#v3`→PremiumV3, `P`→Premium, `F`→Free, `D`→Shared, `I(#v2)`→IsolatedV2/Isolated; unknown falls back to Standard (never forced to Dynamic). An explicitly-provided tier is preserved.

3. **Function PUT-as-update rotated keys and always returned 201.** A re-PUT of an existing function regenerated its `default` key and always answered 201. Now an overwrite preserves the existing function's keys (keys generated only on first create) and returns **200** when the function already existed, **201** when newly created.

4. **Site start/stop returned 405; state pinned to `Running`.** `/stop` and `/start` weren't routed, and `toSiteResource` always reported `State: "Running"`. Added POST `.../sites/{name}/stop` → `Stopped` and `.../start` → `Running`, backed by a `SetSiteState` provider method and a stored `State` (default `Running` on create) that a subsequent GET reflects. `restart` still works.

## Tests

- PATCH: partial update touches only the supplied field and preserves location/kind/runtime/app-settings + an unmodeled overlay property; system-assigned identity assign via PATCH; 404 on a missing site (real armappservice `WebAppsClient` over httptest).
- SKU tier derived for B1/S1/P1v3/Y1/EP1 when tier omitted; explicit tier preserved; table-driven `deriveSKUTier` coverage.
- Function re-PUT preserves the key and returns 200 (raw wire client, since the generated SDK's `BeginCreateFunction` only accepts 201).
- stop → GET reports `Stopped`, start → `Running`.

Gates: `go build ./...`, `go vet`, scoped `go test`, gofmt, `go generate` (zero doc diff), compatgen (exit 0, zero diff), and `golangci-lint --new-from-rev=origin/development` all green.

### DIFF
--- a/providers/azure/functions/app_service_plan.go
+++ b/providers/azure/functions/app_service_plan.go
@@ -30,6 +30,78 @@ type AppServicePlan struct {
 	Tags          map[string]string
 }
 
+// App Service plan pricing tiers, the tier real Azure derives from a SKU name
+// when the caller omits it (azurerm sends only the SKU name + capacity).
+const (
+	tierFree             = "Free"
+	tierShared           = "Shared"
+	tierBasic            = "Basic"
+	tierStandard         = "Standard"
+	tierPremium          = "Premium"
+	tierPremiumV2        = "PremiumV2"
+	tierPremiumV3        = "PremiumV3"
+	tierDynamic          = "Dynamic"
+	tierElasticPremium   = "ElasticPremium"
+	tierIsolated         = "Isolated"
+	tierIsolatedV2       = "IsolatedV2"
+	tierWorkflowStandard = "WorkflowStandard"
+)
+
+// deriveSKUTier maps an App Service plan SKU name to its pricing tier the way
+// real Azure does server-side when a create omits the tier: Y->Dynamic (Consumption),
+// EP->ElasticPremium, WS->WorkflowStandard (Logic Apps), B->Basic, S->Standard,
+// P#v2->PremiumV2, P#v3->PremiumV3, P(other)->Premium, F->Free, D->Shared,
+// I(#v2)->IsolatedV2 else Isolated. An unrecognized name falls back to Standard
+// rather than forcing Dynamic (which would mis-bill a dedicated plan).
+func deriveSKUTier(skuName string) string {
+	name := strings.ToUpper(strings.TrimSpace(skuName))
+	if name == "" {
+		return tierDynamic
+	}
+
+	// Longest prefixes first so EP/WS win over the single-letter matches.
+	prefixes := []struct{ prefix, tier string }{
+		{"EP", tierElasticPremium},
+		{"WS", tierWorkflowStandard},
+		{"Y", tierDynamic},
+		{"F", tierFree},
+		{"D", tierShared},
+		{"B", tierBasic},
+		{"S", tierStandard},
+	}
+	for _, p := range prefixes {
+		if strings.HasPrefix(name, p.prefix) {
+			return p.tier
+		}
+	}
+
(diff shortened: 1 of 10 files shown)


NitinKumar004 (commented):
Reviewed at head 0a8e990. Build/vet/test (incl -race on the provider) green; gofmt clean; 0 new lint; go generate and compatgen produce zero diff; the 10 expected files are the only changes.

PATCH partial-update (WebApps_Update) — the main risk — is correct. The request type is fully pointer/nilable (`patchSiteRequest`/`patchSiteProperties`/`patchSiteConfig`), and every path preserves an omitted field:
- Site metadata: `applySiteMetaPatch` only writes fields whose patch pointer is non-nil, so a PATCH touching only `httpsOnly` leaves location, kind, `linuxFxVersion`, identity and appSettings as stored. Verified against the real SDK, including that an unmodeled overlay property (`clientAffinityEnabled`) survives.
- Function record: `patchFunctionConfig` builds a partial `FunctionConfig` and `applyConfigUpdates` preserves every zero field (runtime/handler/tags/environment), so an omitted runtime or app-settings block is untouched.
- Identity: `identity{type:SystemAssigned}` assigns; `identity{type:None}` detaches to nil; an omitted identity block preserves the attached identity. All three confirmed via the SDK.
- appSettings not leaked: `toSiteResource` still emits `appSettings: null` on the PATCH response so the unmodeled-property overlay treats it as owned; a PATCH that itself carries settings does not echo them on the PATCH response or a plain GET, while the replacement is stored (readable only via config/appsettings/list).
- 404 semantics: PATCH resolves the (scoped) function first and returns 404 before decoding or writing, so a PATCH on a missing site or wrong resource group makes no partial write.

SKU tier derivation: `deriveSKUTier` maps by longest-prefix — B1→Basic, S1→Standard, P1v3→PremiumV3, Y1→Dynamic, EP→ElasticPremium, F→Free — with an explicit tier preserved and an unrecognized name falling back to Standard (no crash/mis-bill).

Function key stability: a first function PUT creates (201, generated default key); a re-PUT updates (200) and `CreateSiteFunction` clones the existing keys when the body omits them, so the key is not rotated on update.

Start/stop state: default Running on create; POST /stop→Stopped, /start→Running reflected on GET; restart preserved; an unknown site sub-resource returns 405, not 500.

No blocking issues. LGTM.

PR #765, opened 2026-08-26
fix(azure-servicebus): SQL filter LIKE/IN/EXISTS/IS NULL, round-trip queue/topic ARM properties, enforce duplicate detection

NitinKumar004:
## Summary

Fixes three confirmed Azure Service Bus correctness/parity issues (all in `server/azure/servicebus/` + `providers/azure/servicebus/`), each with real-user tests.

### 1. SQL filter `LIKE` / `IN` / `EXISTS` / `IS NULL` (wrong-result / message over-delivery)
Unsupported SqlFilter clauses parsed to nothing and fell back to the no-drop `match=true` path, so a rule like `sys.Label LIKE 'order-%'` or `user.Region IN ('us','eu')` matched **every** message. The recursive-descent evaluator now implements:
- `LIKE` / `NOT LIKE` with `%` (any run) and `_` (single char) wildcards, case-sensitive per real SB.
- `IN (list)` / `NOT IN` membership (string and numeric members).
- `IS NULL` / `IS NOT NULL` and `EXISTS(prop)` / `NOT EXISTS(prop)` presence tests (an absent property is NULL / not-EXISTS).

The existing operators, AND/OR precedence and parens are unchanged, and a genuinely-unparseable clause still keeps the safe no-drop fallback.

### 2. Round-trip queue/topic ARM properties (Terraform perpetual diff)
`queueProperties` / `topicProperties` lacked `autoDeleteOnIdle`, `duplicateDetectionHistoryTimeWindow`, `enableBatchedOperations`, plus (queue) `forwardTo` and `enableExpress`, which `azurerm_servicebus_queue`/`_topic` set — causing plan-after-apply drift. These now store and round-trip faithfully with real-SB defaults (`autoDeleteOnIdle` max sentinel, `duplicateDetectionHistoryTimeWindow` `PT10M`, `enableBatchedOperations` true, `enableExpress` false). `enableBatchedOperations` is a `*bool` so an explicit `false` is preserved rather than overwritten by the default.

`forwardTo` is **round-trip-only** (stored/echoed faithfully so TF does not drift); it does not yet actually forward messages — that remains a deferred feature.

### 3. Enforce `RequiresDuplicateDetection` (correctness)
`requiresDuplicateDetection` round-tripped but was never enforced on the send path. The backing store now deduplicates on `MessageId` within `duplicateDetectionHistoryTimeWindow` (default `PT10M`, driven by the injected Clock), mirroring the existing FIFO dedup mechanism: a repeat MessageId inside the window is silently accepted but not re-enqueued; an empty MessageId is never deduped. Topics inherit their duplicate-detection setting onto each subscription's backing store, giving the same observable result as topic-level detection.

## Tests
- Filter: LIKE `order-%` matches `order-1` not `shipment-1`; NOT LIKE; IN/NOT IN membership (string + numeric); IS NULL / IS NOT NULL / EXISTS / NOT EXISTS on present + absent props; composition with AND/OR; unsupported clause keeps safe fallback; direct LIKE-matcher table.
- ARM properties: queue + topic PUT -> GET round-trip and defaults (including explicit `enableBatchedOperations:false` preserved).
- Dedup: same-MessageId re-send dropped inside the window and accepted after it (FakeClock); distinct MessageIds accepted; empty MessageId never deduped; dedup off when not required; topic-level dedup on the fan-out path.

## Gates
`go build ./...`, `go vet`, scoped `go test`, gofmt, `golangci-lint --new-from-rev=origin/development` (0 issues), `go generate` + `compatgen` (zero diff) all green.

### DIFF
--- a/providers/azure/servicebus/servicebus.go
+++ b/providers/azure/servicebus/servicebus.go
@@ -23,6 +23,9 @@ const (
 	defaultVisibilityTimeout = 30
 	maxReceiveMessages       = 10
 	deduplicationWindow      = 5 * time.Minute
+	// defaultDupDetectionWindow is Service Bus' default duplicate-detection
+	// history window when RequiresDuplicateDetection is set without a window.
+	defaultDupDetectionWindow = 10 * time.Minute
 )
 
 // sbMessage represents an internal message stored in a queue.
@@ -59,7 +62,13 @@ type queueData struct {
 	createdAt          time.Time
 	lastModifiedAt     time.Time
 	deduplicationIndex map[string]time.Time
-	dlqConfig          *driver.DeadLetterConfig
+	// requiresDupDetection enables Azure Service Bus MessageId-based duplicate
+	// detection; dedupByMessageID tracks the last-seen time per MessageId and
+	// dupDetectionWindow is the look-back window.
+	requiresDupDetection bool
+	dupDetectionWindow   time.Duration
+	dedupByMessageID     map[string]time.Time
+	dlqConfig            *driver.DeadLetterConfig
 	// deadLetterOnExpiration routes an expired message to the dead-letter queue
 	// instead of dropping it (Service Bus deadLetteringOnMessageExpiration).
 	deadLetterOnExpiration bool
@@ -175,6 +184,11 @@ func (m *Mock) CreateQueue(_ context.Context, cfg driver.QueueConfig) (*driver.Q
 
 	now := m.opts.Clock.Now()
 
+	dupWindow := cfg.DuplicateDetectionWindow
+	if dupWindow <= 0 {
+		dupWindow = defaultDupDetectionWindow
+	}
+
 	qd := &queueData{
 		info:                   info,
 		messages:               make([]*sbMessage, 0),
@@ -185,6 +199,9 @@ func (m *Mock) CreateQueue(_ context.Context, cfg driver.QueueConfig) (*driver.Q
 		createdAt:              now,
 		lastModifiedAt:         now,
 		deduplicationIndex:     make(map[string]time.Time),
+		requiresDupDetection:   cfg.RequiresDuplicateDetection,
+		dupDetectionWindow:     dupWindow,
+		dedupByMessageID:       make(map[string]time.Time),
 		dlqConfig:              cfg.DeadLetterQueue,
 		deadLetterOnExpiration: cfg.DeadLetterOnExpiration,
 		metadata:               metadata,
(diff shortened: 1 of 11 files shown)


NitinKumar004 (commented):
**Verdict: MERGE** — clean build/vet/test/-race on the pushed SHA, no regressions, correctness verified.

**Gates (pushed SHA c02c9c5)**
- `go build ./...`, `go vet ./...` clean; scoped tests green (`server/azure/servicebus`, `providers/azure/servicebus`, `services/messagequeue`); `-race` on `providers/azure/servicebus` green.
- Exactly the expected 11 files changed. gofmt clean; `golangci-lint --new-from-rev=origin/development` 0 issues; `go generate ./...` and `compatgen` produce zero diff.

**Shared driver blast radius (QueueConfig)**
- `RequiresDuplicateDetection` + `DuplicateDetectionWindow` are additive fields, documented as Azure-only and ignored by non-Azure providers. Verified: `server/aws/sqs`, `providers/aws/sqs`, `server/gcp/pubsub`, `providers/gcp/pubsub` all pass — SQS/PubSub simply don't set them, no behavior change.

**SQL filter grammar (filter.go)**
- LIKE: `%` matches any run, `_` a single char, full-value anchored (`Label LIKE 'order'` matches only "order"), case-sensitive; `order-%` matches "order-1" not "shipment-1"; NOT LIKE inverts. Verified via targeted cases.
- IN / NOT IN: string + numeric membership (numeric-aware via `sqlEqual`); NOT IN inverts.
- IS NULL / IS NOT NULL / EXISTS: absent custom prop → IS NULL true / EXISTS false; present → opposite; unknown `sys.*` → no-drop.
- Regression: `=,!=,<>,<,>,<=,>=`, AND/OR precedence, parens, sys./user./bare identifiers, numeric-vs-lexical compare all still correct.
- No-drop invariant holds: tokenize failure, parse error, trailing tokens, unbalanced paren, unterminated string, unknown `sys.*` all fall back to match=true. A well-formed comparison against an absent custom property correctly excludes (pre-existing, matches SB). No unparseable clause drops a message.
- Minor (non-blocking): the DEFER comment says arithmetic falls to the no-drop path, but a bare-custom arithmetic clause like `priority + 1 = 4` parses as an absent-property comparison and excludes rather than delivering. This is pre-existing behavior (the absent-property-excludes path predates this PR) and is a reasonable SB-consistent reading, but the comment slightly over-promises for that shape.

**ARM property round-trip**
- autoDeleteOnIdle, duplicateDetectionHistoryTimeWindow, enableBatchedOperations, forwardTo, enableExpress round-trip PUT→GET→LIST with correct SB defaults. `EnableBatchedOperations` is `*bool`, defaulted only when nil, so an explicit `false` is preserved. forwardTo is stored/echoed only (no actual forwarding) — accepted.

**Duplicate detection (MessageId)**
- Clock-driven (FakeClock): within `DuplicateDetectionWindow` a repeat MessageId is accepted but not re-enqueued; past the window it enqueues; distinct MessageIds always enqueue; empty MessageId never deduped. Window anchored to first send. No interaction with FIFO DeduplicationID dedup (separate maps/keys). Topic-level detection propagates to each subscription backing store via dupConfig.

PR #766, opened 2026-08-26
fix(azure-aks): preserve node pools on cluster PUT, osDiskType/scale-to-zero/UserAssigned identity/inline pool fields

NitinKumar004:
## Summary

Fixes five real AKS control-plane bugs found by a confirming audit, all contained to `providers/azure/aks/` + `server/azure/aks/`. The headline fix is a HIGH-severity data-loss bug.

### BUG-1 [HIGH data-loss] — cluster PUT wiped ALL node pools
`CreateOrUpdateCluster` unconditionally wiped every pool under the cluster and re-added only the inline `agentPoolProfiles`. A cluster PUT that omits `agentPoolProfiles` (the SDK has no PATCH — a tags-only or version-only update is a full PUT) therefore dropped the cluster to **zero pools**, destroying pools created via the standalone AgentPools API (`azurerm_kubernetes_cluster_node_pool`). A nil `properties` block additionally reset kubernetesVersion/dnsPrefix/networkProfile/identity to defaults.

Fix: pools are now reconciled **by name** — the PUT upserts the pools it carries and never removes a pool absent from the body (real AKS never drops a cluster to zero pools). Cluster fields are merged: on an update, any field the request omits is preserved from stored state, so a tags-only or version-only PUT no longer resets networkProfile, identity, dnsPrefix, or the node pools.

### BUG-2 [HIGH] — osDiskType (Ephemeral) always read "Managed"
`osDiskType`/`type` were hardcoded on build and never mapped from input, so the overlay treated them as modeled and the hardcoded value won. Added `OSDiskType`/`Type` to the pool input, mapped both on the inline and standalone paths, and defaulted only when empty (Ephemeral now round-trips; Managed/VirtualMachineScaleSets remain the defaults when unset).

### BUG-3 [MED] — count:0 scale-to-zero impossible
`count` used `omitempty` on a non-pointer, so 0 was indistinguishable from omitted and forced back to the default (3). `count` is now `*int32` end-to-end; an explicit 0 is preserved, and the default applies only when count is truly absent (nil).

### BUG-4 [MED] — UserAssigned identity round-trip lossy
The identity wire shape modeled only Type/PrincipalID/TenantID, and the property overlay does not cover the top-level `identity` block, so `userAssignedIdentities` was dropped. Added `userAssignedIdentities` to the wire model; on a UserAssigned PUT the submitted identity IDs are stored and a deterministic principalId/clientId is synthesized per identity and echoed on GET. SystemAssigned path unchanged.

### BUG-5 [MED TF-drift] — inline default_node_pool fields dropped
Fields submitted inline in `agentPoolProfiles[]` (upgradeSettings.maxSurge, pool tags, enableFIPS, spotMaxPrice, scaleSetEvictionPolicy, nodePublicIPPrefixID, kubeletDiskType, kubeletConfig, linuxOSConfig) were lost because the overlay does not descend into arrays, while the same fields survived on a standalone pool. Modeled these on the pool input + wire structs so an inline submission round-trips identically to a standalone pool. (The alternative — teaching the shared `echo_properties.go` overlay to descend into arrays — is broader/riskier and was intentionally not done here.)

## Tests
Real `armcontainerservice` SDK over httptest:
- standalone-pool-survives-cluster-PUT (the critical data-loss guard) + nil-properties-preserves-fields
- osDiskType Ephemeral round-trips (inline + standalone) + defaults-to-Managed
- count:0 scale-to-zero preserved
- userAssignedIdentities echoed with synthesized principalId/clientId
- inline upgradeSettings/tags/enableFIPS round-trip

### DIFF
--- a/providers/azure/aks/agentpool_cost_test.go
+++ b/providers/azure/aks/agentpool_cost_test.go
@@ -54,7 +54,7 @@ func TestAgentPoolScaleSetPriority(t *testing.T) {
 		pool, err := m.CreateOrUpdateAgentPool(ctx, "rg-1", "k8s-spot", AgentPoolInput{
 			Name:             "spotpool",
 			VMSize:           "Standard_D2s_v3",
-			Count:            3,
+			Count:            int32Ptr(3),
 			ScaleSetPriority: "Spot",
 		})
 		requireNoError(t, err)
@@ -83,7 +83,7 @@ func TestAgentPoolScaleSetPriority(t *testing.T) {
 		pool, err := m.CreateOrUpdateAgentPool(ctx, "rg-1", "k8s-reg", AgentPoolInput{
 			Name:   "regpool",
 			VMSize: "Standard_D2s_v3",
-			Count:  2,
+			Count:  int32Ptr(2),
 		})
 		requireNoError(t, err)
 		assertEqual(t, "Regular", pool.ScaleSetPriority)
--- a/providers/azure/aks/aks.go
+++ b/providers/azure/aks/aks.go
@@ -12,6 +12,7 @@ package aks
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -82,9 +83,20 @@ type ManagedCluster struct {
 	IdentityType string
 	PrincipalID  string
 	TenantID     string
+	// UserAssignedIdentities echoes identity.userAssignedIdentities: the ARM
+	// resource ID of each assigned identity mapped to its synthesized
+	// principal/client pair. Populated only for a UserAssigned identity.
+	UserAssignedIdentities map[string]UserAssignedIdentity
 
 	CreatedAt time.Time
 	UpdatedAt time.Time
+}
+
+// UserAssignedIdentity is the synthesized principal/client pair echoed for a
+// single user-assigned managed identity.
+type UserAssignedIdentity struct {
+	PrincipalID string
+	ClientID    string
 }
 
 // NetworkProfile captures the cluster network configuration echoed back on read.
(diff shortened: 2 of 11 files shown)


NitinKumar004 (commented):
Verified against the pushed head (246d8e6). Build/vet/scoped tests/`-race` on `providers/azure/aks` all green; gofmt clean; 0 new lint (`--new-from-rev`); `go generate` and `compatgen` both zero-diff.

**BUG-1 (data-loss) — correct.** `reconcileInlinePools` upserts inline pools by name and never deletes, so a cluster PUT with no `agentPoolProfiles` leaves every pool (including standalone `azurerm_kubernetes_cluster_node_pool` pools) intact. Reproduced the four cases directly against the Mock:
- version-only PUT with no pools: both pools survive, version updated, `networkProfile` + identity + location preserved.
- PUT with `agentPoolProfiles=[system(count=5)]`: system updated to 5, standalone `userpool` NOT wiped.
- `networkProfile.networkPlugin` change (azure→kubenet): applied, not over-preserved.
- fresh cluster `[a,b]` re-PUT with `[a]`: `b` is preserved (never removed).

The last is a deliberate never-remove choice (documented in the `reconcileInlinePools` comment): pool removal happens via the standalone agentPool DELETE, never by omission from a cluster PUT. This errs toward preservation over data-loss and matches how Terraform splits `default_node_pool` from separate node-pool resources — correct direction. No path lets the cluster drop to zero pools.

`resolveClusterFields`/`mergeStr` cleanly split create-vs-update: submitted non-empty wins, omitted fields preserve stored on update / apply default on create. Tags, EnableRBAC, NetworkProfile, and identity each guarded by `submitted || !existing`, so a partial PUT preserves omitted fields without over-preserving a field the PUT does change.

**BUG-3 (`count *int32` ripple) — safe.** Stored `AgentPool.Count` stays concrete `int32`; `ptrInt32` wraps it on the way out (always non-nil). Every input read (`buildAgentPool`, `totalNodeCount`) nil-checks: nil → `defaultNodeCount`, explicit 0 → scale-to-zero preserved. No non-test call site in `resourcegraph`/`kubernetes_discovery` reads `.Count`; the touched test call-sites are legitimate pointer-wrapping, no nil-deref hidden.

**BUG-2 osDiskType / BUG-4 identity / BUG-5 inline fields — correct.** Ephemeral round-trips inline and standalone; unset defaults to Managed; `Type` defaults correctly. UserAssigned synthesizes a deterministic principal/client pair per ARM ID and is preserved when a later PUT omits identity; SystemAssigned path unchanged. Inline `upgradeSettings`/`tags`/`enableFIPS`/`spotMaxPrice`/etc. round-trip the same as standalone. The shared echo overlay is untouched (diff is confined to the AKS packages + resourcegraph test call-sites), and the prior networkProfile round-trip still holds.

No blocking issues.

PR #767, opened 2026-08-26
fix(azure-sql): round-trip elastic pool sku.capacity, default server version 12.0 when omitted

NitinKumar004:
## What

Two Azure SQL wire-protocol round-trip fixes found by a real-SDK (armsql) audit.

### Elastic pool `sku.capacity` silently dropped (MED/HIGH — Terraform drift)
Creating an elastic pool with `sku={name,tier,capacity:N}` returned `sku.capacity=null` on GET/List. `azurerm_mssql_elasticpool` makes `sku.capacity` REQUIRED, so the missing field caused perpetual Terraform drift. Unlike `properties`, `sku` is a top-level sibling, so it could not fall back to the `echo_properties` overlay. The database path already carried capacity (the merged B1 fix); the pool path was never given the same treatment.

Fixed across all three layers, mirroring the database B1 capacity fix exactly:
- `services/relationaldb/driver/driver.go` — added `SKUCapacity int` to the shared `ElasticPoolConfig` and `ElasticPool` structs (additive; other consumers like RDS/Redshift simply don't set it).
- `server/azure/sql/subresources.go` — `writeElasticPool` now copies `body.SKU.Capacity` (both PUT and PATCH), and `toARMElasticPool` emits it.
- `providers/azure/sql/subresources.go` — `CreateElasticPool` stores it; `UpdateElasticPool` carries it guarded by `!= 0`, so a PATCH omitting capacity preserves the existing value.

### Server `version` default "12.0" not synthesized when omitted (LOW — G3)
A server create that omitted `properties.version` returned an empty version instead of Azure's default `"12.0"`. Now defaulted on create only; the upsert (PUT-on-existing) path still passes the raw request version so a PUT that omits version preserves the existing one rather than clobbering it with the default. An explicit version continues to round-trip unchanged.

## Result
- Pool create with `capacity:4` -> GET/List returns `sku.capacity=4`.
- PATCH changing capacity applies; PATCH omitting capacity preserves.
- Server create omitting version -> GET version `"12.0"`; explicit version preserved.

## Tests
New real-armsql-over-httptest coverage in `server/azure/sql/elasticpool_capacity_sdk_test.go`: pool capacity round-trip (Get + List), PATCH change + PATCH-omit-preserve, and server version default + explicit round-trip.

Gates (scoped): build, vet, `go test` on azure sql + relationaldb + AWS rds/redshift spot-check, gofmt, `go generate` (zero diff), compatgen (exit 0, zero diff), golangci-lint new-from-development (0 issues).

### DIFF
--- a/providers/azure/sql/subresources.go
+++ b/providers/azure/sql/subresources.go
@@ -274,6 +274,7 @@ func (m *Mock) CreateElasticPool(_ context.Context, cfg rdsdriver.ElasticPoolCon
 		Location:     location,
 		SKUName:      cfg.SKUName,
 		SKUTier:      cfg.SKUTier,
+		SKUCapacity:  cfg.SKUCapacity,
 		MaxSizeBytes: cfg.MaxSizeBytes,
 		MinCapacity:  cfg.MinCapacity,
 		MaxCapacity:  cfg.MaxCapacity,
@@ -425,6 +426,10 @@ func (m *Mock) UpdateElasticPool(_ context.Context, cfg rdsdriver.ElasticPoolCon
 
 	if cfg.SKUTier != "" {
 		pool.SKUTier = cfg.SKUTier
+	}
+
+	if cfg.SKUCapacity != 0 {
+		pool.SKUCapacity = cfg.SKUCapacity
 	}
 
 	if cfg.MaxSizeBytes != 0 {
--- a/server/azure/sql/elasticpool_capacity_sdk_test.go
+++ b/server/azure/sql/elasticpool_capacity_sdk_test.go
@@ -1,0 +1,205 @@
+package sql_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/sql/armsql"
+)
+
+const (
+	poolCapacity        = int32(4)
+	poolCapacityUpdated = int32(8)
+)
+
+// createPool creates an elastic pool with the given SKU and waits for the
+// long-running create to finish.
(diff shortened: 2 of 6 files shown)


NitinKumar004 (commented):
Verdict: MERGE.

Verified against the pushed head (b034007). Clean build + vet on the full module; targeted tests green for providers/azure/sql, server/azure/sql, services/relationaldb. Exactly the 6 expected files changed; gofmt clean; 0 new lint; go generate and compatgen both zero-diff.

Shared-struct blast radius (SKUCapacity on relationaldb ElasticPoolConfig/ElasticPool): the change is purely additive, and ElasticPool is consumed only by the Azure SQL provider/server — no AWS RDS or Redshift references. Ran providers/aws/rds, providers/aws/redshift, server/aws/rds, server/aws/redshift tests: all pass (they never set the new field).

Capacity round-trip: writeElasticPool copies body.SKU.Capacity into the config on both PUT and PATCH, CreateElasticPool persists it, and toARMElasticPool emits sku.capacity on GET and List. Confirmed with the real armsql SDK test.

PATCH-preserve: UpdateElasticPool merges capacity under a `!= 0` guard, mirroring the existing SKUName/SKUTier merge pattern, so a PATCH omitting capacity preserves the stored value rather than resetting to 0. The tradeoff — capacity can't be explicitly set to 0 — is harmless since 0 is not a valid pool capacity.

Version default: server create defaults EngineVersion to "12.0" only when the request omits it; an explicit version round-trips. The upsert path (PUT on an existing server) forwards the raw request version (reqVersion), not the create-time default, and ModifyCluster guards empty EngineVersion — so a PUT omitting version preserves the existing value instead of clobbering it to "12.0". Create-only default is correct. (Minor: the upsert-preserve path is verified by code inspection but not directly covered by a test — non-blocking.)

PR #768, opened 2026-08-26
fix(azure-network): idempotent re-PUT for public IP (no duplicate), NSG tags, NAT gateway tags+PIP association

NitinKumar004:
## Summary

Fixes three re-PUT idempotency bugs in Azure networking. Real ARM `PUT` is `CreateOrUpdate`, but these handlers treated a repeated PUT to an existing resource as create-only, so a second apply either duplicated the resource or silently dropped changes.

### Bug 1 (HIGH — duplicate resource + leak): Public IP re-PUT created a duplicate
`createPublicIP` called `AllocateAddress` unconditionally. `AllocateAddress` mints a fresh allocation every call, so a second PUT to the same `publicIPAddresses/{name}` allocated a new hidden address: GET returned the stale one, LIST returned the same name twice, and the old allocation leaked. Added an `upsertPublicIP` path that updates the existing allocation in place (SKU, allocation method, idle timeout, DNS label, tags) via a new Azure-only `UpdateAzurePublicIP`, preserving the allocation id/address. LIST now returns exactly one entry per name.

### Bug 2 (MED): NSG re-PUT dropped tag changes
`upsertNSG` returned the existing NSG unchanged, never applying tag edits. Now calls `UpdateSecurityGroupTags` on the found branch, mirroring `upsertVNet`. Security rules were already rewritten unconditionally and stay correct.

### Bug 3 (MED): NAT gateway re-PUT dropped tags + public-IP re-association
`upsertNATGateway` returned the existing gateway unchanged, discarding the freshly-resolved public-IP binding and tags. Now re-applies both via a new Azure-only `UpdateAzureNATGateway`, which rebinds the public IP (freeing the previous allocation) and updates tags.

## Design

The two new mutation methods are added to the existing Azure-only, type-asserted `AzureNetworkMetadata` surface (AWS/GCP do not implement it), consistent with the peering-subresource CRUD already living there — so the shared cross-cloud `Networking` interface and the AWS/GCP providers are untouched.

## Tests

Real `armnetwork` SDK over `httptest`, each verified fail-before / pass-after:
- Public IP re-PUT updates in place — GET returns new idleTimeout/tags, address stable, and LIST returns exactly ONE entry (the critical duplicate-prevention assertion).
- NSG re-PUT with changed tags reflects on GET while security rules stay correct.
- NAT gateway re-PUT swaps the bound public IP and updates tags; the freed public IP is reusable (no leak).

## Gates
build, vet, scoped tests, gofmt, `go generate` (coverage docs regenerated, idempotent), and compatgen all green; scoped `golangci-lint --new-from-rev` reports 0 issues.

### DIFF
--- a/docs/coverage/aws/vpc.md
+++ b/docs/coverage/aws/vpc.md
@@ -97,6 +97,8 @@ AzureNetworkMetadata is an OPTIONAL, type-asserted capability. The Azure
 | `PutAzureNSGMetadata` |  |
 | `PutAzureVNetMetadata` |  |
 | `SetAzureVNetPeeringState` | SetAzureVNetPeeringState atomically updates just the peeringState field of |
+| `UpdateAzureNATGateway` | UpdateAzureNATGateway re-applies the mutable fields of an existing NAT |
+| `UpdateAzurePublicIP` | UpdateAzurePublicIP overwrites the mutable fields of an existing public IP |
 | `UpsertAzureNSGRule` | UpsertAzureNSGRule creates or replaces a single custom security rule by |
 | `UpsertAzureVNetPeering` | UpsertAzureVNetPeering creates or replaces a single virtualNetworkPeerings |
 
--- a/docs/coverage/azure/vnet.md
+++ b/docs/coverage/azure/vnet.md
@@ -97,6 +97,8 @@ AzureNetworkMetadata is an OPTIONAL, type-asserted capability. The Azure
 | `PutAzureNSGMetadata` |  |
 | `PutAzureVNetMetadata` |  |
 | `SetAzureVNetPeeringState` | SetAzureVNetPeeringState atomically updates just the peeringState field of |
+| `UpdateAzureNATGateway` | UpdateAzureNATGateway re-applies the mutable fields of an existing NAT |
+| `UpdateAzurePublicIP` | UpdateAzurePublicIP overwrites the mutable fields of an existing public IP |
 | `UpsertAzureNSGRule` | UpsertAzureNSGRule creates or replaces a single custom security rule by |
 | `UpsertAzureVNetPeering` | UpsertAzureVNetPeering creates or replaces a single virtualNetworkPeerings |
 
--- a/docs/coverage/coverage.json
+++ b/docs/coverage/coverage.json
@@ -6648,6 +6648,14 @@
             "doc": "SetAzureVNetPeeringState atomically updates just the peeringState field of"
           },
           {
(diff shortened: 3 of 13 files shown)


NitinKumar004 (commented):
Reviewed at head 626fccf. Build, vet, and the targeted tests (server/azure/vnet, providers/azure/vnet, services/networking) all pass, including `-race` on the provider package. gofmt clean; `golangci-lint --new-from-rev=origin/development` reports 0 new issues. `go generate ./...` + `go run ./internal/compatgen` produce zero further diff — docs/coverage is fully regenerated and idempotent.

Interface change is correctly Azure-scoped: the two new methods (`UpdateAzurePublicIP`, `UpdateAzureNATGateway`) land only on the Azure-only `AzureNetworkMetadata` interface, are reached via the type-asserted `azureMeta()` helper, and are implemented only under `providers/azure/vnet`. The shared cross-cloud `Networking` interface and the AWS/GCP providers are untouched — `go build ./...` confirms all implementers still compile.

Public IP re-PUT (the data-corruption bug): `upsertPublicIP` finds the existing PIP by name and mutates it in place via `memstore.Update` keyed by allocation id, so LIST returns exactly one entry per name — no duplicate, no orphaned allocation. `UpdateAzurePublicIP` rewrites only mutable fields (SKU, allocation method, idle timeout, DNS label, tags, zones) and leaves the map key, `PublicIP` address, and `AssociationID` intact, so a Static IP keeps its address across re-PUT. `TestSDKPublicIPRePUTUpdatesInPlace` asserts the single-entry invariant plus stable address and updated idleTimeout/tags.

NSG re-PUT: on the found branch, tags flow through `UpdateSecurityGroupTags` while security rules are applied on their own path (`PutAzureNSGMetadata`), so the tag update does not touch the rules. `TestSDKNSGRePUTUpdatesTags` verifies the changed/added tags reflect on GET and the seeded allow-ssh rule survives.

NAT gateway re-PUT: `UpdateAzureNATGateway` binds the new allocation first (rejecting one already in use) and only then frees the previous binding, so a failed rebind leaves the gateway untouched; the old allocation's `AssociationID` is cleared, making it reusable. `TestSDKNATGatewayRePUTUpdatesTagsAndPIP` swaps pip-a→pip-b, confirms GET reflects the new association and tags, and proves the freed pip-a rebinds to another gateway (no leak, no double-binding). The idleTimeout/SKU/zones echo-only limitation is unchanged.

Verdict: MERGE.

PR #769, opened 2026-08-26
fix(azure-eventgrid): preserve publisher event id + dataVersion, support PATCH on subscriptions/system topics/domains

NitinKumar004:
## What

Fixes three real-cloud-parity gaps in Azure Event Grid found by a confirming audit:

1. **Publisher event `id` was discarded.** `PutEvents` regenerated the event id unconditionally, clobbering the publisher-supplied id that the publish path correctly carried in. Real Event Grid preserves the publisher `id` end-to-end (subscribers dedup on it). Now the id is only synthesized when the publisher omits it.

2. **Publisher `dataVersion` was dropped and `metadataVersion` never set.** The shared `driver.Event` had no `DataVersion`, so the wire `dataVersion` was lost and delivery hardcoded `"1.0"` with an empty `metadataVersion`. Added `DataVersion` to `driver.Event` (additive; Event Grid is the only setter — AWS EventBridge / GCP Eventarc unaffected), threaded it publish -> delivery, and delivery now emits the publisher's `dataVersion` (defaulting to `"1.0"` only when omitted) with `metadataVersion: "1"`.

3. **PATCH (`BeginUpdate`) returned 405** on event subscriptions (topic-scoped and scope-bound), system topics, and domains. `az eventgrid event-subscription update` / SDK `BeginUpdate` send PATCH. Added a PATCH arm to each doing a partial merge (omitted fields preserved — no nil-mask data loss), mirroring the existing topic PATCH. Subscription PATCH merges destination/filter/labels/etc. via a top-level key merge on the round-tripped properties; system-topic/domain PATCH merge tags and mutable properties. 404 on a missing resource before any write. Responses return 201 to satisfy the armeventgrid Update LRO pollers.

## Why

Unmodified apps, CLIs, IaC (Terraform `azurerm_eventgrid_event_subscription` updates), and SDK `BeginUpdate` calls now behave like real Event Grid: publisher ids and data versions survive delivery, and updates no longer 405.

## Tests

Real `armeventgrid` SDK + data-plane publish over `httptest`:
- publisher id `e1` preserved on delivery + in `GetEventHistory`; no-id publish gets a synthesized id
- `dataVersion "2.0"` preserved on delivery + `metadataVersion "1"` (and `"1.0"` default when omitted)
- PATCH on topic + scope-bound event subscriptions updates the supplied field and preserves the omitted one (no nil-mask); 404 on missing
- system topic + domain PATCH merge tags / mutable properties and preserve immutable/pre-existing values

### DIFF
--- a/providers/azure/eventgrid/delivery.go
+++ b/providers/azure/eventgrid/delivery.go
@@ -22,6 +22,24 @@ import (
 // ServiceBus/Functions delivery is left as a follow-up (see PR notes).
 const endpointTypeWebHook = "WebHook"
 
+// defaultDataVersion is the dataVersion Event Grid stamps on a delivered event
+// when the publisher omitted one; metadataVersion is the read-only schema
+// version Event Grid always reports as "1".
+const (
+	defaultDataVersion = "1.0"
+	metadataVersion    = "1"
+)
+
+// dataVersionOrDefault returns the publisher's dataVersion, defaulting to "1.0"
+// only when the publisher omitted it (matching real Event Grid).
+func dataVersionOrDefault(v string) string {
+	if v == "" {
+		return defaultDataVersion
+	}
+
+	return v
+}
+
 // webhookDeliveryTimeout bounds how long PutEvents waits for a single WebHook
 // destination to respond. Real Event Grid waits up to 30s before queuing a
 // retry; this mock uses a shorter bound so a dead test endpoint can't hang a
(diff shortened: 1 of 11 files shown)


NitinKumar004 (commented):
Verdict: MERGE.

Verified against the pushed SHA c0c6d6e (clean checkout, HEAD matches).

Gates (laptop-safe, scoped):
- go build ./... + go vet: clean.
- go test ./server/azure/eventgrid/... ./providers/azure/eventgrid/... ./services/eventbus/...: pass; -race on the provider: pass.
- 11 files changed as expected. gofmt -l clean. golangci-lint --new-from-rev=origin/development: 0 new. go generate + compatgen: zero diff.

Shared driver.Event blast radius:
- DataVersion added to services/eventbus/driver/driver.go is purely additive (new string field, empty for AWS/GCP). EventBridge + Eventarc server/provider suites all pass — no regression.

BUG 1 (id preservation):
- PutEvents now synthesizes an id only when events[i].ID == "" (generate-only-when-empty guard confirmed); a publisher id flows through to storeEvent, GetEventHistory, delivery payload, and result.EventIDs. Covered by TestPutEventsPreservesPublisherID / TestPutEventsSynthesizesMissingID and the SDK end-to-end test.

BUG 2 (dataVersion / metadataVersion):
- Delivery uses dataVersionOrDefault(event.DataVersion) -> publisher value preserved, "1.0" only when omitted; metadataVersion constant "1". Correct per real Event Grid. Covered by delivery tests.

PATCH (nil-mask risk):
- Event subscription, scoped subscription, system topic, and domain all: decode body, lock, look up record, 404 BEFORE any write, then partial merge. mergeRawProperties does a top-level key overlay preserving omitted keys (destination survives a filter-only PATCH); on unmarshal error it returns existing (no data loss). mergeTags preserves pre-existing tags on a new-tag PATCH. Domain inputSchema is immutable (only tags + publicNetworkAccess mutated). The real-SDK tests adversarially assert the no-nil-mask and 404-before-write cases and pass, which is strong evidence the 201 LRO status and top-level (no properties wrapper) update marshalling are correct for armeventgrid.

Topic PATCH (pre-existing) and delivery-time filter matching are unchanged/not regressed.

No blocking issues.

PR #770, opened 2026-08-26
fix(azure-acr): PutImage unions tags on same-digest re-push; regenerateCredential requires admin user

NitinKumar004:
## Summary

Fixes two Azure ACR parity bugs found in a confirming audit, both scoped to `providers/azure/acr/`.

### BUG1 [MED] — PutImage clobbered existing tags on same-digest re-push
`PutImage` built a fresh single-tag `ImageDetail` and `Set` it for the digest, discarding any tags already on that manifest. Pushing `sha256:AAA` as `v1`, then the same content as `latest`, lost `v1` (GetImage `v1` -> 404). Content-addressing parity broke because a manifest can carry multiple tags.

Fix: when the digest already exists, union the incoming tag into the existing tag set (dedup, order-preserving) instead of replacing it with a single-tag detail. New `unionTags` helper.

### BUG2 [LOW] — RegenerateRegistryCredential didn't require admin user
`ListRegistryCredentials` gates on `AdminUserEnabled` (FailedPrecondition when disabled) but `RegenerateRegistryCredential` did not. Added the same guard so credential rotation requires the admin user to be enabled, matching real ACR.

## Tests
- Same-digest re-push under a second tag: both `v1` and `latest` resolve to the one manifest, tags == {v1, latest}, single image in the repo.
- Regenerate with admin disabled -> FailedPrecondition; after enabling -> password rotates.

## Gates
- `go build ./...`, `go vet`, `gofmt` clean
- `go test -p=3 ./providers/azure/acr/... ./server/azure/acr/...` pass
- `go generate ./...` + `go run ./internal/compatgen` produce zero diff
- `golangci-lint --new-from-rev=origin/development ./providers/azure/acr/...` 0 new issues

### DIFF
--- a/providers/azure/acr/acr.go
+++ b/providers/azure/acr/acr.go
@@ -203,11 +203,22 @@ func (m *Mock) PutImage(_ context.Context, manifest *driver.ImageManifest) (*dri
 	now := m.opts.Clock.Now().UTC().Format(time.RFC3339)
 	mediaType := resolveMediaType(manifest.MediaType)
 
+	tags := []string{}
+	if manifest.Tag != "" {
+		tags = append(tags, manifest.Tag)
+	}
+
+	// Content-addressing parity: re-pushing the same digest under a new tag must
+	// preserve the pre-existing tags on that manifest rather than clobbering them.
+	if existing, ok := rd.images.Get(digest); ok {
+		tags = unionTags(existing.detail.Tags, manifest.Tag)
+	}
+
 	detail := driver.ImageDetail{
 		RegistryID: registryName,
 		Repository: manifest.Repository,
 		Digest:     digest,
-		Tags:       []string{manifest.Tag},
+		Tags:       tags,
 		SizeBytes:  manifest.SizeBytes,
 		PushedAt:   now,
 		MediaType:  mediaType,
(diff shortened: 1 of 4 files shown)


NitinKumar004 (commented):
MERGE.

Clean-build on pushed SHA 074814a: go build/vet clean; go test (providers/azure/acr + server/azure/acr) pass; -race on provider passes. Exactly the 4 expected files changed. gofmt clean; golangci-lint --new-from-rev=origin/development on the ACR package reports 0 new issues; go generate and compatgen both produce zero diff.

Tag-union (content-addressing): traced PutImage — pushing digest AAA/v1 then AAA/latest yields both tags on ONE manifest entry (Set on the same digest key), and both resolve via GetImage. unionTags copies existing.detail.Tags, appends only when hasTag is false, so:
- re-push AAA/v1 again -> stays {v1,latest} (no dup, hasTag guard),
- push AAA under 3 tags sequentially -> {v1,v2,v3} all preserved, order kept.
updateTagIndex is untouched; the union only affects the SAME digest's prior tags. Cross-digest retag still works: updateTagIndex still strips the tag from any other digest and moves it to the target (TestTagImage covers this and passes). No regression.

Regen admin guard: RegenerateRegistryCredential now returns FailedPrecondition when AdminUserEnabled is false — byte-for-byte parity with ListRegistryCredentials (same code + message). With admin enabled it rotates only the named password via the password/password2 switch; unknown name still InvalidArgument. New test TestRegenerateRegistryCredentialRequiresAdmin asserts both the FailedPrecondition and the rotation.

No regression to identity/webhook/PATCH-nil-mask paths (untouched). Verdict: MERGE.

PR #771, opened 2026-08-26
fix(gcp-pubsub): round-trip topic fields + enableExactlyOnceDelivery, support subscriptions:detach, validate topic/ackDeadline

NitinKumar004:
## Summary
Fixes GCP Pub/Sub wire-handler bugs found in a confirming audit, in `server/gcp/pubsub/`. All fixes target real-cloud round-trip fidelity against `cloud.google.com/go/pubsub` / `google.golang.org/api/pubsub/v1`.

### Fixed
- **BUG1 (HIGH)** Topic resource dropped every field except name+labels. `topic` now carries `messageRetentionDuration`, `schemaSettings`, `kmsKeyName`, `messageStoragePolicy`, `satisfiesPzs`; these are persisted in `topicState` (mirroring the existing labels/labelsSet override pattern) and echoed on get/list via a new `topicView` helper. `topics.patch` honors `updateMask` for the new fields (only masked fields updated). Prevents the `google_pubsub_topic` perpetual diff.
- **BUG2 (MED-HIGH)** Subscription dropped `enableExactlyOnceDelivery`. Added the field plus a `subMaskSetters` entry so it round-trips on create + patch.
- **BUG3 (MED)** `subscriptions:detach` returned 405. Added a `detach` verb case that sets `Detached=true` and returns 200 with an empty body (matching `DetachSubscriptionResponse`).
- **BUG5 (LOW)** `createSubscription` silently defaulted an empty `topic` to the subscription name. Now rejected with 400 INVALID_ARGUMENT.
- **BUG6 (LOW)** `ackDeadlineSeconds` was unvalidated. Now range-checked (10..600) in create and in `applySubMask` -> 400 out-of-range.

### Deferred (noted, not in this PR)
- **BUG4** ordering-key head-of-line blocking, push delivery, and schema validation — larger behavioral changes, deferred.

### Tests
New `sdk_config_test.go` (real SDK over httptest): topic field round-trip (retention + schemaSettings), topic patch updateMask isolation, `enableExactlyOnceDelivery` create+patch round-trip, detach sets detached=true (200), empty-topic create -> 400, ackDeadline 5/5000 -> 400 and 60 ok.

### Gates
- `go build ./...`, `go vet ./server/gcp/pubsub/...` clean
- `go test -p=3 ./server/gcp/pubsub/... ./providers/gcp/pubsub/...` pass
- `gofmt` clean; `golangci-lint --new-from-rev=origin/development ./server/gcp/pubsub/...` 0 new issues
- `go generate ./...` and `go run ./internal/compatgen` (exit 0) produce zero diff

### DIFF
--- a/server/gcp/pubsub/handler.go
+++ b/server/gcp/pubsub/handler.go
@@ -195,12 +195,14 @@ func (h *Handler) createTopic(w http.ResponseWriter, r *http.Request, project, n
 	ts := h.topicLog(name)
 	ts.labels = copyLabels(info.Tags)
 	ts.labelsSet = true
+	ts.msgRetentionDuration = body.MessageRetentionDuration
+	ts.schemaSettings = body.SchemaSettings
+	ts.kmsKeyName = body.KmsKeyName
+	ts.messageStoragePolicy = body.MessageStoragePolicy
+	ts.satisfiesPzs = body.SatisfiesPzs
 	h.mu.Unlock()
 
-	writeJSON(w, http.StatusOK, topic{
-		Name:   topicName(project, info.Name),
-		Labels: info.Tags,
-	})
+	writeJSON(w, http.StatusOK, h.topicView(project, name, info.Tags))
 }
 
 func (h *Handler) getTopic(w http.ResponseWriter, r *http.Request, project, name string) {
@@ -210,10 +212,7 @@ func (h *Handler) getTopic(w http.ResponseWriter, r *http.Request, project, name
 		return
 	}
 
-	writeJSON(w, http.StatusOK, topic{
-		Name:   topicName(project, q.Name),
-		Labels: h.topicLabels(name, q.Tags),
-	})
+	writeJSON(w, http.StatusOK, h.topicView(project, q.Name, q.Tags))
 }
 
 // patchTopic applies topics.patch: only the fields named by updateMask are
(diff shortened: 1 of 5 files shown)


NitinKumar004 (commented):
Reviewed at pushed SHA 2d3d9d4 (HEAD matches). Verdict: MERGE.

Gates (laptop-safe, GOMAXPROCS=3, -p=3):
- go build ./... + go vet ./server/gcp/pubsub/... clean
- go test ./server/gcp/pubsub/... ./providers/gcp/pubsub/... pass; go test -race ./server/gcp/pubsub/... pass
- gofmt -l clean; golangci-lint --new-from-rev=origin/development: 0 issues
- go generate ./... zero diff; go run ./internal/compatgen exit 0, zero diff

Focus verification:
A. Topic field round-trip: createTopic now persists messageRetentionDuration/schemaSettings/kmsKeyName/messageStoragePolicy/satisfiesPzs into topicState and responds via topicView; get/list/patch all route through topicView. Extended nested fields kept as json.RawMessage so they echo verbatim. topicView preserves the labels override (labelsSet) and never drops labels. Confirmed by TestSDKPubSubTopicFieldsRoundTrip.
B. updateMask isolation (key point): patchTopic initializes ts.labels from driver tags when not yet set, then switches per-mask-path, mutating only the named field; persisted state for unmasked fields is untouched. TestSDKPubSubTopicPatchUpdateMask sends changed labels with mask=["messageRetentionDuration"] and asserts labels stay "core" while retention updates to 7200s — directly proves no clobber and labels-not-dropped.
C. enableExactlyOnceDelivery: added to subscription type + subMaskSetters entry present; round-trips on create and patch (TestSDKPubSubExactlyOnceRoundTrip).
D. detach: subscriptions:detach wired; POST-guarded, sets cfg.Detached=true, 200 empty body, 404 when missing.
E. Validations: empty topic on sub create -> 400 INVALID_ARGUMENT; ackDeadlineSeconds validated 10..600 on both create (non-zero) and patch (applySubMask, only when masked); 0 still defaults. Tests cover 5 and 5000 -> 400, 60 ok.
F. No regression: existing subscription/DLQ/filter/updateMask suite passes; immutable-field guards unchanged.

No blocking issues.

PR #772, opened 2026-08-26
fix(azure-cache): round-trip Redis top-level properties (redisConfiguration, enableNonSslPort, minimumTlsVersion, publicNetworkAccess, redisVersion)

NitinKumar004:
## Summary

Azure Cache for Redis dropped every top-level Redis property except SKU and clustering on round-trip, causing a perpetual `azurerm_redis_cache` diff. The `redisProperties` wire type and the driver `CacheConfig`/`CacheInfo` only modeled SKU + shardCount/replicasPerPrimary; everything else the client sent was silently discarded on Create/Get.

This models the missing properties end to end (wire decode -> driver store -> wire emit):

- **redisConfiguration** (`map[string]string`) — full passthrough, including typed keys (maxmemory-policy, maxmemory-reserved) and unmodeled/custom keys.
- **enableNonSslPort** (`*bool`) — pointer so an unset value is omitted rather than emitted as a spurious `false`.
- **minimumTlsVersion**, **publicNetworkAccess** (strings).
- **redisVersion** — previously hardcoded to `6.0`; now honors the request and falls back to the default only when unset.

## Nil-mask discipline

`UpdateCache` (ARM PATCH / CreateOrUpdate-on-existing) already guarded SKU/clustering behind non-empty checks. The same discipline is extended to all new fields: a tags-only or scale-only PATCH that omits `redisConfiguration`/`enableNonSslPort`/etc. leaves the stored values untouched (nil map / nil `*bool` / empty string = "not supplied").

## Shared-driver safety

Fields are additive to the shared `services/cache/driver`. AWS ElastiCache and GCP Memorystore simply do not set them; their test suites pass unchanged.

## Tests (real armredis SDK over httptest)

- `TestSDKAzureCacheRedisPropertiesRoundTrip` — create with redisConfiguration{maxmemory-policy, maxmemory-reserved, custom passthrough key} + enableNonSslPort=true + minimumTlsVersion=1.2 + publicNetworkAccess=Disabled + redisVersion=6.0; Get echoes all.
- `TestSDKAzureCacheScaleUpdatePreservesProperties` — a capacity-only PATCH preserves the previously-set redisConfiguration and enableNonSslPort (nil-mask).

## Gates

- `go build ./...`, `go vet`, scoped `go test` (server/azure/cache, providers/azure/cache, services/cache) all green.
- Spot-checked `providers/aws/elasticache` + `providers/gcp/memorystore` — unregressed.
- `go generate ./...` and `go run ./internal/compatgen` — zero diff.
- `golangci-lint --new-from-rev=origin/development` on changed packages — 0 new issues; gofmt clean.

### DIFF
--- a/providers/azure/cache/cache.go
+++ b/providers/azure/cache/cache.go
@@ -114,21 +114,26 @@ func (m *Mock) CreateCache(ctx context.Context, cfg driver.CacheConfig) (*driver
 	}
 
 	info := driver.CacheInfo{
-		Name:               cfg.Name,
-		Scope:              cfg.Scope,
-		Location:           cfg.Location,
-		NodeType:           nodeType,
-		Engine:             engine,
-		Status:             "Running",
-		Endpoint:           endpoint,
-		CreatedAt:          m.opts.Clock.Now().UTC().Format(time.RFC3339),
-		Tags:               tags,
-		SKUFamily:          cfg.SKUFamily,
-		SKUCapacity:        cfg.SKUCapacity,
-		ShardCount:         cfg.ShardCount,
-		ReplicasPerPrimary: cfg.ReplicasPerPrimary,
-		PrimaryKey:         generateAccessKey(),
-		SecondaryKey:       generateAccessKey(),
+		Name:                cfg.Name,
+		Scope:               cfg.Scope,
+		Location:            cfg.Location,
+		NodeType:            nodeType,
+		Engine:              engine,
+		Status:              "Running",
+		Endpoint:            endpoint,
+		CreatedAt:           m.opts.Clock.Now().UTC().Format(time.RFC3339),
+		Tags:                tags,
+		SKUFamily:           cfg.SKUFamily,
+		SKUCapacity:         cfg.SKUCapacity,
+		ShardCount:          cfg.ShardCount,
+		ReplicasPerPrimary:  cfg.ReplicasPerPrimary,
+		RedisConfiguration:  cloneStringMap(cfg.RedisConfiguration),
+		EnableNonSSLPort:    cfg.EnableNonSSLPort,
+		MinimumTLSVersion:   cfg.MinimumTLSVersion,
+		PublicNetworkAccess: cfg.PublicNetworkAccess,
+		RedisVersion:        cfg.RedisVersion,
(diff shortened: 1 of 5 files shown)


NitinKumar004 (commented):
Reviewed at pushed SHA e0b554f4. Clean-built and gate matrix green.

Gates
- go build ./... and cache-package go vet clean. (Pre-existing, unrelated vet failure in providers/azure/aks/aks_test.go:76 — *int32 literal — exists on origin/development and is untouched by this PR; not a regression here.)
- go test ./server/azure/cache/... ./providers/azure/cache/... ./services/cache/... pass; -race on providers/azure/cache pass.
- gofmt clean; golangci-lint --new-from-rev=origin/development on the two changed cache trees: 0 issues.
- go generate ./... and go run ./internal/compatgen: zero diff, compatgen exit 0.

Focus findings
- A (shared driver additive): CacheConfig/CacheInfo gains are purely additive; AWS ElastiCache + GCP Memorystore provider and server tests all pass. No regression.
- B (round-trip): create with redisConfiguration (typed + custom passthrough key), enableNonSslPort, minimumTlsVersion=1.2, publicNetworkAccess=Disabled, redisVersion echoes on Get. redisVersion honored via toRedisJSON (defaults only when empty) — not hardcoded. Custom passthrough key survives (flattened into the ARM redisConfiguration map, decoded as map[string]string).
- C (PATCH nil-mask): PATCH routes through createOrUpdateCache -> UpdateCache. Nil-map / nil-*bool / empty-string guards mean a scale-only or tags-only PATCH preserves redisConfiguration + enableNonSslPort + the other new fields. TestSDKAzureCacheScaleUpdatePreservesProperties confirms. Explicit enableNonSslPort=false is representable (the *bool distinguishes false from omitted); wire decode of {"enableNonSslPort":false} yields a non-nil pointer that the driver applies.
- D (map clone): cloneStringMap clones in both CreateCache and UpdateCache; nil stays nil. No aliasing between caches sharing a source config map.

Non-blocking note: no explicit provider-level unit test for the no-aliasing invariant or for the explicit enableNonSslPort=false PATCH path; both are correct by construction and covered indirectly by the SDK server tests. Optional to add.

Verdict: MERGE.

PR #773, opened 2026-08-26
fix(gcp-vpc): 409 on duplicate subnet/firewall, PATCH network/subnet/firewall, firewall egress/advanced round-trip

NitinKumar004:
## Summary

Fixes five GCP VPC networking bugs in the wire handler (`server/gcp/vpc/`), all mirroring the existing networks/addresses/routes/routers patterns. Purely additive — no driver-interface or provider changes (PATCH reuses the existing `UpdateVPCTags`/`UpdateSubnetTags`/`UpdateSecurityGroupTags` methods already on the `Networking` interface).

### Bugs fixed
- **BUG1 (HIGH):** duplicate subnetwork insert (same name+region) now returns `409 alreadyExists` instead of silently shadowing the first (get/delete only ever resolved the first match, leaking the shadow). Mirrors `insertNetwork`'s dup-check.
- **BUG2 (HIGH):** duplicate firewall insert now returns `409 alreadyExists`.
- **BUG3 (MED-HIGH):** added PATCH/PUT for networks (routingMode/mtu/description), subnetworks (privateIpGoogleAccess/secondaryIpRanges/description) and firewalls (merge-patch of allowed/denied/sourceRanges/etc), mirroring `patchRouter`. Update-in-place previously 405'd. The subnet `fingerprint` is now **enforced**: a stale fingerprint on patch returns `412 conditionNotMet`.
- **BUG4 (MED):** firewall EGRESS/advanced fields (`destinationRanges`, `sourceTags`, `sourceServiceAccounts`, `targetServiceAccounts`, `logConfig`, `disabled`) now round-trip instead of being dropped — the spec is stored verbatim as JSON.
- **BUG5 (MED):** a minimal firewall now reads back with GCP defaults — `direction=INGRESS`, `priority=1000` — populated at insert.

### Tests
Real `cloud.google.com/go/compute` SDK over `httptest`:
- duplicate subnet (same region) → 409; duplicate firewall → 409
- PATCH network(routingMode) / subnet(privateIpGoogleAccess) / firewall(allowed) apply and GET reflects
- subnet patch with stale fingerprint → 412
- EGRESS firewall `destinationRanges` + `disabled` + `sourceTags` round-trip
- minimal firewall reads `direction=INGRESS`, `priority=1000`

### Gates
`go build ./...`, `go vet`, scoped `go test`, `gofmt`, `golangci-lint --new-from-rev` (0 new) all green. `go generate ./...` and `go run ./internal/compatgen` produce zero diff.

### DIFF
--- a/server/gcp/vpc/handler.go
+++ b/server/gcp/vpc/handler.go
@@ -70,6 +70,9 @@ const (
 	defaultSubnetPurpose = "PRIVATE"
 	defaultStackType     = "IPV4_ONLY"
 	internalNetCIDR      = "10.0.0.0/16"
+
+	defaultFirewallDirection = "INGRESS"
+	defaultFirewallPriority  = 1000
 )
 
 // nowRFC3339 returns the current time formatted the way GCP stamps
@@ -168,6 +171,8 @@ func (h *Handler) routeNetworks(w http.ResponseWriter, r *http.Request, rp gcpre
 	switch r.Method {
 	case http.MethodGet:
 		h.getNetwork(w, r, rp)
+	case http.MethodPatch, http.MethodPut:
+		h.patchNetwork(w, r, rp)
 	case http.MethodDelete:
 		h.deleteNetwork(w, r, rp)
 	default:
@@ -214,6 +219,8 @@ func (h *Handler) routeSubnetworks(w http.ResponseWriter, r *http.Request, rp gc
 	switch r.Method {
 	case http.MethodGet:
 		h.getSubnetwork(w, r, rp)
+	case http.MethodPatch, http.MethodPut:
+		h.patchSubnetwork(w, r, rp)
 	case http.MethodDelete:
 		h.deleteSubnetwork(w, r, rp)
 	default:
@@ -239,6 +246,8 @@ func (h *Handler) routeFirewalls(w http.ResponseWriter, r *http.Request, rp gcpr
 	switch r.Method {
 	case http.MethodGet:
 		h.getFirewall(w, r, rp)
+	case http.MethodPatch, http.MethodPut:
+		h.patchFirewall(w, r, rp)
 	case http.MethodDelete:
 		h.deleteFirewall(w, r, rp)
 	default:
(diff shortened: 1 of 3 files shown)


NitinKumar004 (commented):
Reviewed at pushed SHA 7d6a2dd. Gates all green: build, vet, `go test` (server+providers vpc), `-race` (server vpc), gofmt clean, golangci-lint `--new-from-rev` 0 new, `go generate` + compatgen both zero-diff. Only 3 files changed (handler.go, types.go, patch_test.go).

Verified against focus areas:

A. Dup-409 / no shadow — insertSubnetwork and insertFirewall now return 409 alreadyExists BEFORE any resolve/create, so the existing resource is never mutated (no shadow leak). Subnet dup-check keys on findSubnetByName(name, region) which filters by AvailabilityZone==region, so a same-name subnet in a different region is still allowed. Networks/addresses/routes/routers dup-checks are unchanged (separate untouched files / pre-existing block).

B. PATCH + fingerprint — patchNetwork/patchSubnetwork/patchFirewall wired for both PATCH and PUT. Subnet patch is fingerprint-guarded via fingerprintOf(name, CIDR), which is exactly what GET stamps, so a stale fingerprint -> 412 conditionNotMet and the read-back value applies. Firewall patch is a true merge-patch: mergeFirewallScalars/Collections overwrite only non-empty scalars and non-nil slices/pointers, so an omitted field is preserved (no nil-mask). Confirmed sourceRanges survives an allowed-only patch.

C. Egress/advanced round-trip — firewallSpec + request/response now carry destinationRanges, sourceTags, sourceServiceAccounts, targetServiceAccounts, logConfig, disabled; toFirewallResponse maps them all. EGRESS + destinationRanges + disabled + sourceTags round-trip on GET (the data-loss fix).

D. Defaults — minimal firewall reads back direction=INGRESS, priority=1000, stamped at insert.

E. No regression to LRO/list/aggregatedList/expandIpCidrRange (untouched).

Tests use the real GCP compute SDK REST clients and cover A-D directly. No blocking issues.

PR #774, opened 2026-08-26
fix(gcp-cloudsql): apply user password update, dataDiskType patch, deletionProtection delete-guard, PUT full-replace semantics

NitinKumar004:
## Summary

Fixes four confirmed GCP Cloud SQL parity bugs, all in the Cloud SQL wire handler, the in-memory provider, and the shared relationaldb driver.

### BUG1 — users.update password change was a silent no-op
`users.update` returned a DONE operation while never applying the new password. The driver `User` had no `Password` field and the provider `UpdateUser` only touched `Host`. Added a write-only `Password` field to `relationaldb/driver.User` and apply it in `CreateUser`/`UpdateUser`. The password is never surfaced on a read (`toWireUser` still omits it), matching Cloud SQL/AWS/Azure semantics.

### BUG2 — settings.dataDiskType patch double-dropped
A patched `dataDiskType` never reached the instance (the wire layer never mapped it onto the modify input, and the backend never applied `input.StorageType`), so a later Get returned the stale value. Now mapped in `modifyInputFromBody` and applied in `ModifyInstance` (guarded non-empty, so patch-preserve holds).

### BUG3 — deletionProtectionEnabled not modeled
`settings.deletionProtectionEnabled` was not on the wire settings, so a protected instance still deleted. Modeled it on the wire settings (pointer, so absent = no-change on patch) and the driver, populate on insert/patch, and guard `DeleteInstance` — returning HTTP 400 FAILED_PRECONDITION when protection is on. Also corrected the wire error mapping so FAILED_PRECONDITION is 400 (Google's canonical mapping), not 409.

### BUG4 — instances.update (PUT) behaved as merge, not full-replace
PUT and PATCH both routed to the merge handler. PUT now uses full-replace semantics: omitted scalar settings (tier, dataDiskSizeGb, dataDiskType, availabilityType, deletionProtectionEnabled, userLabels) revert to their defaults, while PATCH continues to merge.

## Tests
Real `sqladmin` SDK over `httptest`:
- dataDiskType patch reflected on Get (and preserved by an unrelated patch)
- deletionProtectionEnabled=true blocks delete with HTTP 400; clearing it allows delete
- PUT full-replace vs PATCH merge distinction (omitted disk settings revert on PUT, preserved on PATCH)

Provider-level assertion that a user password update is actually applied (not just DONE).

## Notes
- The `Password` field on the shared `relationaldb/driver.User` is additive; RDS/Redshift/Azure-SQL consumers leave it empty and are unaffected (spot-checked `providers/aws/rds` + `providers/azure/sql` tests green).
- PUT full-replace reverts scalar settings; the opaque `databaseFlags`/`backupConfiguration`/`ipConfiguration` blobs remain merge-only (present overwrites, absent no-change).

### DIFF
--- a/providers/gcp/cloudsql/cloudsql.go
+++ b/providers/gcp/cloudsql/cloudsql.go
@@ -277,6 +277,7 @@ func (m *Mock) newInstance(cfg rdsdriver.InstanceConfig) rdsdriver.Instance {
 		Port:               port,
 		State:              rdsdriver.StateAvailable,
 		MultiAZ:            cfg.MultiAZ,
+		DeletionProtection: cfg.DeletionProtection,
 		PubliclyAccessible: cfg.PubliclyAccessible,
 		VPCSecurityGroups:  append([]string(nil), cfg.VPCSecurityGroups...),
 		SubnetGroupName:    cfg.SubnetGroupName,
@@ -396,6 +397,14 @@ func (m *Mock) ModifyInstance(
 		inst.AllocatedStorage = input.AllocatedStorage
 	}
 
+	if input.StorageType != "" {
+		inst.StorageType = input.StorageType
+	}
+
+	if input.DeletionProtection != nil {
+		inst.DeletionProtection = *input.DeletionProtection
+	}
+
 	if input.EngineVersion != "" {
 		// Cloud SQL's databaseVersion is stored in Engine (that is the field
 		// toSQLInstance renders as databaseVersion), so a patched databaseVersion
@@ -445,6 +454,13 @@ func (m *Mock) DeleteInstance(ctx context.Context, id string) error {
 	inst, ok := m.instances.Get(id)
 	if !ok {
 		return cerrors.Newf(cerrors.NotFound, "Cloud SQL instance %q not found", id)
+	}
+
+	// Deletion protection blocks the delete until the guard is cleared, matching
+	// real Cloud SQL (settings.deletionProtectionEnabled).
+	if inst.DeletionProtection {
+		return cerrors.Newf(cerrors.FailedPrecondition,
+			"Cloud SQL instance %q has deletion protection enabled", id)
 	}
 
 	// Tear down the real database backing the instance, if any.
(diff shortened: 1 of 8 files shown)


NitinKumar004 (commented):
Reviewed at pushed SHA 3b3cf68b3cd2110db3087a2555858fcc38dde44e (HEAD confirmed). Verdict: MERGE.

- go build ./... clean; go vet clean.
- go test ./server/gcp/cloudsql/... ./providers/gcp/cloudsql/... ./services/relationaldb/... PASS.
- go test -race ./providers/gcp/cloudsql/... PASS.
- RDS + Azure-SQL regression: go test ./providers/aws/rds/... ./providers/azure/sql/... PASS.
- gofmt clean; golangci-lint --new-from-rev=origin/development 0 new issues.
- go generate zero diff; compatgen exit 0, docs/compat + compat/ zero diff.

Focus findings:
A. Shared driver.User.Password: additive + write-only. toWireUser (server/gcp/cloudsql/subresources.go:326) builds the wire user WITHOUT Password, so it is never echoed on Get/List. RDS/Azure-SQL never populate the field (stays "", omitempty) and no wire read path echoes it - no regression.
B. users.update applies password: UpdateUser now sets user.Password on non-empty (was a silent no-op). Verified applied at driver level (cloudsql_test.go:534) and still stripped on the wire.
C. dataDiskType: mapped in wire (StorageType) and applied in ModifyInstance guarded by != ""; Get reflects it; unrelated patch preserves it (SDK test).
D. deletionProtectionEnabled: modeled as *bool, populated on insert/patch, always emitted on Get; DeleteInstance returns FAILED_PRECONDITION -> HTTP 400 when true.
E. PUT vs PATCH: PATCH preserves omitted scalars (all ModifyInstance applies guarded by non-empty/non-nil), PUT reverts omitted scalar settings to backend defaults via orDefault*. No data-loss regression to the existing correct patch behavior. Opaque blobs (databaseFlags/backupConfiguration/ipConfiguration) remain merge-only on PUT (noted acceptable).

Observation (non-blocking): the writeErr FAILED_PRECONDITION mapping changed 409 -> 400 globally for cloudsql, so restart/transition/failover/promote precondition errors also now return 400. This matches Google's canonical error mapping (FAILED_PRECONDITION -> 400) and all existing tests pass, so it is a correctness improvement, but it is broader than the delete-guard scope in the title - worth a mention.

No blocking issues.

PR #776, opened 2026-08-26
fix(gcp-gce): list filter match-all + labels filter, enforce update fingerprints (412), setMachineType requires TERMINATED

NitinKumar004:
## Summary

Fixes three real-user GCP Compute Engine bugs in the wire handler, confirmed against real `cloud.google.com/go/compute` SDK behavior. All scoped to `server/gcp/compute/`.

### BUG3 [MED-HIGH] — `list --filter` on an unknown field returned an empty list
`filterField` returned `""` for any field it did not model (name/status/machineType/zone), so the predicate compared `"" == value` and excluded **every** instance. `gcloud compute instances list --filter="labels.env=prod"` returned nothing.
- Unrecognized field now **matches all** (mirrors `gcprest.NameMatches` leniency) instead of excluding all.
- Added `labels.<key>=<value>` filter support (reads `resp.Labels[key]`).
- `aggregatedListInstances` now applies the filter too (previously ignored it entirely).

### BUG1 [MED] — fingerprints not enforced on update (lost-update)
`setLabels` / `setMetadata` / `setTags` never checked the incoming fingerprint, so concurrent stale writes silently clobbered each other with no 412 path.
- Each setter now compares the incoming fingerprint (`LabelFingerprint` / `Fingerprint`) against the current computed one and returns **412 conditionNotMet** on mismatch.
- An empty incoming fingerprint skips the check (real GCP leniency).

### BUG2 [MED] — `setMachineType` succeeded on a RUNNING instance
Real GCP requires the instance to be TERMINATED (stopped) before its machine type can change.
- `setMachineType` now rejects a non-TERMINATED instance with **400**.
- Corrected the existing `TestSDKGCEInstanceUpdateVerbs`, which asserted the wrong success-on-running behavior — it now stops the instance first.

### Deferred: BUG4 [LOW] — unknown zone/global operation returns DONE instead of 404
A naive 404 in `serveOperations` breaks VPC and LoadBalancer compat: the compute handler is registered ahead of `vpc`/`loadbalancer` and its `Matches` claims **every** `operations` path, so it is the de-facto operations poller for the whole compute API space. 404ing "unregistered" names 404s siblings' real operations (reproduced: `compatgen` red cells in `compat/gcp` loadbalancer + networking). A correct fix needs a shared operations registry across compute+vpc+loadbalancer (mirroring `server/gcp/lro`), which is a cross-package refactor out of scope here. Deferred to a dedicated change.

## Tests (real compute SDK over httptest)
- `labels.env=prod` filter returns only matching; unknown-field filter matches all.
- `setLabels`/`setMetadata`/`setTags` with a stale fingerprint → 412; with the current one → ok.
- `setMachineType` on a running instance → 400; on a stopped instance → ok.

## Gates
- `go build ./...`, `go vet`, `go test -p=3 ./server/gcp/compute/... ./providers/gcp/compute/...` green
- `gofmt` clean; `golangci-lint --new-from-rev=origin/development ./server/gcp/compute/...` 0 new issues
- `go generate ./...` and `go run ./internal/compatgen` produce zero diff (idempotent, suite green)

### DIFF
--- a/server/gcp/compute/instance_actions.go
+++ b/server/gcp/compute/instance_actions.go
@@ -27,6 +27,11 @@ func (h *Handler) setLabels(w http.ResponseWriter, r *http.Request, rp gcprest.R
 		return
 	}
 
+	if !fingerprintMatches(body.LabelFingerprint, labelFingerprintFor(userLabels(inst.Tags))) {
+		writeConditionNotMet(w, "labelFingerprint")
+		return
+	}
+
 	// Remove existing user labels that the new set drops.
 	var remove []string
 
@@ -54,6 +59,11 @@ func (h *Handler) setMetadata(w http.ResponseWriter, r *http.Request, rp gcprest
 		return
 	}
 
+	if !fingerprintMatches(body.Fingerprint, metadataResponse(decodeMetadata(inst.Tags)).Fingerprint) {
+		writeConditionNotMet(w, "metadata fingerprint")
+		return
+	}
+
 	h.applyMutation(w, r, rp, inst.ID, "setMetadata",
 		map[string]string{keyMetadata: encodeJSON(body.Items)}, nil, "")
 }
@@ -73,6 +83,11 @@ func (h *Handler) setTags(w http.ResponseWriter, r *http.Request, rp gcprest.Res
 		return
 	}
 
+	if !fingerprintMatches(body.Fingerprint, fingerprint(strings.Join(decodeNetTags(inst.Tags), ","))) {
+		writeConditionNotMet(w, "tags fingerprint")
+		return
+	}
+
 	h.applyMutation(w, r, rp, inst.ID, "setTags",
 		map[string]string{keyNetTags: encodeJSON(body.Items)}, nil, "")
 }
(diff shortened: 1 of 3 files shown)


NitinKumar004 (commented):
Reviewed pushed SHA c433a7e (clean checkout, HEAD confirmed). Verdict: MERGE.

Gates (GOMAXPROCS=3, -p=3): go build clean; go vet clean; go test ./server/gcp/compute/... ./providers/gcp/compute/... PASS; go test -race ./providers/gcp/compute/... PASS; gofmt clean on all 3 files; golangci-lint --new-from-rev=origin/development 0 new issues; go generate zero diff; compatgen exit 0 (aws/azure/gcp all green) with zero docs/compat diff. Only 3 files changed as expected.

A. List filter — filterField now returns (value, known); an unknown field returns known=false and the predicate returns true (match ALL) instead of "" excluding everything. labels.<key> is modeled (known=true, resp.Labels[key]) so labels.env=prod filters correctly; a modeled filter (status=RUNNING) still filters. aggregatedListInstances now applies the same predicate. Confirmed unknown fields no longer empty results.

B. Fingerprint enforcement — setMetadata/setTags/setLabels reject a stale fingerprint with 412 conditionNotMet; empty incoming skips (GCP leniency). Round-trip verified: the enforced fingerprint is computed with the exact same helpers the response builder uses (labelFingerprintFor(userLabels), fingerprint(join(decodeNetTags)), metadataResponse(decodeMetadata).Fingerprint), so a fingerprint read from a GET matches on write and a stale one fails.

C. setMachineType — rejects with 400 when gcpStatusFor(state) != TERMINATED; applies once stopped. The corrected TestSDKGCEInstanceUpdateVerbs legitimately Stops the instance before SetMachineType (reflecting the new running->400 rule) and still asserts the machineType updates afterward — not hiding a regression. Its earlier SetLabels/SetTags omit fingerprints (empty -> skip), so they are unaffected.

D. No regression to instance CRUD/list-pagination/LRO — existing tests green; diff scoped to filter + update-verb preconditions.

Nit (non-blocking): the 400 setMachineType uses error reason string "conditionNotMet"; real GCP uses a different reason for the stopped-required 400. Cosmetic only; status code and message are correct.

PR #777, opened 2026-08-26
fix(gcp-gke): honor initialNodeCount=0, cluster labelFingerprint round-trip+enforcement, node-version propagation to pools, scope operation claims

NitinKumar004:
## Summary

Fixes four GKE control-plane fidelity bugs surfaced by a real-user (container SDK) audit. All in `server/gcp/gke/` + `providers/gcp/gke/`.

### BUG1 — `initialNodeCount=0` clobbered to 1 on create
`nodePoolFromSpec` and the default-pool bootstrap defaulted a `0` count to 1, so autoscale-from-zero pools read back 1 and produced a perpetual `google_container_node_pool` diff in Terraform. `InitialNodeCount` is now a `*int64` end to end (wire → provider spec), so an explicit `0` (esp. with `autoscaling.minNodeCount=0`) persists while a genuinely absent field still defaults to the GKE default. Consistent with `setSize 0`, which already worked.

### BUG2 — cluster `labelFingerprint` missing + not enforced
Added `LabelFingerprint` to the cluster model, computed deterministically from the (sorted) resource labels and returned on cluster GET/List. `setResourceLabels` now enforces optimistic concurrency: a non-empty, stale fingerprint fails with `FAILED_PRECONDITION` (HTTP 409); an empty incoming fingerprint skips the check. The fingerprint changes whenever labels change.

### BUG3 — operations list/`:cancel` over-claimed foreign ops
`Matches` now claims a named operation (GET poll or `:cancel`) only when the GKE mock actually recorded it (`HasOperation`), so foreign-service operations fall through to the shared LRO handler instead of being shadowed/`404`'d. The unnamed collection GET (`operations.list`) still stays with GKE.

### BUG4 — cluster `desiredNodeVersion` not propagated to node pools
`UpdateCluster` now rolls the version of the targeted node pool(s) when `desiredNodeVersion` is set — scoped to one pool via `desiredNodePoolId`, else all pools — matching real GKE, where a cluster node-version update propagates to the pools it upgrades.

## Tests (real container SDK over httptest)
- `initialNodeCount=0` node pool and default pool → GET returns 0
- cluster GET returns `labelFingerprint`; stale fingerprint → 409; current fingerprint accepted and rotates
- `desiredNodeVersion` update rolls the pool version
- handler-level `Matches` scoping for own/foreign op GET + cancel + list
- provider-level coverage for pointer semantics, fingerprint enforcement, and version propagation (incl. scoped `NodePoolID`)

`go generate ./...` and `go run ./internal/compatgen` are zero-diff.

### DIFF
--- a/providers/gcp/gke/gke.go
+++ b/providers/gcp/gke/gke.go
@@ -13,6 +13,7 @@ import (
 	"context"
 	"fmt"
 	"hash/fnv"
+	"sort"
 	"sync"
 	"time"
 
@@ -73,6 +74,7 @@ type Cluster struct {
 	NetworkPolicy     bool
 	MasterUsername    string
 	ResourceLabels    map[string]string
+	LabelFingerprint  string // opaque hash of ResourceLabels; computed on read.
 	MaintenanceWindow string // RFC-3339 daily window encoding; empty = none.
 	IPRotationActive  bool
 	NodePoolNames     []string
@@ -249,12 +251,15 @@ func (m *Mock) recordOperation(opType, location, target string) Operation {
 
 // CreateClusterInput captures the subset of CreateCluster we honor.
 type CreateClusterInput struct {
-	Name              string
-	Location          string
-	Description       string
-	Network           string
-	Subnetwork        string
-	InitialNodeCount  int64
+	Name        string
+	Location    string
+	Description string
+	Network     string
+	Subnetwork  string
+	// InitialNodeCount is a pointer so an explicitly-requested 0 (autoscale
+	// from zero) is distinguishable from an absent field. nil = unset →
+	// bootstrap the GKE default; non-nil (including *0) is honored verbatim.
+	InitialNodeCount  *int64
 	LoggingService    string
 	MonitoringService string
 	ResourceLabels    map[string]string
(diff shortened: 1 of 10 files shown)


NitinKumar004 (commented):
MERGE — all gates green on pushed SHA f5c5936.

Verified clean-build on the pushed commit: go build ./..., go vet, targeted tests (server/gcp/gke, providers/gcp/gke, server/gcp/cloudasset, providers/gcp/...), -race on providers/gcp/gke, gofmt clean, 0 new golangci-lint issues (--new-from-rev=origin/development), go generate zero-diff, compatgen exit 0 (aws/azure/gcp suites green) with zero docs/compat diff.

A. initialNodeCount *int64 — pointer threaded end to end (wire gkeCluster/gkeNodePool -> CreateClusterInput/NodePoolSpec -> nodePoolFromSpec). Explicit 0 persists for both the default pool and an added pool; nil (absent) still backfills defaultNodeCount. Covered by provider tests and real container/v1 SDK parity tests (ForceSendFields reproduces the Terraform path). Cross-package consumers: no non-test code outside the gke packages reads InitialNodeCount; cloudasset/kubernetes_discovery changes are mechanical test-callsite updates passing real values — no nil-deref.

B. labelFingerprint — deterministic fnv-64a over sorted key=value pairs, returned on GetCluster/ListClusters. setResourceLabels enforces optimistic concurrency: stale -> FAILED_PRECONDITION (HTTP 409, SDK-verified), empty -> skip, and the fingerprint rotates when labels change. The hand-rolled lock/get/set/recordOperation faithfully mirrors mutateCluster.

C. Operation-claim scoping — Matches now claims a named op (GET poll or :cancel) only when HasOperation is true; foreign-service ops fall through to the shared LRO handler. Narrow fix: the GET path was already HasOperation-gated; only the action (:cancel) over-claim is corrected. GKE's own polling stays intact (own GET + own :cancel both still match, per matches_test).

D. node-version propagation — UpdateCluster desiredNodeVersion rolls pool Version after the cluster mutation (separate lock, no re-entrancy/deadlock); desiredNodePoolId scopes the roll to one pool. Verified at provider and SDK level.

E. No AKS-analog pool-wipe introduced; SetNodePoolSize still allows scale-to-zero (count=0, no floor) and LRO recording is unchanged.

PR #778, opened 2026-08-26
fix(azure-monitor): PATCH on insights resources, fire action groups on alarm breach, model autoscaleSettings

NitinKumar004:
## Summary

Fixes the Azure Monitor (`microsoft.insights`) wire handler and provider so real `armmonitor` SDK clients, `az monitor autoscale`, and `azurerm_monitor_autoscale_setting` work.

### BUG1 — PATCH returned 405 on every insights resource
`ServeHTTP` only switched PUT/GET/DELETE, so `MetricAlerts.Update`, `ActionGroups.Update`, and `ActivityLogAlerts.Update` (all HTTP PATCH) failed. Added a PATCH handler that loads the stored resource and merges non-nil properties + tags over it under nil-mask semantics (omitted fields preserved), re-runs the alarm/action-group side-effect wiring against the merged properties, and returns 200. PATCH on a missing resource is a 404 (real ARM Update contract).

### BUG2 — alarm breach did not fire the action group
Action groups were stored only as opaque ARM resources, disconnected from alarm evaluation. Now the handler registers an action group's receivers into the provider on create/update (and unregisters on delete). On an OK->ALARM transition, `transitionAlarm` resolves each `AlarmActions` id against the registered action groups and delivers to every receiver (email/webhook/sms/azureFunction), recording each delivery so callers can assert it — mirroring the AWS alarm->SNS action wiring. Webhook receivers are additionally POSTed for real when a `WebhookDeliverer` is wired. Clock-driven via `opts.Clock`.

### GAP-A — autoscaleSettings were entirely unserved (501)
Added `autoscaleSettings` as a served `canonicalType`; the generic create/get/list/delete/patch path round-trips the full `properties` (profiles, rules metricTrigger + scaleAction, capacity min/max/default).

### GAP-B — criteria dimensions + scopes ignored on evaluation
`registerAlarm` now maps `criteria.allOf[].dimensions` (Include single-value) onto `AlarmConfig.Dimensions` and pins evaluation to a single `properties.scopes` entry via a `resourceId` dimension, so an alert scoped to one resource no longer fires on another resource's datapoints.

## Tests (real `armmonitor` SDK over httptest)
- PATCH on metricAlert/actionGroup/activityLogAlert returns 200, applies supplied fields, preserves omitted ones.
- Alarm breach fires the action group: asserts recorded email+webhook deliveries and a real webhook receiver hit, clock-driven.
- autoscaleSetting create -> GET round-trip with profiles/rules/capacity intact.
- Dimension/scope-scoped alert does not fire on another resource's data, then fires on its own.

## Gates
- `go build ./...`, `go vet`, scoped `go test` green; `gofmt` clean; `golangci-lint --new-from-rev=origin/development` 0 new issues.
- `go generate ./...` and `go run ./internal/compatgen` both zero-diff.

### DIFF
--- a/providers/azure/monitor/actiongroups.go
+++ b/providers/azure/monitor/actiongroups.go
@@ -1,0 +1,218 @@
+package monitor
+
+import (
+	"context"
+	"strings"
+	"time"
+)
+
+// Receiver type discriminators recorded on a delivery.
+const (
+	receiverEmail         = "email"
+	receiverWebhook       = "webhook"
+	receiverSMS           = "sms"
+	receiverAzureFunction = "azureFunction"
+)
+
+// ActionGroupReceiver is one resolved receiver of an action group. Endpoint is
+// the receiver's address: the email address, webhook URI, phone number, or
+// Azure Function trigger URL, depending on Type.
+type ActionGroupReceiver struct {
+	Type     string
+	Name     string
+	Endpoint string
+}
+
+// ActionGroupDelivery records one action-group receiver fired by an alert
+// breach. It is the observable side effect that lets a caller (or a test)
+// assert an OK->ALARM transition actually reached the action group's receivers,
+// mirroring how an AWS CloudWatch alarm publishes to its SNS topic subscribers.
+type ActionGroupDelivery struct {
+	ActionGroupID string
+	ReceiverType  string
+	ReceiverName  string
+	Endpoint      string
+	AlarmName     string
+	NewState      string
+	Timestamp     time.Time
+}
+
+// actionGroupData is the stored, evaluation-side view of an action group: its
+// resource id and the receivers a breach delivers to.
+type actionGroupData struct {
+	ID        string
+	Enabled   bool
+	Receivers []ActionGroupReceiver
+}
+
+// WebhookDeliverer delivers an alert notification to a webhook receiver's URI.
+// It mirrors the AWS ActionPublisher wiring: nil (the default) records the
+// delivery without a network call, while a wired deliverer additionally
+// performs the real HTTP POST so an end-to-end webhook flow can be exercised.
+type WebhookDeliverer interface {
+	Deliver(ctx context.Context, uri, payload string) error
+}
+
+// SetWebhookDeliverer wires a webhook deliverer so an alert breach that targets
+// an action group with webhook receivers performs the real HTTP POST in
+// addition to recording the delivery. Nil leaves webhook delivery record-only.
+func (m *Mock) SetWebhookDeliverer(d WebhookDeliverer) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	m.webhookDeliverer = d
+}
+
+// RegisterActionGroup registers (or replaces) an action group's receivers under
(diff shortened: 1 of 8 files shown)


NitinKumar004 (commented):
Reviewed at pushed SHA a0badc38. Verdict: MERGE.

- build ./... + vet: clean
- go test ./server/azure/monitor/... ./providers/azure/monitor/... ./services/monitoring/...: ok
- go test -race ./providers/azure/monitor/...: ok
- gofmt: clean; golangci-lint --new-from-rev=origin/development: 0 new issues
- go generate ./...: zero diff; compatgen suite (aws/azure/gcp): green, zero docs/compat diff
- 8 files changed, as expected; worktree clean.

A. PATCH nil-mask (data-loss): mergeResource preserves omitted top-level fields (location/tags/properties); mergeAnyMap/mergeStringMap preserve base keys the overlay omits, on a fresh copy (stored map never mutated pre-commit). 404-before-write on a missing resource (store.get guard precedes any side effect). patch_test.go adversarially proves a severity+one-tag PATCH keeps description, criteria, scopes and the other tag across metricAlerts/actionGroups/activityLogAlerts. Solid.

B. Action-group firing: fires only on the OK/other -> ALARM transition (transitionAlarm: oldState != newState && newState == ALARM). Each AlarmActions id resolves case-insensitively against registered action groups; disabled groups deliver to none. Webhook POST is genuinely off-lock: PutMetricData unlocks before evaluateAlarms; recordDelivery takes/releases the lock, deliverWebhook reads the deliverer under a brief RLock then POSTs after releasing — no eval-lock held across the network call, no goroutine, no deadlock. Firing test asserts a real webhook hit (hits==1 via injectable poster) plus 2 recorded deliveries (email+webhook) at ALARM. No spurious fire: only-on-transition-into-ALARM, and the dimension-scoped test confirms vm2's breach does not drive vm1's alert.

C. autoscaleSettings: new canonicalType routes through the same generic store handlers (PUT/GET/LIST/DELETE/PATCH) with no side effects; switch is on the lowercased type so it does not shadow metricAlerts/actionGroups/activityLogAlerts routing. Round-trip test preserves profiles/rules (metricTrigger+scaleAction)/capacity.

D. dimensions/scopes eval: criterionDimensions maps an Include single-value filter to name->value (multi-value/Exclude intentionally skipped); a single scopes[0] pins resourceId so an alert ignores another resource's datapoints; multi-scope leaves resourceId unset (aggregate). Pre-existing metric/alarm eval and 200-on-create/update unregressed (full server suite passes).

E. Metrics emission + existing monitor tests: unregressed.

Minor (non-blocking):
- actionGroupIDs doc comment still says "Actual delivery ... is not simulated" — now stale; delivery IS simulated.
- Any PATCH on a metricAlert re-runs registerAlarm -> CreateAlarm, resetting alarm State to INSUFFICIENT_DATA. This mirrors existing PUT behavior (PUT already resets), so not a new regression, but a tags-only PATCH transiently clearing a firing alarm's state is a small parity quirk worth a future note.
- OKActions are never delivered (firing is ALARM-only). Safe side (no spurious OK fire); OK-action delivery remains an unimplemented gap.

PR #779, opened 2026-08-26
fix(gcp-gcs): resumable uploads, ranged GET (206), delimiter-list pagination, delete-bucket version guard, metadata-patch updated timestamp

NitinKumar004:
## Summary

Fixes five real-user Google Cloud Storage bugs found by a confirming audit. All changes are in `server/gcp/gcs/` and `providers/gcp/gcs/` (one shared additive field on `services/storage/driver.ObjectInfo`).

### Bugs fixed

1. **Resumable uploads (HIGH)** — objects >16 MiB failed with 400 because `upload()` only handled `uploadType=media|multipart`. The idiomatic `Object.NewWriter` / gsutil path is resumable by default for anything larger than the 16 MiB chunk size. Added a resumable branch: the `uploadType=resumable` POST initialises a session and returns 200 + a `Location` session URI; subsequent chunk uploads carry `Content-Range: bytes X-Y/total`, accumulate in the session buffer, and commit the object through the normal write path on the final chunk. Honors the SDK's `X-GUploader-No-308: yes` (200 + `X-Http-Status-Code-Override: 308`) for the resume-incomplete signal.

2. **Ranged download (HIGH)** — `alt=media` GETs ignored the `Range` header and always returned the full body with 200, so `NewRangeReader(offset>0, ...)` errored. Now parses `Range: bytes=`, emits 206 with `Content-Range` and the sub-slice, 416 on an unsatisfiable range, and sets `Accept-Ranges: bytes`. Mirrors the S3/Azure-blob range handlers.

3. **Delimiter listing pagination (MOD)** — common prefixes were computed over all keys and returned in full on every page, and were not counted toward `maxResults`. Prefixes and objects now fold into one sorted, deduped, lexicographic paginated stream: each prefix appears on exactly one page and counts toward the page size, with position carried in the page token.

4. **DeleteBucket version guard (MOD)** — a bucket holding only noncurrent (archived) versions deleted successfully. Real GCS returns 409 not-empty. Delete now also refuses when any archived version remains.

5. **Metadata-patch `updated` timestamp (MOD)** — a metadata-only patch left `updated` equal to `timeCreated`. Objects now track a distinct creation time; a patch advances `updated` (LastModified) while `timeCreated` stays fixed.

### Tests

Real `cloud.google.com/go/storage` client over `httptest`:
- >16 MiB multi-chunk resumable upload round-trips byte-for-byte
- `NewRangeReader` returns the exact requested slice (206) for mid-range, to-end, and suffix ranges
- delimiter + `MaxSize=1` listing returns each prefix once across pages
- `DeleteBucket` fails with noncurrent versions present

Provider-level (deterministic `FakeClock`):
- metadata patch bumps `updated`, not `timeCreated`
- DeleteBucket refuses a bucket with only noncurrent versions

Existing `TestPaginationTokens` updated to assert the corrected folded-pagination behavior.

### Gates
build ./..., vet, scoped `go test`, gofmt, and `golangci-lint --new-from-rev=origin/development` (0 new) all pass. `go generate ./...` and `go run ./internal/compatgen` produce no diff.

### DIFF
--- a/providers/gcp/gcs/delete_and_timestamp_semantics_test.go
+++ b/providers/gcp/gcs/delete_and_timestamp_semantics_test.go
@@ -1,0 +1,59 @@
+package gcs
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stackshy/cloudemu/v2/config"
+	"github.com/stackshy/cloudemu/v2/services/storage/driver"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestDeleteBucketRefusesNoncurrentVersions checks that a bucket with no live
+// objects but retained noncurrent versions still refuses deletion.
+func TestDeleteBucketRefusesNoncurrentVersions(t *testing.T) {
+	ctx := context.Background()
+	m := newTestMock()
+
+	require.NoError(t, m.CreateBucket(ctx, "b1"))
+	require.NoError(t, m.SetBucketVersioning(ctx, "b1", true))
+
+	require.NoError(t, m.PutObject(ctx, "b1", "k1", []byte("v1"), "text/plain", nil))
+	require.NoError(t, m.PutObject(ctx, "b1", "k1", []byte("v2"), "text/plain", nil))
+
+	// Deleting the live object archives the current generation as noncurrent, so
+	// no live object remains but a noncurrent version does.
+	require.NoError(t, m.DeleteObject(ctx, "b1", "k1"))
+
+	err := m.DeleteBucket(ctx, "b1")
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "not empty")
+}
+
+// TestUpdateObjectBumpsUpdatedNotCreated checks that a metadata-only patch
+// advances the object's updated time while leaving timeCreated fixed.
+func TestUpdateObjectBumpsUpdatedNotCreated(t *testing.T) {
+	ctx := context.Background()
+	m := newTestMock()
+	clk := m.opts.Clock.(*config.FakeClock)
+
+	require.NoError(t, m.CreateBucket(ctx, "b1"))
+
+	created, err := m.PutObjectGCS(ctx, "b1", "k1", []byte("data"), "text/plain", nil, nil, driver.GCSPrecondition{})
+	require.NoError(t, err)
+	require.Equal(t, created.Created, created.LastModified, "a fresh write has equal created and updated")
+
+	clk.Advance(time.Hour)
+
+	val := "bar"
+	patched, err := m.UpdateObjectGCS(ctx, "b1", "k1", driver.GCSObjectUpdate{
+		Metadata: map[string]*string{"foo": &val},
+	}, driver.GCSPrecondition{})
+	require.NoError(t, err)
+
+	assert.Equal(t, created.Created, patched.Created, "timeCreated must stay fixed on a metadata patch")
+	assert.Greater(t, patched.LastModified, created.LastModified, "updated must advance")
+	assert.NotEqual(t, patched.Created, patched.LastModified, "updated must differ from created after patch")
+}
--- a/providers/gcp/gcs/gcs.go
+++ b/providers/gcp/gcs/gcs.go
@@ -50,6 +50,7 @@ type gcsObject struct {
 	ContentType        string
 	ETag               string
 	LastModified       string
+	Created            string
 	Metadata           map[string]string
 	Tags               map[string]string
 	Generation         int64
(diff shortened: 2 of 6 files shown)


NitinKumar004 (commented):
Deep review of cf500f09 (GCS resumable upload / ranged GET / delimiter folding / delete-bucket guard / metadata-patch timestamps).

Gate matrix (on the pushed SHA): build, vet, gofmt clean. `go test ./server/gcp/gcs ./providers/gcp/gcs ./services/storage/...` green; `-race` green on both providers/gcp/gcs and server/gcp/gcs. Shared storage consumers unregressed (`server/aws/s3`, `server/azure/blobstorage` green) after the additive `ObjectInfo.Created` field. `go generate` + `go run ./internal/compatgen` produce zero diff and compatgen exits clean.

What I verified as correct:
- Resumable byte-integrity via the real cloud.google.com/go/storage Writer: 17MiB (default 16MiB chunking, 2 chunks) and an adversarial 256KiB ChunkSize run across 1B / 255KiB / 256KiB / 256KiB+1 / 1MiB+123 / 768KiB objects all reassemble byte-for-byte. Duplicate chunk resend (retry) is correctly rejected by the `start == len(buf)` guard, so no double-append. The 200 + X-Http-Status-Code-Override:308 resume signalling drives the SDK correctly.
- Ranged GET: mid-range, to-end, and suffix all return the exact slice (206). 416 path sets `Content-Range: bytes */total`. Unparseable header falls back to full 200.
- Delimiter list folding: objects carrying the delimiter are excluded from the object list and rolled into the prefix set (dedup by map), then merged into one lexicographically sorted stream and paginated by offset — each prefix appears exactly once with MaxSize=1, no dup/skip. Correct behavior, not a workaround.
- DeleteBucket refuses (409) when only noncurrent versions remain.
- Metadata patch bumps `updated` while `Created` (timeCreated) is preserved via `next := *cur`; overwrite/copy/compose correctly reset Created to the new generation's time.

Two findings:

1. (data-integrity, latent) Resumable finality condition can commit a truncated object on a chunk gap. In `uploadResumableChunk` (handler.go:744):
   `final := total != rangeTotalUnknown && (end == total-1 || buffered >= total)`
   A chunk whose `start != len(sess.buf)` is (correctly) not appended, but the `end == total-1` disjunct still marks the request final regardless of whether the buffer actually holds all bytes. Exact repro against the wire handler: init session; PUT `bytes 0-99/*` (100B); then PUT `bytes 200-299/300` (skipping 100-199). The server appends nothing on the second PUT (start 200 != buffered 100) yet commits, returning 200 with a 100-byte object for a 300-byte declared upload — silent truncation. The conforming Go SDK never produces this (strictly sequential, awaits each 308, resumes from the server-reported offset), so real-SDK uploads are safe today; but the reassembly logic shouldn't accept a short buffer as complete. Minimal fix: drop the `end == total-1` disjunct and gate solely on `buffered >= total` (which is only reached when every byte was appended; still true for the empty-object `bytes */0` and single-chunk cases). Optionally 400 when the final chunk's start leaves a gap.

2. (gate) One new golangci-lint issue vs origin/development: handler.go:601 comment "initialised" flagged by misspell (→ "initialized"). Violates the 0-new-lint gate.

Both fixes are trivial. Real-SDK data paths are proven correct; the finality-condition hole is a robustness/defense-in-depth correctness gap on the data-integrity-critical path plus the lint nit.

NitinKumar004 (commented):
Re-review of the delta (head c9111d23) — verifying only the two fixes; approved parts unchanged.

MERGE. All gates green on the pushed SHA (HEAD matches).

Fix 1 — resumable finality (data-integrity): CORRECT.
- uploadResumableChunk now gates finality solely on `!contiguous || total==unknown || buffered < total`; the old `end == total-1` disjunct is gone. `end` is dropped from the parse. A gapped chunk can never commit a short object.
- New appendChunk(start,data)->(buffered,contiguous): a bytes-carrying chunk with start > len(buf) (gap) appends nothing and returns contiguous=false; start == len(buf) appends; start < len(buf) (duplicate/overlap) and status-probes (no bytes) no-op and stay contiguous. No double-append. Adversarial cases hold: gap-then-fill resumes contiguously once the hole is filled; duplicate/overlap is not re-appended.
- TestResumableGappedChunkDoesNotTruncate drives init -> 'bytes 0-99/*' -> 'bytes 200-299/300' and asserts the gapped chunk is resume-incomplete (200 + X-Http-Status-Code-Override: 308) and the object never materializes. Passes.

Fix 2 — happy path byte-identical: CONFIRMED. TestSDKGCSResumableUpload (17MiB, real SDK Writer) assembles byte-for-byte; the final chunk where buffered reaches total still commits. Passes (incl. -race).

Fix 3 — lint: 'initialised'->'initialized'; header comparison uses canonical 'X-Guploader-No-308' in Header.Get (Go canonicalizes, so SDK's 'X-GUploader-No-308' still matches — runtime interop unchanged).

Approved-unchanged spot-check: the delta commit touches only handler.go (finality + lint) and the test file; ranged GET / list folding / delete-bucket guard / timestamps have no diff.

Gates: build + vet clean; go test and -race green (server/gcp/gcs, providers/gcp/gcs); gofmt -l clean; golangci-lint --new-from-rev=origin/development = 0 issues; go generate zero diff; compatgen exit 0, zero diff.

PR #780, opened 2026-08-26
fix(aws-s3): paginate ListObjectsV2 keys+CommonPrefixes as one MaxKeys-capped stream (no repeated prefixes)

NitinKumar004:
## Summary

`ListObjectsV2`/`ListObjects` with a delimiter repeated every `CommonPrefixes` entry on every page and did not count prefixes toward `MaxKeys`. The provider paginated only the matched object keys while returning the full `commonPrefixes` slice each page, so a delimited listing over a large bucket returned each rolled-up prefix once per page and could exceed `MaxKeys`.

Real S3 caps object keys and rolled-up common prefixes **jointly** by `MaxKeys` over one lexicographically-sorted stream: each prefix appears on exactly one page, no page exceeds `MaxKeys`, and truncation is carried by the continuation token.

## Change

- `providers/aws/s3/s3.go` — `ListObjects` now merges matched object keys and common prefixes into a single lexicographically-sorted stream (`mergeListEntries` / `listEntry`) and paginates that combined stream against `MaxKeys`, carrying the offset in the continuation token. The returned page is split back into `Objects` and `CommonPrefixes`. This mirrors the joint key+prefix treatment already used on the `ListObjectVersions` path. Non-delimited (keys-only) pagination is unchanged in behavior.
- The server handler already computed `KeyCount = len(Objects)+len(CommonPrefixes)` and echoed the provider's prefixes/token, so no handler change was needed — it now yields correct per-page values.

## Tests (real aws-sdk-go-v2)

- `server/aws/s3/list_commonprefixes_pagination_test.go`
  - 2000 keys + 3 common prefixes, delimiter `/`, MaxKeys 1000 via `ListObjectsV2Paginator`: every key and every prefix returned exactly once, no page exceeds MaxKeys, `KeyCount` counts keys+prefixes.
  - >1000 common prefixes (1500 dirs): first page capped at 1000 with `IsTruncated` + continuation token; resume returns the rest with none repeated or dropped.
  - Keys-only pagination (2500 keys) unregressed: 3 pages, each key once.
- Updated `providers/aws/s3/s3_lifecycle_test.go` `TestPaginationTokens`, which previously asserted the buggy "prefixes never paginated" behavior, to assert correct joint pagination.

### DIFF
--- a/providers/aws/s3/s3.go
+++ b/providers/aws/s3/s3.go
@@ -728,27 +728,38 @@ func (m *Mock) ListObjects(_ context.Context, bucket string, opts driver.ListOpt
 
 	matchedObjects, commonPrefixSet := matchListKeys(bkt, allKeys, opts)
 
-	commonPrefixes := make([]string, 0, len(commonPrefixSet))
-	for p := range commonPrefixSet {
-		commonPrefixes = append(commonPrefixes, p)
-	}
-
-	sort.Strings(commonPrefixes)
-
 	maxKeys := opts.MaxKeys
 	if maxKeys <= 0 {
 		maxKeys = s3DefaultMaxKeys
 	}
 
-	page, err := pagination.Paginate(matchedObjects, opts.PageToken, maxKeys)
+	// Real S3 caps object keys and rolled-up common prefixes JOINTLY by
+	// MaxKeys over a single lexicographically-ordered stream: each prefix is
+	// returned on exactly one page, and truncation carries an offset in the
+	// continuation token. Merge the two into one ordered stream, paginate
+	// that, then split the returned page back into keys and prefixes.
+	entries := mergeListEntries(matchedObjects, commonPrefixSet)
+
+	page, err := pagination.Paginate(entries, opts.PageToken, maxKeys)
 	if err != nil {
 		return nil, cerrors.Newf(cerrors.InvalidArgument, "invalid page token: %v", err)
 	}
 
-	// Clone metadata only for the page actually returned — cloning every
-	// match would make a paged scan O(bucket) allocations per request.
+	objects := make([]driver.ObjectInfo, 0, len(page.Items))
+	commonPrefixes := make([]string, 0, len(page.Items))
+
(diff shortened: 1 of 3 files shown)


NitinKumar004 (commented):
Reviewed at pushed SHA 1e6cf75 (HEAD confirmed).

Verdict: MERGE — clean.

Gates (GOMAXPROCS=3, -p=3):
- go build ./... + go vet ./... — exit 0
- go test ./server/aws/s3/... ./providers/aws/s3/... ./services/storage/... — pass
- go test -race ./providers/aws/s3/... — pass
- gofmt -l (3 changed files) — clean
- golangci-lint --new-from-rev=origin/development — 0 new issues
- go generate ./... — zero diff (51 services)
- go run ./internal/compatgen — suite green + zero diff
- Exactly 3 files changed

Correctness:
- Joint MaxKeys cap: keys and rolled-up common prefixes are merged into one lexicographically-sorted stream (mergeListEntries) and paginated together, so no page's (Objects+CommonPrefixes) exceeds MaxKeys. Confirmed via SDK test (2000 keys + prefixes, MaxKeys=1000) — each key/prefix returned exactly once, KeyCount = keys+prefixes.
- No key/prefix collision: matchListKeys 'continue's a key that rolls up into a prefix, so it is excluded from matchedObjects. The two sets are disjoint, making the merged sort total and the offset continuation token stable across pages — an adversarial prefix sorting between two keys lands on the correct boundary deterministically (allKeys sorted, merged stream re-sorted each call).
- Truncation: 1500 prefixes / MaxKeys=1000 -> page1 capped 1000 + IsTruncated + NextContinuationToken; page2 returns remaining 500, none repeated/dropped. Tested.
- Keys-only path unregressed: 2500 keys -> 3 pages, every key once. Tested.
- ListObjectVersions untouched; the updated TestPaginationTokens now asserts real S3 joint-cap behavior (replacing the prior 'prefixes never paginated' survey workaround) — legit.

Metadata cloning preserved (page-only, no O(bucket) regression).

PR #781, opened 2026-08-26
fix(aws-ec2,rds): validate+apply DescribeInstances filters, atomic RunInstances ClientToken dedup, RDS unknown-filter error

NitinKumar004:
## Summary

Fixes three AWS data-integrity/fidelity issues found by a cross-check, all in the EC2/RDS describe-filter and RunInstances idempotency paths.

### BUG1 — DescribeInstances ignored unimplemented filters (over-inclusive)
`providers/aws/ec2` `matchesSingleFilter` match-alled any non-tag filter name it did not implement, so a `vpc-id`/`subnet-id`/etc. filter returned the whole fleet, and a truly-unknown filter name was silently accepted. Adopted the sibling `(matched, known)` pattern used by DescribeVolumes/DescribeSubnets:

- `instanceFilterMatch` now resolves scalar filters via `instanceFilterField` and reports whether each name is modeled.
- Implemented the common instance filters: `vpc-id`, `subnet-id`, `image-id`, `private-ip-address`, `private-dns-name`, `key-name`, `reservation-id`, `availability-zone`, `architecture`, `instance-lifecycle` (plus the existing `instance-id`/`instance-type`/`instance-state-name`), and `tag:`/`tag-key`. Derived values (private-dns-name, availability-zone, architecture) mirror what the wire layer renders so a filter agrees with DescribeInstances output.
- `DescribeInstances` now validates filter names up front and returns `InvalidArgument` (surfaced as `InvalidParameterValue`) for an unknown name.

### BUG2 — RunInstances ClientToken dedup was non-atomic under concurrency
The dup-check (RLock) and record (Lock) were separate lock acquisitions with the whole launch in between, so two concurrent same-token RunInstances could both provision. Made check-and-reserve atomic: the first caller reserves the token under a single lock hold and launches; concurrent callers with the same token find the in-flight reservation and wait on its result; a later retry finds the recorded ids. A token never provisions twice; sequential retries still return the same instances.

### BUG3 — RDS DescribeDBInstances unknown filter matched nothing
`instanceMatchesFilters` returned false for an unrecognized filter name, silently emptying the result. It now validates filter names and returns `InvalidParameterValue` for an unknown name, keeping the modeled `db-instance-id`/`engine`/`db-cluster-id` filters working.

## Tests
- EC2 (aws-sdk-go-v2): `vpc-id`/`subnet-id` filters return only the matching instance; unknown filter -> `InvalidParameterValue`.
- Provider unit: expanded filter-match table; concurrent same-ClientToken -> exactly one instance set (`-race`); sequential retry idempotency + distinct token.
- RDS (aws-sdk-go-v2): unknown filter -> `InvalidParameterValue`; `engine` filter still narrows.

## Gates
build ./..., vet, scoped `go test`, `go test -race ./providers/aws/ec2/...`, gofmt, and `golangci-lint --new-from-rev=origin/development` all green. `go generate ./...` and `go run ./internal/compatgen` produce zero diff.

### DIFF
--- a/providers/aws/ec2/ec2.go
+++ b/providers/aws/ec2/ec2.go
@@ -292,9 +292,9 @@ type Mock struct {
 	// releases it on terminate, so the VPC subnet / security-group delete guards
 	// see a running instance's interface. nil until wired by the provider.
 	networking Networking
-	// mu guards managedResourceVisibility, clientTokens and subnetIPCounters,
-	// which are scalar/map shared state that (unlike the memstores) has no
-	// internal locking of their own.
+	// mu guards managedResourceVisibility, clientTokens, clientTokenInflight and
+	// subnetIPCounters, which are scalar/map shared state that (unlike the
+	// memstores) has no internal locking of their own.
 	mu sync.RWMutex
 	// managedResourceVisibility is "visible" or "hidden". When "hidden",
 	// service-provider-managed instances are omitted from DescribeInstances
@@ -305,6 +305,11 @@ type Mock struct {
 	// of double-provisioning (AWS idempotency). Permanent — an emulator needs no
 	// expiry window.
 	clientTokens map[string][]string
+	// clientTokenInflight tracks ClientTokens whose launch is still provisioning,
+	// so two concurrent RunInstances carrying the same token cannot both provision
+	// (one reserves the token and launches; the others wait on its result). An
+	// entry is removed once the launch completes and its ids move to clientTokens.
+	clientTokenInflight map[string]*clientTokenLaunch
 	// subnetIPCounters is the per-subnet host counter used to allocate private
 	// IPv4 addresses from the subnet's CIDR range.
 	subnetIPCounters map[string]int
@@ -392,6 +397,7 @@ func New(opts *config.Options) *Mock {
 
 		managedResourceVisibility: visibilityVisible,
 		clientTokens:              make(map[string][]string),
+		clientTokenInflight:       make(map[string]*clientTokenLaunch),
 		subnetIPCounters:          make(map[string]int),
 	}
 }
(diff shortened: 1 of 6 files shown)


NitinKumar004 (commented):
Deep review — verdict: MERGE. Clean-built the pushed SHA 3889c11 (detached HEAD confirmed).

- go build ./... clean; go vet target pkgs clean.
- go test ./server/aws/ec2 ./providers/aws/ec2 ./server/aws/rds ./providers/aws/rds: all ok.
- go test -race ./providers/aws/ec2 (incl. TestRunInstancesClientTokenConcurrent x5): clean.
- gofmt: clean. golangci-lint --new-from-rev=origin/development on the three pkgs: 0 issues.
- go generate ./...: zero diff. compatgen (compat/aws|azure|gcp suites): exit 0, zero diff.

A. DescribeInstances filter validate+apply — correct. instanceFilterMatch returns (matched, known); validateInstanceFilters probes a zero instance and consults only `known`, so the accepted set cannot drift from what the matcher honors (mirrors the DescribeVolumes/DescribeSubnets pattern). vpc-id/subnet-id/image-id/private-ip-address/private-dns-name/key-name/reservation-id/availability-zone/architecture all filter to matching instances; an unknown name -> InvalidParameterValue (verified via real aws-sdk-go-v2 e2e test). tag:/tag-key both work. Derived values match the wire render exactly: instanceZone const us-east-1a == wire defaultZone (instances never populate driver.Instance.Zones), instanceArchitecture x86_64 == wire archX86 (rendered as a constant), private-dns-name format identical to server privateDNSName().

B. ClientToken atomicity — correct, no TOCTOU, no deadlock. Single m.mu hold reserves the token (check clientTokens, then clientTokenInflight, then insert launch) and releases before launchInstances (which self-locks). Waiters block on done and read ids/err only after close (happens-before via channel close). Error path leaves clientTokens unset so a retry re-attempts (matches real EC2). 8 concurrent same-token RunInstances -> exactly one instance in the backend, all callers see the same id; sequential retry idempotent; distinct token provisions a distinct instance. -race clean across repeats.

C. RDS unknown-filter — correct. describeDBInstances rejects an unrecognized name with InvalidParameterValue before filtering; db-instance-id/engine/db-cluster-id still filter. instanceMatchesFilters `continue` on known-but-unmatched is safe because unknownInstanceFilter gates up front.

D. No regression. SSM resolveTargets forwards only instance-id and tag:<name> (both known); unsupported target keys return nil early with no DescribeInstances call, so the new validation never fires on that path. SSM (server+provider), topology, chaos suites pass. EC2 conformance tests are provider-specific and untouched.

Updated ec2_test.go filter-match table is legitimate: it asserts real per-field filtering and flips the old "unknown filter passes (match-all)" case to "unknown does not match", reflecting the corrected semantics rather than papering over them.

Minor (non-blocking): the SSM resolveTargets comment (providers/aws/ssm/run_command.go ~L89-91) still describes matchesTagFilter's default as "matches every instance unconditionally", which this PR changes; the early-return behavior it guards remains correct.

PR #782, opened 2026-08-26
fix(gcp-cloudrun): preserve container env ordering and port name round-trip

NitinKumar004:
## Summary

Two GCP Cloud Run container fidelity bugs, fixed end-to-end through the driver, provider, and wire layers in one change.

### BUG1 [MED] — container env vars returned in nondeterministic order
`driver.Container.Env` was a `map[string]string`, so `envToMap`/`envToList` ranged an unordered Go map — env came back reshuffled across GETs and duplicate names silently collapsed. Real Cloud Run env is an ordered `[]EnvVar` that is stable across reads. This caused a perpetual env-ordering diff in `google_cloud_run_v2_service`/`_job`.

Fix: `Container.Env` is now an ordered `[]driver.EnvVar` (name/value pairs) through the driver, the provider `cloneContainers`, and `conv.go` (the map hop is gone). Declaration order is preserved on round-trip.

### BUG2 [LOW-MED] — container port `name` dropped (h2c unconfigurable)
`driver.Container.Ports` was `[]int`, so the wire `containerPort.Name` (e.g. `"h2c"` for HTTP/2 cleartext) had nowhere to live and was discarded.

Fix: `Ports` is now `[]driver.ContainerPort{Name, ContainerPort}` through the driver, conv, and provider clone; the port name round-trips.

## Scope / safety
- `services/cloudrun/driver` change is structural; Cloud Run is the only consumer, so nothing else breaks.
- updateMask / revision / traffic / LRO paths untouched — existing tests remain green.
- `go generate` + `compatgen` produce zero diff.

## Tests (real `google.golang.org/api/run/v2` SDK over httptest)
- `TestSDKServiceEnvOrderStable` — env `[A,B,C]` returns in `[A,B,C]` order stably across 2 GETs.
- `TestSDKServiceContainerPortName` — port `{name:"h2c",containerPort:8080}` round-trips with the name.
- Existing provider tests updated to the new ordered types.

### DIFF
--- a/providers/gcp/cloudrun/cloudrun.go
+++ b/providers/gcp/cloudrun/cloudrun.go
@@ -295,8 +295,8 @@ func cloneContainers(in []driver.Container) []driver.Container {
 			Image:     in[i].Image,
 			Command:   append([]string(nil), in[i].Command...),
 			Args:      append([]string(nil), in[i].Args...),
-			Env:       cloneMap(in[i].Env),
-			Ports:     append([]int(nil), in[i].Ports...),
+			Env:       append([]driver.EnvVar(nil), in[i].Env...),
+			Ports:     append([]driver.ContainerPort(nil), in[i].Ports...),
 			Resources: cloneResources(in[i].Resources),
 		}
 	}
--- a/providers/gcp/cloudrun/cloudrun_test.go
+++ b/providers/gcp/cloudrun/cloudrun_test.go
@@ -65,7 +65,7 @@ func jobCfg() driver.JobConfig {
 			Image:   "busybox",
 			Command: []string{"echo"},
 			Args:    []string{"hi"},
-			Env:     map[string]string{"K": "V"},
+			Env:     []driver.EnvVar{{Name: "K", Value: "V"}},
 		}},
 	}
 }
--- a/providers/gcp/cloudrun/engine.go
+++ b/providers/gcp/cloudrun/engine.go
@@ -197,13 +197,14 @@ func engineContainers(in []driver.Container, taskIndex, taskCount int) []config.
 }
 
 // taskEnv copies a container's env and overlays the CLOUD_RUN_TASK_INDEX and
-// CLOUD_RUN_TASK_COUNT variables the real Cloud Run injects for each task.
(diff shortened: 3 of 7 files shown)


NitinKumar004 (commented):
Reviewed at pushed SHA 2e108f886ab3a02f8835a4ac4ed564f03b72906d (HEAD confirmed).

Verdict: MERGE. Clean, well-scoped fix.

- go build ./... clean; go vet ./... clean.
- go test ./server/gcp/cloudrun/... ./providers/gcp/cloudrun/... ./services/cloudrun/... green; go test -race ./providers/gcp/cloudrun/... green.
- gofmt clean; golangci-lint --new-from-rev=origin/development on both changed packages: 0 issues.
- go generate ./... -> zero diff. compatgen (real SDK suite, all 3 clouds) exit 0 -> zero diff.

Focus findings:
A. Driver change contained. All importers of cloudrun/driver are Cloud Run only (providers/gcp/cloudrun, server/gcp/cloudrun, server/gcp/gcp.go wiring). Container.Env []EnvVar and Ports []ContainerPort are consumed nowhere else; full go build ./... proves no other consumer broke.
B. Env ordering stable. Env is a slice end-to-end (wire toDriverEnv/toWireEnv preserve order; no map on the round-trip path). New TestSDKServiceEnvOrderStable asserts [A,B,C] over 2 GETs. engine.go taskEnv folds the slice into the engine map with last-wins on dup names (documented) and overlays CLOUD_RUN_TASK_INDEX/COUNT correctly. Dup-name handling: wire path preserves the list, engine is last-wins -- reasonable (strictly better than the prior map, which silently dropped order and dups).
C. Port name round-trips. toDriverPorts/toWirePorts carry Name; TestSDKServiceContainerPortName asserts {name:"h2c",containerPort:8080} survives create->get.
D. No regression. cloneContainers deep-copies via append([]EnvVar(nil),...)/append([]ContainerPort(nil),...); EnvVar/ContainerPort are flat value structs so the slice copy is a full deep copy -- no aliasing. updateMask/revision/traffic/LRO untouched (not in diff); existing revision tests green.

PR #783, opened 2026-08-26
fix(azure-lb): case-insensitive LB+child lookup, PATCH UpdateTags, rule/probe defaults

NitinKumar004:
## Summary

Fixes four Azure Load Balancer parity bugs found in a real-SDK (armnetwork) audit of `providers/azure/loadbalancer/` + `server/azure/loadbalancer/`. All whole-LB-replace semantics and in-sync standalone sub-resource CRUD are preserved.

### BUG1 [MED] — case-insensitive native store key
`azureLBKey(rg, name)` was case-SENSITIVE, so a GET/DELETE/sub-resource request using a differently-cased resource group or LB name 404'd, and a re-PUT with different casing created a DUPLICATE store entry. ARM addressing is case-insensitive. Fix: the internal map key is now `strings.ToLower(rg)+"/"+strings.ToLower(name)`. Response body `id`/`name` casing is unchanged — only the map key is normalized. (#762 fixed only the List RG-filter, not this primary key.)

### BUG4 [LOW] — case-insensitive child-name matching
`indexOfString`/`containsString` (backend pools), NAT-rule upsert/delete name matching, and the `findChild` sub-resource helpers (frontend/pool/rule/probe/nat/outbound) used exact `==`. ARM child names are case-insensitive. Fix: `strings.EqualFold` throughout.

### BUG3 [LOW-MED] — PATCH LoadBalancers.UpdateTags
A PATCH on the whole LB previously fell through to 405. Fix: added a `PATCH` branch that MERGES the request's tags into the stored LB (existing tags kept, matching keys overwritten), preserves every child, and returns 200 with the LB.

### BUG5 [LOW] — synthesize omitted rule/probe defaults
Omitted fields read back as Go-zero instead of the defaults real Azure applies. Fix: loadBalancingRule `idleTimeoutInMinutes=4` / `loadDistribution=Default`; probe `intervalInSeconds=15` / `numberOfProbes=2` (inbound NAT rule `idleTimeoutInMinutes=4` applied for consistency).

### Deferred
- BUG2 (`loadBalancerBackendAddresses` round-trip) — DEFERRED: it requires extending the native pool model beyond a `[]string` of names to carry backend-address entries, a larger model change than fits this focused PR.

## Tests
New real-armnetwork tests over `httptest` (`case_patch_defaults_test.go`): cased GET/DELETE + sub-resource CRUD resolve the same LB; cased re-PUT does not duplicate (List stays at 1); UpdateTags merges tags -> 200; a minimal rule/probe reads back idleTimeout=4/loadDistribution=Default/interval=15/count=2.

Gates: build, vet, `go test ./server/azure/loadbalancer/... ./providers/azure/loadbalancer/...`, gofmt, `go generate`/compatgen (zero diff), and `golangci-lint --new-from-rev=origin/development` (0 new) all green.

### DIFF
--- a/providers/azure/loadbalancer/azure_load_balancer.go
+++ b/providers/azure/loadbalancer/azure_load_balancer.go
@@ -12,9 +12,13 @@ import (
 var _ driver.AzureLoadBalancers = (*Mock)(nil)
 
 // azureLBKey keys the native store by (resourceGroup, name), matching ARM's
-// addressing.
+// addressing. ARM resource-group and resource names are case-insensitive, so
+// the key is lower-cased to resolve differently-cased GET/DELETE/sub-resource
+// requests to the same load balancer and keep re-PUT an in-place update rather
+// than a duplicate. Only the internal map key is normalized; the stored body's
+// id/name casing is preserved verbatim.
 func azureLBKey(rg, name string) string {
-	return rg + "/" + name
+	return strings.ToLower(rg) + "/" + strings.ToLower(name)
 }
 
 // CreateOrUpdateAzureLoadBalancer stores the ARM load balancer as a full
@@ -191,7 +195,7 @@ func (m *Mock) UpsertAzureLBNatRule(
 		replaced := false
 
 		for i := range rules {
-			if rules[i].Name == natRuleName {
+			if strings.EqualFold(rules[i].Name, natRuleName) {
 				rules[i] = rule
 				replaced = true
 
@@ -230,7 +234,7 @@ func (m *Mock) DeleteAzureLBNatRule(_ context.Context, rg, name, natRuleName str
 		idx := -1
 
 		for i := range lb.NatRules {
-			if lb.NatRules[i].Name == natRuleName {
+			if strings.EqualFold(lb.NatRules[i].Name, natRuleName) {
 				idx = i
 				break
 			}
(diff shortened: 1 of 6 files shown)


NitinKumar004 (commented):
Reviewed at pushed SHA 6f4df660 in an isolated worktree. Verdict: approve for merge.

- go build ./... clean; go vet ./... clean
- go test ./server/azure/loadbalancer/... ./providers/azure/loadbalancer/... PASS
- go test -race ./providers/azure/loadbalancer/... PASS
- gofmt clean on both changed dirs; golangci-lint --new-from-rev=origin/development = 0 issues
- go generate ./... zero diff; compatgen (go run ./internal/compatgen) exit 0 AND zero diff on docs/compat

Focus verification:
A. Case-insensitive key (BUG1): azureLBKey lower-cases both rg+name; every store op (Set/Get/Delete/CreateOrUpdate/Upsert*/Delete*) routes through it consistently. Cased GET/DELETE/sub-resource CRUD resolve the same LB; cased re-PUT updates in place (List==1). CRITICAL response-casing check: toLBJSON builds id/name from rp.ResourceName/ResourceGroup (the request URL path) verbatim — never lowercased. Only the internal map key is normalized; stored Name/ResourceGroup keep original casing. responseCasingUnchanged holds.

B. Child EqualFold (BUG4): findFrontend/findPool/findRule/findProbe/findNatRule/findOutboundRule use strings.EqualFold; provider indexOfString + NAT-rule match + hasFrontend also EqualFold. Cased child GET/DELETE resolves.

C. PATCH UpdateTags (BUG3): merged := *stored carries all child slices; only merged.Tags is reassigned (mergeTags: base kept, overlay wins, new added; neither input mutated). CreateOrUpdate re-clones children -> no nil-mask wipe. Returns 200. stored comes from a cloned Get, so no store aliasing.

D. Defaults (BUG5): buildRules/buildProbes/buildNatRules apply idleTimeout=4 / loadDistribution=Default / interval=15 / count=2 ONLY when the field is zero/empty; explicit values pass through and round-trip via rulesJSON/probesJSON. Baked at PUT and echoed on GET.

E. No regression: whole-LB full-replace, standalone sub-resource sync, and reference validation paths unchanged. backend-addresses BUG2 correctly out of scope (deferred).

No blocking issues found.

PR #784, opened 2026-08-26
fix(gcp-cloudfunctions): gen2 patch instance-count, gen1 eventTrigger round-trip, gen1 field round-trip, honor updateMask

NitinKumar004:
## Summary

Fixes four GCP Cloud Functions bugs found in a real-SDK audit (apiv1 + apiv2 over httptest). LRO/IAM/upload/pagination paths are untouched.

### BUG1 — gen2 PATCH dropped maxInstanceCount/minInstanceCount
`mergeServiceConfig` omitted both scaling fields, so a `serviceConfig.maxInstanceCount` PATCH never applied (create worked via the wholesale store). Added both to the service-config merge, driven by the update mask.

### BUG2 — gen1 eventTrigger was not modeled
Every gen1 function was returned HTTP-triggered because the v1 struct had no `eventTrigger` and `toCloudFunction` unconditionally attached `httpsTrigger`. Added an `eventTrigger` field to the v1 struct, persisted it in `gen1Meta`, and `toCloudFunction` now emits `httpsTrigger` XOR `eventTrigger` based on what was created.

### BUG3 — gen1 round-trip dropped fields
`description`, `maxInstances`/`minInstances`, `vpcConnector`/`vpcConnectorEgressSettings`, `sourceRepository`, and `sourceArchiveUrl`/`sourceUploadUrl` were accepted but never echoed. `cloudFunction` + `gen1Meta` now carry and echo them.

### BUG4 — updateMask never honored (gen1 + gen2)
Both the gen1 update handler and gen2 `patchV2` ignored the `updateMask` query param and used a non-zero merge, so masked fields could not be cleared; gen2 also replaced the whole `buildConfig` wholesale, wiping runtime/entryPoint on a partial PATCH. Added a shared `updateMask` parser that drives field-level apply/clear in both paths; a partial `buildConfig` mask now merges sub-fields instead of replacing the object.

## Tests

Real functions apiv1 + apiv2 SDK over httptest:
- gen2 PATCH `maxInstanceCount` 7 -> 20 reflects on GET (and unmasked `minInstanceCount` survives)
- gen2 partial `buildConfig` PATCH preserves runtime/entryPoint
- gen1 `eventTrigger` round-trips (GET returns `eventTrigger`, no `httpsTrigger`)
- gen1 `description`/`vpcConnector`/`maxInstances` round-trip
- gen1 `updateMask` clears a masked field while an unmasked field survives

## Notes / deferred

Event-trigger invocation wiring, `:call` code execution, and gen1 list scoping are out of scope and left for follow-up.

### DIFF
--- a/server/gcp/cloudfunctions/gen1meta.go
+++ b/server/gcp/cloudfunctions/gen1meta.go
@@ -18,6 +18,18 @@ type gen1Meta struct {
 	dockerRegistry      string
 	buildID             string
 	versionID           int64
+	// The remaining fields are client-owned inputs a real gen1 client reads back
+	// from Get/List. They have no portable Serverless-driver equivalent, so the
+	// wire handler carries them here and echoes them from toCloudFunction.
+	description                string
+	eventTrigger               *eventTrigger
+	sourceRepository           *sourceRepository
+	sourceArchiveURL           string
+	sourceUploadURL            string
+	maxInstances               int
+	minInstances               int
+	vpcConnector               string
+	vpcConnectorEgressSettings string
 }
 
 // newGen1Meta builds the initial metadata for a freshly created v1 function,
@@ -25,11 +37,20 @@ type gen1Meta struct {
 // Cloud Functions assigns.
 func newGen1Meta(body *cloudFunction, project string) *gen1Meta {
 	m := &gen1Meta{
-		serviceAccountEmail: body.ServiceAccountEmail,
-		ingressSettings:     body.IngressSettings,
-		dockerRegistry:      body.DockerRegistry,
-		buildID:             newBuildID(),
-		versionID:           1,
+		serviceAccountEmail:        body.ServiceAccountEmail,
+		ingressSettings:            body.IngressSettings,
+		dockerRegistry:             body.DockerRegistry,
+		buildID:                    newBuildID(),
+		versionID:                  1,
+		description:                body.Description,
+		eventTrigger:               body.EventTrigger,
+		sourceRepository:           body.SourceRepository,
+		sourceArchiveURL:           body.SourceArchiveURL,
+		sourceUploadURL:            body.SourceUploadURL,
+		maxInstances:               body.MaxInstances,
+		minInstances:               body.MinInstances,
+		vpcConnector:               body.VPCConnector,
+		vpcConnectorEgressSettings: body.VPCConnectorEgressSettings,
 	}
 
 	applyGen1Defaults(m, project)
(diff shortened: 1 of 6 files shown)


NitinKumar004 (commented):
Reviewed at pushed SHA 330c2600551cf676eb3291c3c2563f73a61133c1 (HEAD confirmed). Verdict: MERGE.

Gates (GOMAXPROCS=3, -p=3), all green:
- go build ./... + go vet ./... clean
- go test ./server/gcp/cloudfunctions/... ./providers/gcp/cloudfunctions/... ok
- go test -race ./server/gcp/cloudfunctions/... ok
- gofmt clean; golangci-lint --new-from-rev=origin/development 0 new issues
- go generate: zero tracked diff (coverage regenerated identically)
- compatgen: real SDK suite passes (aws/azure/gcp), zero docs/compat drift
(TestWiringParity_Azure is the known pre-existing dev failure, #785 — out of scope, not run here.)

Correctness verified against the four fixes:

A. updateMask honoring (gen1+gen2). parseUpdateMask/covers/applyMasked* are clean: a masked field is written even when empty (explicit clear); an absent mask falls back to legacy merge-non-zero so an unmasked PATCH never blanks an untouched field. mergeServiceConfig gates every sub-field on covers(), so mask=[serviceConfig.maxInstanceCount] with other fields present changes only maxInstanceCount (TestSDKGen2PatchInstanceCount asserts min survives). mergeBuildConfig merges sub-fields under the mask and does NOT replace the whole object; the output-only build id is recut by computeGen2Outputs, not merged (TestSDKGen2PatchPartialBuildConfig asserts runtime/entryPoint preserved).

B. gen2 maxInstanceCount/minInstanceCount now round-trip on create and reflect a 7->20 PATCH (previously dropped). Covered.

C. gen1 eventTrigger XOR httpsTrigger. toCloudFunction emits EventTrigger when meta.eventTrigger != nil and HTTPSTrigger otherwise, so an event-driven fn returns eventTrigger and no httpsTrigger, while an HTTP fn still gets its trigger URL (TestSDKGen1EventTriggerRoundTrip).

D. gen1 field round-trip. gen1Meta now carries description/max-min instances/vpcConnector(+egress)/sourceRepository/sourceArchiveUrl and echoes them from toCloudFunction (TestSDKGen1FieldRoundTrip).

E. No regression. Diff is confined to server/gcp/cloudfunctions (provider untouched). LRO/IAM/upload/pagination/gen2-create paths unchanged apart from the additive mask parameter and gen1Meta fields; -race clean; clone paths deep-copy nested structs so the in-place patch mutation is never observed mid-write. bumpGen1Meta keeps serviceAccountEmail/ingress/dockerRegistry server-defaulted (set-if-supplied, never cleared) which matches real gen1 semantics.

Minor, non-blocking: with an explicit mask naming buildConfig.runtime/eventTrigger/sourceRepository, if the PATCH body omits the whole buildConfig/eventTrigger object entirely, the field is not cleared (the merge only descends when src.<obj> != nil). Real GCP would clear it. Edge case, not in scope of the reported fixes and no test depends on it.

PR #786, opened 2026-08-26
fix(gcp-firestore): subcollection namespaces, preserve user id field, atomic batch commit, orderBy excludes missing-field docs

NitinKumar004:
## Summary

Fixes four confirmed GCP Firestore bugs in the wire handler and its storage model. All four are covered by new real-SDK (`cloud.google.com/go/firestore` over httptest) tests.

### BUG1 [HIGH] Subcollections modeled as separate namespaces
Previously `cities/SF/landmarks/gg` collapsed into the top-level `cities` table with id `SF/landmarks/gg`, so listing/querying the parent returned subcollection docs (contamination) and the subcollection query 404'd. Now the document's full parent path (`cities/SF/landmarks`) is the driver namespace: `parseFirestorePath`/`splitDocumentName` parse the alternating collection/document segments, and `runQuery`/`listDocuments` scope by the request's parent path (base) joined with the `from` collection id — not the trailing collection id alone. A never-written (sub)collection now lists/queries as empty (cloud-accurate) instead of erroring. Direct-ref CRUD is unchanged.

### BUG2 [HIGH, data-loss] User field named `id` preserved
The reserved doc-id/createTime/updateTime keys now use NUL-prefixed sentinels (`"\x00id"`, etc.) that cannot collide with real field names, and `mapToFields` no longer strips `id`. A user field literally named `id` (or `__createTime__`/`__updateTime__`) round-trips on read and is queryable, while the document id stays independent.

### BUG3 [MED, atomicity] Batch/transaction commit is atomic
`commit` is now two-phase: phase 1 validates every write's precondition against a working snapshot (an overlay reflecting earlier writes in the same batch) and computes each resulting item without persisting; phase 2 applies the staged writes per collection through the atomic `TransactWriteItems`. A precondition failure on write N leaves writes 1..N-1 unpersisted.

### BUG4 [MED-LOW] orderBy excludes docs missing the ordered field
`shapeResults` drops documents that lack any explicit orderBy field before sorting/cursoring (the implicit `__name__` tiebreaker never filters). A present-but-null field is still included.

## Tests
- Subcollection namespace isolation (parent list/query excludes subcol docs; subcol list/query returns its docs; direct-ref CRUD intact)
- User `id` field round-trips and is queryable; doc id independent
- Batch with a failing precondition persists nothing
- orderBy excludes a doc missing the ordered field

## Gates
build ./..., vet, `go test -p=3` (server+provider), `go test -race -p=3` (provider), gofmt, `golangci-lint --new-from-rev=origin/development` (0 new), `go generate ./...` + compatgen (zero diff) — all green.

### DIFF
--- a/server/gcp/firestore/firestore_atomic_commit_test.go
+++ b/server/gcp/firestore/firestore_atomic_commit_test.go
@@ -1,0 +1,48 @@
+package firestore_test
+
+import (
+	"testing"
+
+	"google.golang.org/grpc/codes"
+)
+
+// TestFirestoreBatchAtomicOnPreconditionFailure proves a batch commit is
+// atomic: when a later write fails its precondition, the earlier writes in the
+// same batch are NOT persisted. Previously each write applied immediately, so a
+// failure left preceding writes committed.
+func TestFirestoreBatchAtomicOnPreconditionFailure(t *testing.T) {
+	ctx, client, _ := newDBClient(t, "accounts")
+
+	coll := client.Collection("accounts")
+
+	// Seed an existing document so a later Create (exists=false precondition)
+	// in the batch is guaranteed to fail.
+	if _, err := coll.Doc("existing").Set(ctx, map[string]any{"v": "seed"}); err != nil {
+		t.Fatalf("seed existing: %v", err)
+	}
+
+	// write1 would create a fresh document; write2 Creates over an existing
+	// document and must fail with ALREADY_EXISTS, aborting the whole commit.
+	batch := client.Batch()
+	batch.Set(coll.Doc("fresh"), map[string]any{"v": "one"})
+	batch.Create(coll.Doc("existing"), map[string]any{"v": "two"})
+
+	if _, err := batch.Commit(ctx); dbSDKCode(err) != codes.AlreadyExists {
+		t.Fatalf("batch Commit: code=%v err=%v, want AlreadyExists", dbSDKCode(err), err)
+	}
+
+	// The first write must NOT have been persisted (atomic rollback).
+	if _, err := coll.Doc("fresh").Get(ctx); dbSDKCode(err) != codes.NotFound {
+		t.Errorf("doc 'fresh' should not exist after failed batch, got err=%v", err)
+	}
+
+	// The pre-existing document must be untouched by the failed Create.
+	snap, err := coll.Doc("existing").Get(ctx)
+	if err != nil {
+		t.Fatalf("Get existing: %v", err)
+	}
+
+	if snap.Data()["v"] != "seed" {
+		t.Errorf("existing doc mutated by failed batch: v=%v, want seed", snap.Data()["v"])
+	}
+}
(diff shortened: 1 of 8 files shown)


NitinKumar004 (commented):
Deep adversarial review — verdict: MERGE.

Verified on pushed SHA b1606e39. Build/vet clean, firestore tests pass, -race clean on both providers/gcp/firestore and server/gcp/firestore. gofmt clean, golangci-lint --new-from-rev 0 new issues. go generate zero diff, compatgen exit 0 zero diff. Only server/gcp/firestore files changed.

BUG1 subcollections — SOLID. Subcollections are modeled as separate driver tables keyed on the full parent path (cities/SF/landmarks), and the driver Scan matches table names by exact equality, so there is no prefix-match hazard. Adversarially confirmed: listing/querying "cities" returns [SF, LA] only (no landmark leakage); "cities-2" is fully isolated from "cities" (prefix sibling does not contaminate); SF's landmarks {a,b} are isolated from LA's {c}; direct-ref CRUD on a subcol doc round-trips; deep nesting col/doc/col/doc/col/doc works and does not leak up a level; a query/list on a never-written (sub)collection returns empty (NotFound-as-empty, no 404 crash). runQuery/listDocuments correctly scope by the request parent path (parentPath + from collection id), not the trailing collection id alone.

BUG2 id-field — SOLID. Reserved keys are NUL-prefixed (\x00id / \x00createTime / \x00updateTime); field names are UTF-8 text and cannot contain NUL, so no collision with any legal user field. A doc with user field id:"x" round-trips on Get and is queryable via where id==x, independent of the resource-name id. __name__ still aliases fieldID for ordering/DocumentID. isReservedKey uses exact equality; mapToDocument strips only the three reserved keys.

BUG3 atomic batch — SOLID. Two-phase validate-then-apply: phase 1 validates every write's precondition against a consistent working overlay (staged writes shadow the persisted store) and computes items without persisting; phase 2 applies via per-table TransactWriteItems. A precondition failure on write N aborts before any apply, so 1..N-1 are unpersisted. Adversarially confirmed a multi-collection batch (colA + colB) with a mid-batch Create-over-existing failure persists NOTHING in either collection and leaves the pre-existing doc untouched; a fully-valid multi-collection batch applies all writes. Phase 2 cannot fail on a precondition (ensureCollection precedes each TransactWriteItems, and the driver's only failure mode is a missing table), so the sequential per-table apply cannot leave a partial commit on the failure path.

BUG4 orderBy — SOLID. Docs missing an explicit orderBy field are excluded; present-but-null are included; __name__ tiebreaker never filters. Confirmed with a 3-doc mix (has/missing/null).

Base behavior unregressed: existing lifecycle/query/precondition/transform/cursor/typed-value suites all pass. The harness partition-key change ("id" -> "\x00id") is a legitimate alignment with the new reserved-key storage model, not masking a break.

Note (non-blocking): true cross-collection atomicity within a commit (visibility isolation across the sequential per-table applies) is not provided, consistent with the in-memory store having no isolation levels. The failure-atomicity guarantee this PR targets is fully met.

PR #848, opened 2026-08-28
refactor(server): extract shared server/serverkit assembly from cmd/cloudemu serve

NitinKumar004:
## Summary

Extracts the standalone-server assembly out of `cmd/cloudemu/serve.go` into a new **exported** package `server/serverkit`, so one assembly can back both the shipped `cloudemu serve` binary and (in a later PR) the batteries-included `contrib/server`. This is **PR 0/A** of the Phase 2 program — it is the core extraction only. It does **not** touch `contrib/server`, engines, or Docker.

## What moved

- New `serverkit.Config` (flat data seam) and `serverkit.App` (`New`, `Serve`, `Rebuild`) now own the whole assembly: per-provider build (`cloudemu.New{AWS,Azure,GCP,OCI}` → `<provider>server.DriversFrom` → K8s routing + EKS/GKE/AKS `SetK8sAPI`, OCI excluded), the `admin.Backend` hot-swap control plane, persistence restore/snapshot, init-dir seeding, the `/net/*` + `/cost` admin endpoints, advertise-host/TLS handling, the startup banner + endpoints-file, and the bind → serve → shutdown → snapshot lifecycle. The Azure `WithAccountID(subscription)` last-wins override stays internal, applied to a copy so it never leaks into aws/gcp/oci.
- `cmd/cloudemu/{middleware.go, tlscert.go, endpoints.go}` moved into serverkit (`middleware.go`, `tls.go`, `endpoints.go`) and were deleted from `cmd/cloudemu`.
- `cmd/cloudemu/serve.go` is now a thin flag-parse → `serverkit.Config` → `serverkit.New`/`Serve` wrapper (keeps `parseProviders`).
- The five white-box test files (`advertise`, `persist_serve`, `init`, `net_serve`, `cost_serve`) moved into serverkit as same-package tests; their cmd copies were deleted. The black-box binary tests (`serve`, `lifecycle`, `snapshot`) stay in `cmd/cloudemu` as the byte-identity oracle.

## New capability (not in the old inline code)

serverkit tracks the current live `*Provider` set and `Close()`es the **outgoing** providers after every rebuild swap (`/_cloudemu/reset` and snapshot restore) and on final shutdown **after** the snapshot. This reaps real data-plane engines (embedded Postgres, miniredis, Docker containers) instead of leaking them on every reset — a no-op for cmd/cloudemu's engineless providers. Swap-then-close ordering guarantees a fresh request never routes to a torn-down engine; the in-flight-request-vs-Close best-effort tradeoff is documented in-code as accepted, not an oversight.

## Behavior parity

Serve behavior is **byte-identical** for the `cloudemu serve` binary, guarded by the retained `cmd/cloudemu` black-box tests. New serverkit tests lock: the leak guard (Close fires on reset AND restore, not only shutdown), destructive restore (wipe-to-empty), snapshot-before-close shutdown ordering, the non-loopback admin warning, seed-nil → 501, the 32 MiB fixture body cap, and self-signed + custom-cert TLS.

## Gates

`go build ./...`, `go vet ./...`, `go test ./...`, and `go generate ./...` (no diff) all pass; `golangci-lint run ./server/serverkit/...` reports 0 issues.

### DIFF
--- a/cmd/cloudemu/serve.go
+++ b/cmd/cloudemu/serve.go
@@ -2,45 +2,21 @@ package main
 
 import (
 	"context"
-	"crypto/tls"
-	"encoding/json"
 	"errors"
 	"flag"
 	"fmt"
-	"net"
-	"net/http"
 	"os"
 	"os/signal"
-	"path/filepath"
-	"sort"
-	"strconv"
 	"strings"
-	"sync"
 	"syscall"
 	"time"
 
-	"github.com/stackshy/cloudemu/v2"
 	"github.com/stackshy/cloudemu/v2/config"
-	"github.com/stackshy/cloudemu/v2/features/topology"
-	"github.com/stackshy/cloudemu/v2/persist"
-	eksprov "github.com/stackshy/cloudemu/v2/providers/aws/eks"
-	"github.com/stackshy/cloudemu/v2/seed"
-	"github.com/stackshy/cloudemu/v2/server/admin"
-	awsserver "github.com/stackshy/cloudemu/v2/server/aws"
-	azureserver "github.com/stackshy/cloudemu/v2/server/azure"
-	gcpserver "github.com/stackshy/cloudemu/v2/server/gcp"
-	ociserver "github.com/stackshy/cloudemu/v2/server/oci"
-	"github.com/stackshy/cloudemu/v2/services/kubernetes"
-	"github.com/stackshy/cloudemu/v2/services/pricing"
-	"github.com/stackshy/cloudemu/v2/services/resourcediscovery"
+	"github.com/stackshy/cloudemu/v2/server/serverkit"
 )
 
 // errStateFileRequired is returned when --persist is set without --state-file.
 var errStateFileRequired = errors.New("--persist requires --state-file")
-
-// errUnsupportedSnapshot is returned when a posted snapshot has an unknown
-// schema version.
-var errUnsupportedSnapshot = errors.New("unsupported snapshot schema version")
 
 // serveConfig holds the resolved serve flags.
 type serveConfig struct {
@@ -84,8 +60,8 @@ func (s *stringList) Set(v string) error {
 func runServe(args []string) error {
 	fs := flag.NewFlagSet("serve", flag.ContinueOnError)
 	var c serveConfig
-	// OCI is not started by default: it stays opt-in until its services land,
-	// so the default set never binds a port that serves nothing.
+	// OCI is not started by default: it stays opt-in until its services land, so
+	// the default set never binds a port that serves nothing.
 	fs.StringVar(&c.providers, "providers", "aws,azure,gcp", "comma-separated providers to start: aws,azure,gcp,oci")
 	fs.StringVar(&c.host, "host", "127.0.0.1", "host/interface to bind (use 0.0.0.0 to expose on the network)")
 	fs.StringVar(&c.advertiseHost, "advertise-host", "",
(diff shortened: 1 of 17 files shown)


NitinKumar004 (commented):
Verified against origin/development at head 0ba2e567.

Gates: `go build ./...`, `go vet ./...` clean. `go test -count=1 ./cmd/cloudemu/... ./server/serverkit/...` passes, including `TestServeOutOfProcess` (real AWS/Azure SDK clients driven against the compiled binary) unchanged. `go test -race ./server/serverkit/... ./server/admin/...` clean. `golangci-lint run ./server/serverkit/...` — 0 issues, confirms the claim. `go generate ./...` — no diff.

**Close-tracking (serverkit.go)**
- `Rebuild()` → `swapFresh()` swaps the backends under `rebuildMu` before returning the outgoing `[]closer`; `closeProviders(outgoing)` runs after, outside the lock (serverkit.go:222-233) — correct swap-then-close ordering.
- `shutdown()` calls `Shutdown()` on every `*http.Server` first, writes the persistence snapshot, and only then closes `a.providers` under the lock (serverkit.go:627-657) — engines stay readable through the snapshot.
- `TestProviderClosedOnResetAndRestore` and `TestSnapshotOnShutdownBeforeClose` (serverkit_test.go:51,133) assert this on both `/_cloudemu/reset` and restore, not just shutdown.
- OCI's `builtProvider.provider` is left as a genuine nil `closer` (never assigned), so the `if b.provider != nil` guard (serverkit.go:278) correctly skips it — no nil-interface Close risk.
- For cmd/cloudemu, `Provider.Close()` is always a no-op today (`engineClosers` is only populated via `config.With<X>Engine`, never wired here) — the new capability is inert for the shipped binary and only becomes live for a real-engine caller, not exercised by this PR.
- The comment at serverkit.go:224-231 correctly calls out the one open tradeoff: a request in flight against the outgoing handler at the instant of the swap can still be executing when Close() runs immediately after — no drain window. Not a blocker given today's no-op Close, but it's worth keeping in mind once contrib migrates onto serverkit with real engines wired.

**Behavior parity**
One real deviation from a pure extraction: `ReadHeaderTimeout: serverReadHeaderTimeout` (10s) is now set on every `http.Server` (serverkit.go:589,612); the pre-PR `cmd/cloudemu/serve.go` never set it (unbounded). It's a legitimate Slowloris fix and won't affect any real SDK/CLI caller, but it is new behavior rather than preserved behavior — flagging since it's the one place strict byte-identity doesn't literally hold. Everything else checked out unchanged: destructive restore semantics, the 32MiB/512MiB body caps (unchanged, in the untouched `server/admin` package), the non-loopback admin danger warning, silent first-run restore no-op on `os.ErrNotExist`, partial-bind listener cleanup, seed/snapshot nil→501, k8s `SetBaseURL` 0.0.0.0→127.0.0.1 fallback, `--tls-cert` scoped to the Azure listener only (k8s keeps its own eksprov cert), and the `--shutdown-timeout` default (10s) preserved.

**Move completeness**
No dangling references to moved symbols in the remaining cmd/cloudemu files (env.go, lifecycle.go, net.go, cost.go, main.go); `printBanner`/`endpointSet` have a single definition after the endpoints.go move. Azure's `WithAccountID` override copies into a freshly `make`'d slice (serverkit.go:354-356), so it can't alias into `a.baseOpts` — confirmed non-leaking to aws/gcp/oci. EKS/GKE/AKS get `SetK8sAPI`; OCI is correctly excluded (no managed-Kubernetes service).

**Minor nit**
The package doc comment (serverkit.go:6-8) states serverkit is "the single source of truth consumed by both cmd/cloudemu/serve.go and the batteries-included contrib/server" — `contrib/server` doesn't import serverkit yet (it's a separate go.mod with its own assembly in app.go/engines.go). Not a functional issue, just overstates current wiring.

**nolints**
Only two in the diff. The gosec G706 justification on the opt-in `--log-requests` access-log line (middleware.go) is reasonable. The `forcetypeassert` one is in a test file, not production code. No G112 suppression was added — it was fixed properly (see the ReadHeaderTimeout note above) rather than silenced.

Gates green, Close ordering correct and test-covered, no blockers.

PR #849, opened 2026-08-28
feat(contrib-server): build on shared server/serverkit assembly

NitinKumar004:
## Summary

The batteries-included `contrib/server` now consumes the merged `server/serverkit` package for assembly instead of maintaining its own bind/TLS/shutdown scaffolding, so the two standalone entrypoints (`cmd/cloudemu serve` and the engines server) no longer drift.

### What changed
- **`contrib/server/app.go`** — maps the resolved flag configuration onto a `serverkit.Config` (`Providers: aws,azure,gcp`; `Host`; `Ports`; `AzureSubscription`; `ShutdownTimeout`; `BaseOptions` = the real-engine `config.Option`s from `engines.go`) and holds a `*serverkit.App`. serverkit owns provider construction, listener binding, the shutdown/snapshot lifecycle, and engine teardown on Close.
- **`contrib/server/main.go`** — `run` is now a thin wrapper: wire the signal context, print the engine-selection line, and hand off to `serverkit`'s `Serve` (which prints the endpoint banner and owns graceful shutdown).
- **Deleted** the duplicate `boundServer`/`bindAll`/`bindTLS`/`closeProviders`/graceful-shutdown scaffolding from `app.go` and the whole `contrib/server/tls.go` (`selfSignedTLSConfig`/`selfSignedCert`), now dead once serverkit owns bind + TLS (grep-confirmed no other contrib file referenced them).
- **`engines.go` unchanged** — still returns the per-provider real-engine options (the 5 engines: database/cache/function/compute/container). No new flags; existing `--db/--cache/--functions/--compute/--containers` + `--all-real` and the `CLOUDEMU_*` env vars are untouched.

### Identity-preserving
The swap from `awsserver.NewFromProvider(p)` to `serverkit` (which builds via `awsserver.DriversFrom(p)`) is identity-preserving: `NewFromProvider(p) == New(DriversFrom(p))`, and `DriversFrom` copies `AccountID`/`Region`/`EnforceAuth` verbatim. With Kubernetes disabled in this build, `K8sAPI` is nil on both paths.

A new `TestBatteriesServerIdentityPreserved` guards the field-copy: it boots through serverkit with a distinctive account/region, and asserts the RDS `DBInstanceArn` (built from the provider's configured `AccountID`) reflects the expected account and region — the whole flow succeeding under dummy static credentials also proves enforce-auth stays off. The existing postgres/redis engine e2e (`TestBatteriesServerE2E`) still passes through the new seam, including the embedded-Postgres port-release assertion (serverkit's provider Close reaps the real engines).

### Note
This activates serverkit's provider-Close-on-rebuild for real engines; the documented in-flight-request tradeoff on `/_cloudemu/reset` becomes live for contrib. Expected and accepted. Admin/persist/seed/OCI/k8s/storage flags and the Dockerfile/CI are out of scope here (later PRs).


NitinKumar004 (commented):
Reviewed 06edc6d against origin/development. Diff is contrib/server only (app.go, main.go, main_test.go, tls.go deleted) — server/serverkit, server/aws, server/azure, server/gcp are untouched by this PR.

**Identity-preservation.** `server/aws/aws.go:309-311` shows `NewFromProvider(p) = New(DriversFrom(p))` — this was already true before #848/#849, so contrib's switch to `awsserver.New(awsserver.DriversFrom(p))` is calling the exact same construction path contrib used before, just via serverkit's copy of it. `DriversFrom` (aws.go:253-305) copies every field off the provider including AccountID, Region, EnforceAuth, STS (hardcoded true), ResourceDiscovery, and every service driver — nothing is dropped. Same pattern confirmed for Azure (`server/azure/from_provider.go:18-84`, `NewFromProvider` = `New(DriversFrom(p))`, copies SubscriptionID/EnforceAuth/ResourceDiscovery verbatim). K8sAPI is the only field DriversFrom leaves nil in both old and new paths for contrib (contrib doesn't wire k8s), consistent with the PR's stated scope.

`TestBatteriesServerIdentityPreserved` (main_test.go) is a real exercise of this: it sets a distinct AccountID/region, drives RDS CreateDBInstance/DescribeDBInstances through the emulator, and asserts the returned ARN's account and region fields (parts[4], parts[3]) match — this would fail immediately if AccountID or Region were dropped anywhere in the DriversFrom/serverkit path. It also asserts the call succeeds under dummy static credentials, proving EnforceAuth stays off by default (contrib's `serverkit.Config.EnforceAuth` field is left as its zero value).

**TLS.** contrib deletes its own tls.go; `serverkit.tlsConfig` (server/serverkit/tls.go:24-45) falls back to an in-memory self-signed cert covering `localhost` + the bind host + loopback IPs whenever `Config.TLSCert` is empty, which is exactly contrib's case (never set in `serverkitConfig`, app.go:55-69). The Azure HTTPS listener is unaffected — same self-signed-cert shape as the deleted contrib/tls.go, just centralized.

**Admin plane scope.** `serverkit.Config.Admin` is never set in `serverkitConfig` (app.go), so it's `false`. `handlerFor` (serverkit.go:499-505) returns the bare backend when Admin is off — no `/_cloudemu` control plane mounted. Correctly scoped for PR B.

**Real-engine Close ordering.** `App.shutdown` (serverkit.go:627-657) does HTTP `Shutdown` → persistence snapshot (skipped here, `Persist` is false) → `closeProviders`, in that order — engines are torn down only after the listeners have fully drained, matching contrib's prior behavior. `TestBatteriesServerE2E` exercises this for real (embedded Postgres + miniredis) and asserts the Postgres port is released after `stop()` cancels the serve context.

**Nothing lost.** Banner still prints via `serverkit.printBanner` (contrib's `Config.Quiet` defaults false); engine selection banner (`printEngines`) still prints from contrib/main.go. go.mod/go.sum are unchanged in this diff — no new dependencies pulled into either module.

**Gates — both green.**
- Root: `go build ./...`, `go vet ./...` clean; `go test -count=1 ./server/serverkit/... ./cmd/cloudemu/...` → both packages `ok`.
- `contrib/server`: `go build ./...`, `go vet ./...` clean; `go test -count=1 ./...` → `TestBatteriesServerE2E` (real embedded-Postgres + miniredis round trip, asserts port released post-shutdown) and `TestBatteriesServerIdentityPreserved` both PASS.
- `golangci-lint run --timeout=9m ./...` from contrib/server: 0 issues.

No blockers found.

PR #850, opened 2026-08-28
feat(contrib-server): admin plane + persistence + init-dir via serverkit flags

NitinKumar004:
## Summary

Brings the batteries-included `contrib/server` to parity with `cloudemu serve` for the admin control plane, persistence, and boot-time fixtures. serverkit already implements all of this — this PR only exposes the flags and threads them into `serverkit.Config`.

## Flags added (matching `cloudemu serve` names/defaults)

- `--admin` (default `true`) — mount the `/_cloudemu` control plane (reset/health/seed/snapshot)
- `--persist` (default `false`) + `--state-file` — restore state on startup, snapshot on shutdown (`--persist` requires `--state-file`)
- `--persist-metadata-only` (default `false`) — omit object bodies from the snapshot
- `--init-dir` — apply every `*.json` seed fixture in a directory on boot

Each is mapped onto the corresponding `serverkit.Config` field (`Admin`/`Persist`/`StateFile`/`PersistMetadataOnly`/`InitDir`); serverkit owns the control-plane mount, restore-on-boot, snapshot-on-shutdown, and init-dir application. No behavior is re-implemented in contrib.

## Tests

New `contrib/server/admin_persist_test.go` (hand-rolled helpers, no testify):
- `TestAdminReset` — with `--admin`, POST `/_cloudemu/reset` returns OK and clears seeded state
- `TestPersistRoundTrip` — create a resource -> shutdown (snapshot) -> fresh app restores it from the state file
- `TestInitDir` — a fixture JSON in `--init-dir` is seeded on boot

Engines are off for these, so they are fast and need no Docker/downloads. Existing engine e2e tests still pass.

## Docs

`contrib/server/README.md` gains an "Admin, persistence & seeding" section documenting the new flags.

### DIFF
--- a/contrib/server/README.md
+++ b/contrib/server/README.md
@@ -85,6 +85,34 @@ not on `PATH`.
 | `--project-id`        | `cloudemu-local`                         | GCP project id                             |
 | `--shutdown-timeout`  | `10s`                                    | grace period for in-flight requests        |
 
+## Admin, persistence & seeding
+
+At parity with `cloudemu serve`, these flags are threaded through to the shared
+`server/serverkit` assembly:
+
+| Flag                     | Default | Meaning                                                            |
+|--------------------------|---------|--------------------------------------------------------------------|
+| `--admin`                | `true`  | mount the `/_cloudemu` control plane (`reset`, `health`, `seed`, `snapshot`) for test isolation |
+| `--persist`              | `false` | restore state from `--state-file` on startup and save it on shutdown (includes object bodies) |
+| `--state-file`           | *(none)*| path to the JSON state snapshot (**required** with `--persist`)     |
+| `--persist-metadata-only`| `false` | persist resource structure but omit object bodies (smaller snapshot)|
+| `--init-dir`             | *(none)*| apply every `*.json` seed fixture in this directory on startup      |
+
+```bash
+# Isolated test runs: POST /_cloudemu/reset between suites to wipe state.
+go run . --admin
+
+# Persist state across restarts:
+go run . --persist --state-file ./state.json
+
+# Seed fixtures on boot:
+go run . --init-dir ./fixtures
+```
+
+`--admin` is on by default; on a non-loopback `--host` it prints a warning, since
+`POST /_cloudemu/reset` wipes all state and `GET /_cloudemu/snapshot` dumps it
+(secrets included) to any caller. Pass `--admin=false` to disable it.
+
 The Azure endpoint serves HTTPS with an in-memory self-signed certificate
 (covering `localhost` and the loopback IPs); clients must trust it or skip
 verification.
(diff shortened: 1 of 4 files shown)


NitinKumar004 (commented):
Reviewed at 8283168 (mechanical flag-wiring on top of the already-reviewed serverkit implementation from #848).

**Scope**: only `contrib/server/{main.go,app.go,README.md,admin_persist_test.go}` touched — matches expectation, no k8s/oci/tls/storage leak-in.

**Flag -> Config wiring**: all five flags (`--admin`, `--persist`, `--state-file`, `--persist-metadata-only`, `--init-dir`) are bound in `main.go` and passed straight through to `serverkit.Config` in `app.go`'s `serverkitConfig()` — field names and defaults (`admin=true`, `persist=false`) match `cmd/cloudemu/serve.go` exactly. `--persist` without `--state-file` is rejected with `errStateFileRequired = errors.New("--persist requires --state-file")`, byte-for-byte identical to serve.go's validation and error text.

**No reimplementation**: `app.go` only assembles `serverkit.Config` fields; admin mount, restore/snapshot, and init-dir seeding all live in `server/serverkit` (confirmed `Admin`/`Persist`/`StateFile`/`PersistMetadataOnly`/`InitDir` are consumed only inside `serverkit.go`/`init.go`, not duplicated in contrib).

**Tests are non-tautological**:
- `TestAdminReset` seeds via `/_cloudemu/seed`, confirms presence via `/_cloudemu/snapshot`, POSTs `/_cloudemu/reset`, confirms absence — would fail if `--admin` weren't wired (the endpoints wouldn't mount).
- `TestPersistRoundTrip` genuinely round-trips through a real file: seeds a bucket, calls the blocking `stop()` (which cancels ctx and waits on `Serve`'s return), then boots a second app on the same `stateFile` and asserts the bucket is present. Verified `serverkit.App.Serve` returns `a.shutdown(servers)` directly and `shutdown()` calls `snapshotState` synchronously before returning — so `stop()` not returning until the snapshot is written is real, not a race.
- `TestInitDir` writes a fixture file and confirms it's seeded on boot via the snapshot check — exercises `--init-dir` through to `applyInitDir`.

**nolint justification**: both `//nolint:noctx` are on test-only `http.Get`/`http.Post` helper calls (`getSnapshot`, `post`) against an in-process test server — standard, not masking anything.

**README**: new table's defaults match the flags; the non-loopback admin warning documented in the README matches the actual `isLoopbackHost` gate in `serverkit.go`.

**Gates** (GOMAXPROCS=3):
- Core: `go build ./...` OK, `go vet ./...` OK, `go test -count=1 ./server/serverkit/... ./cmd/cloudemu/...` → both `ok`
- Contrib: `go build ./...` OK, `go vet ./...` OK, `go test -count=1 ./...` → `ok` (7.6s)
- `golangci-lint run --timeout=9m ./...` from `contrib/server` → `0 issues`

No blockers found. MERGE.

PR #851, opened 2026-08-28
feat(contrib/server): full flag-parity with cloudemu serve (OCI, k8s, storage engine, socket-degrade, TLS/latency/logging/auth)

NitinKumar004:
Brings the batteries-included `contrib/server` to full flag-parity with `cmd/cloudemu serve`. serverkit already wires OCI, the shared Kubernetes data-plane, TLS, latency, request logging, and auth enforcement — this PR only exposes the flags and threads them into `serverkit.Config`. Flag names and defaults mirror `serve.go` exactly.

Batches plan steps **D + E + F** (kept separate for reviewability in the plan; done as one PR since they are the same low-risk pattern). Does **not** touch the Dockerfile / docker.yml / ci.yml — that is the separate final PR G.

## D — OCI + Kubernetes
- `--oci-port` (default 4571), `--k8s-port` (default 4570, empty disables).
- `--providers` accepts `oci` (opt-in), parsed with a mirror of serve.go's `parseProviders`.
- Ports threaded into `serverkit.Config` (`Ports["oci"]`, `K8sPort`). serverkit already mirrors serve.go's OCI (no `SetK8sAPI`) and k8s wiring.

## E — Storage engine + socket-degrade + MODE banner
- `--storage=off|localfs` + `--storage-dir` → `config.WithStorageEngine(blobstore.New(root))` (6th engine); `--all-real` enables it.
- Docker-socket **degrade-to-memory**: when a docker-backed engine (`mysql`/`both`/`compute`/`containers`) is selected but the socket is absent, that capability drops its real option and records a `fell-back-to-memory (no docker socket)` MODE row + one stderr warning, instead of hard-failing. Flag-overrides-env precedence preserved.
- `engines.go` now returns the per-provider `[]config.Option` **and** a per-capability MODE table; `main.go` prints the MODE banner at startup (suppressed under `--quiet`; degrade warnings always shown).

## F — TLS / latency / logging / auth / advertise
- `--tls-cert` / `--tls-key` / `--tls-host` (repeatable), `--latency`, `--log-requests`, `--quiet`, `--enforce-auth`, `--advertise-host`, `--endpoints-file` → matching `serverkit.Config` fields (`Latency`/`EnforceAuth` via the Config fields serverkit maps to `WithLatency`/`WithEnforceAuth`).

## Tests (hand-rolled helpers, no new SDK deps)
- OCI endpoint binds and answers (`--providers aws,oci`).
- Kubernetes HTTPS endpoint reachable when `--k8s-port` set.
- Socket-degrade: forced probe seam asserts db=mysql/both, compute=docker, containers=docker degrade (MODE row + no real engine wired), the server still boots, and the degraded capability works in-memory — deterministic regardless of whether docker is present.
- enforce-auth: with `--enforce-auth` an unregistered-key request is rejected (403); without it dummy creds pass.
- StorageEngine: unit assertion that `--storage=localfs` wires `WithStorageEngine` into the built options (no S3 SDK dep).

README updated: new flags, MODE banner, socket-degrade behavior, and socket-mount security note.

### DIFF
--- a/contrib/server/README.md
+++ b/contrib/server/README.md
@@ -23,6 +23,12 @@ go run .
 # Real Postgres + real Redis (no Docker needed — embedded Postgres + miniredis):
 go run . --db=postgres --cache=redis
 
+# Persist object-storage bytes to a real local filesystem (no Docker):
+go run . --storage=localfs --storage-dir ./objects
+
+# Add the OCI endpoint and the shared Kubernetes data-plane:
+go run . --providers aws,azure,gcp,oci --k8s-port 4570
+
 # Everything real (Docker required for compute/containers):
 go run . --all-real
 ```
@@ -54,6 +60,17 @@ docker run --rm -p 4566:4566 -p 4568:4568 -p 4569:4569 \
   -e CLOUDEMU_DB=mysql -e CLOUDEMU_CONTAINERS=docker cloudemu-server
 ```
 
+Without the socket mount, Docker-backed engines **degrade to in-memory** (see the
+MODE banner section) rather than failing — so the same `docker run` works with or
+without the mount.
+
+> **Security note:** mounting `/var/run/docker.sock` grants the container
+> root-equivalent control of the host Docker daemon. Do it only for local dev on
+> a machine you trust; never mount the raw socket in shared or multi-tenant CI —
+> use a socket-proxy that restricts the API surface instead. If you cannot mount
+> a socket, just leave the Docker-backed engines off (or let them degrade) and use
+> the no-Docker engines (`postgres`, `redis`, `subprocess`, `localfs`).
+
 ## Engine flags / env
 
 Each capability is independently selectable; all default to `off` (in-memory).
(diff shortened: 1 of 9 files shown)


NitinKumar004 (commented):
Reviewed at 4cd92c26 (diff vs `development` touches only `contrib/server/**`, 9 files, +733/-114 — matches the PR description).

**Gates — all green**
- Core: `go build ./...`, `go vet ./...` clean; `go test -count=1 ./server/serverkit/... ./cmd/cloudemu/...` → both `ok`.
- Contrib: `go build ./...`, `go vet ./...` clean; `go test -count=1 ./...` → `ok` (10.3s, embedded-Postgres round-trip + Redis + port-release assertion all genuinely exercised); `go test -count=1 -race ./...` → `ok` (33.5s, clean).
- `golangci-lint run --timeout=9m ./...` from `contrib/server` (v2.11.4) → **0 issues**.
- `gofmt -l .` → clean. `go.mod`/`go.sum` are untouched by this diff (not part of the 9 changed files), so no tidy delta to evaluate here.

**Socket-degrade logic (engines.go)** — verified correct. Only `dbMySQL`/`dbBoth`/`backingDocker` (compute, containers) branches call `dockerAvailable()`; `dbPostgres`, `cacheRedis`, `functionsSubprocess`, `storageLocalFS` never gate on it, so the no-Docker engines can't be wrongly degraded. `errDockerUnavailable` is gone — every degrade path returns `(nil option, engineMode{status: modeMemory, detail: reasonNoDocker}, nil error)`, never an error, so a socket-less run still boots (confirmed live via `TestDegradedEngineStillBoots`, which creates+describes an RDS instance against the degraded in-memory db). Found no path where a real engine is dropped while Docker *is* available — every `modeReal` branch is reached only after a successful `dockerAvailable()` check (or needs no Docker at all). The probe is a plain `func() bool` seam: production wires the real `dockerAvailable()` (PATH lookup), tests inject deterministic `noDocker`/`withDocker`. Flag-overrides-env precedence is preserved by ordinary `flag.Parse` semantics (`engineEnvOr` only supplies the *default*).

**MODE table** — can't diverge from what's wired: each `<capability>Option` function returns the `config.Option` and its `engineMode` from the same switch arm, so "real" is never reported alongside a dropped option or vice versa.

**enforce-auth** — `TestEnforceAuthRejectsInvalidSignature` genuinely asserts a 403/`InvalidClientTokenId` rejection for an unregistered SigV4 key with `--enforce-auth` on, and a clean pass with it off; both subtests pass. Wiring (`Config.EnforceAuth` → `config.WithEnforceAuth()`) lives in `server/serverkit` and is unchanged by this PR.

**StorageEngine teardown** — `config.Options.EngineClosers()` (`config/engine.go:15`) already includes `StorageEngine`, and `providers/aws/aws.go` wires `engineClosers: o.EngineClosers()` into `Provider.Close()` — the cascade reaches blobstore with no code change needed here. `TestStorageEngineOffByDefault`/`TestStorageEngineWiresOption` confirm off-by-default and correct wiring when `--storage=localfs` is set.

**OCI/k8s** — `--oci-port`/`--k8s-port` thread into `serverkit.Config.Ports["oci"]`/`K8sPort` correctly; `--providers` accepts `oci` via a `parseProviders` that mirrors `cmd/cloudemu/serve.go`'s byte-for-byte. `TestOCIEndpointBinds` and `TestKubernetesEndpointReachable` both bind and answer for real (HTTP OCI listener, HTTPS k8s listener with the self-signed serving cert).

**Flag parity** — diffed every flag against `cmd/cloudemu/serve.go`: names and defaults match exactly (providers, host, advertise-host, aws/azure/gcp/oci/k8s ports, account-id, azure-subscription, region, project-id, latency, tls-cert/key/host, admin, log-requests, quiet, enforce-auth, persist/state-file/persist-metadata-only, init-dir, endpoints-file, shutdown-timeout). Only cosmetic gap: `--enforce-auth`'s help string in `contrib/server/main.go` is a one-liner vs. `serve.go`'s detailed AWS/Azure semantics paragraph — doc-only, no behavioral difference since both map to the same `Config.EnforceAuth` field.

**Deviations** — MODE banner printing before `Serve()` is the only correct option: `serverkit.App.Serve` blocks and prints its own endpoint banner from inside that call, so this is the last synchronous point available (correctly documented in the code). `newApp` wrapper removal is clean — `main.go`'s `run()` now inlines exactly what `newApp` used to do (`buildOptions` → `newAppFromOptions`), no behavior lost. `orOff()` only affects a directly-constructed `engineSelection` (e.g. tests) or an explicit `--db=""`, which is indistinguishable from omitting the flag — a genuinely invalid value like `--db=bogus` still hits the error path (verified: `invalid --db value (want off|postgres|mysql|both): "bogus"`).

No High or Medium findings. Verified clean — approving.

PR #852, opened 2026-08-28
ci(contrib-server): publish cloudemu:engines batteries image + contrib CI test/smoke build

NitinKumar004:
Final step of the Docker/library parity work — publishes the batteries-included contrib/server as `ghcr.io/stackshy/cloudemu:engines`.

## Changes (all additive — existing :latest publish + CI jobs untouched)
- **contrib/server/Dockerfile** — bundles `python3`+`nodejs` (for the subprocess functions engine), EXPOSEs 4570 (k8s) + 4571 (oci), adds `CLOUDEMU_STORAGE=off`. Engines default off, so a no-arg run is pure in-memory. Build context stays repo root (relative-replace + serverkit import).
- **.github/workflows/docker.yml** — new job publishing the engines image on `v*` tags, with its own `permissions: {contents: read, packages: write}`, explicit `type=raw,value=engines` tag (+ `:vX.Y.Z-engines`/`:X.Y-engines` via semver suffix), separate cache scope. No plain `latest` from this job.
- **.github/workflows/ci.yml** — adds an advisory `go test` (with embedded-postgres cache) + a paths-gated Docker smoke build to the existing contrib-server job.

Now `docker run -v /var/run/docker.sock:/var/run/docker.sock ghcr.io/stackshy/cloudemu:engines` gives real Postgres/Redis/functions/containers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,11 +156,39 @@ jobs:
         run: |
           go build ./...
           go vet ./...
+      # embedded-postgres downloads its Postgres binary on first run; cache it so
+      # the CI-safe test tier (no docker socket needed — the E2E uses embedded
+      # Postgres, the degrade tests run WITHOUT docker) doesn't refetch each run.
+      - uses: actions/cache@v4
+        with:
+          path: ~/.embedded-postgres-go
+          key: embedded-postgres-${{ runner.os }}
+      # Advisory (report-only), matching the contrib-* convention: these tests
+      # need no docker socket but the embedded-Postgres binary download can flake;
+      # promote to blocking once proven stable.
+      - name: go test (report-only)
+        working-directory: contrib/server
+        run: "go test -count=1 ./... || echo '::warning::contrib/server go test reported findings (advisory)'"
       - name: install golangci-lint v2
         run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b "$(go env GOPATH)/bin"
       - name: golangci-lint run (report-only)
         working-directory: contrib/server
         run: "$(go env GOPATH)/bin/golangci-lint run --timeout=9m ./... || echo '::warning::golangci-lint reported findings (advisory)'"
+      # Only run the heavy multi-stage engines image build when the trees it is
+      # built from change, so it doesn't fire on every repo PR.
+      - name: filter engines-image paths
+        id: engines_paths
+        uses: dorny/paths-filter@v3
+        with:
+          filters: |
+            engines:
+              - 'contrib/server/**'
+              - 'contrib/realengine/**'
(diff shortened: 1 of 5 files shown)


NitinKumar004 (commented):
Reviewed at c657d5b0 in an isolated worktree, diffed against origin/development merge-base (d787f98f). Diff touches exactly the 4 files listed in the PR description.

**Existing jobs untouched.** Both `docker.yml`'s existing `publish` job and `ci.yml`'s existing `contrib-server` steps (`build & vet`, `golangci-lint run`) are unchanged except for purely additive hunks — confirmed via `git diff` against the merge-base.

**`publish-engines` job (docker.yml).** Own `permissions: {contents: read, packages: write}` block. Tags use an explicit `type=raw,value=engines` + `type=semver,...,suffix=-engines` list rather than a global `flavor.suffix` (which would only ever produce `latest-engines`, never a bare `engines` tag) — correct call. `file: contrib/server/Dockerfile`, `context: .` (repo root, required for the module's relative `replace` directives). Shares the workflow's `on: push: tags: v*` trigger. Separate GHA cache scope (`scope=engines`). Does not emit a plain `latest`, so no collision with the root image's tag.

**ci.yml additions.** The `go test` step for `contrib/server` is report-only (`|| echo '::warning::...'`), matching the existing `golangci-lint run (report-only)` convention already in that job, with an embedded-postgres binary cache. The Docker smoke-build step is correctly paths-gated (`dorny/paths-filter@v3` on `contrib/server/**`, `contrib/realengine/**`, `contrib/dockerengine/**`, `server/serverkit/**`) so it won't fire on unrelated PRs. Existing steps in the job stay unconditional.

**Dockerfile.** `python3`/`nodejs` added via `--no-install-recommends` with `rm -rf /var/lib/apt/lists/*`; comment correctly points at `contrib/realengine/functions/functions_runtime.go` as the consumer. `EXPOSE` adds 4570 (k8s) + 4571 (oci). `CLOUDEMU_STORAGE=off` added alongside the other all-off engine defaults, so a no-arg run stays pure in-memory. `debian:12-slim` runtime base retained (needed for embedded-postgres glibc binaries). `COPY . .` from repo root preserved.

**actionlint.** `docker.yml`: exit 0, clean. `ci.yml`: exit 1, but the 6 shellcheck findings (SC2046 unquoted `$(...)`, SC2044 for-loop-over-find) are pre-existing — ran actionlint against the pre-PR baseline and got the identical 6 findings with the same content, just shifted line numbers matching the inserted line count. No new findings from this PR.

**Go gates.** Core `go build ./...` and `contrib/server`'s `go build ./...` + `go vet ./...` all pass.

Local `docker build` wasn't run — Docker was unresponsive in the review environment; not blocking since CI's own smoke-build step (gated on the paths this PR touches) covers it.

No blockers found. Scope is clean, additive-only to shared CI, and the new job/steps match the stated design intent.

Issue #578, opened 2026-08-25
Azure: deferred real-user e2e follow-ups


NitinKumar004:
Deferred items from the Azure real-user e2e work, to pick up.

- [ ] KeyVault Certificates lifecycle (create/import/get/list)
- [ ] KeyVault Vault ARM management (create/list vaults)
- [ ] Service Bus DLQ sub-queue semantics + renew-lock

NitinKumar004:
Closed by #853: KeyVault Vault ARM management (Microsoft.KeyVault/vaults CreateOrUpdate/Get/List/Delete with scope enforcement + Resource Graph discovery + persistence) added end-to-end. (Certs lifecycle and ServiceBus DLQ were already done in prior work.)

Issue #409, opened 2026-08-20
Resource Graph returns no ARM-created resources (breaks ARG-based discovery) + Azure model coverage gaps


NitinKumar004:
Environment: ghcr.io/stackshy/cloudemu:2.5 (image revision f147a914), standalone server; Azure endpoint :4568, AWS :4566.
▎
▎ Evaluating cloudemu as a local Azure emulator for a cost/discovery pipeline that reads inventory via Azure Resource Graph. Hit one blocking bug and a few coverage/fidelity gaps — details + repro below. Happy to contribute PRs.
▎
▎ 1. 🔴 Bug (blocking): ARM-created resources never appear in Resource Graph
▎
▎ PUT https://localhost:4568/subscriptions/sub1/resourceGroups/rg1/providers/Microsoft.Compute/virtualMachines/vm1?api-version=2023-07-01
▎   {"location":"eastus","properties":{"hardwareProfile":{"vmSize":"Standard_D2s_v5"}}}
▎   → 200, {"provisioningState":"Succeeded", ...}   # created fine
▎
▎ POST https://localhost:4568/providers/Microsoft.ResourceGraph/resources?api-version=2021-03-01
▎   {"subscriptions":["sub1"],"query":"Resources | project id,type | limit 20"}
▎   → {"count":0,"data":[],"totalRecords":0}         # unfiltered — still empty
▎ The ARM data plane and the Resource Graph query engine appear decoupled: a resource created via the ARM REST API is not indexed into ARG, even for a fully-modeled type (Microsoft.Compute/virtualMachines) that is in the ARG type map. Any tool that discovers inventory via Microsoft.ResourceGraph/resources sees nothing after creating resources via ARM. Expected: ARG returns resources created via the ARM API.
▎
▎ 2. Coverage: MySQL/PostgreSQL Flexible Server missing highAvailability
▎
▎ server/azure/mysqlflex/types.go models availabilityZone but no highAvailability / mode. Real servers expose properties.highAvailability.mode (Disabled/SameZone/ZoneRedundant), which drives HA (2×) cost. A PUT including highAvailability.mode returns 200 with the field silently dropped.
▎
▎ 3. Coverage: Microsoft.Cache/redis missing clustering fields
▎
▎ server/azure/cache/types.go — SKU Capacity/Family are stub defaults and there's no shardCount / replicasPerShard. Premium clustered caches can't be represented (these drive node-count cost).
▎
[...]

PR #853, opened 2026-08-28
feat(azure/keyvault): Microsoft.KeyVault/vaults ARM control plane (#578)

NitinKumar004:
## Summary

Finishes the open **#578** item "KeyVault Vault ARM management" by modeling the Key Vault *vault* resource (`Microsoft.KeyVault/vaults`) end-to-end, and confirms **#409 gap 4** (ARM PUT dropping unmodeled properties) is already handled.

### #578 — Vault ARM control plane (added end-to-end)

Previously only the per-vault **data plane** (secrets/keys/certs) existed; a vault name was accepted implicitly in data-plane URLs, with no resource-manager record. This adds the control plane:

- **Driver surface** — new `secretsdriver.KeyVaultVaults` optional interface (`CreateOrUpdateVault` / `GetVault` / `ListVaults(scope)` / `DeleteVault`) with the ARM vault property model (tenantId, SKU, access policies or RBAC, enabledForDeployment/DiskEncryption/TemplateDeployment, soft-delete/purge-protection, public network access, vaultUri).
- **Provider store** — in-memory store on the Azure Key Vault mock, keyed by (globally-unique) vault name, recording the request scope; deep-copied in/out so callers never alias the store. Included in the mock **snapshot/restore** for persistence.
- **ARM wire handler** — `Microsoft.KeyVault/vaults` handler over the shared `azurearm` codec: `PUT` (create 201 / replace 200, LRO completes inline), `GET`, `DELETE`, and RG-scoped + subscription-scoped list. Scope is enforced (a vault created in one resource group 404s when addressed under another).
- **Wiring** — registered in the Azure server factory (unique ARM provider name, before the blob fallback) via a type-assertion on the KeyVault backend; the provider factory feeds the same backend into the discovery engine.
- **Resource Graph** — new `walkKeyVaults` discovery walker plus the `microsoft.keyvault/vaults` ⇄ `secrets/Vault` type mapping, so an ARM-created vault surfaces in ARG queries.
- Regenerated `docs/coverage`.

### #409 gap 4 — unmodeled ARM PUT properties (already fixed)

Assessment: this is **already implemented** the architecturally-preferred way (Option A) via the `echoUnmodeledProperties` overlay in `server/azure/echo_properties.go`, wired at the server-assembly level so it benefits **all** ARM resource handlers generically — it captures request properties a handler didn't model, merges them back on the create response and later GET/LIST, handles nested objects/arrays and PATCH-vs-PUT semantics, evicts on delete, and sanitizes write-only secrets. No codec change (and no `DisallowUnknownFields`) is needed. This PR adds a test asserting an unmodeled property round-trips through the new vault resource.

## Tests

- Provider level (`providers/azure/keyvault`): create/replace round-trip, returned-copy isolation, list by RG and by subscription, delete, name validation.
- Server level (`server/azure/keyvault`): full ARM lifecycle (create 201 → get → replace 200 → list ×2 → delete → 404), cross-resource-group scope 404, unmodeled-property preservation (#409 gap 4), and vault surfacing in Resource Graph.

Gates: `go build ./...`, `go vet`, `go test` (server/azure, providers/azure, services/resourcediscovery, persist) all green; `go generate` committed; golangci-lint clean of new gating findings.


NitinKumar004 (commented):
Reviewed against `origin/development` (merge-base `d787f98f`) in an isolated worktree at head `99d29c79`.

**Gates:** `go build ./...` clean · `go vet ./...` clean · `go test -count=1 ./server/azure/... ./providers/azure/... ./services/resourcediscovery/... ./persist/...` all green · full `go test -count=1 ./...` green, no failures · `go generate ./...` produces no diff (docs/coverage regenerated correctly and matches the driver) · `go mod tidy` clean · `golangci-lint` on the touched packages: only the two expected `hugeParam` advisories in `vault_arm_azure.go` (matching the existing `CreateSecret` convention), no other new findings on any new file.

Architecture checks out: `KeyVaultVaults` is a type-asserted optional capability (`providers/azure/azure.go:247`, `server/azure/azure.go:521`), doesn't force AWS/GCP or other Azure services to implement it. Provider store deep-copies on every read/write (`cloneVaultInfo`/`copyVaultProperties`), verified by `TestGetVaultReturnsCopy`. `armVaults` is included in snapshot/restore (`providers/azure/keyvault/snapshot.go`) and `persist/completeness_test.go` passes. ARM handler registration is a unique provider name (`Microsoft.KeyVault/vaults`), disjoint from every other Azure handler and from the data-plane secrets/keys/certs handlers on this same backend — registration order relative to them is genuinely unconstrained (confirmed by reading `azurearm.ParsePath` dispatch). `walkKeyVaults` + the `microsoft.keyvault/vaults` ⇄ `secrets/Vault` type mapping are wired correctly into Resource Graph.

Two issues found by tracing behavior end-to-end:

**[High] `server/azure/keyvault/vault_arm_test.go` never runs on any real build.** After stripping `_test.go`, the remaining basename `vault_arm` ends in `arm`, which Go's filename-based implicit build-constraint convention (`*_GOARCH.go`) interprets as "only build for `GOARCH=arm`" (32-bit ARM). Verified with `go list -json ./server/azure/keyvault/` on darwin/arm64 **and** cross-compiled for `GOOS=linux GOARCH=amd64` (the normal CI target): both report `"IgnoredGoFiles": ["vault_arm_test.go"]`. `go test -run TestVaultARM ./server/azure/keyvault/...` reports "no tests to run." None of `TestVaultARMLifecycle` (PUT 201/200, GET, list×2, DELETE, 404), `TestVaultARMScopeMismatch` (cross-RG 404), `TestVaultARMUnmodeledPropertyPreserved` (the #409 gap-4 proof), or `TestVaultInResourceGraph` are ever compiled or executed — the package reports `ok` only because it silently has zero tests from this file, not because they passed. The PR's central verification claims (full ARM lifecycle, scope enforcement, and the #409 gap-4 round-trip) are currently unverified by `go test ./...` on any normal machine or CI runner. Fix: rename the file so its basename doesn't end in a GOOS/GOARCH token, e.g. `vault_arm_handler_test.go` or `vault_control_plane_test.go`.

**[Medium] PUT does not enforce resource-group scope, unlike GET/DELETE.** `createOrUpdateVault` (`server/azure/keyvault/vault_arm_operations.go:14-35`) determines create-vs-replace purely from `GetVault(name)` by name, with no check that the existing record's resource group matches the URL's resource group — unlike `getVault`/`deleteVault`, which both explicitly compare `info.Scope.Matches(...)`. Reproduced: create `vault4` under `sub-kv/rg-kv` (201), then PUT the same name under `sub-kv/other-rg-4` (200 "replace") — the stored scope is silently overwritten: a subsequent GET under `rg-kv` now 404s, and GET under `other-rg-4` now succeeds. Real Key Vault vault names are globally unique and fixed to their original resource group; a PUT under a different resource group with an existing name returns `409 Conflict` (`VaultAlreadyExists`), not a silent relocation. This also means the PR description's claim ("a vault created in one resource group 404s when addressed under another") only holds for GET/DELETE, not PUT.

Minor, non-blocking: `DeleteVault` is a hard purge with no soft-delete/retention state for the vault resource itself (unlike this same package's existing secret soft-delete semantics), so a delete immediately frees the name for reuse — a documented-limitation comment would help, but this matches the #578 scope as described (Create/Get/List/Delete only).

Issue #493, opened 2026-08-24
Enforce request auth: verify SigV4 + gate on IAM policy evaluation


NitinKumar004:
Today the wire layer accepts any credentials — SigV4 is parsed but not verified, and IAM policy evaluation (CheckPermission) exists in the backend but does not gate incoming requests. Fine for dev/test; cloudemu is not currently an auth boundary.

Goal: opt-in enforcement so requests are authenticated + authorized like real AWS (default stays permissive so existing dev/test flows are unaffected).

- Verify SigV4 signatures against registered access keys
- Resolve the principal from the credential scope and pass it to handlers
- Gate requests through the existing IAM CheckPermission(principal, action, resource)
- Return real AWS deny shapes (AccessDenied / UnauthorizedOperation)
- Config flag to toggle enforcement (off by default)

NitinKumar004:
Shipped: opt-in AWS request auth — SigV4 signature verification against registered access keys (#838) + action-level IAM CheckPermission authorization gate with real AWS deny shapes (#840), behind WithEnforceAuth (default off); Azure claims-based authN in #842. Remaining hardening (ASIA verification, body-binding, expiry) and IAM NotAction/Condition are tracked separately in #839 and #841.

PR #854, opened 2026-08-28
fix(iam): NotAction/NotResource/Condition evaluation + resource-scoped authz (#841)

### LINKED ISSUE #841: IAM policy evaluation: support NotAction/NotResource/Condition (now on the authz path via #493)
The IAM policy evaluator (providers/aws/iam evaluatePolicy) only reads Action/Resource. Statements using NotAction, NotResource, or Condition are parsed with a nil Action and silently skipped. Consequences now that CheckPermission gates the wire (#493 PR2):
- A Deny with NotAction is under-enforced (over-permit).
- A conditional allow/deny is applied unconditionally (Condition ignored).

Add NotAction/NotResource matching (inverse semantics) and basic Condition-key evaluation so the authorization decision matches real IAM. Pre-existing; low blast radius while auth is opt-in.

NitinKumar004:
## Summary

Fixes #841. IAM policy evaluation only read `Action`/`Resource`, silently dropping `NotAction`, `NotResource` and `Condition`, and the authorization gate hardcoded `resource="*"`. Now that `CheckPermission` gates the wire (#493), those made allow/deny decisions wrong (a `NotAction` Deny under-enforced, conditional statements applied unconditionally, resource-scoped policies never bit). This makes the evaluator match real IAM semantics end-to-end.

## What changed

**Statement model** — added `NotAction`, `NotResource` and `Condition` to the policy statement. `Action`/`NotAction` (and `Resource`/`NotResource`) are mutually exclusive: `NotAction` matches every action except those listed, so a `Deny` with `NotAction` fails closed. Explicit-Deny-wins and wildcard matching are preserved.

**Reusable condition evaluator** (`providers/aws/iam/condition.go`) — a focused, per-operator evaluator decomposed by family (no `nolint`), evaluated against a request condition context with real AWS absent-key rules (plain condition → no match, `...IfExists` → match).

Operators: `StringEquals` / `StringNotEquals` / `StringEqualsIgnoreCase` / `StringNotEqualsIgnoreCase` / `StringLike` / `StringNotLike`, `Bool`, `NumericEquals` / `NotEquals` / `LessThan(Equals)` / `GreaterThan(Equals)`, `DateEquals` / `NotEquals` / `LessThan(Equals)` / `GreaterThan(Equals)`, `ArnEquals` / `ArnLike` / `ArnNotEquals` / `ArnNotLike`, `IpAddress` / `NotIpAddress` (CIDR), `Null`, and the `...IfExists` variants of all.

**Context keys threaded from the request**: `aws:SourceIp`, `aws:CurrentTime`, `aws:SecureTransport`, `aws:PrincipalArn`, `aws:username`, `aws:RequestedRegion`.

**Resource-scoped authz** (`server/aws/authzgate.go`) — derives the target resource ARN from the request (DynamoDB table from the JSON body today; `*` fallback for actions whose resource can't be derived) and threads the condition context via the new `ContextualAuthorizer` capability. Public API stays backward-compatible: `CheckPermission` delegates to `CheckPermissionWithContext` with an empty context (today's behavior); `SimulatePrincipalPolicy` / `SimulateCustomPolicy` honor `ContextEntries`.

## Tests

- Provider (`providers/aws/iam`): `NotAction` Deny blocks the complement / allows the listed (fail-closed), `NotResource` likewise, each condition operator allow/deny against a supplied context, IfExists present-vs-absent, `Null`, multi-key AND / multi-value OR, `CheckPermissionWithContext`, resource-scoped allow/deny, and `SimulatePrincipalPolicy` reflecting conditions.
- Server (`server/aws`): real SigV4-signed DynamoDB request through the authz gate — a resource-scoped policy allows the matching table and 403s a non-matching one.

## Deferred

Resource derivation currently covers the DynamoDB table ARN (the JSON-RPC service with resource-level policies and the enforced protocol); other actions fall back to `*`. Additional per-service derivations can follow incrementally.

### DIFF
--- a/docs/coverage/aws/iam.md
+++ b/docs/coverage/aws/iam.md
@@ -60,6 +60,14 @@ AccessKeyResolver is an optional capability: an IAM implementation that can
 | --- | --- |
 | `AccessKeyByID` |  |
 
+### ContextualAuthorizer
+
+ContextualAuthorizer is an optional capability: an IAM implementation that can
+
+| Operation | Description |
+| --- | --- |
+| `CheckPermissionWithContext` |  |
+
 ### PolicyInspector
 
 PolicyInspector is an optional capability: an IAM implementation that can
--- a/docs/coverage/azure/iam.md
+++ b/docs/coverage/azure/iam.md
@@ -60,6 +60,14 @@ AccessKeyResolver is an optional capability: an IAM implementation that can
 | --- | --- |
 | `AccessKeyByID` |  |
 
+### ContextualAuthorizer
+
+ContextualAuthorizer is an optional capability: an IAM implementation that can
+
+| Operation | Description |
+| --- | --- |
+| `CheckPermissionWithContext` |  |
+
 ### PolicyInspector
 
 PolicyInspector is an optional capability: an IAM implementation that can
--- a/docs/coverage/coverage.json
+++ b/docs/coverage/coverage.json
@@ -5325,6 +5325,15 @@
         ]
       },
       {
+        "name": "ContextualAuthorizer",
+        "doc": "ContextualAuthorizer is an optional capability: an IAM implementation that can",
+        "operations": [
+          {
+            "name": "CheckPermissionWithContext"
+          }
+        ]
+      },
+      {
         "name": "PolicyInspector",
         "doc": "PolicyInspector is an optional capability: an IAM implementation that can",
         "operations": [
--- a/docs/coverage/gcp/iam.md
+++ b/docs/coverage/gcp/iam.md
@@ -60,6 +60,14 @@ AccessKeyResolver is an optional capability: an IAM implementation that can
 | --- | --- |
 | `AccessKeyByID` |  |
 
+### ContextualAuthorizer
+
+ContextualAuthorizer is an optional capability: an IAM implementation that can
+
+| Operation | Description |
+| --- | --- |
+| `CheckPermissionWithContext` |  |
+
 ### PolicyInspector
 
 PolicyInspector is an optional capability: an IAM implementation that can
--- a/docs/coverage/oci/identity.md
+++ b/docs/coverage/oci/identity.md
@@ -60,6 +60,14 @@ AccessKeyResolver is an optional capability: an IAM implementation that can
 | --- | --- |
 | `AccessKeyByID` |  |
 
+### ContextualAuthorizer
+
+ContextualAuthorizer is an optional capability: an IAM implementation that can
+
+| Operation | Description |
+| --- | --- |
+| `CheckPermissionWithContext` |  |
+
 ### PolicyInspector
 
 PolicyInspector is an optional capability: an IAM implementation that can
--- a/providers/aws/iam/condition.go
+++ b/providers/aws/iam/condition.go
@@ -1,0 +1,332 @@
+package iam
+
+import (
+	"net"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// ConditionContext carries the request condition keys available for policy
+// evaluation (e.g. "aws:SourceIp", "aws:CurrentTime", "aws:SecureTransport",
+// "aws:PrincipalArn", "aws:RequestedRegion"). AWS condition keys are matched
+// case-insensitively, so lookups normalize the key. A nil or empty context means
+// no keys are available: a plain condition referencing a key then fails to
+// match, while its ...IfExists variant passes (real IAM rules).
+type ConditionContext map[string]string
+
+// get resolves a condition key case-insensitively, reporting whether it was
+// present in the request context.
+func (c ConditionContext) get(key string) (string, bool) {
+	if c == nil {
+		return "", false
+	}
+
+	if v, ok := c[key]; ok {
+		return v, true
+	}
+
+	for k, v := range c {
+		if strings.EqualFold(k, key) {
+			return v, true
+		}
+	}
+
+	return "", false
+}
+
+// evaluateConditions reports whether every operator/key in a statement's
+// Condition block is satisfied against the request context. AWS combines them
+// with AND across operators and AND across keys within an operator; within a
+// single key the listed values combine with OR (a negative operator requires
+// that none match). An empty condition block is vacuously true.
+func evaluateConditions(conds map[string]map[string]any, cctx ConditionContext) bool {
+	for rawOp, keyVals := range conds {
+		for key, raw := range keyVals {
+			if !evaluateConditionKey(rawOp, key, toStringSlice(raw), cctx) {
+				return false
+			}
+		}
+	}
+
+	return true
+}
+
+// evaluateConditionKey evaluates one operator against one condition key and its
+// policy-supplied values. It resolves the request value from the context,
+// applies the ...IfExists rule for an absent key, and dispatches to the operator
+// family. The Null operator is handled first because it is defined in terms of
+// key presence, not the key's value.
+func evaluateConditionKey(rawOp, key string, values []string, cctx ConditionContext) bool {
+	base, ifExists := splitIfExists(rawOp)
+
+	ctxVal, present := cctx.get(key)
+
+	if strings.EqualFold(base, "Null") {
+		return evalNull(present, values)
+	}
+
+	if !present {
+		// A missing key never matches a plain condition; the ...IfExists variant
+		// passes so the statement is gated only when the key is actually supplied.
+		return ifExists
+	}
+
+	return evalPresentOperator(base, ctxVal, values)
+}
+
+// evalPresentOperator evaluates an operator whose key is present. String-shaped
+// operators (String*, Arn*, Ip*, Bool) share the value-comparator path; the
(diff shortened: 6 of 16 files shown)


NitinKumar004 (commented):
## Security-focused review: NotAction/NotResource/Condition + resource-scoped authz

Verified the fail-closed properties by reasoning through the concrete allow/deny scenarios, then ran the gate matrix on the pushed commit (29ad50d). Findings below.

### Fail-closed verification (all confirmed correct)

**`actionMatches`/`resourceMatches` (providers/aws/iam/iam.go:770-793)**: Action/NotAction and Resource/NotResource are correctly mutually exclusive via the `switch` — a statement with neither returns `false`, so it can never spuriously match. Traced the canonical allow-list-via-deny case:
```json
{"Effect":"Deny","NotAction":["s3:GetObject"],"Resource":"*"}
```
`s3:PutObject` → `matchesAction(["s3:GetObject"], "PutObject")=false` → `actionMatches=true` → Deny fires (blocked, correct). `s3:GetObject` → `actionMatches=false` → Deny doesn't apply (exempted, correct). This is fail-closed as intended — "deny everything except the listed action" actually denies everything except the listed action.

**Explicit-deny-wins** (`decide` in providers/aws/iam/simulate.go:70-92, unchanged by this PR): deny short-circuits allow regardless of statement order across documents; `CheckPermissionWithContext` treats anything `!= decisionAllowed` (both explicit and implicit deny) as denied. Confirmed intact through the boundary-intersection path too (iam.go:915-924) — a boundary that doesn't explicitly allow still denies even when the identity policy allows.

**Condition operators** — spot-checked the negative/IfExists/Null families against condition.go:
- `StringNotEquals`/`ArnNotLike`/`NotIpAddress` all route through `matchAny(values, negate=true, ...)` (condition.go:112-120), which requires *none* of the values to match, not "any" — correct.
- `...IfExists`: absent key → `ifExists` (true, condition passes); plain operator + absent key → `false` (condition fails). Verified this holds for negated `...IfExists` variants too (`splitIfExists` strips the suffix before the present/absent branch, so `StringNotEqualsIfExists` with an absent key still returns `true`, matching AWS).
- `Null:true` → key must be absent; `Null:false` → key must be present; multi-value OR confirmed in `evalNull`.
- Multi-key AND / multi-value OR: confirmed via `evaluateConditions` (AND across operator+key pairs) and `matchAny` (OR within a key's values) — matches `TestConditionMultipleKeysAnd`/`TestConditionValueListOr`.
- `IpAddress` uses real CIDR containment via `net.ParseCIDR`/`Contains`, not string prefix matching.
- Numeric/date parse failures on the **request** value don't fail open: `evalNumeric`/`evalDate` return `false` on an unparseable `ctxVal` regardless of whether the operator is negated (condition.go:233-236, 273-276) — so a malformed date/number can't be turned into a match via a `NotEquals` variant.

One narrower edge case worth a look: for `NotIpAddress`, `ipMatch` re-parses `ctxVal` inside the per-value predicate rather than up front. If `ctxVal` itself is unparseable, every value's predicate returns `false`, and `matchAny`'s negate branch then returns `true` (condition "satisfied", since none of the CIDRs matched). So a garbage `aws:SourceIp` would pass a `NotIpAddress` guard rather than failing it. Not reachable through the wire gate today (see below — the gate always supplies a real parsed IP), but it is reachable through `CheckPermissionWithContext`/`SimulatePolicy` directly if a caller supplies a bogus value. Low severity, but the numeric/date functions already get this right by checking `ctxVal` parseability once up front before the per-value loop — `ipMatch` could do the same for consistency.

### Resource-scoped authz (server/aws/authzgate.go)

`deriveResource` correctly returns `"*"` for every non-DynamoDB service, identical to the pre-PR hardcoded `"*"` — verified no other currently-enforced action's resource check changes. For DynamoDB, this is actually a **fix** to a previously-silent bug: before this PR, `CheckPermission(principal, action, "*")` meant a table-scoped `Deny` (`Resource: arn:...:table/X`) could never match the literal resource `"*"` being checked, so table-scoped Denies were silent no-ops. Confirmed via `TestAuthzGateResourceScoped` (server/aws/authzgate_test.go) that the fix works end-to-end over real SigV4-signed requests.

`CheckPermission` (no context) delegates to `CheckPermissionWithContext(ctx, principal, action, resource, nil)` — byte-identical decision for any policy with no `Condition` block (`evaluateConditions` on a nil map short-circuits to `true`), confirmed by `TestCheckPermissionWithContext`'s last assertion (plain `CheckPermission` sees a Condition-guarded Allow as unmatched).

### Finding: `aws:SourceIp` is derived from an unauthenticated, caller-controlled header

`clientIP` (server/aws/authzgate.go:169-183) prefers the first hop of `X-Forwarded-For` over `r.RemoteAddr` unconditionally:
```go
if fwd := r.Header.Get("X-Forwarded-For"); fwd != "" {
    if i := strings.IndexByte(fwd, ','); i >= 0 {
        return strings.TrimSpace(fwd[:i])
    }
    return strings.TrimSpace(fwd)
}
```
cloudemu's wire server has no reverse proxy or trusted-hop concept anywhere in the codebase (grepped for `X-Forwarded-For`/`TrustedProxy` — this is the only occurrence). Any direct caller can set this header on the request they're sending, so `aws:SourceIp` is fully attacker-controlled input, not a verified property of the connection. That means every `IpAddress`/`NotIpAddress` condition this PR adds support for is bypassable by simply setting `X-Forwarded-For` to a value inside (or outside, for `NotIpAddress`) the target CIDR — e.g. a `Deny` gated on `NotIpAddress: officeCIDR` (a common "block everyone except the office network" pattern) is defeated by `curl -H "X-Forwarded-For: <office-ip>" ...`. Since this is new code introduced by this PR (not pre-existing), and IP-conditioned policies are exactly the capability being shipped, I'd suggest either dropping the `X-Forwarded-For` fallback (use `RemoteAddr` only, which is unspoofable for a direct connection) or gating it behind an explicit trusted-proxy config the way real front-door services do. As written, IP-based Allow/Deny conditions provide no actual restriction against a caller.

### Note: `chaos.WrapIAM` silently drops Condition enforcement

`features/chaos/wrappers_iam.go`'s `chaosIAM` embeds `iamdriver.IAM` (the interface), not the concrete `*iam.Mock`, and doesn't implement `CheckPermissionWithContext`. If a library caller does `chaos.WrapIAM(provider.IAM, engine)` and passes the wrapped driver into `server.New(...)` as `Drivers.IAM`, `authorize()`'s type assertion for `iamdriver.ContextualAuthorizer` (server/aws/authzgate.go:127) fails, and it silently falls back to `CheckPermission` with a `nil` context — every `Condition` block in every policy (not just IP ones) stops being evaluated against the real request, so a Condition-guarded `Deny` (e.g. "deny unless MFA present") would never fire. This doesn't affect the default/documented wire-server wiring (`providers/aws.Provider.IAM` is passed through unwrapped, confirmed via `server/aws/aws.go:265`/`providers/aws/aws.go:143`), so it's not a regression of anything in the standard path, but it's a silent capability gap for anyone composing `chaos.WrapIAM` with `EnforceAuth`. Might be worth a comment on `ContextualAuthorizer`'s doc, or forwarding the method through `chaosIAM`, so it doesn't surprise someone later.

### Minor / non-blocking

- `deriveResource` only derives DynamoDB's single `TableName` field; batch ops (`BatchGetItem`/`BatchWriteItem`) address tables via a `RequestItems` map and fall back to `"*"`, so a table-scoped policy won't restrict those calls specifically. Reasonable scope for this PR given the docstring says so explicitly; flagging for awareness only.
- `parseContextEntries` (server/aws/iam/simulate.go) takes only the first value of a multi-valued `ContextKeyValues.member`, so `SimulatePolicy` calls with multi-valued context keys won't fully reflect AWS's OR semantics. `SimulatePolicy`-only, doesn't affect live enforcement.

### Gates (commit 29ad50db85fbd8a545a251a6c69cfa76e2774ed5)

- `go build ./...` — clean
- `go vet ./...` — clean
- `go test -count=1 ./providers/aws/iam/... ./server/aws/... ./services/iam/... ./server/wire/sigv4/...` — all pass
- `golangci-lint run --timeout=9m ./providers/aws/iam/... ./server/aws/...` — 122 pre-existing issues across the package tree, none on the lines this PR touches; the only two hits inside touched files (`simulate.go` goconst on `gatherPrincipalDocs`, `iam.go` hugeParam on `CreateRole`) are on unmodified pre-existing code, not this diff.

Tests are non-tautological — `TestNotActionDeny`/`TestNotActionAllow`/`TestNotResource` each assert both the exempted and non-exempted sides, and `TestCheckPermissionWithContext`/`TestAuthzGateResourceScoped` prove the integration path end-to-end (the latter over a real signed HTTP request), so an inverted `negate` or a dropped resource-derivation would fail them.

**Verdict: the core NotAction/NotResource/Condition evaluation is fail-closed and correct.** The `X-Forwarded-For` source-IP spoofing issue undermines the IP-conditioned-policy feature this PR adds, so I'd like that addressed (or at minimum explicitly acknowledged/scoped) before merge.

PR #855, opened 2026-08-28
fix(auth): harden SigV4 enforce-auth path — verify ASIA temp creds, enforce expiry/skew, bind content-sha256 (#839)

### LINKED ISSUE #839: Auth hardening follow-ups (after #493 PR1 SigV4 authn)
Deferred from the SigV4 authentication foundation (#493 PR1, #838):
- Verify STS temporary (ASIA) credential signatures instead of passing them through unverified (closes the forged-ASIA gap).
- Bind the S3 `x-amz-content-sha256` header to the actual body (re-hash + compare) so a body swap that preserves the signed header is rejected (real AWS: XAmzContentSHA256Mismatch).
- Enforce request timestamp expiry / clock skew on x-amz-date (currently a captured signature replays).

NitinKumar004:
Closes #839. Auth hardening follow-ups after the SigV4 authN foundation (#838). Every change is gated on `--enforce-auth`; with it off (the default) behavior is byte-for-byte unchanged.

## 1. ASIA temporary credentials are now verified (was: trusted unverified)
STS previously returned a single constant synthetic ASIA credential and the gate trusted any ASIA-prefixed key with no signature check. STS now records the temporary credentials it mints in a per-server session store (`server/aws/sts/sessions.go`), created only when enforce-auth is on. Each `AssumeRole` / `GetSessionToken` / `GetFederationToken` / web-identity / SAML call mints a unique ASIA key id with a fresh high-entropy secret and stores `id -> {secret, expiration}`. The auth gate resolves the ASIA key's secret from that store and SigV4-verifies the signature exactly like a long-term AKIA key.

- Valid signature made with the STS-issued secret: accepted.
- Forged/wrong secret: `SignatureDoesNotMatch`.
- Unknown ASIA key: `InvalidClientTokenId`.
- Expired session: `ExpiredToken`.
- No store wired (shouldn't happen when enforce-auth is on): fails closed.

## 2. Timestamp expiry / clock skew enforced
`sigv4.Verify` now uses the injected `config.Clock` (was ignored). Header-signed requests must be within 5 minutes of server time in either direction (`RequestTimeTooSkewed`); presigned URLs must be inside their `X-Amz-Expires` window and not future-dated beyond the skew (`AccessDenied` / "Request has expired"). This stops a captured signature from replaying. `FakeClock` makes it deterministic in tests.

## 3. x-amz-content-sha256 bound to the body
When the header is a real hex SHA-256 (not a sentinel), the body is re-hashed and a mismatch is rejected with `XAmzContentSHA256Mismatch` — so a body swap that keeps the signed header fails. `UNSIGNED-PAYLOAD` and the `STREAMING-*` sentinels carry no body hash and are skipped (recognized as not being a 64-char hex digest). Genuine signed requests where the header matches the body are unaffected.

## Tests
`server/wire/sigv4/hardening_test.go` (skew, presigned expiry, body-swap rejection, UNSIGNED-PAYLOAD passthrough, sentinel discriminator) and `server/aws/authgate_temp_test.go` (temp-credential valid/forged/unknown/expired via FakeClock; real-SDK end-to-end AssumeRole -> signed request proving STS and the gate share the store; enforce-auth-off returns the fixed synthetic credential). All use hand-rolled helpers; the real aws-sdk-go-v2 v4 signer produces the signatures.

Gates: `go build`, `go vet`, `go test` (server + sigv4 + iam), `go generate` (no doc diff), and `golangci-lint` on the touched packages all pass with no new issues.

### DIFF
--- a/compat/aws/auth_sigv4_compat_test.go
+++ b/compat/aws/auth_sigv4_compat_test.go
@@ -132,10 +132,15 @@ func TestCompatAWSSigV4AuthEnabled(t *testing.T) {
 		}
 	})
 
-	t.Run("TempCredentialAccepted", func(t *testing.T) {
+	t.Run("ForgedTempCredentialRejected", func(t *testing.T) {
+		// ASIA temporary credentials are now verified against STS-issued
+		// sessions, so a forged one that STS never minted is rejected rather
+		// than trusted (a real assumed-role credential is covered by the STS
+		// session-store tests).
 		client := iamClientAt(t, sess.Endpoint(), "ASIATEMPCREDENTIAL00", "any-synthetic-secret", "session-token")
-		if _, err := client.ListUsers(ctx, &iam.ListUsersInput{}); err != nil {
-			t.Fatalf("STS-style temporary (ASIA) credential should pass through, got: %v", err)
+		_, err := client.ListUsers(ctx, &iam.ListUsersInput{})
+		if code := apiErrorCode(t, err); code != "InvalidClientTokenId" {
+			t.Fatalf("forged temp credential: want InvalidClientTokenId, got %q", code)
 		}
 	})
 
--- a/providers/aws/aws.go
+++ b/providers/aws/aws.go
@@ -183,6 +183,10 @@ type Provider struct {
 	// EnforceAuth mirrors config.Options.EnforceAuth: when true the AWS wire
 	// server verifies each request's SigV4 signature. Off by default.
 	EnforceAuth bool
+	// Clock mirrors config.Options.Clock: it drives SigV4 timestamp-expiry and
+	// STS temporary-credential expiry when EnforceAuth is on. Threaded to the
+	// wire server so a FakeClock stays deterministic there too.
+	Clock config.Clock
 
 	// engineClosers holds any wired real engines that implement io.Closer, so
 	// Close can cascade teardown to them. Empty for the in-memory default.
@@ -239,6 +243,7 @@ func New(opts ...config.Option) *Provider {
 		AccountID:           o.AccountID,
 		Region:              o.Region,
 		EnforceAuth:         o.EnforceAuth,
+		Clock:               o.Clock,
 		engineClosers:       o.EngineClosers(),
 	}
 	p.EC2.SetMonitoring(p.CloudWatch)
--- a/server/aws/authgate.go
+++ b/server/aws/authgate.go
@@ -8,31 +8,34 @@ import (
 
 	"github.com/stackshy/cloudemu/v2/config"
 	"github.com/stackshy/cloudemu/v2/server/authctx"
+	stssrv "github.com/stackshy/cloudemu/v2/server/aws/sts"
 	"github.com/stackshy/cloudemu/v2/server/wire"
 	"github.com/stackshy/cloudemu/v2/server/wire/awsquery"
 	"github.com/stackshy/cloudemu/v2/server/wire/sigv4"
 	iamdriver "github.com/stackshy/cloudemu/v2/services/iam/driver"
 )
 
-// tempCredentialPrefix marks STS-issued temporary access keys. Their secret is
-// synthetic (derived by STS, not stored as an IAM access key), so their SigV4
-// signature cannot be verified here; such requests are treated as authenticated
-// pass-throughs in this revision, and verifying temporary-credential signatures
-// is a follow-up. Note the limitation this leaves: unsigned or malformed
-// requests are still rejected, but ANY credential whose scope carries an
-// ASIA-prefixed key id is trusted without a signature check — so this is not yet
-// a hard boundary against a forged ASIA credential. That is acceptable for a
-// local dev emulator (rejecting all ASIA would break legitimate STS/IRSA/
-// instance-profile flows) and is closed when temp-credential verification lands.
+// tempCredentialPrefix marks STS-issued temporary access keys (ASIA). Their
+// secret is minted by STS, which — when EnforceAuth is on — records it in the
+// session store so the signature made with it can be SigV4-verified here exactly
+// like a long-term AKIA key, and a forged ASIA credential (unknown/wrong secret)
+// is rejected.
 const tempCredentialPrefix = "ASIA"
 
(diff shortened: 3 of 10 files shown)


NitinKumar004 (commented):
Security review of the SigV4 hardening (ASIA verification, expiry/skew, content-sha256 binding), focused on fail-open/regression risk.

**Enforce-auth OFF is byte-identical.** Confirmed. `server/aws/aws.go` only constructs `stssrv.NewSessionStore` and `authGate` when `d.EnforceAuth`; with it off, `stsHandler.sessions` stays nil, so `synthCredentials` (server/aws/sts/operations.go) returns the same fixed `ASIACLOUDEMU000000000` / `cloudemuSecretAccessKey...` values as before, and no pre-dispatch auth hook runs at all. `TestSTSCredentialsGatedByEnforceAuth` asserts the fixed AccessKeyId with EnforceAuth off.

**Shared session store wiring.** Confirmed single instance: `server/aws/aws.go` builds one `stsSessions := stssrv.NewSessionStore(authClock)` and passes the same pointer to both `stsHandler.SetSessions(stsSessions)` (mint side) and `newAuthGate(d.IAM, d.AccountID, stsSessions, authClock)` (verify side) — not two independently-constructed stores. `TestAuthGateVerifiesAssumedRoleCredential` proves this end-to-end with the real AWS SDK: `AssumeRole` mints a credential, a request signed with it verifies against the same server, and a tampered secret is rejected with 403.

**ASIA verification (server/aws/sts/sessions.go, server/aws/authgate.go).** `Mint` draws the access key id, secret, and session token from `crypto/rand` (not `math/rand`) via `randUpperAlnum`, keyed id→`Session{secret, expiration}` in a `sync.RWMutex`-guarded map. `verifyTempCredential` resolves the secret and delegates to `sigv4.Verify`, which compares via `hmac.Equal` (constant-time) — not `==`. Outcomes match spec: valid → pass; forged secret → `SignatureDoesNotMatch`; unknown key → `InvalidClientTokenId`; expired session (checked against the injected clock before signature verification) → `ExpiredToken`; nil store → `InvalidClientTokenId` (fail closed). All five covered by `TestVerifyTempCredential` subtests.

**Expiry/skew (server/wire/sigv4/sigv4.go).** `checkExpiry` runs after signature match, using the caller-injected `config.Clock`. Header-signed: >5min skew either direction → `RequestTimeTooSkewed` (tested both directions). Presigned: elapsed `X-Amz-Expires` → `AccessDenied`/"Request has expired" (matches real S3's error shape); future-dated beyond skew → `RequestTimeTooSkewed`. STS session expiry is a separate, clock-based comparison against the stored `Expiration` in `verifyTempCredential`, independent of `checkExpiry`.

Two non-blocking edge cases worth noting, both cryptographically inert against an attacker who doesn't hold the secret (X-Amz-Date and X-Amz-Expires are covered by the signature — via the string-to-sign and canonical query respectively — so neither can be altered post-hoc without invalidating the signature; a real SDK never omits or malforms either):
- `checkExpiry`/`parseAmzDate`: an X-Amz-Date that parses to neither the ISO-8601 basic form nor RFC1123 silently skips the skew/expiry check rather than rejecting (the comment in sigv4.go acknowledges this). Suggest failing closed here for defense-in-depth even though it's not currently reachable via an unauthenticated forger.
- `parsePresigned`: when `X-Amz-Expires` is absent or non-numeric, `in.expires` stays 0 and the "elapsed" branch of `checkExpiry` is skipped entirely (only the future-skew bound still applies) — again only reachable by the credential holder's own presigner, not a third party.

**Body-binding (server/wire/sigv4/sigv4.go).** `checkContentSHA256` re-hashes the actual body and rejects a mismatch against a real 64-hex `x-amz-content-sha256` header (`XAmzContentSHA256Mismatch`, 400) before the signature check runs. `isHexSHA256` correctly excludes `UNSIGNED-PAYLOAD` and `STREAMING-*` sentinels (length/charset discriminator, tested including uppercase-hex and truncated-length cases) and an absent header. Canonicalization (`canonicalHeaders`) still hashes the client's raw header value from `r.Header` for the signature — the re-hash is a distinct comparison, not a substitution — so a genuine matching request still verifies (`TestVerifyContentSHA256BodySwap`, `TestVerifyUnsignedPayloadSkipsBodyBinding`).

**No fail-open.** All error paths that matter under EnforceAuth deny rather than allow: missing SigV4 material → `MissingAuthenticationToken`; unresolvable AKID → `InvalidClientTokenId`; nil session store → `InvalidClientTokenId`; the two date/expires edge cases above are the only "skip the check" branches, and neither is reachable without already holding the secret (see above).

**Minor nits (non-blocking):**
- `server/aws/aws.go`'s `EnforceAuth` field doc comment ("STS temporary (ASIA) credentials are accepted without signature verification (a follow-up)") is now stale — this PR is that follow-up. Worth a one-line update in a fast-follow.
- `randUpperAlnum`'s `crypto/rand.Read` error fallback fills the buffer with a fixed repeated character rather than erroring — practically unreachable, but produces a fully predictable secret/key-id in the (essentially impossible) case it triggers.

**Gates on b5958272:**
- `go build ./...` — clean
- `go vet ./...` — clean
- `go test -count=1 ./server/... ./providers/aws/iam/... ./config/...` — all packages pass, including `server/aws` (`TestVerifyTempCredential`, `TestAuthGateVerifiesAssumedRoleCredential`, `TestSTSCredentialsGatedByEnforceAuth` all PASS) and `server/wire/sigv4`
- `go generate ./...` — no diff (coveragegen: 76 services, clean tree)
- `golangci-lint run --timeout=9m ./server/wire/sigv4/... ./server/aws/...` — 118 pre-existing issues across the tree (none in this PR's changed files: `authgate.go`, `aws.go`, `sts/sessions.go` (new), `sts/handler.go`, `sts/operations.go`, `sigv4.go`, and both new test files are all lint-clean); 0 new gating issues

**Verdict: MERGE.** The default (EnforceAuth off) path is unchanged, the session store is correctly shared between mint and verify, verification uses constant-time comparison and crypto-random secrets, and every failure path denies. The two date/expires parsing edge cases are worth hardening as defense-in-depth but are not exploitable without already holding the credential's secret, so they don't block.

Issue #577, opened 2026-08-25
AWS: deferred real-user e2e follow-ups


NitinKumar004:
Deferred items from the AWS real-user e2e passes, to pick up.

- [ ] Kinesis → Lambda ESM delivery (wire the source-agnostic DeliverEventSourceBatch to Kinesis; correct event shape; reuse recursion guard)
- [ ] Validate KmsKeyId exists on SecretsManager CreateSecret
- [ ] Lambda DryRun Invoke should validate Qualifier (unknown alias/version → ResourceNotFoundException, not 204)
- [ ] SQS→Lambda ESM should honor DelaySeconds before delivery
- [ ] SQS→Lambda ESM synchronous retry loop can block SendMessage (revisit async delivery)
- [ ] DeliverEventSourceBatch should not abort sibling mappings on first invoke error (aggregate/log)
- [ ] Lambda alias RoutingConfig.AdditionalVersionWeights ignored (weighted-alias routing)
- [ ] Move BuildLambdaStreamEvent out of the cross-cloud services/database/driver package into an AWS-scoped location

NitinKumar004:
5/6 items done via #856 (Kinesis→Lambda ESM, SQS DelaySeconds, weighted-alias routing, DryRun qualifier, SecretsManager KmsKeyId validation). Remaining: (a) SQS synchronous retry loop → async delivery; (b) relocate BuildLambdaStreamEvent out of the cross-cloud services/database/driver package. Keeping open for those two.

NitinKumar004:
Final follow-ups resolved in #931: `BuildLambdaStreamEvent` relocated out of the cross-cloud `services/database/driver` into `providers/aws/dynamodb` (shared driver no longer carries AWS-Lambda-specific logic), and the sibling-abort + DryRun-qualifier behaviors were verified correct with locking tests. The earlier items on this issue (Kinesis→Lambda ESM, KmsKeyId validation, SQS ESM DelaySeconds, weighted-alias routing) had already landed. Closing as complete.

PR #856, opened 2026-08-28
fix(aws): ESM + Lambda routing fidelity follow-ups (#577)

NitinKumar004:
Closes the event-source-mapping and Lambda routing fidelity gaps from #577.

- Kinesis PutRecord/PutRecords invoke mapped Lambdas (aws:kinesis event batch via the existing DeliverEventSourceBatch choke point)
- SQS→Lambda ESM honors DelaySeconds (visibility-gated sweep, mirrors the poll path)
- Lambda weighted-alias routing: full AdditionalVersionWeights map preserved end-to-end + deterministic weighted selection
- Lambda DryRun validates Qualifier (unknown alias/version → ResourceNotFoundException)
- SecretsManager CreateSecret validates KmsKeyId against KMS

5/6 checklist items done (+ sibling-mapping resilience already in code). Deferred: SQS async-delivery revisit; relocating BuildLambdaStreamEvent out of the cross-cloud services/database/driver package.

### DIFF
--- a/providers/aws/aws.go
+++ b/providers/aws/aws.go
@@ -268,6 +268,9 @@ func New(opts ...config.Option) *Provider {
 	// An IamInstanceProfile passed to RunInstances resolves through IAM so the
 	// role->profile->instance chain reads back on DescribeInstances.
 	p.EC2.SetInstanceProfileResolver(p.IAM)
+	// A KmsKeyId passed to CreateSecret is validated against KMS, so a secret
+	// can't reference a key that doesn't exist.
+	p.SecretsManager.SetKMSKeyResolver(p.KMS)
 	p.SSM.SetInstanceResolver(p.EC2)
 	// ECS-registered container instances surface as managed EC2 instances, so
 	// #159 (ECS) composes with #300 (EC2 managed-resource visibility).
@@ -314,6 +317,10 @@ func New(opts ...config.Option) *Provider {
 	// event-source-mapping target(s), deleting the message on success or
 	// leaving it for DLQ redrive on failure (mirrors the DynamoDB Streams wiring).
 	p.SQS.SetEventSourceInvoker(p.Lambda)
+	// Kinesis -> Lambda: records written to a stream (PutRecord/PutRecords)
+	// deliver a Kinesis-shaped event batch to the stream's event-source-mapping
+	// target(s) (mirrors the DynamoDB Streams -> Lambda wiring).
+	p.Kinesis.SetLambdaInvoker(p.Lambda)
 	// CloudWatch Logs subscription filters -> Lambda: log events matching a
 	// subscription filter's pattern are delivered (gzipped awslogs payload) to
 	// the filter's Lambda destination on PutLogEvents.
--- a/providers/aws/kinesis/event_source_mapping_test.go
+++ b/providers/aws/kinesis/event_source_mapping_test.go
@@ -1,0 +1,161 @@
+package kinesis_test
+
+import (
+	"context"
+	"encoding/base64"
+	"encoding/json"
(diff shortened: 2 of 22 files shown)


NitinKumar004 (commented):
Reviewed against a clean checkout of 119f9273 (isolated worktree, merge-base d787f98f).

**Gates (all green):** `go build ./...`, `go vet ./...`, full `go test ./...`, `go test -race` on the Bedrock set + on all PR-touched packages (lambda/kinesis/sqs/secretsmanager/server-aws-lambda), `go generate ./...` (no diff), `go mod tidy` (no diff). `golangci-lint --new-from-rev` against the merge-base: 0 new issues (2 pre-existing `wsl` hits live in `providers/aws/sqs/movetask.go`, untouched by this PR).

**Focus-area verification:**
- Driver change (`AliasRoutingConfig.AdditionalVersionWeights`): Azure and GCP `aliases.go` both mirror the exact same map deep-copy as AWS; `server/aws/lambda/handler.go` `toRoutingConfig`/`toAliasResponse` preserve the full map in both directions (no collapse-to-first); migrated tests assert the map correctly, not weakened.
- Kinesis→Lambda ESM: event envelope matches the real Kinesis event shape; `deliverToLambda`'s recursion-guard carry-forward (`recursionguard.WithDepth(context.Background(), recursionguard.Depth(callerCtx))` on a fresh background context, after releasing the stream lock) is identical to DynamoDB Streams' `flushStreamDeliveries`; nil-safe when no invoker is wired.
- SQS DelaySeconds gating: `deliverToESM`'s sweep correctly excludes messages with `VisibleAt.After(now)`; verified a delayed message is skipped and picked up on the next `SendMessage` after the delay elapses; DLQ-redrive and visibility-timeout paths unaffected.
- Weighted alias selection: deterministic (SHA-256 hash of the invoke payload → `[0,1)`, sorted cumulative bands); tests assert distribution without flakiness.
- DryRun + KmsKeyId: DryRun 404s on an unknown qualifier via `resolveQualifier` before the 204 path; `SetKMSKeyResolver(p.KMS)` wiring in `providers/aws/aws.go` is correct and nil-safe; confirmed via a real-SDK round-trip test for both the 204 and `ResourceNotFoundException` cases.
- Scope: only the named files + Azure/GCP `aliases.go` touched.

**Findings (neither blocks this PR):**

1. **[Medium] Pre-existing data race on concurrent alias updates** — `providers/aws/lambda/aliases.go` (`CreateAlias`/`UpdateAlias`, ~L69/88/95/98). `fd.aliases.Get(name)` hands out a shared `*aliasData` pointer with no per-alias lock; `UpdateAlias` mutates `ad.alias.*` fields in place. Reproduced with a throwaway `-race` test (concurrent `Invoke` + `UpdateAlias` on the same alias). Confirmed pre-existing — `CreateAlias`/`UpdateAlias` bodies are unchanged by this PR (only the `copyRoutingConfig`/`validateRoutingConfig` helpers changed), so the old single-field `AdditionalVersion`/`Weight` shape had the same exposure. Flagging because this PR makes weighted aliases a real, Terraform/CDK-reachable feature, raising the odds a concurrent alias update trips it. Suggest a follow-up ticket.

2. **[Low] No bounds/sum validation on `AdditionalVersionWeights`** — `validateRoutingConfig` in the same file. Real Lambda rejects a weight outside `[0,1)` or an additional-weights sum `>=1.0` with `InvalidParameterValueException`; this mock accepts any float64 (degrades gracefully rather than crashing, but silently diverges from real validation). Also pre-existing (the old single-weight shape had the identical gap) — worth a follow-up ticket, not a blocker.

**Verdict: MERGE.**

Issue #581, opened 2026-08-25
AWS: Step Functions ASL execution + KMS↔Secrets/SSM real encryption


NitinKumar004:
Two AWS features surfaced by the real-user e2e passes — currently behave incompletely rather than being broken.

- [ ] Step Functions: actually interpret the ASL definition on StartExecution (run Task/Choice/Pass/Wait/Succeed/Fail states, invoke Lambda Task states, sequential Next chaining, basic Retry/Catch) instead of returning a canned SUCCEEDED echo
- [ ] KMS ↔ SecretsManager/SSM: make KmsKeyId real — encrypt secret/SecureString values with the referenced KMS key and decrypt on read (envelope crypto), instead of KmsKeyId being a cosmetic string

NitinKumar004:
KMS↔Secrets/SSM real encryption landed in #857 (envelope encryption via KMS GenerateDataKey + AES-256-GCM; a disabled/deleted key now fails reads). The remaining part of this ticket — the Step Functions ASL execution interpreter — is in progress as a phased 3-PR series (PR1 core walker + I/O + Choice/Wait now building).

NitinKumar004:
Both parts complete: (1) real KMS↔Secrets/SSM encryption shipped in #857; (2) the Step Functions ASL execution interpreter shipped as a 3-PR series — core walker + I/O pipeline + Choice/Wait/Pass (#859), Task→Lambda with recursion guard + Retry/Catch (#860), and Parallel/Map + catchable state errors (#861). Executions now really run the state graph (JSONPath, intrinsics, $$ context, real history) instead of auto-succeeding.

PR #857, opened 2026-08-28
fix(kms): encrypt Secrets Manager & SSM SecureString through real KMS (#581)

NitinKumar004:
## Summary

Part of #581 (the KMS↔Secrets/SSM cosmetic-crypto gap). Secrets Manager and SSM SecureString values are now encrypted through **real KMS** instead of being stored verbatim, so their at-rest form is genuine ciphertext and their KMS-failure paths are real.

KMS's own crypto (`providers/aws/kms`) is used unchanged.

## What changed

- **New crypto seam** `providers/aws/kmscrypto` (`Envelope`): `Encrypt(keyRef, plaintext)` mints a KMS data key (wrapped under the addressed key), seals the value locally with AES-256-GCM, and returns a self-describing blob; `Decrypt` unwraps the data key back through KMS — the step that surfaces a disabled/deleted key — then opens the ciphertext. An unresolvable managed-key reference (the reserved `alias/aws/secretsmanager` / `alias/aws/ssm`, which KMS refuses to let callers alias) is created and cached on demand.
- **Secrets Manager**: `CreateSecret` / `PutSecretValue` / `UpdateSecret` encrypt under the explicit `KmsKeyId` (validated to exist) or the default `aws/secretsmanager` key; every read path (`GetSecretValue`, `GetSecretValueStage`) decrypts. `KmsKeyId` reflection, versioning, and `VersionStages` are preserved. `ClientRequestToken` idempotency now compares plaintext.
- **SSM**: `PutParameter(Type=SecureString)` stores ciphertext (base64); `GetParameter`/`GetParameters`/`GetParametersByPath` with `WithDecryption=true` decrypt, `WithDecryption=false` returns the opaque blob (matching AWS). String/StringList params are untouched.
- **Wiring**: both services share one `Envelope` over the KMS backend in `providers/aws/aws.go`.
- **Fallback**: with no KMS wired (library users who didn't create KMS), both services keep today's plaintext behavior — nil-safe, nothing breaks.

## Tests

Round-trip create→get with a real key; `GetSecretValue` fails after the key is disabled; SecureString put→get(WithDecryption true returns value, false does not return plaintext); String params unaffected; the no-KMS plaintext fallback still works; plus unit tests for the `Envelope` adapter.

## Gates

`go build`/`go vet`, `go test ./providers/aws/... ./server/aws/... ./compat/aws/`, `go generate ./...` (no doc changes), and golangci-lint on the touched packages all green — 0 new gating issues.

### DIFF
--- a/providers/aws/aws.go
+++ b/providers/aws/aws.go
@@ -35,6 +35,7 @@ import (
 	"github.com/stackshy/cloudemu/v2/providers/aws/keyspaces"
 	"github.com/stackshy/cloudemu/v2/providers/aws/kinesis"
 	"github.com/stackshy/cloudemu/v2/providers/aws/kms"
+	"github.com/stackshy/cloudemu/v2/providers/aws/kmscrypto"
 	"github.com/stackshy/cloudemu/v2/providers/aws/lambda"
 	"github.com/stackshy/cloudemu/v2/providers/aws/memorydb"
 	"github.com/stackshy/cloudemu/v2/providers/aws/networkfirewall"
@@ -273,9 +274,13 @@ func New(opts ...config.Option) *Provider {
 	// An IamInstanceProfile passed to RunInstances resolves through IAM so the
 	// role->profile->instance chain reads back on DescribeInstances.
 	p.EC2.SetInstanceProfileResolver(p.IAM)
-	// A KmsKeyId passed to CreateSecret is validated against KMS, so a secret
-	// can't reference a key that doesn't exist.
-	p.SecretsManager.SetKMSKeyResolver(p.KMS)
+	// Secrets Manager and SSM SecureString values are encrypted through real KMS
+	// (envelope encryption): a KmsKeyId is validated to exist, the value is stored
+	// as genuine ciphertext, and a later disabled/deleted key makes reads fail as
+	// in AWS. Both services share one Envelope over the KMS backend.
+	kmsCrypto := kmscrypto.New(p.KMS)
+	p.SecretsManager.SetKMSCrypto(kmsCrypto)
+	p.SSM.SetKMSCrypto(kmsCrypto)
 	p.SSM.SetInstanceResolver(p.EC2)
 	// ECS-registered container instances surface as managed EC2 instances, so
 	// #159 (ECS) composes with #300 (EC2 managed-resource visibility).
--- a/providers/aws/kmscrypto/kmscrypto.go
+++ b/providers/aws/kmscrypto/kmscrypto.go
@@ -1,0 +1,193 @@
+// Package kmscrypto routes Secrets Manager and SSM SecureString values through
+// real KMS envelope encryption so their at-rest form is genuine ciphertext and
+// their KMS-failure paths are real: a disabled or deleted key makes a read fail
+// exactly as it does in AWS.
+//
+// Encrypt asks KMS for a data key (wrapped under the addressed KMS key), seals
+// the value locally with AES-256-GCM under that data key, and returns a
+// self-describing blob carrying the wrapped key. Decrypt unwraps the data key
+// back through KMS — the step that surfaces a disabled/deleted key — then opens
+// the AES-GCM ciphertext. KMS's own crypto (providers/aws/kms) is used as-is and
+// never modified.
+package kmscrypto
+
+import (
+	"context"
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
+	"encoding/binary"
+	"sync"
+
+	"github.com/stackshy/cloudemu/v2/errors"
+	kmsdriver "github.com/stackshy/cloudemu/v2/services/kms/driver"
+)
+
+// Envelope blob layout: magic(1) | wrappedKeyLen(uint32 BE) | wrappedKey |
+// nonce(12) | AES-256-GCM ciphertext. The wrapped key is a KMS ciphertext blob
+// that names its own key, so Decrypt needs no key id from the caller.
+const (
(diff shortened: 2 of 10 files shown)


NitinKumar004 (commented):
**Verdict: MERGE**

Reviewed at head `4937166b`. Verified the crypto seam, the default (no-KMS) fallback, and semantics preservation.

**Envelope crypto (`providers/aws/kmscrypto/kmscrypto.go`)**
- `Encrypt`: `GenerateDataKey` → AES-256-GCM `Seal` with a fresh 12-byte `crypto/rand` nonce per call, wrapped-key + nonce + ciphertext packed into a self-describing blob (magic/len-prefixed). `Decrypt` unwraps the data key via `KMS.Decrypt` *before* opening GCM, so a disabled/deleted key surfaces the KMS error at that step, matching real Secrets Manager/SSM read failures — confirmed by `TestEnvelopeDecryptFailsAfterKeyDisabled` and mirrored at the service level (`TestGetSecretValueFailsAfterKeyDisabled`, `TestSecureStringFailsAfterKeyDisabled`).
- `Decrypt` bounds-checks `len(blob) < headerSize`, the magic byte, and `len(rest) < wrappedLen+nonceSize` before any slicing — `wrappedLen` comes from a `uint32` so it's always non-negative, and on 64-bit `wrappedLen+nonceSize` can't overflow `int`. No panic path on truncated/garbage input; covered by `TestEnvelopeDecryptRejectsGarbage`.
- `providers/aws/kms/` (KMS's own crypto) has a zero-line diff — confirmed untouched.
- Data-key plaintext is wiped (`defer wipe(...)`) after building the AES cipher in both `Encrypt` and `Decrypt`.

**Default (no-KMS) path**
- `providers/aws/secretsmanager.encrypt/decrypt` and `providers/aws/ssm.sealValue/revealValue` all nil-check `m.kmsCrypto` first and return the value verbatim — no panic, no behavior change from `origin/development` for library users who don't wire KMS. Covered by `TestSecretNoKMSFallbackStoresPlaintext` and `TestSecureStringNoKMSFallback`.
- In practice this fallback is close to unreachable for standard usage: `providers/aws/aws.go`'s `New()` (used by both `cloudemu.NewAWS()` and the `cloudemu serve` binary) always wires `kmscrypto.New(p.KMS)` into both services, so real deployments always get real envelope encryption. The fallback only applies to direct construction of `secretsmanager.New`/`ssm.New` bypassing the factory. Worth calling out in case that's not the intended default-on posture, but it's consistent with existing `SetMonitoring`-style wiring elsewhere in the factory.

**Semantics preserved**
- Secrets create→get round-trip, `PutSecretValue`/`UpdateSecret`/rotation/versioning/`VersionStages` all route through `encrypt`/`decrypt` correctly (`secretsmanager.go`, `staging.go`, `update.go`).
- `ClientRequestToken` idempotency (`reusedTokenVersion`) now decrypts the existing stored ciphertext to compare *plaintext* before deciding same-token/same-content vs. `ResourceExistsException`. This is correct and necessary — a byte-comparison on ciphertext would falsely reject identical resubmitted content, since each `Encrypt` call mints a new data key + nonce and produces different ciphertext for identical plaintext. Not a regression.
- SSM: `SecureString` stores base64 ciphertext; `WithDecryption=true` decrypts via `revealValue`; `=false` returns the opaque blob unchanged; `String`/`StringList` bypass `sealValue`/`revealValue` entirely (type-gated) and are unaffected — confirmed by `TestStringParameterNotEncrypted`.
- `RotateSecret` carries the current version's already-encrypted bytes forward under a new version id without re-encrypting — correct, since it's the same envelope blob and decrypts fine regardless of version id.

**No plaintext leak**
- `storedValue` test helper confirms at-rest ciphertext for Secrets Manager (`TestSecretEncryptedAtRestAndRoundTrips`); SSM's `TestSecureStringEncryptedAtRest` confirms `WithDecryption=false` returns neither the plaintext nor an empty string.
- `secretsmanager.ListSecretVersions` only returns metadata (no `Value` field), so no incidental leak there.

**One gap worth a follow-up ticket (not a blocker, not a security regression — fails closed)**: `SSM.GetParameterHistory` (`providers/aws/ssm/ssm.go`) has no `withDecryption` parameter at all — `server/aws/ssm/operations.go:235` parses `req.WithDecryption` off the wire request but never passes it through. Pre-PR this was moot since everything was already stored in the clear; post-PR, a `GetParameterHistory` call against a `SecureString` parameter with `WithDecryption=true` will get back the opaque base64 ciphertext blob instead of the decrypted value, diverging from real SSM. Direction is safe (ciphertext instead of secret, not the other way around), but it's an accuracy gap the PR's own doc comment doesn't mention.

**Gates (all green on `4937166b`, GOMAXPROCS=2)**
- `go build ./...` — clean
- `go vet ./...` — clean
- `go test -count=1 ./providers/aws/... ./server/aws/... ./compat/aws/...` — all packages `ok`, including `providers/aws/kmscrypto`, `providers/aws/secretsmanager`, `providers/aws/ssm`, `server/aws/secretsmanager`, `server/aws/ssm`
- `go generate ./...` — no diff (coveragegen wrote 76 services, unchanged)
- `golangci-lint run ./providers/aws/secretsmanager/... ./providers/aws/ssm/... ./providers/aws/kms/... ./providers/aws/kmscrypto/...` — 5 findings, all pre-existing on `origin/development` (verified byte-identical function bodies for `DeleteParameters`, `LabelParameterVersion`, `ensurePublicParameter`, and the two `wsl` cuddling spots in `GetParameters`/`GetParameterHistory`, none touched by this diff). Zero new issues.

Issue #847, opened 2026-08-28
Auto-telemetry parity: Azure/GCP audit logs + function execution logs


NitinKumar004:
**Goal:** match real-cloud auto-telemetry. CloudEmu already auto-emits metrics + evaluates alarms on all 3 clouds, and auto-records CloudTrail on AWS. Two things real clouds auto-generate are still missing.

**Ops:**
- Azure **Activity Log** + GCP **Cloud Audit Logs**: auto-record ARM/API operations for both providers. Mirror the AWS CloudTrail server observer (`server/aws/aws.go` SetObserver + `server/aws/cloudtrail_recording.go`) in `server/azure` and `server/gcp`.
- **Function execution logs**: Lambda / Cloud Functions / Azure Functions invoke currently emits metrics only — no log lines. Auto-write invocation stdout/stderr to the log service (CloudWatch Logs / Cloud Logging / Log Analytics), especially on the real-engine path. Template: the ECS `awslogs` sink (`providers/aws/ecs/engine.go` `surfaceLogs`/`pushAwslogs` + `SetLogSink` in `providers/aws/aws.go`).

NitinKumar004:
Implemented in #858: Azure Activity Log + GCP Cloud Audit Log auto-recording (mirroring the AWS CloudTrail observer), and auto execution logs for Lambda/Cloud Functions/Azure Functions (mirroring the ECS awslogs sink). Metrics + alarms were already at parity across all 3 clouds.

PR #858, opened 2026-08-28
feat(telemetry): Azure/GCP audit logs + function execution logs (#847)

NitinKumar004:
Closes the auto-telemetry parity gaps in #847.

**Audit logs (mirror AWS CloudTrail's observer):**
- Azure Activity Log — server/azure auto-records mutating ARM operations, queryable via the activity-log API
- GCP Cloud Audit Logs — server/gcp auto-records API operations, surfaced through Cloud Logging

**Function execution logs (mirror the ECS awslogs sink):**
- Lambda, Cloud Functions, Azure Functions invokes auto-write execution log lines (and real-engine stdout/stderr) to CloudWatch Logs / Cloud Logging / Log Analytics; conventional log group/stream auto-created.

No-observer / no-sink default paths are byte-unchanged. Metrics + alarms were already at parity across all 3 clouds.

### DIFF
--- a/docs/coverage/aws/cloudwatch.md
+++ b/docs/coverage/aws/cloudwatch.md
@@ -20,6 +20,19 @@ AWS's `monitoring` service · portable interface `driver.Monitoring` · [AWS ind
 | `PutMetricData` |  |
 | `SetAlarmState` |  |
 
+## Optional capabilities
+
+Discovered by type assertion; only some providers implement these.
+
+### ActivityLogRecorder
+
+ActivityLogRecorder is an OPTIONAL capability, discovered by type assertion
+
+| Operation | Description |
+| --- | --- |
+| `ListActivityLogEvents` |  |
+| `RecordActivityLogEvent` |  |
+
 ## Not in scope
 
 _Not documented yet. See the [emulator boundary](../../../README.md) for cloudemu-wide non-goals._
--- a/docs/coverage/azure/monitor.md
+++ b/docs/coverage/azure/monitor.md
@@ -20,6 +20,19 @@ Azure's `monitoring` service · portable interface `driver.Monitoring` · [Azure
 | `PutMetricData` |  |
 | `SetAlarmState` |  |
 
+## Optional capabilities
+
+Discovered by type assertion; only some providers implement these.
+
+### ActivityLogRecorder
+
+ActivityLogRecorder is an OPTIONAL capability, discovered by type assertion
+
+| Operation | Description |
+| --- | --- |
+| `ListActivityLogEvents` |  |
+| `RecordActivityLogEvent` |  |
+
 ## Not in scope
 
 _Not documented yet. See the [emulator boundary](../../../README.md) for cloudemu-wide non-goals._
--- a/docs/coverage/coverage.json
+++ b/docs/coverage/coverage.json
@@ -6698,6 +6698,20 @@
         "name": "SetAlarmState"
       }
     ],
+    "optionalCapabilities": [
+      {
+        "name": "ActivityLogRecorder",
+        "doc": "ActivityLogRecorder is an OPTIONAL capability, discovered by type assertion",
+        "operations": [
+          {
+            "name": "ListActivityLogEvents"
+          },
+          {
+            "name": "RecordActivityLogEvent"
+          }
+        ]
+      }
+    ],
     "providers": {
       "aws": "CloudWatch",
       "azure": "Monitor",
--- a/docs/coverage/gcp/cloudmonitoring.md
+++ b/docs/coverage/gcp/cloudmonitoring.md
@@ -20,6 +20,19 @@ GCP's `monitoring` service · portable interface `driver.Monitoring` · [GCP ind
 | `PutMetricData` |  |
 | `SetAlarmState` |  |
 
+## Optional capabilities
+
+Discovered by type assertion; only some providers implement these.
+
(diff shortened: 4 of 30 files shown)


NitinKumar004 (commented):
Reviewed at 9e462add.

**Audit-log observers (Azure Activity Log / GCP Cloud Audit Log)**

Both mirror the AWS CloudTrail observer pattern (`server/aws/aws.go` `SetObserver`) correctly:

- Azure (`server/azure/activitylog_recording.go`): `writeVerb()` gates on PUT/PATCH/DELETE/POST only, skips GET/HEAD; further skips any request whose ARM path resolves to provider `Microsoft.Insights` (its own read API), so no self-recording loop is possible.
- GCP (`server/gcp/auditlog_recording.go`): gates on non-GET/HEAD, and `isLoggingOperation()` skips any path containing `/entries:` or `/logs` so Cloud Logging's own writes never get audit-recorded. In both cases the writer path (`RecordActivityLogEvent` for Azure, `logs.PutLogEvents` for GCP) doesn't itself go back through `ServeHTTP`/the observer, so there's no possible feedback loop architecturally, not just by the filter.
- Both observers are wired with a single `SetObserver` call, conditioned on the capability being present (`d.Monitor.(mondriver.ActivityLogRecorder)` for Azure, `d.CloudLogging != nil` for GCP) — nil-safe, no panic when absent, and doesn't clobber any other observer registration (grepped — one `SetObserver` call per server).
- Azure Activity Log is bounded at `maxActivityLogEvents = 2000` with amortized trim (`providers/azure/monitor/activitylog.go`). Queryable via the real `Microsoft.Insights/eventtypes/management/values` API (`server/azure/monitor/activitylog.go`) and GCP's via standard Cloud Logging `entries:list` against log name `cloudaudit.googleapis.com/activity` — both verified end-to-end against real SDK/REST clients in `server/azure/activitylog_test.go` and `server/gcp/auditlog_test.go`.
- `ActivityLogRecorder` is declared in `services/monitoring/driver/activitylog.go` as an optional capability discovered by type assertion, not added to the mandatory `Monitoring` interface — AWS/GCP/OCI monitoring drivers are untouched and don't need to implement it (same established pattern as IAM's AWS-only `AccessKeyResolver`).

**Function execution logs (Lambda / Cloud Functions / Azure Functions)**

- `SetLogSink` wired in all three provider factories (`providers/aws/aws.go`, `providers/azure/azure.go`, `providers/gcp/gcp.go`), mirroring ECS's `SetLogSink(p.CloudWatchLogs)` wiring for `awslogs`.
- Confirmed the log-leak concern directly by reading all three wire Invoke handlers: `server/aws/lambda/handler.go` `invoke()` writes only `out.Payload` / `out.Error` to the response; `server/gcp/cloudfunctions/handler.go` sets only `resp.Result = string(out.Payload)` / `resp.Error`; `server/azure/functions/handler.go` writes only `out.Payload` / `out.Error`. None of the three reference `out.Logs` — grepped `server/azure/functions/*.go` and `server/gcp/cloudfunctions/*.go` for `.Logs` with zero hits. `driver.InvokeOutput.Logs` is explicitly documented as "surfaced to the log service... rather than returned on the wire."
- Default (no sink wired) path is a no-op: all three `surfaceInvokeLogs` implementations start with `if m.logs == nil { return }`, and each has a `TestInvokeNoLogSinkDefault` test.
- Each auto-creates the conventional log group/stream (`/aws/lambda/<name>`, `cloudfunctions.googleapis.com/cloud-functions`, `FunctionAppLogs`) and writes START/END/REPORT-equivalent bookend lines plus engine stdout/stderr when present.

**Determinism**

- `providers/azure/monitor/activitylog.go` `ListActivityLogEvents`: reverses to insertion order, then `sort.SliceStable` descending by timestamp — ties (same FakeClock instant) resolve to insertion-order-newest-first deterministically. Confirmed by `providers/azure/monitor/activitylog_test.go`. Good fix versus a random-UUID tiebreak.
- One inconsistency worth a note (non-blocking): `server/gcp/auditlog_recording.go:59` uses `time.Now()` directly for the audit entry timestamp instead of routing through `m.opts.Clock`, unlike every other new timestamp in this PR (Lambda/Functions/Cloud Functions execution logs all use `m.opts.Clock.Now()`, and the Azure Activity Log recorder uses the provider's injected clock). It doesn't break anything today — no test asserts exact audit-log timestamps — but it means GCP audit-log entries won't respect `FakeClock` the way everything else in this PR does.

**No regressions / scope**

- `go build ./...`, `go vet ./...` clean.
- `go test -count=1 ./server/... ./providers/...` — all green (150 packages, exit 0).
- `go generate ./...` — no diff; docs/coverage was already regenerated and committed correctly, including the new `ActivityLogRecorder` optional-capability section on all 4 clouds' monitor pages (expected/consistent with the existing generator behavior for shared-package optional capabilities, e.g. IAM's `AccessKeyResolver`).
- `golangci-lint run --timeout=9m --new-from-rev=origin/development` over the whole repo: 0 new issues. A scoped run over the touched packages surfaces 70 pre-existing issues, none in any file this PR touches (verified — zero hits grepping for activitylog/auditlog/execlogs filenames in the lint output).
- `providers/azure/monitor/monitor.go` only adds two new fields (`activityMu`, `activityLog`) to `Mock`; alarm/metric evaluation code paths are untouched.

Verdict: MERGE. No blockers found.

PR #859, opened 2026-08-28
feat(sfn): real ASL interpreter core (PR1/3) for Step Functions (#581)

NitinKumar004:
## Summary

PR1 of 3 replacing the Step Functions auto-succeed stub with a real Amazon States Language (ASL) interpreter (#581). This PR ships the **core walker + I/O pipeline + history plumbing**; Task→Lambda lands in PR2, Parallel/Map in PR3.

New internal package `providers/aws/sfn/asl` walks the state graph from `StartAt` following `Next`/`End`, computes the true terminal status/output, and records a real per-state event history.

### What's live
- **State types:** Pass, Choice, Wait, Succeed, Fail. Task/Parallel/Map parse (accepted at create time) but fail loudly at run time until their handlers land.
- **Choice comparators:** String (incl. glob `StringMatches`), Numeric, Boolean, Timestamp, type-checks (`IsNull`/`IsPresent`/`IsNumeric`/`IsString`/`IsBoolean`/`IsTimestamp`), And/Or/Not, and `...Path` variants.
- **I/O pipeline:** `InputPath → Parameters → [Result] → ResultPath → OutputPath`, applied per state only where ASL permits.
- **JSONPath:** reference/selection subset (`$`, `$.a.b`, `$[0]`, `$.a.b[2]`); filters/wildcards/recursive-descent are rejected, not silently mis-run.
- **`$$` context object** + first-wave intrinsics (`Format`, `Array`, `ArrayGetItem`, `StringToJson`, `JsonToString`, `UUID`).
- **History:** monotonic IDs with a chained `PreviousEventID`, persisted on `driver.Execution.History` so it snapshot/restores for free; `ExecutionFailed`/`ExecutionAborted` carry Error/Cause.

### Corrected semantics
- **ResultPath merges onto the RAW input** (the document handed to the state, *before* InputPath filtered it) — so InputPath-narrow + ResultPath-reattach preserves the sibling fields InputPath excluded. `ResultPath:null` passes the raw input through; default `$` replaces the document. Explicitly tested.
- **ctx is threaded end-to-end:** `runExecution` now takes `ctx`; `StartExecution`/`StartSyncExecution` forward it instead of discarding it (plumbing for the PR2 Task→Lambda recursion guard, even though no PR1 handler consumes it).

### Validation
Stricter `CreateStateMachine`/`UpdateStateMachine`/`ValidateStateMachineDefinition`/`TestState`: unknown state type, dangling `Next`, `Choice` without `Choices`, unsupported fields on Wait/Choice/Succeed/Fail, and `QueryLanguage: "JSONata"` are rejected with InvalidDefinition. Definitions are still stored verbatim (snapshot/persist untouched).

### Execution model
Synchronous whole-graph execution preserved; the `settle.Window` overlay still gates RUNNING→terminal observability under `AsyncSettle` (off by default → instant), extended by accumulated `Wait` durations. `GetExecutionHistory` time-truncates events while unsettled.

## Tests
- Pass/Wait/Choice/Succeed/Fail run to correct terminal output; a Choice branch per comparator family.
- **Dedicated ResultPath test** proving InputPath-narrow + ResultPath-reattach preserves the excluded sibling field, plus `ResultPath:null` and default `$`.
- OutputPath filtering; malformed ASL / JSONata / unsupported-JSONPath rejection at create/validate time.
- `GetExecutionHistory` real event sequence with a valid `PreviousEventID` chain (forward + reversed); snapshot/restore round-trips history byte-identically.
- Wait under `AsyncSettle` + FakeClock (RUNNING+truncated → SUCCEEDED+full) and instant without it; StopExecution surfaces Error/Cause on ABORTED; maxSteps Choice/Next cycle fails with `CloudEmu.StateTransitionLimitExceeded`; ctx-forwarding entry point.
- asl-package unit tests for glob, JSONPath, intrinsics, payload template, parser.

## Gates (GOMAXPROCS=2)
`go build ./...`, `go vet`, `go test ./...` (full suite), `go generate ./...` (coverage regen committed), and `golangci-lint run` on the sfn packages — all green, 0 lint issues, no `nolint` added.

## Follow-ups
- **PR2:** Task(Lambda) + recursion-guarded `LambdaSyncInvoker` seam + Retry/Catch + `LambdaFunction*` sub-events.
- **PR3:** Parallel (sequential branches) + Map (sequential iterations).

### DIFF
--- a/docs/coverage/aws/sfn.md
+++ b/docs/coverage/aws/sfn.md
@@ -36,7 +36,7 @@ AWS's `sfn` service · portable interface `driver.SFN` · [AWS index](./README.m
 | `SendTaskSuccess` |  |
 | `StartExecution` | Executions. |
 | `StartSyncExecution` |  |
-| `StopExecution` |  |
+| `StopExecution` | StopExecution aborts a still-settling execution, persisting the caller's |
 | `TagResource` | Tags. |
 | `TestState` | Definition tooling. |
 | `UntagResource` |  |
--- a/docs/coverage/coverage.json
+++ b/docs/coverage/coverage.json
@@ -11093,7 +11093,8 @@
         "name": "StartSyncExecution"
       },
       {
-        "name": "StopExecution"
+        "name": "StopExecution",
+        "doc": "StopExecution aborts a still-settling execution, persisting the caller's"
       },
       {
         "name": "TagResource",
--- a/providers/aws/sfn/asl/asl_test.go
+++ b/providers/aws/sfn/asl/asl_test.go
@@ -1,0 +1,175 @@
+package asl
+
+import (
+	"encoding/json"
+	"reflect"
+	"testing"
+	"time"
+)
+
+func TestGlobMatch(t *testing.T) {
+	cases := []struct {
+		pattern, s string
+		want       bool
+	}{
+		{"foo*bar", "foobar", true},
+		{"foo*bar", "fooXXbar", true},
+		{"foo*bar", "foobaz", false},
+		{"*", "anything", true},
+		{"a*b*c", "axxbyyc", true},
+		{"a*b*c", "axxc", false},
+		{"literal", "literal", true},
+		{"literal", "other", false},
+		{`foo\*bar`, "foo*bar", true},
+		{`foo\*bar`, "fooXbar", false},
+	}
+
+	for _, c := range cases {
+		if got := globMatch(c.pattern, c.s); got != c.want {
+			t.Errorf("globMatch(%q,%q) = %v, want %v", c.pattern, c.s, got, c.want)
+		}
+	}
+}
+
+func TestEvalPath(t *testing.T) {
+	root := map[string]any{
+		"a": map[string]any{"b": []any{float64(10), float64(20), float64(30)}},
+		"s": "hi",
+	}
+
+	cases := []struct {
+		path    string
+		want    any
+		present bool
+	}{
+		{"$", root, true},
+		{"$.s", "hi", true},
+		{"$.a.b[2]", float64(30), true},
+		{"$['s']", "hi", true},
+		{"$.missing", nil, false},
+		{"$.a.b[9]", nil, false},
+	}
+
+	for _, c := range cases {
+		got, present, err := evalPath(c.path, root)
+		if err != nil {
+			t.Fatalf("evalPath(%q) error: %v", c.path, err)
+		}
+
+		if present != c.present {
+			t.Errorf("evalPath(%q) present = %v, want %v", c.path, present, c.present)
+		}
+
+		if present && !reflect.DeepEqual(got, c.want) {
+			t.Errorf("evalPath(%q) = %v, want %v", c.path, got, c.want)
+		}
+	}
+}
+
+func TestEvalPathUnsupportedFailsLoudly(t *testing.T) {
+	for _, p := range []string{"$.a[*]", "$..b", "$.a[?(@.x)]"} {
+		if _, _, err := evalPath(p, map[string]any{}); err == nil {
+			t.Errorf("evalPath(%q) should reject unsupported syntax", p)
+		}
+	}
+}
+
+func TestIntrinsics(t *testing.T) {
+	it := &interp{}
+	it.buildContext(&RunInput{ExecName: "run"}, nil)
+
+	input := map[string]any{"name": "Ada", "n": float64(2)}
+
+	got, err := it.evalIntrinsic(`States.Format('Hi {}, item {}', $.name, $.n)`, input)
+	if err != nil {
+		t.Fatalf("Format: %v", err)
+	}
+
+	if got != "Hi Ada, item 2" {
+		t.Errorf("Format = %q", got)
+	}
+
+	arr, err := it.evalIntrinsic(`States.Array($.name, 1, 'x')`, input)
+	if err != nil {
+		t.Fatalf("Array: %v", err)
+	}
+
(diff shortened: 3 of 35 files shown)


NitinKumar004 (commented):
Reviewed at head `a97fc596` in an isolated worktree vs `origin/development` merge-base `d7d3ca5`.

## Gates
- `go build ./...`, `go vet ./...`: clean
- `go test -count=1 ./providers/aws/sfn/... ./server/aws/sfn/... ./services/sfn/... ./compat/aws/...`: all `ok`
- `go test ./persist/... ./cmd/... .`: all `ok` (no regression from the new `Execution.History` field)
- `go generate ./...`: no diff (`coveragegen: wrote coverage for 76 services`)
- `golangci-lint run --timeout=9m ./providers/aws/sfn/...` (v2.11.4): 0 issues

## ResultPath-onto-raw: confirmed correct
`applyResultPath` (`providers/aws/sfn/asl/io.go:54-69`) splices the result onto a `deepCopy` of the state's pre-InputPath raw document, not the InputPath-filtered one. `TestInterpreterResultPathMergesOntoRawInput` genuinely proves this — I patched the merge base locally to the InputPath-filtered document and reran the test: it fails exactly as expected (loses `detail`/`meta`, only produces `{"id":1,"result":{"ok":true}}`), confirming the test would catch a regression to the old semantic. `ResultPath:null` (raw passthrough) and the default `$` (result replaces the document) are separately pinned and correct.

## ctx-forwarding: NOT actually fixed — High
`runExecution` → `interpret()` → `asl.Run(ctx, ...)` correctly forwards the caller's context up to the package boundary. But inside the package, `Run` discards it:

```go
// providers/aws/sfn/asl/interpret.go:84
func Run(_ context.Context, def *StateMachineDef, in *RunInput) *RunResult {
    ...
    res := it.walk(context.Background(), input)   // line 106 — fresh background context, not the caller's ctx
```

Every handler (`choiceHandler`, `passHandler`, etc.) receives `context.Background()`, never the context `StartExecution`/`StartSyncExecution` were called with. The PR's own stated purpose for this plumbing is "the PR2 Task→Lambda recursion guard" — if that guard is a `context.WithValue` depth counter set on the incoming ctx (mirroring the existing SQS/SNS/Lambda `StartExternal` recursion chokepoints elsewhere in the codebase), it will be invisible from inside `walk()`/handlers once PR2 lands, because `Run` fabricates a new context before it ever reaches them.

`TestInterpreterStartExecutionForwardsContext` (`interpreter_history_test.go:229-250`) is the only test touching this, and it only asserts `exec.Status != ""` on a value-carrying context — it never asserts the value is observable from inside the interpreter, so it passes identically whether `Run` forwards or discards ctx. This needs a fix (`it.walk(ctx, input)` at minimum) plus a test that actually observes a context value from inside a handler before this is safe plumbing for PR2.

## ResultSelector accepted on Pass but never applied — Medium
`validateFieldApplicability` (`validate.go:40-48`) rejects `Parameters`/`Result`/`ResultSelector`/`ResultPath` on Wait/Choice/Succeed/Fail, but not on Pass — which also doesn't support `ResultSelector` in real Step Functions. No code in the package ever reads `st.ResultSelector` (grep confirms it appears only in the struct field, doc comments, and that one rejection list). So `{"Type":"Pass","ResultSelector":{"x.$":"$.y"},"End":true}` is accepted at `CreateStateMachine`/`ValidateStateMachineDefinition` and the field is silently no-op'd at runtime — the "silently mis-run" failure mode the PR otherwise avoids. Task/Parallel/Map are unaffected since `unsupportedHandler` fails those states outright regardless of fields present.

## Minor / non-blocking
- Unsupported JSONPath syntax (wildcards/filters/recursive-descent) in `InputPath`/`OutputPath`/`ResultPath`/Choice `Variable`/comparand paths is only caught at run time (`jsonpath.go:45-51`'s `rejectUnsupportedPath`, invoked from `evalPath`), not at `CreateStateMachine`/`ValidateStateMachineDefinition` time — unlike `QueryLanguage: "JSONata"`, which is correctly rejected at parse time (`parse.go:117`). Still loud, never a wrong silent result — just a later validation point than JSONata gets. Worth tightening in a follow-up.
- `intrinsicFormat` (`intrinsics.go:154`) reuses the `emptyObject` (`"{}"`) constant from `errors.go` as the `States.Format` placeholder token. Correct today only because both roles share the literal `"{}"` — a naming trap if `emptyObject`'s value ever changes for its original purpose.

## What else checked out
Choice comparator families (String incl. glob `StringMatches`, Numeric, Boolean, Timestamp, type-checks, And/Or/Not, `...Path` variants) match ASL semantics on inspection. Task/Parallel/Map parse/validate at create time but fail loudly (`States.Runtime`, `unsupportedHandler`) at run time, never a silent success. History has monotonic IDs and a correct `PreviousEventID` chain, `ExecutionStarted → ... → ExecutionSucceeded/Failed`, `StopExecution → ABORTED` with `Error`/`Cause` and history trimmed to an `ExecutionAborted` terminal, snapshot/restore round-trips history byte-identically, and the `maxSteps` guard trips correctly on a Choice/Next cycle. The `TestTestState` correction is a genuine fix, not a weakened test: `TestState` takes a single bare ASL state object per the real API, not a whole `{"StartAt":...,"States":{...}}` machine as the old test passed — confirmed against the pre-image. No regressions in sfn CRUD, the wire handler, or the broader suite; `services/sfn/driver/driver.go` changes are additive only.

**Verdict: request changes** on the ctx-discard (High) before merge — it's the one piece of plumbing this PR exists partly to lay down for PR2, and it doesn't reach past the package boundary. The ResultSelector gap (Medium) should be closed too. Everything else here is solid.

Issue #826, opened 2026-08-27
GCP: external L7 HTTP(S) Load Balancer front-end unimplemented


NitinKumar004:
The external L7 HTTP(S) LB chain is not served — a real user cannot stand up a working L7 LB.

Ops returning 501/405 today:
- targetHttpProxies / targetHttpsProxies (insert/get/list/delete)
- sslCertificates (insert/get/list/delete)
- instanceGroups / regionInstanceGroups (insert/get/list/delete)
- backendServices.getHealth

Chain to make functional: forwardingRule → targetHttp(s)Proxy → urlMap → backendService → instanceGroup. Known gap flagged in the pre-release acceptance audit.

NitinKumar004:
Implemented in #863: targetHttpProxies/targetHttpsProxies, sslCertificates, instanceGroups/regionInstanceGroups (+ addInstances/listInstances) wired, so the forwardingRule→targetHttpProxy→urlMap→backendService→instanceGroup chain resolves end-to-end and backendServices.getHealth reflects group membership. Managed instance groups (instanceGroupManagers) are a separate follow-up.

PR #863, opened 2026-08-28
fix(gcp-lb): serve external L7 HTTP(S) LB front-end chain (#826)

NitinKumar004:
## Summary

Completes the GCP external L7 HTTP(S) Load Balancer front-end (#826). Before this change `server/gcp/loadbalancer` only routed `backendServices` and `forwardingRules`; the target-proxy, certificate and instance-group resource types were unserved, so the `forwardingRule → targetHttpProxy → urlMap → backendService → instanceGroup` chain was severed (501/405 for a real user).

## What changed

New resource types, served through the existing opaque `GCPComputeResourceStore` (no new driver surface, so `docs/coverage` is unchanged):

- **targetHttpProxies / targetHttpsProxies** — insert/get/list/delete, plus the `setUrlMap` (both) and `setSslCertificates` (https) action verbs.
- **sslCertificates** — insert/get/list/delete (self-managed cert model).
- **instanceGroups** (zonal, unmanaged) and **regionInstanceGroups** — insert/get/list/delete, plus `addInstances` / `removeInstances` / `listInstances`. Membership is stored in the resource and drives the group's reported `size`.

Wiring:

- The chain now resolves end-to-end: a forwarding rule's `target` → proxy, the proxy's `urlMap` → url map, the url map's `defaultService` → backend service, the backend service's `backends[].group` → instance group. Every reference round-trips verbatim so a client can walk the whole chain.
- `backendServices.getHealth` reflects the referenced instance group's registered members (HEALTHY per instance), falling back to the group reference itself when empty.
- Delete is guarded by reference integrity across the chain (url-map pinned by a proxy, certificate pinned by an https proxy, instance group pinned by a backend service, proxy pinned by a forwarding rule) — real GCP's `resourceInUseByAnotherResource`.
- The compute API addresses `setUrlMap` / `setSslCertificates` with a scopeless-global URL that the shared strict parser rejects; a local `parseLBPath` fallback handles just that form for the two global target-proxy types, leaving the shared parser untouched.

Managed instance groups (`instanceGroupManagers`) are intentionally out of scope — a documented follow-up; what `backendServices` reference (unmanaged instance groups) is fully wired.

## Tests

- `TestSDKGCPL7Chain` — provisions the full chain via the real compute SDK, walks it end-to-end, registers an instance and asserts `listInstances` + `getHealth` reflect membership, checks the in-use delete guard, then tears down in order.
- `TestSDKGCPTargetHTTPSProxyWithCertificate` — HTTPS proxy bound to an SSL certificate via `setSslCertificates`, with the certificate delete-guard.
- `TestGCPRegionInstanceGroupCRUD` — regional instance-group CRUD over raw HTTP (the SDK's RegionInstanceGroups client has no Insert).

### DIFF
--- a/server/gcp/loadbalancer/handler.go
+++ b/server/gcp/loadbalancer/handler.go
@@ -34,6 +34,7 @@ package loadbalancer
 
 import (
 	"net/http"
+	"strings"
 
 	"github.com/stackshy/cloudemu/v2/server/wire/gcprest"
 	lbdriver "github.com/stackshy/cloudemu/v2/services/loadbalancer/driver"
@@ -73,12 +74,14 @@ func New(lb lbdriver.LoadBalancer) *Handler {
 // the compute handler's /operations poll route.
 func (h *Handler) SetOperationRegistry(reg *gcprest.OperationRegistry) { h.ops = reg }
 
-// Matches returns true for /compute/v1/.../backendServices|forwardingRules
-// URLs. Disjoint from the compute (instances/operations/disks/…) and networks
-// (networks/subnetworks/firewalls) handlers, so registration order is
-// unconstrained.
+// Matches returns true for the load-balancing resource types — backendServices,
+// forwardingRules, healthChecks, targetPools, urlMaps, the L7 front-end chain
+// (targetHttpProxies, targetHttpsProxies, sslCertificates) and instanceGroups /
+// regionInstanceGroups. Disjoint from the compute (instances/operations/disks/…)
+// and networks (networks/subnetworks/firewalls) handlers, so registration order
+// is unconstrained.
 func (*Handler) Matches(r *http.Request) bool {
-	rp, ok := gcprest.ParsePath(r.URL.Path)
+	rp, ok := parseLBPath(r.URL.Path)
 	if !ok {
 		return false
 	}
@@ -92,7 +95,9 @@ func (*Handler) Matches(r *http.Request) bool {
 
 	switch rp.ResourceType {
 	case resourceBackendServices, resourceForwardingRules,
-		resourceHealthChecks, resourceTargetPools, resourceURLMaps:
+		resourceHealthChecks, resourceTargetPools, resourceURLMaps,
+		resourceTargetHTTPProxies, resourceTargetHTTPSProxies, resourceSslCertificates,
+		resourceInstanceGroups, resourceRegionInstanceGroups:
 		return true
 	}
 
@@ -101,7 +106,7 @@ func (*Handler) Matches(r *http.Request) bool {
 
 // ServeHTTP routes the request based on resource type and method.
 func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	rp, ok := gcprest.ParsePath(r.URL.Path)
+	rp, ok := parseLBPath(r.URL.Path)
 	if !ok {
 		gcprest.WriteError(w, http.StatusBadRequest, "invalid", "malformed path")
 		return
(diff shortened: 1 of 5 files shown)


NitinKumar004 (commented):
Reviewed at b3b76073 (worktree checkout, diff vs origin/development).

**Gates**: `go build ./...`, `go vet ./...` clean. `go test -count=1 ./server/gcp/... ./providers/gcp/... ./compat/gcp/...` all pass, including the new `TestSDKGCPL7Chain`, `TestSDKGCPTargetHTTPSProxyWithCertificate`, `TestGCPRegionInstanceGroupCRUD`. `go generate ./...` produces no diff. `golangci-lint run ./server/gcp/loadbalancer/...` reports exactly the 2 `nolintlint`/`dupl` hits the author called out at operations.go:262/413 — diffed those two function bodies (`deleteBackendService`, `deleteForwardingRule`) byte-for-byte against origin/development and they're unchanged, just shifted down by the new code above them. 0 new lint issues.

**Chain resolution**: confirmed end-to-end via `TestSDKGCPL7Chain` using the real `cloud.google.com/go/compute/apiv1` REST clients — forwardingRule.target → proxy → urlMap.defaultService → backendService.backends[].group → instanceGroup all round-trip and are walkable, `addInstances`/`listInstances`/`getHealth` reflect real membership, and the in-use delete guard fires across the whole chain (proxy pinned by forwarding rule, url map by proxy, cert by https proxy, instance group by backend service).

**parseLBPath (handler.go:195-225)**: narrow and safe. It only activates after the shared `gcprest.ParsePath` fails, requires the `/compute/v1/projects/` prefix, and only matches when `parts[1]` is exactly `targetHttpProxies` or `targetHttpsProxies` — every other resource type still 404s through the strict parser. `gcprest_test.go` (untouched by this diff) still passes, confirming the shared parser's strictness wasn't touched. One thing worth noting: the fallback isn't restricted to the `setUrlMap`/`setSslCertificates` action forms the PR description says it's "just for" — it also accepts a bare scopeless `GET`/`DELETE` on a named proxy (no action segment, `len(parts)==3`), which real GCP would 404. Harmless (it's strictly more permissive, not less strict), but worth either tightening to the two action verbs or updating the comment to match actual scope.

**Two real bugs found** (both reproduced with scratch tests against the pushed commit, not present in your test suite):

1. `backendServiceRefGroup` (server/gcp/loadbalancer/l7frontend.go:421-440), used by the instance-group delete guard, matches a referenced group purely by trailing name (`lastPathSegment(backends[j].Group) == groupName`) with no scope check. Repro: create instance group `ig` in zone `us-central1-a`, reference it from backend service `bs1`; create an unrelated instance group also named `ig` in zone `us-central1-b`; `Delete` on the zone-B group fails with `resourceInUseByAnotherResource` even though nothing references it. This is a real regression from the pattern already established two functions away in the same PR — `healthCheckInUse` (resources.go:328-354) correctly scopes via `tgs[i].Tags[bsScopeTag] != scope` before matching. `backendServiceRefGroup` should do the equivalent (compare the group reference's full zone/region + name, or at minimum the scope segment, not just the trailing name) before reporting a same-named group in a different zone/region as blocking.

2. `listScopeSegment` (server/gcp/loadbalancer/resources.go:370-376) only special-cases `ScopeGlobal`, defaulting everything else to `"regions/" + rp.ScopeName`. That was correct before this PR (only `regions` scope existed for opaque resources besides global), but this PR adds zonal `instanceGroups`, and the zonal list envelope's `id` field now reads `projects/{p}/regions/{zone}/instanceGroups` instead of `.../zones/{zone}/instanceGroups`. Confirmed via a scratch list call — item-level `selfLink`/`zone` are correct (they go through `gcprest.SelfLink`, which is scope-aware), only the collection envelope's `id` is wrong. Low real-world impact (clients don't generally parse this field) but should switch on `rp.Scope` (zones/regions/global) rather than binary global-vs-not.

Everything else — dispatch registration (registered before the compute fallback catch-all per gcp.go:202-214, disjoint resource-type set, no shadowing), regional vs. zonal scoping elsewhere, membership storage driving `size`/`getHealth`, and the documented deferral of managed instance groups (confirmed `instanceGroupManagers` isn't claimed by any handler's `Matches`, so it falls through to the compute fallback's structured error rather than being half-wired) — checks out.

Requesting changes for finding #1 (functional false-positive on a delete guard); #2 is a should-fix but non-blocking.

Issue #448, opened 2026-08-23
CloudFormation stack support


NitinKumar004:
cloudemu already serves the underlying AWS APIs (S3, EC2, DynamoDB, IAM, …) through the first-match dispatcher in `server/server.go`. A CloudFormation handler would deploy templates by calling those existing services — and it unblocks AWS CDK, which synthesizes to CloudFormation.

## What to build
- A new AWS handler (`server/aws/cloudformation`) with `CreateStack` / `UpdateStack` / `DeleteStack` / `DescribeStacks` / `DescribeStackResources`.
- Parse a template's `Resources`, resolve the common intrinsics (`Ref`, `Fn::GetAtt`), and create/delete each resource via the emulated services; track stack and per-resource state and status transitions.
- An example project and a tested create → assert → delete flow.

## Where
`server/aws/` (new handler), `server/server.go` (register it).

NitinKumar004:
Implemented in #864: CloudFormation stack orchestrator — CreateStack/UpdateStack/DeleteStack/DescribeStacks/GetTemplate provision real resources via the existing service drivers (S3/DynamoDB/SQS/SNS/Lambda/IAM/SecretsManager/SSM), with Ref/GetAtt/Sub/Join intrinsics, dependency ordering, real rollback on failure, and replacement propagation. JSON templates (YAML deferred).

PR #864, opened 2026-08-28
feat(cloudformation): AWS CloudFormation stack orchestrator (#448)

NitinKumar004:
## Summary

Adds a CloudFormation service (#448 — previously entirely absent, blocking SAM/CDK/CFN users). CloudFormation is implemented as an **orchestrator**: it provisions each stack resource by calling the **existing** AWS service driver for that type, so a stack's resources really exist in their own backends (S3, DynamoDB, …) and are queryable through those services' own SDKs. It does **not** store a copy of each resource — the same pattern as `features/topology` and `services/resourcediscovery`, which consume driver interfaces.

## Design

- **`services/cloudformation`** (provider-neutral, cross-service engine — no `driver/` subpackage, like `resourcediscovery`): JSON template parser (Parameters / Resources / Outputs), the common intrinsics, resource dependency ordering (topological sort with cycle detection), and the `Provisioner` contract + `Registry` (resource `Type` → provisioner).
- **`providers/aws/cloudformation`**: the in-memory stack store and the orchestration — CreateStack / UpdateStack / DeleteStack driving the registry, the status lifecycle, per-resource `StackEvents`, logical→physical mapping, and rollback on failure. Implements `snapshot.Snapshottable` so stacks persist across a snapshot/restore.
- **Provisioner registry** (`providers/aws/cloudformation_provisioners.go`, in the factory package like the `resourcediscovery` adapters): maps each resource type to a provisioner built from the **Layer-2 service driver interfaces** — so the orchestrator stays free of concrete-mock imports.
- **`server/aws/cloudformation`**: AWS query-protocol wire handler (registered before the EC2 catch-all), so the real `aws-sdk-go-v2` CloudFormation client / CDK / SAM / CFN work against it.

## Operations

CreateStack, UpdateStack, DeleteStack, DescribeStacks, DescribeStackEvents, ListStacks, DescribeStackResources, ListStackResources, GetTemplate.

Status lifecycle: `CREATE_IN_PROGRESS → CREATE_COMPLETE`, `UPDATE_*`, `DELETE_*`, and `ROLLBACK_COMPLETE` / `UPDATE_ROLLBACK_COMPLETE` on failure. Synchronous completion (like the rest of the emulator), but real statuses + a recorded event stream.

## Resource types (first wave)

`AWS::S3::Bucket`, `AWS::DynamoDB::Table`, `AWS::SQS::Queue`, `AWS::SNS::Topic`, `AWS::Lambda::Function`, `AWS::IAM::Role`, `AWS::SecretsManager::Secret`, `AWS::SSM::Parameter` — each provisioned via its existing driver, with logical→physical mapping and `Fn::GetAtt` attributes (e.g. `Arn`). An **unsupported resource type fails the stack and rolls it back** (AWS-accurate: "Resource type X is not supported"), rather than silently skipping.

## Intrinsics

`Ref`, `Fn::GetAtt`, `Fn::Sub` (both string and `[template, {vars}]` forms, plus `${!Literal}` escaping), `Fn::Join`, and pseudo-parameters (`AWS::Region`, `AWS::AccountId`, `AWS::StackName`, `AWS::Partition`, …).

## Deferred

- **YAML templates** (JSON only for now — YAML additionally needs the short-form `!Ref`/`!GetAtt` tags; noted as the clear next step).
- Additional resource types beyond the first wave.
- Update is replacement-based for changed resources (delete + create); unchanged resources are kept in place.

## Tests

- **Real-user end-to-end (`server/aws/cloudformation`, real SDK):** a CloudFormation client deploys a template that creates an S3 bucket + DynamoDB table; the resources are then queried through the real **S3 and DynamoDB** clients (proving they were provisioned in their own backends), the stack reports `CREATE_COMPLETE` with resolved `Ref`/`GetAtt` outputs, `GetTemplate` returns the body, events are recorded, and `DeleteStack` removes the resources. Plus update → `UPDATE_COMPLETE` and unsupported-type → `ROLLBACK_COMPLETE`.
- **Provider orchestration (`providers/aws/cloudformation`):** dependency ordering, output resolution, replacement update, teardown, rollback (unsupported + failing resource), missing-parameter rejection, duplicate-stack rejection, and snapshot/restore round-trip.
- **Unit (`services/cloudformation`):** parser, resolver (Ref/GetAtt/Sub/Join/pseudo), topological order + cycle detection.
- **Compat (`compat/aws`):** full lifecycle across every operation via the real SDK.

`docs/coverage/` regenerated (`go generate`).


NitinKumar004 (commented):
Re-reviewed the fix commit (069016c0) in isolation (`git diff e2472c63..069016c0`) against the three previously-flagged bugs.

**1. UpdateStack rollback (High, data-loss) — fixed correctly.**
`rollbackUpdate` (orchestrate.go) now: tears down everything the forward update created (`m.teardown(ctx, sd, created)`, safe against partial forward failure since it only touches ids actually in `sd.provisionOrder`), re-provisions everything it deleted/replaced from the captured `oldT` template (`m.provision(ctx, sd, oldT, resolver, removed)`, resolver correctly seeded with pre-update param values + surviving `keep` resources), then reverts template/params/description/outputs via `revertStackMeta`. `stackData.priorState()` copies `TemplateBody` (string, immutable), and fresh slices of `Parameter`/`Output` (both flat string-only structs, no shared mutable state) — genuine deep copy, not an alias.

Traced `TestUpdateStackFailureRollsBack` by hand against the `diffResources`/`classifyChanges` split: `NewThing` (created) and `MyTopic`'s v2 (created+replaced) both land in `created` and get torn down; `MyTopic`'s v1 lands in `removed` and gets re-provisioned from `oldT`; `Keeper` (untouched) is never touched by either step. Assertions correctly check all three: new resource gone, replaced resource restored, unrelated resource survives, template body reverted. `go test -race` clean.

Two non-blocking gaps worth a follow-up, not this commit's stated scope:
- `applyStackMeta` also overwrites `sd.stack.Tags`/`Capabilities` when the update input sets them, but `priorState`/`revertStackMeta` don't capture or restore those two fields — a failed update that changed tags/capabilities leaves the new tags/capabilities in place post-rollback.
- The restore step is explicit best-effort (`_ = m.provision(...)`, orchestrate.go): if re-provisioning from `oldT` itself errors (e.g. a physical-name collision), the stack still lands `UPDATE_ROLLBACK_COMPLETE` with a resource silently missing, rather than `UPDATE_ROLLBACK_FAILED`. Narrow (only hit on a second, nested failure) but real.

**2. CreateStack race (High) — fixed correctly.**
`claimStackSlot` does `SetIfAbsent` first, and only falls back to a store-locked `Update` (which itself is check-and-set under the store's single `sync.Mutex`-guarded `Lock()`) when the key already exists. Since `DeleteStack` never removes the map entry (only flips status to `DELETE_COMPLETE`), there's no window where a key present at `SetIfAbsent`-fail time could vanish before `Update` runs — no remaining check-then-act gap. `TestConcurrentCreateStackSameName` (20 goroutines) passes, and `go test -race -count=1 ./providers/aws/cloudformation/...` is clean.

**3. Replacement propagation (Medium) — fixed correctly.**
`directlyChanged` → `propagateChanges` → `classifyChanges` is a clean split. `propagateChanges` walks resources in `OrderResources` (topological) order and marks a resource changed if any of its direct deps is changed; since dependencies are always processed before dependents in that order, a single pass reaches the transitive fixpoint (no multi-pass needed, no risk of infinite loop — `OrderResources` already errors out on a cycle before `diffResources` does any mutation, so `UpdateStack` returns cleanly with no partial state change on a cyclic template). `TestUpdateReplacementPropagates` confirms `Bucket` v1→v2 propagates through `Fn::Sub "${Bucket}-events"` into `Topic`, landing on `bucket-v2-events` with the stale `bucket-v1`/`bucket-v1-events` gone.

**Gates (069016c0, GOMAXPROCS=2):**
- `go test -count=1 ./services/cloudformation/... ./providers/aws/cloudformation/... ./server/aws/... ./compat/aws/ ./providers/ ./persist/` — all pass (includes wiring-parity, `cloudformation_before_ec2` dispatch-ordering, snapshot completeness, real-SDK roundtrip, compat suite)
- `go test -race -count=1 ./providers/aws/cloudformation/...` — pass, clean
- `golangci-lint run --timeout=9m ./services/cloudformation/... ./providers/aws/cloudformation/... ./server/aws/cloudformation/... ./providers/aws/` — 0 new issues (one pre-existing `gocyclo` hit in `providers/aws/rds_discovery.go`, unrelated file/commit, unchanged by this PR)

All three targeted bugs are genuinely fixed with correct, well-isolated changes. The two rollback gaps above (tags/capabilities not reverted, best-effort restore can mask a nested failure) are real but narrow — worth a follow-up ticket, not blocking.

Issue #590, opened 2026-08-25
Architecture Theme 2: enforce dispatch/registration invariants as tests


NitinKumar004:
Architecture-audit Theme 2. The wire dispatcher is first-match-wins over a hand-ordered registration list; correctness depends on 15+ 'register before the catch-all' prose comments (RDS-before-EC2, GuardDuty-before-S3, Cosmos-before-BlobStorage, ...) with no test enforcing order. coveragegen also doesn't verify a wire handler is actually registered.

- [ ] Handler-ordering/ambiguity regression test per cloud (server/{aws,azure,gcp}): build the full Server, feed the known-ambiguous request shapes the comments already name, assert the correct handler wins
- [ ] REST-JSON path-prefix collision test: assert no two handlers claim overlapping reserved top-level path segments
- [ ] Extend coveragegen (or a sibling check) to cross-reference driver-discovered services against the Drivers-struct fields registered in server/<cloud>/<cloud>.go New() — flag driver+field-but-no-registered-handler
- [ ] Extend the Structure CI job with the STRUCTURE.md §3 same-filename-across-layers check

NitinKumar004:
Done: per-cloud dispatcher first-match-wins ordering regression tests (server/{aws,azure,gcp}/dispatch_ordering_test.go) in #592, plus coveragegen cross-check flagging dead/unregistered wire handlers and the same-filename-across-layers CI check in #603.

Issue #587, opened 2026-08-25
Architecture Theme 1: concurrency safety of mutable store entities + race CI


NitinKumar004:
Architecture-audit Theme 1. memstore.Store[V] makes only MAP ops atomic; ~71/96 stores hold *T pointers whose fields are mutated with no per-entity sync, and `go test -race` runs on only one package in CI — so the race class is invisible. (Already caught live once: PR #583's LB sub-resource race.)

- [ ] Add a non-blocking (advisory) `go test -race ./...` CI job so races surface
- [ ] Document the canonical idiom in CLAUDE.md Thread-Safety: a pointer entity stored in memstore and mutated after creation carries its own sync.Mutex (see SQS queueData.mu); read-modify-write uses memstore.Update (never Get+Set)
- [ ] Fix hotspots the race job flags, EC2 first; collapse EC2's duplicate State field into the statemachine
- [ ] Roll the idiom out provider-by-provider; flip the race job to blocking once green

NitinKumar004:
Landed in #866: the one real data race (Lambda alias get-then-mutate) is fixed with the per-entity-mutex idiom, concurrency stress tests added for Lambda/SQS/DynamoDB, the full tree is 'go test -race ./...' clean (306 ok, 0 races), and the advisory Race CI job is now BLOCKING (enforced). The broad provider-by-provider per-entity-lock refactor is unnecessary for correctness (coarse per-service locks are race-free); closing the correctness scope. Reopen/track separately if a scalability-oriented per-entity rollout is later wanted.

PR #865, opened 2026-08-28
chore(fidelity): cleanup batch — SFN/CloudTrail docs, SigV4 fail-closed, SSM history decryption, audit clock, ASL/API-GW/Lambda validation

NitinKumar004:
## Summary

A batch of small, independent fidelity/correctness follow-ups surfaced by prior reviews. Each touches disjoint files; each carries a focused test (or doc verification).

### 1. docs — stale "not implemented" claims (SFN + CloudTrail)
- **Step Functions**: rewrote the section that still claimed *"No ASL interpreter — the emulator does not execute the Amazon States Language"*. The interpreter shipped (#859/#860/#861) and really runs Pass/Choice/Wait/Task(→Lambda)/Parallel/Map with Retry/Catch, JSONPath I/O processing, intrinsics, the `$$` context object, real execution history, and create-time validation. Honestly notes the deferrals (JSONata, `.sync`/`.waitForTaskToken`, non-Lambda service integrations including Activity tasks, distributed-map runs) and the nesting/step guards.
- **CloudTrail**: fixed the now-false *"records no audit events / `LookupEvents` returns empty pages"* claim — management events are recorded from served requests (#502) and `LookupEvents` reflects them; only the Lake analytics/Insights surfaces remain synthesized.

### 2. SigV4 hardening follow-ups (#839)
- `checkExpiry` now **fails closed** on an unparseable `X-Amz-Date` (reject, not silently skip the skew/expiry check). `Verify` runs only under EnforceAuth, so the auth-off path is unchanged.
- STS session `Mint` **returns an error on crypto/rand failure** instead of a predictable, forgeable fallback credential; the five STS handlers surface it as `InternalFailure`.

### 3. SSM `GetParameterHistory` decryption (#857)
Forwards `WithDecryption` to the value read (was returning the opaque ciphertext even when `WithDecryption=true`), mirroring `GetParameter`. Driver signature updated across the portable wrapper and server.

### 4. GCP audit-log clock (#858)
`auditlog_recording.go` uses the injected clock (threaded from the provider) instead of `time.Now()`, so FakeClock tests are deterministic.

### 5. ASL Retrier bound + nesting test (#861)
`Retrier.IntervalSeconds` must be `> 0` (positive) per the ASL spec (only `MaxAttempts` is non-negative). Adds a test that trips the `CloudEmu.ExecutionNestingLimitExceeded` guard via deeply nested Parallel-in-Parallel.

### 6. API Gateway sibling validation + dispatch guard (#862)
`CreateResource` rejects a second variable sibling with a **different** path-parameter name (`{id}` then `{name}`) as a ConflictException — real API Gateway behavior; two variable siblings made `matchRoute` depend on Go map order. Also adds an `apigateway_before_s3` case to the dispatch-ordering table (execute-api handler stays ahead of the S3 catch-all, the #590 pattern).

### 7. Lambda alias weight bounds (#856)
Validates `AdditionalVersionWeights`: each weight in `[0.0, 1.0]` and the additional weights sum to at most `1.0`, rejected at PutAlias/UpdateAlias. (The concurrent-alias data race remains deferred to #587.)

## Testing
- `go build ./... && go vet ./...` — clean
- `go test ./server/... ./providers/... ./services/...` — pass
- `go generate ./...` — no regen (interface signature change did not alter the coverage listing)
- `golangci-lint run` on every touched package — 0 new issues vs `development`

New tests: SFN retrier/nesting, SigV4 fail-closed, STS Mint entropy, SSM history decryption, GCP audit clock determinism, API Gateway sibling rejection, Lambda weight bounds.

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -2627,25 +2627,50 @@ returned unchanged by `DescribeStateMachine`.
 | Definition tooling | TestState, ValidateStateMachineDefinition |
 | Tags | TagResource, UntagResource, ListTagsForResource |
 
-**No ASL interpreter — a deliberate simplification.** The emulator does not
-execute the Amazon States Language. `StartExecution` (and `StartSyncExecution`)
-complete the execution immediately: it transitions `RUNNING → SUCCEEDED` with
-output echoing the input, and `GetExecutionHistory` synthesises a minimal but
-valid event list (`ExecutionStarted → PassStateEntered → PassStateExited →
-ExecutionSucceeded`). Because nothing schedules real work, `GetActivityTask`
-returns an empty task token (the same long-poll-empty result real clients see),
-and `SendTaskSuccess/Failure/Heartbeat` reject any token as `InvalidToken`. This
-keeps behaviour deterministic and dependency-free while preserving the SDK wire
-shapes for build-and-orchestrate testing.
+**A real ASL interpreter runs executions.** `StartExecution` and
+`StartSyncExecution` actually interpret the Amazon States Language definition
+against the input rather than echoing it. The interpreter executes **Pass,
+Choice, Wait, Task, Parallel, Map, Succeed, and Fail** states, following
+`Next`/`End` transitions to a terminal state and producing the real
+`GetExecutionHistory` event sequence for the path taken (state Entered/Exited,
+Task Scheduled/Started/Succeeded/Failed, Parallel/Map iteration events, and the
+terminal `ExecutionSucceeded`/`ExecutionFailed`). Supported semantics:
 
-Consistent with the no-interpreter model: `RedriveExecution` records a fresh
-redrive date on an existing execution rather than re-running it; `TestState`
-succeeds for any non-empty definition and echoes the input as output;
-`ValidateStateMachineDefinition` returns `OK` for a non-empty, valid-JSON
-definition and `FAIL` with a diagnostic otherwise (no semantic ASL validation).
-Distributed-map Map Runs are never produced by execution, so `ListMapRuns`
-returns empty for ordinary executions; `DescribeMapRun`/`UpdateMapRun` operate on
-Map Run records seeded through the provider's `SeedMapRun` helper.
+- **I/O processing** — `InputPath`, `Parameters`, `ResultSelector`,
+  `ResultPath`, and `OutputPath` with JSONPath evaluation, including the `.$`
+  reference form and merge-onto-input `ResultPath` semantics.
+- **Choice rules** — string/numeric/boolean/timestamp comparators and the
+  `And`/`Or`/`Not` combinators (including the `...Path` variants), with `Default`.
+- **Error handling** — `Retry` (with `IntervalSeconds`/`MaxAttempts`/`BackoffRate`
+  and `ErrorEquals` matching) and `Catch` (routing to a fallback state with the
+  error carried on `ResultPath`).
+- **Intrinsic functions** — `States.Format`, `States.Array`,
+  `States.ArrayGetItem`, `States.StringToJson`, `States.JsonToString`, and
+  `States.UUID`.
+- **The context object** — `$$` resolves the execution/state context
+  (`Execution.Name`, `StateMachine`, `State.EnteredTime`, …).
+- **Task → Lambda** — a `Task` whose `Resource` is `arn:aws:states:::lambda:invoke`
+  (payload from `Parameters.Payload`) or a direct Lambda function ARN really
+  invokes the wired Lambda backend; a `FunctionError` feeds `Retry`/`Catch`.
+
+Definitions are **validated at create time**: `CreateStateMachine` (and
+`ValidateStateMachineDefinition`) parse the ASL and reject structural errors —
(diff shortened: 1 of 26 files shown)


NitinKumar004 (commented):
Reviewed at b1b2d88e (head). Verified all 7 fixes individually; gates green (build/vet clean, `go test -count=1 ./server/... ./providers/... ./services/...` — 281 packages ok, 0 fail; `go generate ./...` no diff; `golangci-lint --new-from-rev=origin/development` across every touched package — 0 issues).

1. **docs/services.md (SFN/CloudTrail)** — checked every capability claim against code, all accurate:
   - ASL: Pass/Choice/Wait/Task/Parallel/Map/Succeed/Fail, intrinsics (Format/Array/ArrayGetItem/StringToJson/JsonToString/UUID) all present in `providers/aws/sfn/asl/intrinsics.go`; `$$` context object (`Execution.Name/StateMachine/State.EnteredTime`) built in `interpret.go:buildContext/enterStateContext`; `arn:aws:states:::lambda:invoke` + bare-ARN Task integration in `task.go`; `CreateStateMachine` calls `validateASL` at create time (`statemachines.go:53`). Deferred list (JSONata, `.sync`/`.waitForTaskToken`, non-Lambda integrations, Activity tasks) matches what's actually absent.
   - CloudTrail: `server/aws/cloudtrail_recording.go` (pre-existing, not part of this diff) derives management events from `X-Amz-Target`/`Action` + SigV4 credential scope, skips CloudTrail's own read-only calls; `LookupEvents` (`providers/aws/cloudtrail/events.go:102`) sorts newest-first with an id tiebreak and paginates by cursor. Doc claim matches. No stale "no interpreter"/"no real event stream" language left elsewhere in the doc or repo.

2. **SigV4 fail-closed (#839)** — `checkExpiry` now returns `AccessDenied` on an unparseable `X-Amz-Date` instead of `nil`. Confirmed `sigv4.Verify`/`checkExpiry` are only reachable through `newAuthGate`, which is wired in `aws.go:723` only when `d.EnforceAuth` is true — enforce-auth-off path never calls this code, so it's byte-identical as claimed. `randUpperAlnum`/`Mint`/`synthCredentials` now propagate a `crypto/rand` failure up to `mintCredentials`, which writes `InternalFailure` instead of minting a predictable credential; all 5 STS handler call sites updated. `synthCredentials`'s "no sessions wired" branch (auth-off default) still returns the fixed synthetic values with a nil error, so behavior there is unchanged.

3. **SSM GetParameterHistory decryption (#857)** — signature change threaded correctly: `driver.ParameterStore` interface → `services/parameterstore/parameterstore.go` → `providers/aws/ssm/ssm.go` (now calls `revealValue` per version) → `server/aws/ssm/operations.go` passes `req.WithDecryption` (matching the sibling `GetParameter`/`GetParameters` calls in the same file, so no default mismatch). Only one implementor of the interface exists (AWS SSM); `go generate` clean.

4. **GCP audit clock (#858)** — `Provider.Clock` exposed from `providers/gcp/gcp.go`, threaded through `DriversFrom` → `server/gcp/gcp.go Drivers.Clock` → `installAuditObserver` → `recordAuditLogEvent(logs, r, clock)`, replacing `time.Now()`. `config.Option.Clock` defaults to `RealClock{}` (never nil from `New(opts...)`), and `installAuditObserver` itself nil-guards to `RealClock{}` as a second line of defense. New test asserts the recorded timestamp equals the injected `FakeClock` time exactly.

5. **ASL Retrier IntervalSeconds (#861)** — bound changed from `< 0` to `<= 0` (must be `> 0`), matching the real ASL spec (only `MaxAttempts` may be 0). Nesting test builds an actual 40-deep nested-Parallel definition against `defaultMaxNesting = 32` (`interpret.go:21`) and asserts the specific `CloudEmu.ExecutionNestingLimitExceeded` error — not tautological, it runs the real interpreter to the guard.

6. **API Gateway variable-sibling conflict (#862)** — `findVariableSibling` correctly scoped to `ParentID` match; rejects a second variable child with a different name while still allowing a literal sibling alongside the existing variable child, matching real API Gateway's route-matching constraint. `apigateway_before_s3` dispatch-ordering case added and passes.

7. **Lambda AdditionalVersionWeights bounds (#856)** — per-weight `[0.0,1.0]` check plus sum-`<=1.0` check added to the existing `$LATEST`/version-exists validation, applied uniformly since `CreateAlias`/`UpdateAlias` share `validateRoutingConfig`. Test covers over-1.0, negative, sum-over-1.0-on-update, and a valid in-bounds case — good coverage, not tautological.

Non-blocking observation: Azure Functions and GCP Cloud Functions alias weighted-routing (`providers/azure/functions/aliases.go`, `providers/gcp/cloudfunctions/aliases.go`) still don't `validateRoutingConfig` at all — no `$LATEST` rejection, no weight bounds. Pre-existing gap, not introduced by this PR, but the "mirror across all 3 providers" behavior described in the project docs doesn't yet hold for weighted aliases outside AWS.

No blockers found. All 7 fixes are individually correct and tests are non-tautological.

PR #866, opened 2026-08-28
fix(lambda): guard alias in-place mutation; enforce -race CI gate (#587)

NitinKumar004:
## Summary

Fixes the one real data race in the mutable-store concurrency class (#587) and, now that the tree is race-clean, promotes the advisory `-race` CI job to a blocking gate.

### The race
`UpdateAlias` fetched the shared `*aliasData` from the store via `fd.aliases.Get()` and mutated its `alias` fields (`FunctionVersion`, `Description`, `RoutingConfig`, `RevisionID`) with no lock — a classic get-then-mutate race against concurrent `UpdateAlias`/`GetAlias`/`ListAliases`. The new stress test flags 33 `DATA RACE` reports without the fix.

### Fix (follows the documented `queueData.mu` / `instanceData.mu` idiom)
- `aliasData` gains its own `sync.Mutex` guarding `alias`; held around every read/write in `UpdateAlias`, `GetAlias`, `ListAliases`. `CreateAlias` builds the entity fully before `Set`.
- Dropped the redundant `fd.aliases.Set(name, ad)` in `UpdateAlias` (ad is the shared pointer already in the store).

### Stress tests (surface latent races under `-race`)
- `providers/aws/lambda`: `TestConcurrentAliasRaceFree` — N goroutines hammering Update/Get/List on one alias.
- `providers/aws/sqs`: `TestConcurrentQueueRaceFree` — Send/Receive/Set/Get on one queue.
- `providers/aws/dynamodb`: `TestConcurrentItemRaceFree` — Put/Update/Get/Scan on one item (guards the copy-on-write invariant).

### CI gate
`go test -race ./...` is now blocking (`Race` job). Verified locally: full tree `go test -race ./...` = **306 ok, 0 DATA RACE, 0 FAIL**. `docs/architecture.md` updated to say the gate is enforced.

No other real races were found by the detector across the whole tree.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -252,7 +252,7 @@ jobs:
       - run: go test -race ./providers/aws/bedrock/... ./server/aws/bedrock/... ./providers/aws/bedrockagent/... ./server/aws/bedrockagent/... ./server/aws/bedrockagentruntime/...
 
   race:
-    name: Race (advisory)
+    name: Race
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
@@ -268,15 +268,13 @@ jobs:
           key: go-${{ runner.os }}-${{ hashFiles('**/go.sum') }}
           restore-keys: |
             go-${{ runner.os }}-
-      # Advisory whole-tree race detector. -race is 2-10x slower and much of the
-      # codebase is not race-clean yet (mutable entities held as *T behind
-      # memstore.Store have their fields mutated without per-entity sync), so this
-      # is non-blocking for now: it surfaces data races as warnings while the
-      # per-entity-mutex idiom (see SQS queueData.mu; use memstore.Update, never
-      # Get+Set) is rolled out provider-by-provider. Flip to blocking once green.
-      # Tracking: github issue "Architecture Theme 1".
-      - name: go test -race ./... (advisory)
-        run: go test -race -timeout 20m ./... || echo '::warning::go test -race reported findings (advisory; concurrency hardening in progress)'
+      # Blocking whole-tree race detector. The per-entity-mutex idiom (see SQS
+      # queueData.mu / EC2 instanceData.mu; use memstore.Update, never Get+Set) is
+      # rolled out across the mutable-store entities, so the tree is race-clean and
+      # any newly introduced data race must fail CI. -race is 2-10x slower, hence
+      # the generous timeout.
+      - name: go test -race ./...
+        run: go test -race -timeout 20m ./...
 
   tidy:
     name: Tidy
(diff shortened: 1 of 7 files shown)


NitinKumar004 (commented):
Reviewed at dcd67d5d.

**Alias race fix (providers/aws/lambda/aliases.go, lambda.go)**

- `aliasData.mu` guards every read/write of the mutated fields: `UpdateAlias` locks before computing `effectiveVersion` and holds through all three mutations (`FunctionVersion`, `Description`, `RoutingConfig`, `RevisionID`) via `defer ad.mu.Unlock()`; `GetAlias` and `ListAliases` lock around the `result := ad.alias` copy. `CreateAlias` builds the `driver.Alias` value fully before the entity is ever inserted into the store (`fd.aliases.Set(cfg.Name, &aliasData{alias: a})`), so it's not racing anything — correct not to need the lock there. `DeleteAlias` never touches `ad.alias` fields, just removes the store entry, so no lock needed there either.
- No deadlock / lock-ordering issue: in every path, `fd.aliases.Get`/`.All()` (which takes and releases memstore's internal lock) completes and returns *before* `ad.mu.Lock()` is ever taken — the store lock and the entity lock are never nested in either order. The redundant `fd.aliases.Set` removed from `UpdateAlias` was correct to drop since `ad` is the same pointer already in the store; removing it also removes what would otherwise be a store-lock-under-entity-lock call. Confirmed by reading `internal/memstore/store.go`: `Get`/`Set`/`Has`/`All` each lock/unlock internally per call, no callback held under the lock.
- The shallow copy returned by `GetAlias`/`ListAliases`/`UpdateAlias` is safe even for the `*driver.AliasRoutingConfig` pointer field: `copyRoutingConfig` always allocates a fresh struct+map on every set, and the old one is never mutated in place afterward — it's replaced wholesale, not updated in place — so a caller holding a stale copy's pointer never observes a later mutation and never races on it.

**Stress tests are real**

Verified by reverting just the aliases.go/lambda.go fix and re-running `TestConcurrentAliasRaceFree` under `-race`: it reliably reports data races (Update vs Get, Update vs List, Update vs Update) on the exact fields the fix protects. With the fix restored, `go test -race -count=1 ./providers/aws/lambda/... ./providers/aws/sqs/... ./providers/aws/dynamodb/...` is clean (0 races). Also spot-checked `-race` on `./providers/aws/ec2/...` and `./services/serverless/...` — clean.

**CI gate flip (.github/workflows/ci.yml)**

Job id stays `race` (only `name:` changed from "Race (advisory)" to "Race"), no other job has a `needs:` on it, and the diff is purely: drop the `|| echo ::warning::` swallow and the advisory comment block. `actionlint` reports no new issues on this job (only pre-existing, unrelated shellcheck warnings on other jobs). `docs/architecture.md` correctly updated to describe the gate as blocking now.

**Full gate run** (GOMAXPROCS=2, dcd67d5d)

- `go test -count=1 ./...` — 306 ok, 0 FAIL
- `go test -count=1 ./providers/aws/lambda/... ./providers/aws/sqs/... ./providers/aws/dynamodb/... ./server/aws/...` — all ok
- `go test -race -count=1 ./providers/aws/lambda/... ./providers/aws/sqs/... ./providers/aws/dynamodb/...` — 0 races
- `golangci-lint run --new-from-rev=origin/development` on the three touched provider packages — 0 issues

No deadlock/lock-ordering issue found, and the CI flip is a clean gate promotion with no behavioral change to what's run. MERGE.

Issue #877, opened 2026-08-29
Kubernetes: reach real-cluster (minikube) parity — tracking


NitinKumar004:
Tracking issue for closing the gap between CloudEmu's Kubernetes data-plane and a real cluster / minikube.

## Where we are (audit verdict)
CloudEmu k8s is **not** a thin CRUD mock — it's a *synchronously-reconciled emulated cluster*: creating a Deployment really produces a ReplicaSet → N Running Pods with IPs on a Node; Services get Endpoints/EndpointSlices from Pod labels; Jobs run to Succeeded; StatefulSets get ordinals + Bound PVCs; cascade-GC works; CRDs + custom resources work; there's a real watch stream. The full `kubectl create → scale → set-image → expose → apply → delete` lifecycle runs end-to-end. But it converges **instantly and synchronously** — there is no real control plane of async loops (no scheduler, kubelet lifecycle, etcd-with-time, or events).

## Gaps to 100% real-cluster parity (priority order)
1. #871 — server-side **Table printing** (kubectl get shows only NAME+AGE) — highest user-visible.
2. #872 — typed lists omit **list-level resourceVersion** → `kubectl rollout status` hangs / informers weak (+ STS/DS updateStrategy defaulting).
3. #873 — reconciler emits **no Events** (get events / describe empty).
4. #874 — **pod lifecycle progression** (pods teleport to Running) + object-fidelity nits (Job labels, Node creationTimestamp, materialized-pod RV reset).
5. #875 — **scheduler semantics** (single fixed node, every pod force-assigned).
6. #876 — **async/time progression** + empty kube-system (CronJob needs manual tick; no coredns/kube-proxy).

## Related
- #868 — **persistence**: the k8s data-plane has no Snapshot/Restore surface and is excluded from `persist.ExportAll`; state is worth persisting (real materialized children) but greenfield. Cluster-UID indirection (kubeconfigs point at `/k8s/<uid>`) must round-trip.

Suggested sequencing: #871 + #872 first (they're the difference between 'looks broken' and 'looks like a real cluster'), then #873, then persistence #868, then the deeper loops #874–#876.

NitinKumar004:
[...]

Issue #875, opened 2026-08-29
Kubernetes: no scheduler semantics (single fixed node, every pod force-assigned)


NitinKumar004:
Every Pod is stamped `spec.nodeName=cloudemu-node-0` at materialization (`services/kubernetes/reconcile.go:59-61`) against one bootstrapped Ready node (`state.go:154`). No scheduling loop: no unschedulable Pods, no resource requests/limits accounting, no taints/tolerations, node affinity/anti-affinity, or nodeSelector-respecting placement. Multi-node and 'Pending because unschedulable' scenarios can't be reproduced.

## What to build (deterministic emulation, not a real scheduler)
- Support multiple registered Nodes; honor `nodeName`, `nodeSelector`, basic taints/tolerations and requests-vs-allocatable when assigning.
- Leave a Pod Pending/Unschedulable (with a condition + Event) when nothing fits.

## Where
`services/kubernetes/reconcile.go`, `state.go`.

Issue #874, opened 2026-08-29
Kubernetes: pods teleport to Running (no lifecycle progression) + object-fidelity nits


NitinKumar004:
Materialized Pods jump straight to `Running` with all conditions true (`services/kubernetes/reconcile.go:43-90`) — no Pending→ContainerCreating→Running progression, no restarts, no readiness/liveness probe gating, no crash/backoff states. Controller-materialized Pods also reset `resourceVersion` to '1' (`reconcile.go:118`). Minor object-fidelity gaps from the same audit: Job Pods lack the real `job-name`/`controller-uid` labels; Node has no `creationTimestamp` (kubectl AGE shows `<unknown>`).

## What to build (deterministic, no real containers)
- Optional state-machined pod phase progression (Pending→ContainerCreating→Running) driven by the reconcile/clock rather than instant, keeping determinism for tests.
- Preserve a monotonic `resourceVersion` on materialized pods instead of resetting to '1'.
- Add `job-name`/`controller-uid` labels to Job pods; stamp Node `creationTimestamp`.

## Where
`services/kubernetes/reconcile.go`.

NitinKumar004:
Done — opt-in pod lifecycle progression + object-fidelity nits in #878; Job pod labels (batch.kubernetes.io/job-name, controller-uid) in #885.

Issue #873, opened 2026-08-29
Kubernetes: reconciler emits no Events (kubectl get events / describe always empty)


NitinKumar004:
`Event` is a stored kind with `involvedObject` field selectors (`services/kubernetes/registry.go:825-861`), but no reconciler ever emits one (no `Eventf`/`recordEvent` anywhere), so `kubectl get events` and the Events section of `kubectl describe` are always empty.

## Real k8s
Controllers emit Events: scheduler `Scheduled`; kubelet `Pulling`/`Pulled`/`Created`/`Started`; deployment `ScalingReplicaSet`; replicaset `SuccessfulCreate`; job `SuccessfulCreate`/`Completed`; etc.

## What to build
- An event recorder used by the reconcile hooks; emit the common Events on create/scale/materialize/complete/delete.
- Store them as core/v1 Events (involvedObject, reason, message, type, count, timestamps from `config.Clock`).

## Where
`services/kubernetes/reconcile.go` (emit points) + an event helper.

NitinKumar004:
Done — reconciler emits core/v1 Events (ScalingReplicaSet/SuccessfulCreate/Scheduled/…) in #878; kubectl describe/get events populated.

PR #878, opened 2026-08-29
feat(kubernetes): real-cluster parity — Table printing, list resourceVersion, events, node/kube-system, opt-in pod lifecycle (#877)

NitinKumar004:
Closes the Kubernetes real-cluster parity epic (#877): #871, #872, #873, plus #875/#876 (scoped) — makes the emulated k8s data-plane behave and *display* like a real cluster to `kubectl`. North-star model is KWOK (status-rewriting controller, no real kubelet).

## What now works end-to-end
- **Server-side Table printing (#871):** `kubectl get pods/deployments/services/nodes/...` shows real columns (READY/STATUS/UP-TO-DATE/AVAILABLE, `-o wide` extras) instead of raw JSON. `meta.k8s.io/v1 Table` on `Accept: as=Table`; unknown kinds fall back to NAME/AGE.
- **List-level resourceVersion (#872):** a single monotonic cluster RV counter replaces the previous three uncoordinated schemes (per-object `bumpResourceVersion`, `registryStore.rv`, hardcoded `"1"`); every list carries a collection RV, so `kubectl rollout status` and client-go reflectors complete their List→Watch handshake instead of hanging. Dry-run keeps a non-mutating RV peek (no side effect). Typed watches emit a post-sync BOOKMARK.
- **Events (#873):** controllers emit `core/v1` Events (`ScalingReplicaSet` on real replica-count change, `SuccessfulCreate` for RS/STS/DS/Job/CronJob pods, `Scheduled`, `Started`, `Completed`), deduped by (involvedObject, reason, message, type) with a rolling `count` — `kubectl describe`/`get events` show real activity.
- **Node + kube-system (#875/#876 scoped):** the node reports Ready + capacity + a `coordination.k8s.io` Lease in `kube-node-lease`; seeded kube-system components; pods routed through a scheduler stamp emitting `Scheduled`.
- **Opt-in Pod lifecycle progression (#874/#876):** default = pods instant-Running (unchanged, deterministic). `SetLifecycleProgression(true)` / `cloudemu serve --k8s-progression` starts pods `Pending` and advances `Pending→ContainerCreating→Running` (→Terminating on delete) on a logical clock, emitting the kubelet event sequence. Tests drive it via `Tick()` (deterministic under FakeClock); live serve runs a real-time ticker (`--k8s-progression-interval`, default 1s).

## Tests
New `k8s_parity_test.go`: one end-to-end flow (create Deployment → Table columns → rollout status → events → scale → expose→endpoints → Job on a tick → describe → watch-resume → bare Pod Pending→Running) plus an invariant guard (progression off → immediate Running). Existing k8s suite stays green (progression is default-off).

## Deferred (with reasons)
True multi-node scheduling (DaemonSet fan-out/endpoint node-stamping is invasive), watch resume-from-arbitrary-RV + stale-410 (needs an event-history ring-buffer), KWOK `Stage` CRD (chaos), and k8s persistence (#868).


NitinKumar004 (commented):
Gates on the pushed commit (`ba9476f`): `go build ./...`, `go vet ./...`, `go test ./...` (scoped: `services/kubernetes`, `server/serverkit`, `cmd/cloudemu`, root), `go test -race ./services/kubernetes/...`, `gofmt`, `go mod tidy`, `go generate` (docs), and `contrib/server` build+test all pass. `golangci-lint` scoped to `services/kubernetes/...` shows only the pre-existing `admission.go:351` gosec finding — no new lint issues from this PR.

## High

**Dry-run (and admission/quota-denied) writes on every hand-written typed resource silently advance the global cluster resourceVersion — contradicts the PR's own documented invariant.**

`state.go:212-217` documents `peekClusterRVLocked` as existing because dry-run echo paths "must have zero side effects," and the registry-backed (unstructured) kinds honor that correctly (`registry.go:358,444,525,610` use the peek). But the 8 hand-written typed-resource files never got the same treatment:

- `pod.go:332` (`updatePod`) / `:392` (`patchPod`) — `in.ResourceVersion = s.nextClusterRVLocked()` / `patched.ResourceVersion = s.nextClusterRVLocked()` run *before* the `isDryRun(r)` check a few lines below.
- Same pattern in `deployment.go:262,310` (+ create at `:147`), `service.go:269,327` (+ `:137`), `secret.go:246,300` (+ `:133`), `configmap.go:249,299` (+ `:137`), `serviceaccount.go:234,269` (+ `:127`), `namespace.go:195,240` (+ `newNamespaceObject`, `:363-374`), `pdb.go:161` (+ `:107`).
- The shared `stamp()` helper (`configmap.go:371-375`, used by every `createX`) runs before `admit()` too, so an admission-*denied* create also burns an RV tick even though nothing persists.

I verified this is real, not a misreading: a throwaway test that did a dry-run PUT on ConfigMap `cm1` and then a real create of an unrelated `cm2` got `cm2` RV 13, vs. RV 12 with no intervening dry-run. No existing test — including this PR's own `k8s_parity_test.go` / `phase1_fidelity_test.go` — asserts the counter is unchanged by a dry-run; the existing tests only check "not persisted."

Concrete impact: `kubectl apply --dry-run=server` / `kubectl diff` / GitOps preview steps / IDE live validation against Pods, Deployments, Services, Secrets, ConfigMaps, ServiceAccounts, Namespaces, or PDBs — i.e. most of what a real cluster user touches — advance the cluster RV with zero-effect requests, directly contradicting this PR's own claim that "Dry-run keeps a non-mutating RV peek (no side effect)." Delete-dry-run is unaffected on every file (checked) — only create/update/patch on typed kinds leak.

Fix: split `stamp()` into a metadata-only stamp (UID + CreationTimestamp, needed before admission runs) and defer the resourceVersion to `peekClusterRVLocked()` in the dry-run echo branch / `nextClusterRVLocked()` immediately before the real persist — mirroring `registry.go`'s already-correct create/update/patch flow — across all 8 typed handler files.

## Medium

**Claimed progression-delete ("→Terminating on delete") and Service/Namespace Table rendering ship with zero test coverage.**

`lifecycle.go:161` (`reapTerminatingLocked`) and `:176` (`beginTerminatingLocked`) are 0% covered — no test deletes a Pod under progression, despite the PR body explicitly claiming "(→Terminating on delete) on a logical clock." `table_columns.go:191` (`serviceTableProjector`) and `:211` (`namespaceTableProjector`) are also 0% covered, despite the PR body's headline claim that `kubectl get pods/deployments/services/nodes/...` shows real columns. (`reconcile.go:488` `reconcileStatefulSet` and `registry_ops.go:139` `registryStatus` are similarly untouched by any test, pre-existing or new.)

I wrote a throwaway test driving delete-under-progression end to end (Running → DELETE → still 200/Terminating → Tick → 404) and it passed — the code is functionally correct today, so this is a coverage gap rather than a live bug, but a future regression to either path would go uncaught by CI.

Fix: extend `k8s_parity_test.go` (or a new `table_test.go`) with a pod-delete-under-progression assertion and Service/Namespace Table-column assertions.

## Low

**Two independently-written `envBoolOr` helpers added in the same PR** — `cmd/cloudemu/serve.go:249` and `contrib/server/main.go:301` — same new `--k8s-progression` flag, two separate implementations (different signatures) across the two modules. Not a lint violation (separate `go.mod`s) but worth consolidating if there's a natural shared-util home.

## Verified clean

RV-refactor sweep is otherwise complete — no leftover `bumpResourceVersion` calls or hardcoded `"1"` resourceVersion literals anywhere in the package. Controller-built Pods in `reconcile.go` correctly source `nextClusterRVLocked`. The pre-existing k8s suite is fully preserved (progression-off invariant intact; the `pod_test.go`/`secret_test.go`/`serviceaccount_test.go` diffs are legitimate adaptations to the new monotonic-counter scheme, not weakenings). Table column sets match real kubectl output (Deployment correctly gets READY/UP-TO-DATE/AVAILABLE, not ReplicaSet's DESIRED/CURRENT). `wantsTable` Accept-header negotiation is correctly scoped, and scale/status/SAR/eviction/discovery payloads correctly bypass the Table wrapper. Events are properly delta-gated (`ScalingReplicaSet` only on an actual replica-count change) and deduped with a capped, bounded store; STS/DS pods get `SuccessfulCreate` via the shared `syncStablePods` path. The real-time progression ticker (`server/serverkit/serverkit.go`) is serve-only (never started in tests), lifecycle-managed (started in `Serve`, stopped+drained in `shutdown`), and snapshots the cluster map under `RLock` before ticking each cluster outside the lock — no concurrent-map race, confirmed under `-race`. No new `time.Now()` was introduced; all new timestamps route through `config.Clock`.

services/kubernetes/pod.go
```diff
@@ -315,7 +329,7 @@ func (s *ClusterState) updatePod(w http.ResponseWriter, r *http.Request, namespa
 	in.Namespace = namespace
 	in.UID = cur.UID
 	in.CreationTimestamp = cur.CreationTimestamp
-	in.ResourceVersion = bumpResourceVersion(cur.ResourceVersion)
+	in.ResourceVersion = s.nextClusterRVLocked()
```
NitinKumar004: This runs before the `isDryRun(r)` check a few lines below (line 341), so a dry-run PUT still advances the shared cluster resourceVersion counter even though nothing is persisted. `registry.go`'s equivalent update path (`registryUpdate`, line ~444) correctly uses `peekClusterRVLocked()` in the dry-run branch instead. Same pattern repeats in `deployment.go`, `service.go`, `secret.go`, `configmap.go`, `serviceaccount.go`, `namespace.go`, `pdb.go` — see the summary comment for the full list.

services/kubernetes/pod.go
```diff
@@ -375,7 +389,7 @@ func (s *ClusterState) patchPod(w http.ResponseWriter, r *http.Request, namespac
 		return
 	}
 
-	patched.ResourceVersion = bumpResourceVersion(cur.ResourceVersion)
+	patched.ResourceVersion = s.nextClusterRVLocked()
```
NitinKumar004: Same issue as the `updatePod` RV stamp above: this mutates the global counter before the `isDryRun(r)` check at line 404, so a dry-run PATCH also leaks a tick. Should defer to `peekClusterRVLocked()` for the dry-run echo, mirroring `registryPatch`.

NitinKumar004 (commented):
Inline anchors for the High finding in the summary review (dry-run RV leak on typed handlers).

Issue #868, opened 2026-08-29
Kubernetes data-plane state is not persisted (--persist / snapshot omits it)


NitinKumar004:
`--persist` and `/_cloudemu/snapshot` only cover the 4 cloud providers (each `Provider.SnapshotServices()` in providers/{aws,azure,gcp,oci}). The shared Kubernetes data-plane (`services/kubernetes`, mounted as the k8s backend in serverkit) has no `Snapshottable` surface, so pods/deployments/services created via EKS/AKS/GKE are lost on restart even with `--persist` on.

## What to do
- Give the kubernetes data-plane store a `Snapshottable` (Snapshot/Restore) surface (identity-preserving, per the #582 pattern).
- Include it in the persisted snapshot — extend the serverkit persist wiring to cover the shared k8s backend alongside the 4 providers.
- Extend `persist/completeness_test.go` so a stateful k8s store that isn't Snapshottable fails CI.
- Round-trip test: create k8s objects, snapshot, restart, assert survival.

NitinKumar004:
Done (#879) — the Kubernetes data-plane is persisted with the 4 cloud providers; clusters/pods/CRDs survive --persist restart and old kubeconfigs stay valid.

PR #879, opened 2026-08-29
feat(persist): persist the Kubernetes data-plane — clusters/pods/CRDs survive restart with kubeconfigs valid (#868)

NitinKumar004:
## What & why

`--persist` restored the four cloud providers but **not** the shared Kubernetes data-plane. After a restart the EKS/AKS/GKE `DescribeCluster` already returned the correct `…/k8s/<uid>` endpoint (providers persist their cluster-name→UID mapping), but `APIServer.clusters` was empty, so that endpoint answered `404 unknown cluster <uid>` and every pod/deployment was gone. #868 implements the missing half: capture `APIServer` state keyed by UID and reinstate each `ClusterState` under those UIDs, so a kubeconfig captured before the restart stays valid **verbatim**.

## What survives a restart

Everything a client created against a cluster's kubeconfig endpoint, **keyed by the same cluster UID the kubeconfig embeds**:

- The nine typed core/apps/policy stores (namespaces, pods, services + endpoints, configmaps, secrets, serviceaccounts, deployments, PDBs).
- Every registry-backed kind (ReplicaSets, StatefulSets, Jobs, Ingresses, HPAs, PV/PVC, the synthetic Node + node-lease, …) **and CustomResourceDefinitions with their custom resources**.
- The cluster-wide `resourceVersion` high-water mark (so client-go watches still establish) and the Service-ClusterIP / Pod-IP allocators (so a post-restore create advances rather than colliding).

## How the UID round-trip works

The two maps — the provider's `k8sUIDs` and the data-plane's `clusters` — both restore from the **same snapshot file**, so they're automatically consistent; the k8s side never re-registers through providers (that would couple ordering and duplicate identity). A new top-level `Snapshot.Kubernetes json.RawMessage` field carries the data plane (schema bumped 3→4). `*APIServer` implements `snapshot.Snapshottable`; `Restore` rebuilds each `ClusterState` via the same `newClusterState(...)` path `RegisterCluster` uses (config re-injected), loads the persisted stores wholesale over the fresh bootstrap seed (so no wrong-UID seeded copy survives), reconstructs CRD stores from the restored CRD objects **before** their CRs load, rebuilds all broadcasters fresh (live channels are never serialized), and rebuilds the event-dedup index from the restored Event store. It writes `s.clusters[uid]` directly rather than widening `RegisterCluster`'s random-UID-only API.

**Export ordering (correctness):** serverkit captures `snapTargets` **and** `a.k8s` in one `rebuildMu` section, then exports **providers before k8s** (shared `exportSnapshot` helper, used by both the flusher save and the admin `/_cloudemu/snapshot`), bounding the CreateCluster race to the harmless direction (an inert orphan `ClusterState`, never a dangling provider UID → 404).

**Dirty seam:** the k8s handler is now wrapped in `wrapDirty` at backend-swap time (before the admin Control), so a pure-`kubectl` mutation that never touches a provider port still marks state dirty; health probes fronted by Control don't. The progression ticker (which bypasses the HTTP seam) now marks dirty only when a Pod actually advances a stage (`Tick`/`TickAll` return a "changed" signal).

**Completeness guard:** a k8s-side field guard reflects over **both** `ClusterState` and `APIServer`, classifying every field as persisted or runtime/config — a new unclassified field fails CI (the provider guard is structurally blind to these plain-map types). A byte-identical Snapshot→Restore→Snapshot round-trip exercises one object of every registered kind + a CRD/CR + non-zero allocators/rv.

## Tests

- **kill-9 e2e** (`cmd/cloudemu`, real subprocess + real EKS SDK, under both `scheduled` and `on-request`): create cluster → apply Deployment+Service over the kubeconfig endpoint → SIGKILL → restart → same `/k8s/<uid>`, old endpoint still serves, Deployment/Pods/Service/Endpoints survive.
- rv-monotonicity + allocator advance-from-restored, CRD round-trip, incomplete-snapshot rejection, export-parity (flusher vs admin byte-identical `kubernetes` payload), restore-through-App, and the ticker "changed" dirty signal.

## Deferred (non-goals)

- Persisting live watch streams / resume-from-old-RV across a restart — a restart severs TLS so client-go relists (the correct behavior); no cross-restart event ring buffer (same deferral as #877).
- Per-cluster selective snapshot (all clusters snapshot together, matching provider behavior).


NitinKumar004 (commented):
Deep review of `3359fd9` (HEAD matches `gh pr view`).

**Gates.** `go build`, `go vet`, `go test ./...` (144/144 packages), `go test -race` on services/kubernetes + server/serverkit, and a `-count=2` flake check all pass clean. `golangci-lint` scoped to persist/, services/kubernetes/, server/serverkit/, cmd/cloudemu/ is byte-identical to the `development` baseline — zero new issues. `go mod tidy` / `gofmt` / generated-docs are clean. Coverage on the touched packages (persist 69.9%, serverkit 73.3%, kubernetes 80.2%) is unchanged-to-slightly-better than baseline, not a regression.

**Verified by reading + reproducing, not by trusting the description:**

- **Export ordering.** `exportSnapshot` (server/serverkit/persist.go) captures providers via `persist.ExportAll` before `k8s.Snapshot()`, and both call sites (flusher save, admin snapshot) grab `a.snapTargets` + `a.k8s` inside one `rebuildMu` critical section first. Confirmed the direction is right: a `CreateCluster` racing the export can only produce a harmless orphan `ClusterState`, never a dangling provider UID that 404s.
- **Restore completeness.** Every `ClusterState`/`APIServer` field is classified persisted-or-runtime and enforced by `TestSnapshotFieldGuard`. I added a throwaway unclassified field to `ClusterState`, ran the guard, watched it fail naming the field, then reverted — the guard is real, not decorative.
- **UID round-trip.** `swapFresh` builds exactly one `kubernetes.APIServer` and wires the same pointer into every provider's `SetK8sAPI` and into `a.k8s`; restore runs against that same instance after it's fully wired. Bootstrap-seeded Node/Lease/namespaces get wholesale-overwritten because `Snapshot()` always emits an entry for every registered store (even empty ones), so restore always finds and replaces them — no fresh-UID leftovers survive.
- **CRD ordering.** `restoreRegistryStores` reconstructs CRD-defined stores via `crdResourceDefs`+`addStore` before the generic item-fill loop — same order `reconcileCRD` uses live.
- **RV/allocators.** `TestSnapshotRestorePreservesAllocatorsAndRV` proves the high-water mark and both IP allocators come back exactly and the next write/allocation advances from them, not from zero.
- **Dirty seam.** `wrapDirty` is applied inside `swapFresh`, before `admin.NewControl` fronts the backend — confirmed via `handlerFor`. `TestDirtySeamIncludesKubernetes` is a genuine inversion of the old exclusion test (asserts dirty=true on a kubectl mutation, false on a Control-fronted health probe), not just a rename.
- **Kill-9 test.** `killHard` sends a real SIGKILL to a subprocess built from source; ran `TestK8sDataPlanePersistsAcrossCrash` standalone under both persist strategies — passes reliably, and it could not compile against pre-#868 `serverkit` (the `restoreState`/`k8s` signature didn't exist), so it's provably tied to this feature.
- **Schema v3→4.** `persist.ReadFile` rejects a version mismatch with a clear error; the server-startup path treats that as "warn + start empty," not a silent mis-restore or a crash. The `cloudemu snapshot save/load` CLI round-trips the new `kubernetes` field transparently via `json.RawMessage` passthrough through the already-covered admin snapshot/restore path.
- **Determinism.** No new `time.Now()`/`rand` in the diff.

One non-blocking note: `clusterSnapshot.validate()` only asserts the four system namespaces as its "complete snapshot" marker and doesn't independently check `Registry` is populated. In the real export path this is unreachable (`Snapshot()` always populates every registered store's entry), so it only matters for a hand-corrupted file, which the schema-version gate already catches earlier. Not worth changing.

No high or medium findings. Approving.

Issue #445, opened 2026-08-23
Enable real engines from `cloudemu serve`


NitinKumar004:
The opt-in real data-plane engines are wired only through the in-process library. In `cmd/cloudemu/serve.go` the providers are built with `cloudemu.NewAWS(opts...)` (and the Azure/GCP/OCI equivalents), but the engine options — `config.WithDatabaseEngine`, `WithCacheEngine`, `WithFunctionEngine`, `WithComputeEngine`, `WithContainerEngine`, `WithStorageEngine` — are never passed there, so a running `cloudemu serve` is always pure in-memory.

## What to build
- Flags/env on `cloudemu serve` to enable each engine (e.g. `--engine-db`, `--engine-cache`, … or `CLOUDEMU_ENGINE_*`).
- Translate them into the matching `config.With*Engine(...)` options and pass them into the `New{AWS,Azure,GCP}(opts...)` calls in `serve.go`.
- Document which flags need Docker: the no-Docker engines live in `contrib/realengine`, the Docker-backed ones in `contrib/dockerengine`.

## Where
`cmd/cloudemu/serve.go`, `config/`, `contrib/realengine`, `contrib/dockerengine`. Related: #428.

PR #884, opened 2026-08-30
refactor(serve): extract shared serve flags into server/serveflags so the lean and :engines binaries can't drift (#445)

NitinKumar004:
## What & why

The two standalone-server entrypoints — the lean `cmd/cloudemu serve` (CGO-off, zero-dep static binary) and the batteries-included `contrib/server` (the `:engines` image) — hand-duplicated ~30 command-line flags. They were kept in lockstep only by a comment and a persist-only mirror test, so any new or renamed flag could silently drift between the two mains.

This PR (part of #445) extracts the shared flags into a new dependency-light core package, `server/serveflags`, that both mains build from. The refactor is behavior-preserving: for identical args the resulting `serverkit.Config` is unchanged.

## New package: `server/serveflags`

Imports only `flag`, `fmt`, `os`, `strings`, `time`, `config`, and `serverkit` — never an engine constructor or a contrib package, so the lean binary stays free of the heavy engine deps (embedded-postgres, miniredis, go-redis, Docker SDK).

- `CommonConfig` — the full shared flag set (providers, host/advertise-host, ports, identity, latency, TLS, persistence, init-dir, enforce-auth, k8s-progression, …).
- `RegisterCommon(fs, *CommonConfig, getenv)` — registers all common flags; `getenv` is injectable so the env-backed defaults stay testable (cmd passes `os.Getenv`).
- `(*CommonConfig).Validate()` — owns the cross-field rules both mains shared: `--tls-cert`/`--tls-key` must be paired, `--persist` requires `--state-file`.
- `(*CommonConfig).ToServerkitConfig(providers)` — maps the flags onto `serverkit.Config`, seeding `BaseOptions` with the three identity options.
- `ParseProviders(string)` — the folded provider parser.
- `StringList` — the repeatable `--tls-host` flag value.
- `EngineFlags` / `IsEngineFlag` — the single shared list of real-engine selectors (db/cache/functions/compute/containers/storage/storage-dir/all-real) that both sides range over.

## Both mains slimmed onto it

- `cmd/cloudemu/serve.go` registers its common flags via `RegisterCommon`, validates via `Validate`, and builds the config via `ToServerkitConfig`. The PR-1 engine-discoverability detector now reads `serveflags.EngineFlags` / `IsEngineFlag` (local duplicate removed). Behavior identical.
- `contrib/server/main.go` registers only the engine selectors plus `RegisterCommon`.
- `contrib/server/app.go` control-flow change: identity options now build inside `ToServerkitConfig`, so the engine options are appended to `cfg.BaseOptions` **after** calling it (rather than merging identity+engine before). The engine wiring — `buildEngineOptions`, the MODE table, and the Docker degrade path — is untouched.

## Guards

- **Drift guard** — `serveflags` pins the registered common-flag name set + shared-constant defaults; each main has a mirror test asserting it builds its FlagSet from `RegisterCommon` (names + defaults), and contrib asserts its engine-flag names equal `EngineFlags`. A common flag added/renamed in only one place fails a test.
- **Dep guard** — a test shells `go list -deps .` for the lean binary and fails if any of `embedded-postgres`, `miniredis`, `go-redis`, or `docker/docker` appears, proving the lean binary and `serveflags` stay engine-dep-free.

Zero change to the option/engine/teardown machinery (`config/engine.go`, `EngineClosers`, `Provider.Close`, `serverkit.closeProviders`, the `exportMu` guard).

Note: the only intended user-visible change is `contrib/server`'s `--enforce-auth --help` text, now unified to the fuller SigV4/Bearer usage string (both mains share one usage string).

### DIFF
--- a/cmd/cloudemu/lean_deps_test.go
+++ b/cmd/cloudemu/lean_deps_test.go
@@ -1,0 +1,39 @@
+package main
+
+import (
+	"os/exec"
+	"strings"
+	"testing"
+)
+
+// engineModulePaths are the heavy real-engine dependencies that must never reach
+// the lean cloudemu binary. They live only in the :engines image (contrib/server);
+// the shared server/serveflags package the lean binary now builds from is
+// deliberately dep-light so pulling it in cannot drag these along. This is the
+// structural guard against the "compile engines into the lean binary" trap.
+//
+//nolint:gochecknoglobals // test fixture: the forbidden engine module paths
+var engineModulePaths = []string{
+	"github.com/fergusstrange/embedded-postgres",
+	"github.com/alicebob/miniredis",
+	"github.com/redis/go-redis",
+	"github.com/docker/docker",
+}
+
+// TestLeanBinaryHasNoEngineDeps shells `go list -deps .` for the lean binary and
+// fails if any forbidden engine module appears in its transitive dependency
+// graph. It proves both the lean binary and server/serveflags stay engine-free.
+func TestLeanBinaryHasNoEngineDeps(t *testing.T) {
+	out, err := exec.Command("go", "list", "-deps", ".").CombinedOutput()
+	if err != nil {
+		t.Fatalf("go list -deps .: %v\n%s", err, out)
+	}
+
+	deps := string(out)
+
+	for _, mod := range engineModulePaths {
+		if strings.Contains(deps, mod) {
+			t.Errorf("lean binary transitively depends on engine module %q — it must stay engine-free", mod)
+		}
+	}
+}
--- a/cmd/cloudemu/persist_flags_test.go
+++ b/cmd/cloudemu/persist_flags_test.go
@@ -3,6 +3,7 @@ package main
 import (
 	"testing"
 
+	"github.com/stackshy/cloudemu/v2/server/serveflags"
 	"github.com/stackshy/cloudemu/v2/server/serverkit"
 )
 
@@ -13,7 +14,7 @@ func TestServePersistFlagDefaults(t *testing.T) {
 	t.Setenv("CLOUDEMU_PERSIST_STRATEGY", "")
 	t.Setenv("CLOUDEMU_PERSIST_INTERVAL", "")
 
-	var c serveConfig
+	var c serveflags.CommonConfig
 	fs := newServeFlagSet(&c)
 
 	strat := fs.Lookup("persist-strategy")
@@ -39,11 +40,11 @@ func TestServePersistFlagDefaults(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if c.persistStrategy != "on-request" {
-		t.Fatalf("parsed strategy = %q, want on-request", c.persistStrategy)
+	if c.PersistStrategy != "on-request" {
+		t.Fatalf("parsed strategy = %q, want on-request", c.PersistStrategy)
 	}
 
-	if c.persistInterval.String() != "3s" {
-		t.Fatalf("parsed interval = %q, want 3s", c.persistInterval)
+	if c.PersistInterval.String() != "3s" {
+		t.Fatalf("parsed interval = %q, want 3s", c.PersistInterval)
 	}
 }
--- a/cmd/cloudemu/serve.go
+++ b/cmd/cloudemu/serve.go
@@ -7,16 +7,11 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
-	"strings"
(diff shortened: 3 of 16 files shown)


NitinKumar004 (commented):
Reviewed the pushed commit (e27b3f09) directly, not the description.

**Gates (clean-built the pushed commit):**
- `go test -count=1 ./server/serveflags/... ./cmd/... .` — pass
- `go test ./server/...` — all packages pass, including `server/serveflags` and `server/serverkit`
- `contrib/server`: `go build ./...` and `go test ./...` — pass, including the Postgres/Redis e2e (`TestBatteriesServerE2E`) and the identity-preservation e2e (`TestBatteriesServerIdentityPreserved`)
- `golangci-lint` on `./server/serveflags/...` and `./cmd/cloudemu/...`, and separately on `./contrib/server/...`: 0 new issues. The 14 issues golangci-lint reports under `cmd/cloudemu` are pre-existing (`main.go`/`lifecycle.go`/`net.go`/`snapshot.go`, none of which this PR touches) and byte-identical to `development`.

**Behavior-preservation (the core bar):** Diffed the old `cmd/cloudemu/serve.go:toServerkitConfig` and old `contrib/server/app.go:serverkitConfig` against the new `serveflags.CommonConfig.ToServerkitConfig` field by field — every `serverkit.Config` field is carried over identically (ports map, K8s progression, persist strategy/interval, TLS, admin, enforce-auth, endpoints file, shutdown timeout). Identity `BaseOptions` order is still account→region→project in both entrypoints; `contrib/server/app.go` appends engine options to `BaseOptions` *after* `ToServerkitConfig` builds it, which yields the same final order (identity, then engine) as the old pre-merged slice — confirmed by `TestBatteriesServerIdentityPreserved`. `Out` is correctly overridden to `cfg.out` in `contrib/server/app.go:27` after the shared config builds it (which otherwise hardcodes `os.Stdout`), so contrib's injectable banner sink for tests still works.

**Dep-light constraint:** `go list -deps ./server/serveflags` and `go list -deps ./cmd/cloudemu` show zero occurrences of `embedded-postgres`, `miniredis`, `go-redis`, or `docker/docker` (the only "postgres" hits are the unrelated `providers/azure/postgresflex` mock service). `cmd/cloudemu/lean_deps_test.go` genuinely shells `go list -deps .` and greps precise module paths — not a stub.

**Drift guards:** `server/serveflags/serveflags_test.go:TestRegisterCommonNameSet` pins the flag-name set as an independent literal list (not derived from `RegisterCommon` itself), so a flag added/renamed/dropped inside `RegisterCommon` without updating the test fails it — this is a real guard, not tautological. `TestBatteriesEngineFlagNamesMatchSharedList` cross-checks contrib's hardcoded `registerEngineFlags` literals against `serveflags.EngineFlags` independently, so an engine-flag rename in only one place fails it. The two per-main mirror tests (`TestServeRegistersEveryCommonFlag`, `TestBatteriesRegistersEveryCommonFlag`) only assert that each main's FlagSet is a superset of `RegisterCommon`'s names/defaults — they wouldn't catch a stray flag hand-added directly in one main's registration code outside `RegisterCommon`. Minor, non-blocking: the refactor structurally eliminates most of that risk (there's no duplicated implementation left to drift from), but the guard isn't symmetric.

**Validate()/engine-detection/getenv:** `--tls-cert`/`--tls-key` XOR and `--persist`⇒`--state-file` checks are identical to the old inline checks in both mains, same order of precedence per binary as before. PR-1's stub-flags + `fs.Visit` engine detection is preserved verbatim (just reads `serveflags.EngineFlags`/`IsEngineFlag`); `TestServeAccountIDValueIsNotAnEngineFlag` (the `--account-id --db` false-positive guard) still passes. `contrib/server/engines.go` (buildEngineOptions/MODE table/degrade/`--all-real`) is untouched by this diff. `RegisterCommon`'s injected `getenv` is threaded correctly — cmd passes `os.Getenv`, contrib's tests pass stubs, and `TestRegisterCommonEnvFallback` exercises persist/k8s-progression env fallbacks.

**Layering:** confirmed one-directional — `server/serverkit` has zero references to `server/serveflags`; only `serveflags.go`/`serveflags_test.go` import `serverkit`.

One user-visible change, correctly called out in the PR description: `contrib/server`'s `--enforce-auth` usage text is now the fuller SigV4/Bearer string (previously only `cmd/cloudemu` had it) — no test pins the old shorter string, so nothing breaks.

No blocking or major issues found. Approving.

PR #885, opened 2026-08-30
fix(kubernetes): stamp job-name/controller-uid labels on Job pods (#874)

NitinKumar004:
## Summary

Job Pods created by the in-memory Kubernetes reconciler were missing the standard Job-controller identity labels that real clusters stamp on every Job Pod. This is the residual item from #874 (the `creationTimestamp`/`resourceVersion` nits were already handled in `buildControllerPod`).

Each Job Pod now carries the four real upstream labels:

- `batch.kubernetes.io/job-name` = Job `metadata.name` (GA since k8s 1.27)
- `job-name` = Job name (legacy alias)
- `batch.kubernetes.io/controller-uid` = Job `metadata.uid`
- `controller-uid` = Job UID (legacy alias)

## Placement

The labels are stamped in `reconcileJob`'s top-up loop, immediately after `buildControllerPod`, **not** inside `buildControllerPod` itself. `buildControllerPod` is shared by ReplicaSet/StatefulSet/DaemonSet reconciliation — stamping Job labels there would wrongly label their Pods. Values come straight from the Job object at the call site: `obj.GetName()` and `owner.UID` (= `obj.GetUID()`, already populated on create and relied on by `podsOwnedByLocked`). Label keys are small named constants.

## Blast radius: none (additive)

- Job Pod ownership is by `ownerReferences` UID (`podsOwnedByLocked`), not label selector — added labels cannot affect which Pods a Job reconciles.
- Endpoints only match `PodRunning`; Job Pods are driven to `PodSucceeded`, so they never enter Endpoints regardless of labels.
- `labelsMatch` is selector⊆labels, so extra labels can only add matches, never remove one.

## Tests

Adds `TestJobPodLabels`: creates a `batch/v1` Job, reconciles, fetches the owned Pod, and asserts all four keys are present, the two `job-name` aliases agree and equal the Job name, and the two `controller-uid` aliases agree and equal the Job UID. Purely additive — no existing test asserted these labels.

Gates (GOMAXPROCS=3): `go build ./...`, `go vet`, `go test`, `go test -race`, and `golangci-lint` for `services/kubernetes/...` all clean (only the pre-existing `admission.go` gosec baseline remains). `go generate ./...` produces no diff.

### DIFF
--- a/services/kubernetes/phase1_fidelity_test.go
+++ b/services/kubernetes/phase1_fidelity_test.go
@@ -267,6 +267,73 @@ func TestPodExec_TypedNotImplemented(t *testing.T) {
 	}
 }
 
+func TestJobPodLabels(t *testing.T) {
+	base, done := newFixture(t)
+	defer done()
+
+	body := mustJSON(t, map[string]any{
+		"apiVersion": "batch/v1", "kind": "Job",
+		"metadata": map[string]any{"name": "labeljob"},
+		"spec": map[string]any{
+			"completions": 1,
+			"template": map[string]any{"spec": map[string]any{
+				"containers": []any{map[string]any{"name": "c", "image": "img"}},
+			}},
+		},
+	})
+
+	resp := do(t, http.MethodPost, base+"/apis/batch/v1/namespaces/default/jobs", body)
+	created := decodeMap(t, resp.Body)
+	resp.Body.Close()
+
+	createdMeta, _ := created["metadata"].(map[string]any)
+	jobUID, _ := createdMeta["uid"].(string)
+	if jobUID == "" {
+		t.Fatalf("created Job has empty metadata.uid")
+	}
+
+	pods := do(t, http.MethodGet, base+"/api/v1/namespaces/default/pods", nil)
+	defer pods.Body.Close()
+
+	list := decodeMap(t, pods.Body)
+	items, _ := list["items"].([]any)
+	if len(items) != 1 {
+		t.Fatalf("job pod count: got %d, want 1", len(items))
+	}
+
+	pod, _ := items[0].(map[string]any)
+	podMeta, _ := pod["metadata"].(map[string]any)
+	rawLabels, _ := podMeta["labels"].(map[string]any)
+
+	labels := map[string]string{}
+	for k, v := range rawLabels {
+		s, _ := v.(string)
+		labels[k] = s
+	}
+
+	want := map[string]string{
+		"batch.kubernetes.io/job-name":       "labeljob",
+		"job-name":                           "labeljob",
+		"batch.kubernetes.io/controller-uid": jobUID,
+		"controller-uid":                     jobUID,
+	}
+	for k, v := range want {
+		if labels[k] != v {
+			t.Fatalf("pod label %q: got %q, want %q", k, labels[k], v)
+		}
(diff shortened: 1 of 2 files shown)


NitinKumar004 (commented):
Adversarial review, verified against pushed commit 2cc5534 in an isolated worktree.

**Gates (GOMAXPROCS=3, services/kubernetes/...):** `go build ./...`, `go vet ./...`, `go test -count=1`, `go test -race -count=1`, `golangci-lint run` all clean. The one lint hit (`admission.go:351` gosec G704 SSRF) is pre-existing baseline, untouched by this diff (confirmed `git diff` on that file is empty against development). `go generate ./...` produces no diff. Diff scope is exactly the two claimed files (reconcile.go +20, phase1_fidelity_test.go +67), single commit.

**1. Placement — verified correct.** The four labels are stamped in `reconcileJob`'s top-up loop (reconcile.go:641-644), strictly after the `buildControllerPod` call, not inside it. Traced all three callers of `buildControllerPod`: `syncScaledPods` (RS/Deployment, line 199), `syncStablePods` (STS/DS, line 235), and `reconcileJob` (line 636) — only the `reconcileJob` call site has the post-stamp. RS/STS/DS pods cannot get `job-name`/`controller-uid` labels.

**2. Label keys/values — correct.** `jobName := obj.GetName()` and `jobUID := string(owner.UID)` where `owner := ownerRefOf(obj)` sets `UID: obj.GetUID()`. Confirmed the Job's UID is populated before `reconcileJob` runs in both invocation paths: the generic create/update handlers in registry.go call `obj.SetUID(...)` before `st.def.reconcile(s, obj)` (lines 354→391, 610→641), and the CronJob path explicitly does `job.SetUID(...)` (cronjob.go:225) before `reconcileJob(s, job)` (cronjob.go:231). UID is never empty at the stamping site.

**3. Blast radius — none found.** `podsOwnedByLocked` filters strictly by `OwnerReferences` UID (reconcile.go:159-165), never touches labels. Endpoint matching only selects `PodRunning` (reconcile.go:289); Job pods are driven straight to `PodSucceeded` via `markPodSucceededLocked`, so they were already excluded from Endpoints regardless of labels. `labelsMatch` (reconcile.go:414-422) iterates the selector and checks `labels[k] == v` — extra keys in the pod's label map are inert for matching (selector⊆labels semantics hold).

**4. Non-nil Labels — confirmed.** `buildControllerPod` always does `labels := make(map[string]string, len(tmpl.Labels))` (line 107) before assigning to `pod.Labels`, so the map is never nil at the stamping site — no panic risk even for a Job with an empty pod template.

**5. Test teeth — good.** `TestJobPodLabels` POSTs a real `batch/v1` Job, reads back the created object's real `metadata.uid` (fails fast if empty), lists the resulting Pod via the actual HTTP list endpoint, and asserts all four label keys equal the real job name / real UID, plus that the GA/legacy aliases agree with each other. This would fail if any label were dropped, misnamed, or given a hardcoded/wrong value — it's reading live values end-to-end through the wire API, not internal state.

**6. Existing suite — green.** `TestJobReconcile_ShrinkCorrectsSucceeded` and the rest of `services/kubernetes/...` pass unchanged under `-race`.

No blocking or major findings. Approving.

Issue #498, opened 2026-08-24
coveragegen: suppress AWS-only optional capabilities from non-AWS provider pages


NitinKumar004:
Optional type-asserted interfaces that live in a shared `services/*/driver` package (e.g. compute's ImageRegistrar/KeyPairImporter/LaunchTemplateVersioner/SnapshotCopier/VolumeModifier, and ConsoleReader) are AWS-only, but coveragegen emits their capability sections onto the Azure and GCP provider pages too (docs/coverage/azure/*, docs/coverage/gcp/*), with 'AWS-only' wording that reads as though GCE/Azure VMs support RegisterImage/ModifyVolume/etc.

Have the generator either scope AWS-only optional capabilities to the AWS provider page, or qualify per-provider so unimplemented ops aren't listed under clouds that don't implement them.

NitinKumar004:
Done (#886) — same per-provider gate suppresses AWS-only optional capabilities from non-AWS provider pages (a genuine method-set gate, not an AWS-page filter — ConsoleReader correctly stays on all three).

PR #886, opened 2026-08-30
fix(coverage): gate optional capabilities to providers that implement them (#394, #498)

### LINKED ISSUE #394: coveragegen: a shared driver's provider-specific optional capabilities are attributed to every provider

NitinKumar004:
## Summary

`internal/coveragegen` collected each service's optional capabilities service-wide (no provider dimension) and then rendered **all** of them on **every** provider page. So the "authoritative" `docs/coverage/` made false claims: `oci/vcn.md` listed AWS-only `TransitGateways`/`IPAM`, `gcp/gce.md` listed AWS-only `ImageRegistrar`/`VolumeModifier`/`SnapshotCopier` and Azure-only `AzureVMController`, and `aws/vpc.md` listed `AzureNetworkInterfaces`. Fixes #394 and #498.

## Fix

Gate each optional capability to the providers whose concrete mock actually implements it — using **full method-set coverage** (a type satisfies an optional interface iff its mock declares every method of that interface), the same `covers()` primitive the generator already uses to resolve primary interfaces.

- Add an unexported `providerMethods map[string]map[string]bool` (provider -> mock method-set) to `Service`. It is unexported, so it is **not** serialized — `coverage.json` (which the compat matrix reads) is byte-identical.
- Populate it per (service, provider). AWS/Azure/GCP fields are typed `*pkg.Mock`, so the mock dir comes straight from the field type; OCI's fields are **bare driver interfaces** whose concrete mock is visible only in the constructor (`p.VCN = vcnprovider.New(o)`), so a new `constructorMockPkgs` parses the `New()` body for `p.Field = pkg.NewXxx(...)` assignments — modeled on `native.go`'s `handlerCallPkg`. Both styles fold into one `providerMockDirsByField` helper.
- `renderProviderPage` now gates each capability: `if !covers(svc.providerMethods[prov], capability.Operations) { continue }`. A single-method probe would be wrong — `DeleteNetworkInterface` appears on both aws/vpc and oci/vcn from other interfaces; only full coverage is correct.

## Why full coverage (not an AWS-page filter)

`ConsoleReader` (`GetConsoleOutput`) is implemented by aws, azure, **and** gcp, so it must remain on all three compute pages. #498's crude "AWS-page-only" filter would have dropped it. The per-provider method-set gate keeps exactly what each mock implements.

## Verified capability diffs

- `oci/vcn.md`: `TransitGateways`/`IPAM`/`AzureNetworkInterfaces` (and 24 other AWS-only network caps) gone; genuine cross-cloud caps the OCI mock covers (`NetworkInterfaces`, `VPCAttributes`, `NetworkInterfaceCreator`) remain.
- `gcp/gce.md`: `ImageRegistrar`/`VolumeModifier`/`SnapshotCopier`/`AzureVMController` gone.
- `azure/virtualmachines.md`: `ImageRegistrar`/`VolumeModifier`/`SnapshotCopier` gone; `AzureVMController` + `ConsoleReader` remain.
- `aws/vpc.md`: `AzureNetworkInterfaces` gone; `IPAM`/`TransitGateways` remain.
- `coverage.json`: unchanged (public schema untouched).

## Tests

`TestOptionalCapabilitiesGatedToImplementers` renders every service page and asserts the rendered capability set equals exactly what the provider's mock covers (catches any un-gating regression). `TestOptionalCapabilityVanishRemainAnchors` pins the specific vanish/remain pairs above, including `ConsoleReader`-on-all-three.

Regenerated `docs/coverage/` committed.


NitinKumar004 (commented):
Verified against the pushed commit (ff08aa4e) in an isolated worktree.

**Gates**
- `go build ./...`, `go vet ./...`: clean.
- `go test -count=1 ./internal/coveragegen/...`: all 12 tests pass, including the two new gate tests (`TestOptionalCapabilitiesGatedToImplementers`, `TestOptionalCapabilityVanishRemainAnchors`).
- `golangci-lint run ./internal/coveragegen/...`: 0 issues.
- `git diff development...HEAD -- docs/coverage/coverage.json`: empty. `providerMethods` is correctly unexported and never touches the serialized schema the compat matrix reads.
- `go generate ./...` on the branch reproduces the committed `docs/coverage/` byte-for-byte (`git status --porcelain docs/coverage` clean afterward) — the regeneration isn't stale.

**No over-suppression (re-derived independently, not just trusting the diff)**
Spot-checked removed vs. kept capabilities against the actual mock source across services beyond the PR's own anchors:
- AWS EC2 / Azure VM / GCP GCE: confirmed by grepping mock methods directly — Azure/GCP genuinely lack `RegisterImage`/`CopySnapshot`/`ModifyVolume` (ImageRegistrar/SnapshotCopier/VolumeModifier correctly vanish), all three genuinely have `GetConsoleOutput` (ConsoleReader correctly stays on all three pages).
- OCI identity.md / monitoring.md: `AccessKeyByID`, `CheckPermissionWithContext`, `PrincipalHasPolicies`, `ListActivityLogEvents`, `RecordActivityLogEvent` are absent from `providers/oci/identity` and `providers/oci/monitoring` — correctly removed, and the page cleanly drops the whole "## Optional capabilities" section (no dangling heading).
- OCI vcn.md: the interesting case. `NetworkInterfaces` and `AzureNetworkInterfaces` both declare a method named `DeleteNetworkInterface` — exactly the collision the render.go comment calls out. OCI's mock (`providers/oci/vcn/vnic.go`) implements all 3 of `NetworkInterfaces`'s methods but only 1 of `AzureNetworkInterfaces`'s 4 methods (`CreateOrUpdateNetworkInterface`/`GetNetworkInterface`/`ListNetworkInterfaces` are absent). The gate correctly keeps `NetworkInterfaces` and drops `AzureNetworkInterfaces` — confirms it's doing full-method-set matching, not name-substring/single-method matching.
- Cross-checked the "kept" side too: Azure's vnet mock has all 4 `AzureNetworkInterfaces` methods (`providers/azure/vnet/network_interface.go`), GCP's alloydb mock has all 4 `Databases` methods, AWS redshift's mock has none of `AddTagsToResource`/`CreateDBSubnetGroup`/`CreateDBParameterGroup` (all 29 relationaldb capabilities correctly vanish from its page — Redshift's mock is a narrow, separate implementation from the RDS/Aurora mock).
- Partial suppression works too: aws/elasticache.md drops `AccessKeys` (Azure-only) but correctly keeps `ReplicationGroups`, which AWS's mock implements — confirms the section isn't all-or-nothing.

**OCI constructor-parse**
`providers/oci/oci.go`'s `New()` only actually constructs 3 of its ~17 declared fields (`Identity`, `VCN`, `Monitoring` — everything else stays a nil driver interface). Field types are bare driver interfaces, so the struct-literal pass alone can't find the mock; `constructorMockPkgs` correctly parses `p.VCN = vcnprovider.New(o)` / `p.Identity = identity.New(o)` and resolves to `providers/oci/vcn` / `providers/oci/identity`, which is exactly what the OCI pages now render (real caps, not empty, not AWS-only leakage).

**Tests have teeth**
Reasoned through both new tests against the gate being reverted (rendering every capability unconditionally, the old behavior): `TestOptionalCapabilitiesGatedToImplementers` directly compares `covers(providerMethods, ops)` against the parsed rendered output per provider/capability pair — with the gate removed, every currently-vanished pair (OCI TransitGateways/IPAM, GCP/Azure ImageRegistrar, etc.) would flip `rendered=true` against `covered=false` and fail. `TestOptionalCapabilityVanishRemainAnchors` pins 16 concrete cases the same way. Neither test could pass against un-gated rendering.

**Minor observation (non-blocking)**
`mockMethods` (providers.go:389) collects methods from every receiver type declared in the mock package directory, not just the primary `Mock` struct (e.g. `providers/gcp/alloydb` also has a `recordingEngine` type in the same dir). This is a pre-existing pattern already relied on by `implementedService`, not something this PR introduces, and none of the spot-checked pages show a false positive from it — but it's worth knowing that a future helper type whose method set happens to fully match some capability's method names would be silently credited.

No blocking or major issues found.

PR #887, opened 2026-08-30
feat(cli): add 'cloudemu doctor' preflight check (ports, version, docker) (#454)

### LINKED ISSUE #454: `cloudemu doctor` — preflight environment check
cloudemu has `serve` / `start` / `stop` / `status` / `logs` / `snapshot` / `env` / `version`, but no single command to confirm the environment is ready before starting.

## What to build
- A `cloudemu doctor` subcommand that checks: the ports are free, the binary version, and — for the real engines — whether Docker is available.
- Print a clear pass/fail summary with fix hints.

## Where
`cmd/cloudemu/main.go` (register the subcommand), a new `cmd/cloudemu/doctor.go`.

NitinKumar004:
## What

Adds a `cloudemu doctor` subcommand — a fast, dependency-free preflight check you run before `cloudemu serve` (closes #454).

## Checks it runs

- **Build version** — prints the ldflag-stamped `version/commit/date/builtBy`.
- **Default ports free** — for each port `serve` binds by default (read from `server/serveflags` so it can't drift: AWS 4566, Azure 4568, GCP 4569, Kubernetes 4570, OCI 4571), tries to `Listen` then closes. Required ports (aws/azure/gcp/k8s) in use are blockers; OCI is opt-in so it only warns. Accepts `--host` (default `127.0.0.1`) mirroring `cloudemu env`. This is a point-in-time probe, so output says "free (at check time)" — not a guarantee (TOCTOU-honest).
- **Docker** — `exec.LookPath("docker")` (cross-platform stdlib). Framed as "only needed for the :engines image" — a warning when absent, never a hard failure.
- **Engines note** — states the real engines aren't compiled into the lean binary and points at the `:engines` image using the exact same string `serve` prints: `docker run -p 4566:4566 ghcr.io/stackshy/cloudemu:engines --all-real`.

Exit code is non-zero only when a required port is in use.

## Sample output

```
cloudemu doctor — preflight check

[ ok ] version dev (commit none, built unknown by unknown)

Ports on 127.0.0.1 (free = free at check time; a port can still be taken before serve binds it):
  [ ok ] AWS            127.0.0.1:4566  free (at check time)
  [ ok ] Azure          127.0.0.1:4568  free (at check time)
  [ ok ] GCP            127.0.0.1:4569  free (at check time)
  [ ok ] Kubernetes     127.0.0.1:4570  free (at check time)
  [ ok ] OCI (opt-in)   127.0.0.1:4571  free (at check time)

[ ok ] docker found at /usr/local/bin/docker (only needed for the :engines image)

Real engines (postgres/redis/subprocess/docker/localfs) are NOT compiled into this lean
binary. To run them, use the :engines image:
  docker run -p 4566:4566 ghcr.io/stackshy/cloudemu:engines --all-real

[ ok ] all preflight checks passed — ready to `cloudemu serve`.
```

## Notes

- New file `cmd/cloudemu/doctor.go` is **cross-platform** — no `//go:build` tag (verified with `GOOS=windows go build ./cmd/cloudemu`).
- The Docker check is behind an injected `lookPath` param and the port checker takes an explicit `host:port`, so both are unit-testable without a real docker/server (mirrors serve's injected-getenv pattern). Tests are hand-rolled asserts (no testify).

## Gates

`go build ./...`, `go vet ./cmd/cloudemu/...`, `go test ./cmd/cloudemu/...`, `golangci-lint` (zero new vs baseline), and `go generate ./...` (no diff) all pass; Windows cross-compile succeeds.

### DIFF
--- a/cmd/cloudemu/doctor.go
+++ b/cmd/cloudemu/doctor.go
@@ -25,10 +25,8 @@ const (
 	markFail = "[fail]"
 )
 
-// enginesImagePointer is the one-line pointer at the :engines image, kept
-// identical to the message serve prints (errEnginesNotInLeanBinary) so the CLI
-// speaks with one voice about where the real engines live.
-const enginesImagePointer = "docker run -p 4566:4566 ghcr.io/stackshy/cloudemu:engines --all-real"
+// The doctor echoes serve's canonical enginesImagePointer (defined in serve.go)
+// so the two commands can't drift on where the real engines live.
 
 // portCheck is one default endpoint the emulator would bind. required marks a
 // port the default provider set (aws,azure,gcp) plus the k8s data-plane bind on
--- a/cmd/cloudemu/doctor.go
+++ b/cmd/cloudemu/doctor.go
@@ -1,0 +1,188 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"flag"
+	"fmt"
+	"io"
+	"net"
+	"os"
+	"os/exec"
+
+	"github.com/stackshy/cloudemu/v2/server/serveflags"
+)
+
+// errPreflightBlocker is returned when a required port is in use, so runDoctor
+// exits non-zero after printing the report.
+var errPreflightBlocker = errors.New("preflight found a blocker (see above)")
+
+// Status markers for the doctor report. ASCII-only so they render the same on
+// every platform (this command is deliberately not Unix-tagged).
+const (
+	markPass = "[ ok ]"
+	markWarn = "[warn]"
+	markFail = "[fail]"
+)
+
+// enginesImagePointer is the one-line pointer at the :engines image, kept
+// identical to the message serve prints (errEnginesNotInLeanBinary) so the CLI
+// speaks with one voice about where the real engines live.
+const enginesImagePointer = "docker run -p 4566:4566 ghcr.io/stackshy/cloudemu:engines --all-real"
+
+// portCheck is one default endpoint the emulator would bind. required marks a
+// port the default provider set (aws,azure,gcp) plus the k8s data-plane bind on
+// startup — an in-use required port is a genuine blocker. OCI is opt-in, so a
+// busy OCI port is only a warning.
+type portCheck struct {
+	label    string
+	port     string
+	required bool
+}
+
+// defaultPortChecks returns the ports `cloudemu serve` binds by default, read
+// from serveflags so the doctor can never drift from the real defaults.
+func defaultPortChecks() []portCheck {
+	var c serveflags.CommonConfig
+
+	fs := flag.NewFlagSet("doctor-defaults", flag.ContinueOnError)
+	serveflags.RegisterCommon(fs, &c, func(string) string { return "" })
+	_ = fs.Parse(nil)
+
+	return []portCheck{
+		{"AWS", c.AWSPort, true},
+		{"Azure", c.AzurePort, true},
+		{"GCP", c.GCPPort, true},
+		{"Kubernetes", c.K8sPort, true},
+		{"OCI (opt-in)", c.OCIPort, false},
+	}
+}
+
+// checkPortFree reports whether a TCP listener can be bound at hostPort right
+// now. This is a point-in-time probe: a "free" port can still be taken before
+// `cloudemu serve` binds it (TOCTOU), which the report wording makes explicit.
+func checkPortFree(hostPort string) bool {
+	var lc net.ListenConfig
+
+	ln, err := lc.Listen(context.Background(), "tcp", hostPort)
+	if err != nil {
+		return false
+	}
+
+	_ = ln.Close()
+
+	return true
+}
+
+// dockerLine reports whether a `docker` binary is on PATH. Docker is needed only
+// to run the :engines image, so its absence is a warning, never a failure.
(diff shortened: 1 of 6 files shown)


NitinKumar004 (commented):
## Review: `cloudemu doctor` preflight command

Checked out PR head `5468c3b3` against merge-base `e314e955` in an isolated worktree (diff confirmed identical to the PR: `cmd/cloudemu/doctor.go` +188, `doctor_test.go` +157, `main.go` +7).

### Gate matrix

| Gate | Result |
|---|---|
| `go build ./...` | PASS |
| `go vet ./...` | PASS |
| `go test -count=1 -v ./cmd/cloudemu/...` | PASS — all 5 new doctor tests + full existing suite (32 tests) green |
| `GOOS=windows go build ./cmd/cloudemu` | PASS |
| `GOOS=darwin GOARCH=arm64 go build ./cmd/cloudemu` | PASS |
| `golangci-lint run ./cmd/cloudemu/...` (v2.11.4) | 14 issues — identical set to the merge-base baseline (same rule/line for `gochecknoglobals`×3, `gosec`×6, `mnd`×3, `gocyclo`×2). None touch `doctor.go`/`doctor_test.go`. Zero new issues. Local gate per repo convention, not CI-blocking. |
| `go generate ./...` | clean, no diff |
| `go mod tidy` | clean, no diff |

### Verified

- **Cross-platform**: no `//go:build` tag; imports only `context/errors/flag/fmt/io/net/os/os/exec` + `server/serveflags`. Windows and Darwin cross-compiles both succeed.
- **Port-check correctness**: `checkPortFree` (`doctor.go:64`) binds via `net.ListenConfig.Listen` then closes; no listener leak. `--host` defaults to `127.0.0.1` and is threaded consistently into the probe, so there's no 0.0.0.0/127.0.0.1 mismatch as long as the user passes the same `--host` to both `doctor` and `serve`. TOCTOU is explicitly called out in the output wording ("free (at check time)... a port can still be taken before serve binds it").
- **Default ports source**: `defaultPortChecks()` (`doctor.go:45-59`) reads the values back from `serveflags.RegisterCommon` against an empty `FlagSet`, so it can't drift from `serve`'s real defaults. Confirmed against `server/serveflags/serveflags.go:110-114`: aws 4566, azure 4568, gcp 4569, k8s 4570, oci 4571 — all match.
- **Exit semantics**: `runDoctor` returns a non-nil error only when `blockers>0`, and `blockers` only increments for `p.required` ports (`doctor.go:117-118`). Docker absence and an in-use OCI port never touch `blockers`. `TestWriteDoctorReportBlockingPort`/`TestWriteDoctorReportAllClear` would fail if this classification regressed.
- **Docker check**: `dockerLine` takes an injected `lookPath func(string)(string,error)`; `TestDockerLine` exercises both found/not-found paths with no real docker binary required. `enginesImagePointer` (`doctor.go:31`) is a byte-for-byte match of the strings already in `serve.go:23,90`.
- **main.go wiring**: `cmdDoctor` const, switch case, and usage line follow the existing pattern (`cmdEnv`/`cmdCost`/`cmdNet`) exactly, dispatching `runDoctor(os.Args[2:])` and printing/exiting(1) on error.
- **Tests have teeth**: flipping the required/warn classification, the blocker count, or the engines-pointer string would fail the existing tests.

### Findings

- **[Medium] Missing docs section in `docs/standalone-server.md`.** Every sibling subcommand has a dedicated section there (`## Background mode`, `## Named snapshots`, `## Network reachability`, `## Cost estimate`, `## Pointing SDKs at it` for `env`) — `doctor` gets none, and isn't cross-referenced even from the adjacent `## Ports` section. A user reading the canonical CLI reference has no way to discover `cloudemu doctor` exists.
- **[Low] Triplicated engines-pointer string, no coupling test.** `"docker run -p 4566:4566 ghcr.io/stackshy/cloudemu:engines --all-real"` now lives in `serve.go:23`, `serve.go:90`, and `doctor.go:31`. `TestWriteDoctorReportAllClear` only asserts doctor's own constant appears in its own output, not equality against `serve.go`'s string, so the two can silently drift on a future edit to either file. (Pre-existing duplication pattern in `serve.go` itself, so not a regression, but a missed chance to extract a shared constant.)
- **[Low] `main()` cyclomatic complexity 16→18** (`main.go:60`), already over the lint threshold (10) at baseline — not a new golangci-lint issue (same reported line, identical 14-issue count before/after), but each new subcommand case pushes an existing violation further.
- **[Low] `runDoctor` and the non-AWS `portFlagHint` branches are untested** (0% / 33% coverage) — consistent with the repo's existing convention of not testing CLI entrypoint wrappers (`runEnv`/`runCost`/`runNet` aren't tested either either), so not a deviation, just worth noting.
- **[Low/observation] No `--providers` flag on doctor.** It always treats aws/azure/gcp/k8s as required regardless of how the user intends to invoke `serve --providers=...`. This matches the documented scope ("ports `serve` binds by default"), so not a bug — noting only as a possible future enhancement.

No High findings. All CI-blocking gates (build/vet/test/tidy) are green, and the local golangci-lint gate shows zero net-new issues.

**Verdict: approving.** The Medium docs gap and the triplicate-string nit are being addressed in a follow-up commit on this branch.

Issue #408, opened 2026-08-20
End-to-end app testing gaps: no port-forward/exec, synthetic pods, PVC GC on uninstall (from testing a real Helm chart)


NitinKumar004:
## Summary

While validating a real Helm chart (a Qdrant vector-DB `StatefulSet` chart) end-to-end against cloudemu, the **control-plane half worked great** but I hit a clear, consistent set of gaps that make **full end-to-end application testing** (the "does the workload actually work" half) impossible today. Filing this as a consolidated, chart-driven companion to the (now-closed) #312, which already documents that the data plane emulates the **Kubernetes API + built-in controllers but does not run containers**.

This isn't a bug report — everything cloudemu claims to do worked. It's a scoped feature request: the delta between "the API behaves like a cluster" and "I can test my app end-to-end like I would on minikube."

## What worked (control plane — excellent)

Running an EKS-issued kubeconfig against the in-memory k8s API (`:4570`), a full `helm install` of a non-trivial chart round-tripped cleanly:

- `helm install` / `upgrade` / `uninstall` all succeeded via real Helm + `client-go`.
- All objects accepted and created: `StatefulSet`, `Pod`, `PVC` (from `volumeClaimTemplates`), two `Service`s, `Secret`, `ConfigMap`.
- **Custom CRDs** (`ServiceMonitor` + `PrometheusRule`, `monitoring.coreos.com/v1`) registered and their CRs were accepted.
- StatefulSet materialized its pod to `Running`/`ready`, PVC named correctly (`<vol>-<sts>-0`).
- `helm upgrade` preserved a `lookup`-generated Secret (idempotent credential pattern) — great.
- Label-selector-driven `ServiceMonitor` targeting resolved correctly at the object level.

## What blocks full end-to-end testing (the gaps)

These are the things I could **not** verify on cloudemu and had to fall back to minikube for:

1. **No real container runtime — pods are synthetic.**
   Pod comes up `Running` with `imageID: cloudemu://...` and a "synthetic log stream". The app inside (`qdrant/qdrant:v1.19.0`) never actually executes, so nothing serves on its ports. This is the root of items 2–4.

2. **`pods/portforward` not implemented.**
   ```
   error: error upgrading connection: k8s api: pods/portforward requires a
   streaming connection upgrade cloudemu does not implement
   ```
[...]

PR #888, opened 2026-08-30
fix(kubernetes): honor persistentVolumeClaimRetentionPolicy + cascade registry stores on namespace delete (#408)

NitinKumar004:
## Summary

StatefulSet `volumeClaimTemplate` PVCs were **unconditionally** stamped with the StatefulSet as owner in `syncStatefulSetPVCsLocked`. Because `garbageCollectLocked` cascades on an owner's delete, deleting a StatefulSet (a `helm uninstall`) reaped every PVC — a **data-loss bug**. Real Kubernetes defaults `persistentVolumeClaimRetentionPolicy` to `Retain/Retain`, where the PVCs are ownerless and survive both delete and scale-down.

This PR makes cloudemu honor `spec.persistentVolumeClaimRetentionPolicy` and, paired with it, fixes a pre-existing namespace-cascade leak.

### 1. PVC retention policy (`reconcile.go`)
- **Retain/Retain (default)** → **no owner ref**. PVCs outlive the StatefulSet, so a helm uninstall no longer loses data.
- **`whenDeleted=Delete`** → set the StatefulSet owner ref, so the STS-delete GC cascade reaps the PVC (the one case where the owner-ref mechanism actually works).
- **`whenScaled=Delete`** → **explicit out-of-range-ordinal reap** in the PVC-sync path. StatefulSet scale-down deletes Pods with a bare `delete(s.pods,…)` that never invokes `garbageCollectLocked`, so a Pod owner ref would be inert. Instead, after computing the new replica count, any `volumeClaimTemplate` PVC whose ordinal `>= replicas` is deleted directly (finalizer-gated, quota-accurate — same teardown as a direct DELETE).

### 2. Namespace cascade fix (`namespace.go`)
`deleteNamespaceLocked` swept only the 7 typed maps and never `s.reg.stores`, so a namespace delete **leaked** all registry-backed namespaced objects (PVCs, StatefulSets, Ingresses, custom resources). Added a namespace-scoped sweep of `s.reg.stores` (finalizer-gated, mirroring the typed cascade; cluster-scoped objects never match) so a namespace delete removes everything inside it, matching real k8s.

## Tests (`statefulset_pvc_retention_test.go`, HTTP-fixture style)
- STS + volumeClaimTemplate → PVC created + Bound, **no owner ref**; delete STS → PVC **survives** (default Retain).
- `whenDeleted=Delete` → delete STS → PVC removed.
- `whenScaled=Delete`, replicas 3 → 3 PVCs; scale to 1 → ordinals 1,2 removed, ordinal 0 kept.
- Namespace delete with a PVC + StatefulSet + CRD/CR → all removed.

## Validation (GOMAXPROCS=3)
- `go build ./...` clean
- `go vet ./services/kubernetes/...` clean
- `go test -count=1 ./services/kubernetes/...` pass
- `go test -race -count=1 ./services/kubernetes/...` pass
- `golangci-lint run ./services/kubernetes/...` — only the pre-existing `admission.go:351` gosec (baseline), exit 0

### DIFF
--- a/services/kubernetes/namespace.go
+++ b/services/kubernetes/namespace.go
@@ -314,6 +314,41 @@ func (s *ClusterState) deleteNamespaceLocked(ns *corev1.Namespace) {
 	cascadeDeleteWithEvents(s, s.services, prefix, name, s.wServices)
 	cascadeDeleteWithEvents(s, s.deployments, prefix, name, s.wDeployments)
 	cascadeDeleteWithEvents(s, s.endpoints, prefix, name, s.wEndpoints)
+
+	// The typed maps above hold only 7 built-in kinds. Registry-backed namespaced
+	// objects (PVCs, StatefulSets, Ingresses, custom resources, …) live in
+	// s.reg.stores and were previously leaked on namespace delete. Sweep them too
+	// so a namespace delete removes everything inside it, matching real k8s.
+	s.cascadeRegistryStoresLocked(name)
+}
+
+// cascadeRegistryStoresLocked deletes every registry-backed namespaced object in
+// the named namespace, mirroring the typed cascade above (finalizer-gated:
+// finalizer-bearing objects go Terminating, the rest are removed with a DELETED
+// event). Cluster-scoped registry objects (PVs, Nodes, CRDs) carry an empty
+// namespace and never match. Callers hold s.mu — the set of stores only changes
+// under s.mu, so ranging s.reg.stores directly is safe (see garbageCollectLocked).
+func (s *ClusterState) cascadeRegistryStoresLocked(namespace string) {
+	prefix := namespace + "/"
+
+	for _, st := range s.reg.stores {
+		for key, obj := range st.items {
+			if !strings.HasPrefix(key, prefix) || obj.GetNamespace() != namespace {
+				continue
+			}
+
+			if s.markForDeletionUnstructured(obj) {
+				s.stampRegistryRVLocked(obj)
+				st.watch.publish(EventModified, namespace, *obj.DeepCopy())
+
+				continue
+			}
+
+			s.stampRegistryRVLocked(obj)
+			delete(st.items, key)
+			st.watch.publish(EventDeleted, namespace, *obj.DeepCopy())
+		}
+	}
 }
 
 // deepCopier constrains the element type of a per-resource map: it must be
--- a/services/kubernetes/reconcile.go
+++ b/services/kubernetes/reconcile.go
@@ -7,6 +7,7 @@ import (
 	"reflect"
 	"sort"
 	"strconv"
+	"strings"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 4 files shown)


NitinKumar004 (commented):
Verified against 79ddeb6 (fix/k8s-pvc-retention).

Gates (GOMAXPROCS=3): `go build ./...` clean, `go vet ./...` clean, `go test -count=1 ./services/kubernetes/...` pass, `go test -race -count=1 ./services/kubernetes/...` pass, `go generate ./...` no diff. `golangci-lint run ./services/kubernetes/...` reports only the pre-existing `admission.go:351` gosec finding — identical on `development`, not introduced here.

Adversarially verified each claim by mutation-testing the actual mechanism (reverting the guard/branch and confirming the corresponding test fails) plus two throwaway reproductions:

- **Default Retain → no owner ref**: confirmed. Reverting the `whenDeleted == pvcRetentionDelete` guard back to an unconditional owner-ref stamp makes `TestStatefulSetPVC_SurvivesDelete_DefaultRetain` fail immediately (`data-web-0 has owner refs under Retain`). A real STS delete under the default leaves the PVC in place.
- **whenDeleted=Delete → GC cascade**: traced the path — `registryDelete` → `teardownRegistryObjectLocked` → `garbageCollectLocked(sts.UID)` walks `s.reg.stores` matching `ownedBy(pvc.OwnerReferences, uid)`, and only this policy stamps that owner ref.
- **whenScaled=Delete reap**: `reapScaledStatefulSetPVCsLocked`'s ordinal parse (`strconv.Atoi` on the name suffix after the `<template>-<sts>-` prefix) safely skips any non-numeric remainder rather than deleting it, so a name that merely looks similar is never touched. Reverting the reap call to a no-op makes `TestStatefulSetPVC_WhenScaledDelete_ReapsOrdinals` fail (`data-web-1 survived`). Additionally verified scale-*up* (3 replicas via both a full PUT and the `/scale` subresource, which also runs `reconcileStatefulSet`) creates the newly in-range ordinals and never touches existing ones — no deletion fires on scale-up.
- **Namespace cascade scoping**: added and ran a throwaway test creating a cluster-scoped PersistentVolume named `doomed-pv` (namespace being deleted is `doomed`) — it survives the namespace delete. The `objKey(ns, name) = ns+"/"+name` format plus k8s names disallowing `/` means the `strings.HasPrefix(key, prefix)` check can't accidentally match a cluster-scoped key (`"/doomed-pv"` vs prefix `"doomed/"`), and the `obj.GetNamespace() != namespace` check is a correct (if redundant) second guard.
- **Finalizer-gating**: present and correct in all three new/changed delete paths (`reapRegistryObjectLocked`, `cascadeRegistryStoresLocked`, and the scaled-PVC reap that calls into `reapRegistryObjectLocked`) — each checks `markForDeletionUnstructured` before hard-deleting.
- **Blast radius of `reapRegistryObjectLocked`**: purely additive — `registryDelete`'s existing inline finalizer/teardown logic is untouched, and nothing else in the pre-existing codebase calls the new helper.

One robustness note, not blocking: `cascadeRegistryStoresLocked` (namespace.go) re-implements the finalizer-check-then-delete pattern inline instead of calling the new `reapRegistryObjectLocked`, and its delete branch is a strict subset of `teardownRegistryObjectLocked` — it skips `garbageCollectLocked`, any kind-specific `onDelete` hook, and `releaseQuotaLocked`. Today that's harmless (the only `onDelete` hook, the CRD one, is cluster-scoped and excluded by the same namespace guard; every namespaced child in the doomed namespace is independently swept by the same loop regardless of ownership; and quota state disappears along with the ResourceQuota object itself). But it's a latent trap the day a namespaced kind grows an `onDelete` hook — the namespace-delete path would silently skip it while a direct DELETE or the STS PVC reap would not. Worth having it call `reapRegistryObjectLocked` directly instead of duplicating the branch.

Everything else checked out. Approving.

PR #889, opened 2026-08-30
feat(kubernetes): serve pods/exec and pods/attach over WebSocket (deterministic session) + pin k8s listener to http/1.1 (#408)

NitinKumar004:
## Summary

`kubectl exec` / `kubectl attach` against the emulated Kubernetes data plane no longer return `501 NotImplemented`. `pods/exec` and `pods/attach` now complete a real WebSocket upgrade (remotecommand **v5**/v4 channel protocol) and run a **deterministic synthetic session**:

- First stdout bytes are a **self-describing banner** (mirroring the `pods/log` synthetic marker): `cloudemu: synthetic exec session for pod ns/name — no real container runtime; command echoed, no on-disk state`.
- For `exec`, the `command` query params are echoed to stdout.
- stdin/resize are drained (no container to receive them); `tty`/`container`/`stdin` are honored cosmetically.
- A `metav1.Status{Status: "Success"}` is written to the **error channel (3)** = exit 0, so `kubectl exec` returns 0. Bytes are byte-identical on every run (determinism).

This covers kubectl ≥ 1.29 and Helm v4, which negotiate the WebSocket v5 CLOSE protocol.

### Hand-registered v4/v5 subprotocols
`wsstream.NewDefaultChannelProtocols` registers only `""` / `channel.k8s.io` / `base64.channel.k8s.io` — it does **not** include `v4.channel.k8s.io` / `v5.channel.k8s.io`. We start from the default map and add both (names from `remotecommand.StreamProtocolV4Name/V5Name`), honoring the client's `Sec-WebSocket-Protocol`.

### Mandatory http/1.1 ALPN pin (h2 Hijack panic)
The k8s listener's serving TLS config (`internal/k8spki.ServingTLSConfig`) previously set no `NextProtos`, so `net/http` auto-negotiates HTTP/2 via ALPN. `golang.org/x/net/websocket` does a naked `http.Hijacker.Hijack()` with no fallback, which **panics** under an h2 connection. We pin `NextProtos: []string{"http/1.1"}` there. This config is k8s-listener-only (Azure uses a separate TLS config), and watch-streaming is plain chunked HTTP/1.1, so it is unaffected. A test proves that a client offering `h2` first in ALPN still negotiates `http/1.1`.

## Deferred
- **portforward**: still returns the typed `501` Status (needs an in-process pod:port handler / SPDY-over-WebSocket tunneling — tracked separately).
- **Real container runtime** workloads, SPDY (non-WebSocket) exec clients, probe/metric actuation.

## Tests
- WebSocket exec/attach happy paths (v5 and v4): banner on stdout, echoed command, Success Status on the error channel, determinism across two runs.
- Non-WebSocket exec → clean typed `400` Status (not a raw 501, no panic).
- portforward still returns a typed `501`.
- `k8spki`: `NextProtos` is `http/1.1`-first with no `h2`; and an ALPN e2e proving `http/1.1` is negotiated when the client offers `h2`.
- Existing route/subresource/log tests stay green.

Gates: `go build ./...`, `go vet`, `go test` (incl. `-race`) on kubernetes + k8spki + serverkit, `golangci-lint` (zero new), `go mod tidy` (x/net flips indirect→direct), `go generate` (no diff) all clean.


NitinKumar004 (commented):
## Gate matrix (clean checkout at `0d449b87`, GOMAXPROCS=3)

| Gate | Result |
|---|---|
| `go build ./...` | pass |
| `go vet ./...` | pass |
| `go test -count=1 ./services/kubernetes/... ./internal/k8spki/... ./server/serverkit/...` | pass |
| `go test -race -count=1 ./services/kubernetes/...` | pass |
| `go test -race -count=1 ./internal/k8spki/...` | pass |
| `go mod tidy` | no diff (x/net stays direct as described) |
| `gofmt -l` on changed files | clean |
| `go generate ./...` | no diff (coverage docs unaffected — no driver-capability change) |
| `golangci-lint run ./services/kubernetes/... ./internal/k8spki/...` (v2.11.4) | 1 pre-existing hit in `admission.go` (untouched by this diff, gosec G704) — **zero new issues** |

Verified independently (not just trusting the PR description):
- **ALPN pin has teeth.** Temporarily reverted the `NextProtos` line in `internal/k8spki/pki.go` and reran `TestServingTLSConfig_PinsHTTP11` / `TestServingTLSConfig_NegotiatesHTTP11UnderH2Offer` — both fail without the pin (negotiated ALPN comes back `"h2"`), confirming Go's `http.Server.ServeTLS` really does auto-enable h2 here and the test would catch a regression. Restored the line; diff matches the original exactly afterward.
- **`ServingTLSConfig` is the single shared k8s data-plane TLS config**, used by `server/serverkit/serverkit.go:814` for the one `a.k8sBackend` listener that serves EKS/AKS/GKE together (Azure's ARM listener at `providerAzure` uses the unrelated `tlsConfig()` helper). No other code path in the repo negotiates or depends on h2 against this listener (checked `server/{aws,azure,gcp}/{eks,aks,gke}` and `providers/aws/eks`) — pinning http/1.1-only doesn't regress watch streaming (plain chunked HTTP/1.1) or anything else on this port.
- **Wire shape for the error channel is correct.** Cross-checked against `client-go@v0.32.0/tools/remotecommand/v4.go` (`errorDecoderV4.decode`) and `errorstream.go` (`watchErrorStream`): the client `io.ReadAll`s the error channel until EOF and does `json.Unmarshal` into `metav1.Status`, treating `Status: "Success"` as exit 0. `pod_exec.go`'s `writeExecStatus` + the deferred `conn.Close()` (which EOFs the channel) reproduces this exactly.
- **v4/v5 channel index constants match upstream** (`remotecommand.StreamStdIn/Out/Err/Err/Resize = 0..4`), and `wsstream.NewDefaultChannelProtocols` was confirmed (by reading `apimachinery@v0.32.0/pkg/util/httpstream/wsstream/conn.go`) to register only `""`/`channel.k8s.io`/`base64.channel.k8s.io` — the hand-added v4/v5 binary entries are the only way kubectl ≥1.29/Helm v4 negotiate at all.
- **Non-WebSocket exec / portforward**: read the full `servePodSubresource` switch — no panic path; `writeStreamingUnsupported` for portforward is untouched, and `serveExecAttach` returns a typed 400 before ever touching `wsstream` when `IsWebSocketRequest` is false.
- **Determinism**: no `time.Now()`/`rand` anywhere in `pod_exec.go`; `TestPodExec_Deterministic` diffs two independent sessions' stdout/error-channel bytes.

## Findings

**[High] Unbounded per-request goroutine leak on any failed WS handshake — `services/kubernetes/pod_exec.go:63-73`**

`runSyntheticExecSession` does:
```go
conn := wsstream.NewConn(execChannelProtocols())
conn.SetIdleTimeout(execStreamIdleTimeout)
defer conn.Close()                    // line 66

_, streams, err := conn.Open(w, r)    // line 68
if err != nil {
    return                            // line 72
}
```
`wsstream.Conn.Close()` is `<-conn.ready; return conn.closeNonThreadSafe()` — `conn.ready` is only closed by `Conn.initialize()`, which only runs if the RFC6455 handshake inside `golang.org/x/net/websocket`'s `serveWebSocket` succeeds. Whenever `conn.Open()` returns an error — an unsupported `Sec-WebSocket-Protocol`, or *any* malformed/incomplete WebSocket handshake (missing `Sec-WebSocket-Key`, bad `Sec-WebSocket-Version`, etc.) that nonetheless has `Upgrade: websocket`+`Connection: Upgrade` and so passes `wsstream.IsWebSocketRequest` — `conn.ready` is never closed, and the deferred `conn.Close()` at line 66 blocks forever on `<-conn.ready`. The goroutine handling that HTTP request is permanently parked; this is invisible to the client (the TCP connection is already closed by `x/net/websocket`'s own `defer rwc.Close()`), so nothing looks wrong from a client's point of view while the server silently accumulates stuck goroutines.

Reproduced with a throwaway test (not part of this review's diff): 20 requests against `pods/.../exec` with `cfg.Protocol = []string{"totally-bogus-protocol"}` left exactly 20 extra live goroutines, all blocked at `pod_exec.go:72` inside `wsstream.(*Conn).Close` (`conn.go:311`, waiting on `<-conn.ready`), confirmed via a `runtime.Stack` dump. This is trivially reachable — no exploit needed, just a client (or health-check/probe/old client) that hits `pods/exec` or `pods/attach` with a WS-looking-but-invalid handshake — and each attempt leaks one goroutine forever with no cap, which is a real resource-exhaustion path against the k8s data-plane listener.

Fix: only defer/call `conn.Close()` on the success path, e.g.
```go
conn := wsstream.NewConn(execChannelProtocols())
conn.SetIdleTimeout(execStreamIdleTimeout)

_, streams, err := conn.Open(w, r)
if err != nil {
    return
}
defer conn.Close()
```
(`wsstream.Open` already calls its own `closeNonThreadSafe()` internally on the failure path, so nothing extra is needed there.)

**[Medium] Docs not updated for the new exec/attach behavior — `docs/services.md:1848`, `docs/services.md:1872`, `docs/sdk-server.md:337`**

All three still read "`exec`/`attach`/`portforward` return a typed 501" (or equivalent), which is now false for `exec`/`attach`. This PR touches no files under `docs/`. `docs/services.md` is called out in the project's own conventions as the source of truth for what's implemented vs. a documented limitation; leaving it stating the old (pre-PR) behavior will mislead anyone reading it about what `kubectl exec`/`attach` now do against the emulator.

## Verdict

Recommend **request changes** for the goroutine-leak finding — straightforward one-line-reorder fix, but it's a real, reproducible DoS surface on a network-facing listener and should not merge as-is. The h2/ALPN pin, the v4/v5 protocol registration, the wire shape on the error channel, and the WebSocket happy-path tests are all correct and independently verified (including making the ALPN test fail by reverting the fix). The docs-sync gap is a should-fix, not a blocker.

Issue #876, opened 2026-08-29
Kubernetes: no async/time progression + empty kube-system


NitinKumar004:
The reconciler is synchronous/always-converged: no async controllers, no rollout surge/maxUnavailable pacing, and CronJobs only advance via an explicit `TickCronJobs` (no background clock loop). The cluster also has an empty `kube-system` — no coredns/kube-proxy/system pods. Rollouts complete instantly and time-based behavior (CronJob schedules, pod age) can't be observed without manual ticks.

## What to build (opt-in, deterministic)
- A driven clock/tick that can advance CronJobs and any time-based progression, kept deterministic/testable (wired to `config.Clock` + an admin tick or interval).
- Optionally seed representative `kube-system` components (coredns, kube-proxy) so `kubectl get pods -n kube-system` looks real.

## Where
`services/kubernetes/` (cronjob tick loop, bootstrap).

NitinKumar004:
Done (#890) — seeded kube-system (coredns Deployment+Service, kube-proxy DaemonSet) default-on; CronJobs fire on the opt-in --k8s-progression ticker with successful/failed history limits.

PR #890, opened 2026-08-30
feat(kubernetes): seed kube-system (coredns, kube-proxy) + drive CronJobs on the opt-in ticker with history limits (#876)

NitinKumar004:
## Summary

Part of the Tier-2 Kubernetes real-cluster epic (#876). Makes a fresh cluster look like a managed cluster and gives CronJobs a real firing loop, without breaking CloudEmu's determinism pillar (static realism default-on; time-driven behavior opt-in behind `--k8s-progression`).

### 1. Static kube-system seeds (DEFAULT-ON, deterministic)
A fresh cluster now boots kube-system with the standard add-ons — `kubectl -n kube-system get pods/deploy/ds/svc` looks like EKS/AKS/GKE instead of an empty object store. All are plain objects that ride the existing generic reconcile (no bespoke controller):

- **coredns Deployment** (apps/v1, replicas 2, `k8s-app=kube-dns`) → the existing Deployment reconcile yields 2 Running coredns pods (via a real ReplicaSet).
- **kube-dns Service** (v1, ports 53/UDP + 53/TCP + 9153/TCP) with its **ClusterIP pinned to `10.96.0.10`**; the monotonic ClusterIP allocator is **bumped past .10** so a client Service never collides with (or re-hands out) it. Endpoints populate from the coredns pods via the existing resync.
- **kube-proxy DaemonSet** (apps/v1, `k8s-app=kube-proxy`) seeded as a **real DaemonSet object** (not a hand-written static pod), so the DaemonSet reconcile yields 1 pod on the single node today and auto-scales to N once multi-node scheduling lands — zero further code needed.

Control-plane static pods (apiserver/scheduler/etc.) are intentionally omitted for the managed-cluster look. coredns is cosmetic (no real DNS resolver) — it just makes the cluster look real to tooling. No new snapshot `validate()` requirements: the seeds live in the normal typed maps / registry stores and snapshot cleanly.

### 2. CronJob firing on the opt-in ticker
`TickCronJobs()` (previously called only from tests) is now wired into `APIServer.TickAll()`, which the existing real-time serve ticker drives — gated on `--k8s-progression`. A **defense-in-depth self-gate** (`if !s.lifecycleProgression { return }`, mirroring `Tick()`) was added inside `TickCronJobs()` so it stays a no-op unless progression is on, regardless of caller. Tests still drive it directly under a FakeClock.

### 3. Job history-limit pruning
After each fire, finished Jobs beyond `successfulJobsHistoryLimit` (default 3) / `failedJobsHistoryLimit` (default 1) are deleted (oldest-first, deterministic), so `kubectl get jobs` self-trims and a long-lived `serve` doesn't grow unbounded.

## Tests
- Fresh cluster kube-system: coredns Deployment (2 Running pods), kube-proxy DaemonSet (1 pod), kube-dns Service @ 10.96.0.10 with populated endpoints.
- CronJob under progression + FakeClock fires on schedule; successful history trims to ≤3, failed to ≤1.
- `TickCronJobs` with progression off is a no-op (the self-gate).
- Updated `service_test.go` first-client-IP expectations (now `.11/.12/.13` — kube-dns consumed .10 and bumped the counter) plus the all-namespace pod/service/deployment/endpoint count assertions that shift with the seeds.

All gates green: `go build ./...`, `go vet`, `go test` (full suite) + `-race`, `golangci-lint` (0 new — the one gosec G704 is pre-existing in the untouched admission.go), `go generate` no diff.

### DIFF
--- a/services/kubernetes/apiserver.go
+++ b/services/kubernetes/apiserver.go
@@ -120,6 +120,11 @@ func (s *APIServer) TickAll() bool {
 	changed := false
 
 	for _, st := range states {
+		// CronJob firing is opt-in time-driven behavior like the staged Pod
+		// lifecycle — both are gated on progression (Tick and TickCronJobs
+		// self-gate), so the real-time serve ticker drives them together.
+		st.TickCronJobs()
+
 		if st.Tick() {
 			changed = true
 		}
--- a/services/kubernetes/cronjob.go
+++ b/services/kubernetes/cronjob.go
@@ -1,6 +1,7 @@
 package kubernetes
 
 import (
+	"sort"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -34,6 +35,13 @@ const (
 func (s *ClusterState) TickCronJobs() {
 	s.mu.Lock()
 	defer s.mu.Unlock()
+
+	// Defense-in-depth self-gate (mirrors Tick's at lifecycle.go): CronJob firing
+	// is opt-in time-driven behavior, so it stays a no-op unless progression is on
+	// — even if a caller other than the gated serve ticker reaches this.
+	if !s.lifecycleProgression {
+		return
+	}
 
 	cronStore := s.reg.getStore(apiGroupBatch, "v1", "cronjobs")
 	jobStore := s.reg.getStore(apiGroupBatch, "v1", "jobs")
@@ -178,22 +186,7 @@ func activeJobsFor(cj *unstructured.Unstructured, jobStore *registryStore) []*un
 
 // jobTerminal reports whether a Job has finished (Complete or Failed True).
 func jobTerminal(job *unstructured.Unstructured) bool {
-	conds, _, _ := unstructured.NestedSlice(job.Object, "status", "conditions")
-	for _, raw := range conds {
-		cond, ok := raw.(map[string]any)
-		if !ok {
-			continue
-		}
-
-		ctype, _, _ := unstructured.NestedString(cond, "type")
-		cstatus, _, _ := unstructured.NestedString(cond, "status")
-
-		if cstatus == "True" && (ctype == "Complete" || ctype == "Failed") {
-			return true
-		}
-	}
-
-	return false
+	return jobHasCondition(job, "Complete") || jobHasCondition(job, "Failed")
 }
 
 // deleteJobLocked removes a Job and cascade-collects its Pods (Replace policy).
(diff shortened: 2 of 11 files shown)


NitinKumar004 (commented):
Adversarial review, clean build (`go build ./...`, `go vet ./...` — 0 issues), targeted tests (`services/kubernetes`, `server/serverkit`, root) and `-race` on `services/kubernetes` all green, `golangci-lint run ./services/kubernetes/...` shows the same single pre-existing `gosec G704` finding in `admission.go:351` (unrelated file, present on `development` baseline too) — no new lint issues.

Verified each load-bearing claim:

1. **kube-proxy is a real DaemonSet, not a static pod.** `seedKubeProxyDaemonSetLocked` (kube_system.go:141-172) writes into the actual `apiGroupApps/v1/daemonsets` registry store and calls `reconcileDaemonSet` directly — the same function `registry_defs.go:69` wires as that resource's generic reconcile hook. `reconcileDaemonSet` (reconcile.go:506) computes pod names via `syncStablePods` with `obj.GetName()+"-"+nodeName` (deterministic, not the random `shortID()` used for Deployment-owned pods), and since the seeded DaemonSet spec carries no `nodeSelector`, `daemonSetSchedulesToNode` always matches the single synthetic node — exactly 1 pod today, and this is the same path PR4's multi-node scheduling will scale.

2. **ClusterIP pin + allocator bump.** `nextClusterIP` starts at `firstClusterIPOffset=1` and nothing else allocates before kube-system seeding (no bootstrap `kubernetes` Service in this codebase), so `seedKubeDNSServiceLocked`'s `if s.nextClusterIP <= 10 { s.nextClusterIP = 11 }` bump is unconditionally correct on a fresh cluster — no off-by-one, no collision, no reuse of `.10`. `TestService_ClusterIPAllocatedSequentially` correctly asserts `.11/.12/.13` for the first three client Services.

3. **Determinism / snapshot round-trip.** Object names/UIDs in the seed path use the same `newUID()`/`shortID()` calls used everywhere else in the package for Deployment-owned pods — not a new non-determinism introduced by this PR (existing framework-wide pattern). Snapshot correctness isn't actually affected: `restoreClusterLocked` wholesale-replaces every typed map and every registry store (including the new `daemonsets` store) from the snapshot, discarding whatever the fresh `newClusterState` seeded — the exact same pattern already used for namespaces/SAs/Node/Lease. `TestSnapshotRoundTrip`'s byte-equality assertion (`bytes.Equal(raw, raw2)`) passed against a cluster that now includes the new kube-system objects, empirically confirming lossless round-trip.

4. **CronJob opt-in gating.** `TickCronJobs` self-gates on `s.lifecycleProgression` (cronjob.go:42-44) and is only invoked from `TickAll`, which is only invoked from `server/serverkit/serverkit.go`'s ticker goroutine gated on `a.cfg.K8sProgression` (i.e. `--k8s-progression`/`CLOUDEMU_K8S_PROGRESSION`). No default test path can reach it. Both existing direct callers (`cronjob_test.go`'s `newCronFixture`, `cronjob_internal_test.go`'s `TestTickCronJobs_MaterializesJob`) were correctly updated to call `SetLifecycleProgression(true)` first — without that the self-gate would have turned them into silent no-ops, and the new `TestTickCronJobs_NoOpWithoutProgression` explicitly locks in the gate.

5. **History pruning.** `pruneJobHistoryLocked` scopes to `ownedBy(job.GetOwnerReferences(), cj.GetUID())` (only this CronJob's Jobs, not cluster-wide), buckets only by `jobHasCondition(..., "Complete"/"Failed")` (active/non-terminal Jobs never enter either bucket, never touched), and `trimFinishedJobsLocked` sorts oldest-first by `(creationTimestamp, name)` before dropping everything beyond the limit — deterministic. Defaults (3 successful / 1 failed) and the negative-clamps-to-0 handling match `batch/v1` semantics.

6. **Test count updates are arithmetically correct, not fudged.** With coredns (2 pods, 1 Deployment) + kube-proxy (1 pod, 1 DaemonSet) + kube-dns (1 Service, 1 Endpoints) seeded: pod all-ns count 2→5 (2 user + 3 system), Service count 2→3, Deployment count 2→3, Endpoints count 2→3 all check out exactly against the seed shape.

7. **kube-dns endpoints ordering.** `seedKubeSystemLocked` seeds coredns before kube-dns; `reconcileDeploymentLocked` synchronously materializes and marks the 2 coredns pods Running before returning, so by the time `seedKubeDNSServiceLocked`'s `reconcileServiceEndpointsLocked` scans `s.pods` for the `k8s-app=kube-dns` selector, both pods are already there — `TestKubeSystem_KubeDNSServicePinnedClusterIP` confirms 2 endpoint addresses.

No blocking or major issues found. One informational (non-blocking) note: coredns/kube-proxy pod names and object UIDs are randomly generated per cluster boot (same pattern as every other Deployment/DaemonSet in this codebase), so kube-system pod identity isn't stable across `serve` restarts absent a snapshot restore — this is pre-existing framework behavior, not something this PR introduces, and snapshot restore (as verified above) preserves whatever identity a client already saw.

PR #891, opened 2026-08-30
feat(kubernetes): opt-in multi-node scheduler — nodeSelector/taints/requests first-fit + per-node DaemonSet fan-out (#875)

### LINKED ISSUE #875: Kubernetes: no scheduler semantics (single fixed node, every pod force-assigned)

NitinKumar004:
## Summary

Closes #875 — the last Tier-2 Kubernetes PR. Adds an **opt-in, fixed-at-seed multi-node scheduler** to the in-memory data plane. Default behavior (single node `cloudemu-node-0`, pods instantly Running) is unchanged; every existing single-node test passes untouched.

## What changed

**Opt-in `--k8s-nodes N` (fixed at seed, immutable).** A cluster seeds N synthetic nodes once at creation — node membership never changes for its lifetime (no Node reconcile hook, deleting a Node object is a documented no-op). N=1 (default) is a single control-plane node; N>1 seeds one control-plane node tainted `node-role.kubernetes.io/control-plane:NoSchedule` plus N-1 workers, each with its own InternalIP (`10.0.0.{i+1}`) and allocatable capacity. Wired through `serveflags` → `serverkit` → `APIServer.SetNodeCount`.

**Deterministic first-fit scheduler** (`scheduleNodeLocked`): honors `spec.nodeName` (bypass), then filters name-sorted candidates by nodeSelector, taints/tolerations (NoSchedule/NoExecute repel; Equal/Exists match), and CPU/memory requests ≤ allocatable (summed over pods already on the node). First feasible node wins; `status.HostIP` = that node's InternalIP. No fit → **Pending + `PodScheduled=False`/Unschedulable + `FailedScheduling` event**, never force-Running.

**Structural N=1 back-compat gate:** the request-feasibility/taint filtering runs *only* when node count > 1 (read from the node store, so it survives snapshot/restore). At N=1 scheduling is unconditionally back-compat regardless of a manifest's requests — the invariant is structural, not "small requests happen to fit."

**Correctness fixes that fall out of fallible scheduling:**
- `markPodSucceededLocked` no longer forces `Succeeded` on an unscheduled pod (checks `spec.nodeName != ""`); `reconcileJob` counts only actually-Succeeded pods and marks the Job `Complete` only when all completions succeed — an unschedulable Job is not Complete.
- Workload `readyReplicas`/`availableReplicas` (Deployment/RS/StatefulSet) are recomputed from actual Running pod phase, not mere existence, so a Deployment with infeasible requests reports fewer ready replicas while pods sit Pending. `replicas` still reflects the materialized count.

**Per-node DaemonSet fan-out** (`reconcileDaemonSet`): one pinned pod (`<ds>-<node>`) per node whose labels match the template's nodeSelector and whose taints the template tolerates; `spec.nodeName` set before scheduling. The seeded **kube-proxy** DaemonSet now tolerates every taint (real kube-proxy behavior), so it scales to N pods (control-plane included) when `--k8s-nodes>1` — zero pods on the control-plane for a DaemonSet without a matching toleration.

**Endpoints** carry each pod's actual `spec.nodeName` (typed Endpoints + EndpointSlice). Node metrics iterate all nodes.

## Deferred (tracked follow-ups)

Node/inter-pod affinity, topology spread, scoring/bin-packing, `tolerationSeconds`/live eviction, and dynamic node add/remove — fixed-at-seed is v1.

## Tests

New `multinode_test.go`: control-plane/worker seeding + taints; single-node untainted; deterministic first-fit spread; nodeSelector placement; taint repels/tolerates; unschedulable-requests → Pending + FailedScheduling; Deployment readyReplicas reflects Pending; Job that can't schedule is not Succeeded; DaemonSet fan-out (workers only); kube-proxy scales to N; endpoints carry real node. Default N=1 suite runs unchanged.

Gates (GOMAXPROCS=3): `go build ./...`, `go vet`, `go test ./...` (full module), `go test -race ./services/kubernetes/...`, `golangci-lint` (zero new vs baseline), contrib/server build+test, `go generate` (no diff) — all green.


NitinKumar004 (commented):
Deep review of `5fcec723` against `development` (merge-base `c2e4196`). Full gate matrix green: `go build`/`go vet`/`go test` (including `./services/kubernetes/... ./server/serverkit/... ./server/serveflags/... ./cmd/... .`), `go test -race ./services/kubernetes/...`, contrib/server build+test, `go generate` (no diff, expected — no driver-interface change). `golangci-lint` on `./services/kubernetes/...` and `./cmd/cloudemu/...` reports 0 issues in any file this PR touches (the 14 pre-existing hits are all in `cmd/cloudemu/main.go|lifecycle.go|net.go`, none of which are in the diff); `./server/serveflags/... ./server/serverkit/...` is clean. `gofmt` flags one pre-existing misalignment in `serveflags_test.go` that's already present at the merge-base, unrelated to this PR's edits.

Adversarially verified the load-bearing claims:

- **N=1 structural back-compat**: the gate is `services/kubernetes/node.go:308` (`if len(nodes) <= 1`), which unconditionally places the pod with no taint/nodeSelector/request-feasibility check — not "requests happen to fit." Confirmed `state.go:198` seeds node-0 untainted whenever `nodeCount <= 1` (`tainted := s.nodeCount > 1`). Full existing k8s suite passes unchanged at default.
- **`markPodSucceededLocked`**: now checks `pod.Spec.NodeName == ""` before forcing Succeeded (`reconcile.go:772`); `reconcileJob` counts only pods actually at `PodSucceeded` and gates `Complete` on `succeeded >= completions`. Confirmed with the PR's own `TestMultiNode_JobUnschedulableNotSucceeded` — an infeasible-request Job correctly reports `succeeded=0`, no `Complete` condition, and the pod stays Pending, not Succeeded. Under the old code (unconditional `Succeeded` + `succeeded := len(owned)`) this would have falsely reported the Job complete.
- **Status counts**: `countRunningOwnedLocked` (actual `PodRunning`) feeds `readyReplicas`/`availableReplicas`, while `replicas` reflects the materialized count from `syncScaledPods`/`syncStablePods`/`syncDeploymentReplicaSetLocked`/`reconcileReplicaSet`. Traced the Deployment→RS→Pod threading — no double counting, the Deployment just mirrors its RS's already-correct `status.replicas`/`readyReplicas`.
- **Determinism**: `nodesLocked()` sorts by name after building from the map, so first-fit is reproducible regardless of Go map iteration order; `consumedOnNodeLocked` sums are order-independent (pure addition over already-committed `NodeName`s). Sequential pod creation in the top-up loops means scheduling order is loop order, not map order.
- **Taints/tolerations**: `tolerationMatchesTaint` correctly mirrors upstream `ToleratesTaint` (effect-match-or-empty, `Exists`-with-empty-key-tolerates-all, else key+value `Equal`); `podToleratesTaints` only hard-repels on `NoSchedule`/`NoExecute`, leaving `PreferNoSchedule` soft. Verified with the PR's tainted-control-plane tests plus an independent snapshot round-trip (below).
- **DaemonSet fan-out**: one pod per node passing `labelsMatch` + `podToleratesTaints`, named `<ds>-<node>`, `spec.nodeName` pre-set so `scheduleNodeLocked`'s explicit-nodeName bypass fires — confirmed this bypass also skips resource-fit for DaemonSet pods specifically (matches the historical DaemonSet-controller behavior, and is consistent since cloudemu's synthetic nodes never carry resource-pressure taints). `buildControllerPod`/`newControllerPod` split is clean: RS/STS/Deployment pods still go through the unchanged `buildControllerPod` path.
- **Endpoints**: `matchingEndpointAddressesLocked` stamps `NodeName: strPtr(pod.Spec.NodeName)` from the real scheduled node, gated by the existing `PodRunning` + non-nil-`PodIP` filter, so an unscheduled/Pending pod can never leak into Endpoints.
- **Snapshot round-trip**: wrote and ran a throwaway test — seeded 3 nodes, scheduled a pod onto a worker via `nodeSelector`, snapshotted, restored into a *fresh* `APIServer` configured with `nodeCount=1`, and confirmed the restored cluster still reports 3 nodes, the control-plane taint, and the pod still `Running` on its original worker. `restoreClusterLocked` wholesale-replaces the registry (including the `nodes` store) from the snapshot, so the persisted topology — not the live config — is authoritative on restore, exactly as the `snapshot_test.go` field-guard comment claims.
- **`markPodRunningLocked` fallibility**: traced every caller (`pod.go` create/PUT, `buildControllerPod`/`newControllerPod`, the Job path) — none crash or nil-deref on a failed schedule; `markPodUnschedulableLocked` only touches `Phase`/`Conditions`, leaving `PodIP`/`HostIP` at their zero value.
- **serveflags/serverkit wiring**: both `cmd/cloudemu` and `contrib/server` route through the same `serveflags.RegisterCommon`, so `--k8s-nodes` lands in both automatically; the flag-name drift tests derive their expected set from `RegisterCommon` itself rather than a hardcoded list, so there's no drift surface here.

Two things worth addressing:

**[Medium] `docs/services.md` now contradicts the shipped capability — lines 1834 and 1872.** The "Workload controllers" line still says DaemonSet fan-out is "one Pod per node... the synthetic node" (singular), and the "Emulation boundaries" paragraph explicitly states *"no real scheduling beyond the single synthetic node (DaemonSet `nodeSelector` is honored, but affinity/taints/resource-fit are not)"* — both now false once `--k8s-nodes>1` is set. `docs/services.md` is the authoritative capability doc; a reader deciding whether cloudemu can validate taint-based placement will conclude it can't. Suggest updating those two sentences the same way `--k8s-progression` is documented two paragraphs above (a short "opt-in multi-node scheduler" callout).

**[Low, informational, not this PR's scope]** `pod.go`'s Pod PUT/replace path doesn't enforce `spec.nodeName` immutability (pre-existing, untouched by this diff). At N=1 that was harmless — every scheduling attempt unconditionally succeeded. Now that scheduling is fallible, a PUT that omits `spec.nodeName` on an already-Running pod in a multi-node cluster will silently re-run the scheduler and can flip it to Pending if it no longer fits. Not a regression to fix here, but worth a tracked follow-up given the new scheduler makes it observable.

Everything else — the scheduler, the Job/status fixes, taints/tolerations, DaemonSet fan-out, snapshot fidelity, and the flag wiring — checks out.

PR #978, opened 2026-09-03
feat(azure-blob): soft-delete + undelete

NitinKumar004:
## Summary

Adds Azure Blob **soft delete** to the blob data plane, gated on the account-level delete-retention policy (`deleteRetentionPolicy.enabled/days`, already round-tripped via the ARM `blobServices/default` sub-resource).

When soft delete is in effect:
- **Delete Blob** retains the blob (bytes and all) instead of removing it.
- A normal **List Blobs** omits it; **List Blobs `include=deleted`** surfaces it with `Deleted=true`, a `DeletedTime`, and a `RemainingRetentionDays` countdown.
- **Undelete Blob** (`PUT ?comp=undelete`) restores it to active.
- After the retention window elapses the blob is purged lazily (via `config.Clock`, FakeClock-deterministic).

With soft delete **disabled**, Delete Blob remains a permanent hard delete (unchanged).

## Design
- New optional `AzureSoftDeleteBlob` driver capability, discovered by type assertion — mirrors the #968 `AzureVersionedBlob` pattern. S3/GCS are untouched.
- Soft-deleted blobs live in a per-container `softDeleted` store (persisted in Snapshot/Restore).

## Coexistence with versioning (#968)
Soft delete engages only when the retention policy is on **and** account versioning is off. With versioning enabled, a delete keeps the retained versions as the recovery mechanism (unchanged), rather than diverting to soft delete — verified by a dedicated test.

## Tests
Real `azblob` SDK wire tests (delete→list include=deleted→undelete; hard-delete when disabled; versioning coexistence) plus provider-level tests covering retention-window expiry, persist round-trip, and a concurrent `-race` delete/undelete sweep. Full `go test ./...` green; `docs/coverage` regenerated.

### DIFF
--- a/docs/coverage/azure/blobstorage.md
+++ b/docs/coverage/azure/blobstorage.md
@@ -85,6 +85,16 @@ AzureBlobExtensions is an OPTIONAL Azure-specific blob data-plane capability,
 | `SetContainerMetadata` | SetContainerMetadata replaces a container's metadata (Set Container |
 | `StageBlock` | StageBlock buffers an uncommitted block (Put Block, ?comp=block) for a blob |
 
+### AzureSoftDeleteBlob
+
+AzureSoftDeleteBlob is an OPTIONAL Azure-specific capability, discovered by
+
+| Operation | Description |
+| --- | --- |
+| `ListDeletedBlobs` | ListDeletedBlobs returns the soft-deleted blobs matching opts, so a List |
+| `SoftDeleteEnabled` | SoftDeleteEnabled reports whether soft delete is currently in effect for |
+| `UndeleteBlob` | UndeleteBlob restores a soft-deleted blob to active (PUT ?comp=undelete). |
+
 ### AzureVersionedBlob
 
 AzureVersionedBlob is an OPTIONAL Azure-specific capability, discovered by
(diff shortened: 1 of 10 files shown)


NitinKumar004:
Follow-up (review round 2): fixed the critical cross-store data-loss race in `c4f396d6`.

- The `objects`↔`softDeleted` transition is now one critical section under `ctr.mu` in BOTH directions (soft-delete and Undelete), so a blob is always in exactly one store — never dropped from both under any interleaving. Undelete sets `objects` before deleting `softDeleted` so a concurrent reader never observes it in neither.
- Added a same-key Delete/Undelete stress test asserting exactly-one-store after every round. Verified it FAILS without the lock (reproduces the `live=false deleted=false` permanent-loss state) and PASSES with it.
- Gates re-run: `go test ./providers/azure/blobstorage/... ./server/azure/blobstorage/... ./persist/... -race -count=2` green, broad `./server/azure/...` green, full `go test ./...` exit 0, `golangci-lint --new-from-rev=origin/development` 0 issues, coverage/mod-tidy clean.

Secondary (deferred, documented in `appendDeletedBlobs`): `include=deleted` appends the soft-deleted set on top of the maxresults-bounded live page rather than folding it into one merged, uniformly paginated live+deleted stream. Correct for the common small-container case; full merged-stream pagination needs a marker scheme encoding both stream positions and is left as future work.

PR #1041, opened 2026-09-04
fix(ec2): real-user e2e divergences from AWS

NitinKumar004:
## Summary
Found via a black-box audit driving cloudemu (real `serve` process) through the aws CLI and Terraform (hashicorp/aws ~> 5.0), then comparing against documented real AWS EC2/VPC behavior.

- Wire error messages leaked the internal cloudemu error-taxonomy code as a prefix (e.g. `AlreadyExists: key pair "x" already exists` instead of the clean human message real EC2 returns). Fixed ~20 call sites across the ec2 handlers to use `cerrors.Message(err)` instead of `err.Error()`.
- `DescribeVolumes` rejected the `attachment.*` filters (`attachment.device`, `attachment.instance-id`, `attachment.status`, `attachment.delete-on-termination`), breaking Terraform's `aws_volume_attachment` resource outright.
- `DescribeNetworkInterfaces` rejected `attachment.instance-id`, `attachment.device-index`, `attachment.attachment-id`, `attachment.status`, and `source-dest-check`, breaking `aws_instance`'s primary-ENI lookup.
- `DescribeAddressesAttribute` was entirely unimplemented ("unknown action"), breaking Terraform's `aws_eip` resource outright.
- `NetworkInterfaces[].SourceDestCheck` was always missing from `DescribeInstances`/`RunInstances` responses even though the underlying ENI record carries it.

## Test plan
- [x] `go build ./...`
- [x] `go test ./providers/aws/ec2/... ./providers/aws/vpc/... ./server/aws/ec2/... -race`
- [x] `go test ./...` (full suite, no regressions)
- [x] `gofmt -l` clean on changed files
- [x] `golangci-lint run` — 0 new issues (pre-existing baseline issues unchanged/reduced)
- [x] Reproduced each bug live against `cloudemu serve` with aws-cli and Terraform, and confirmed the fix with a full `terraform init/apply/destroy` cycle (VPC + subnet + IGW + route table + security group + key pair + instance + EBS volume + attachment + EIP)
- [x] Added regression tests: `TestDeleteVolumeAttachedNoLeakedPrefix`, `TestCreateKeyPairDuplicateNoLeakedPrefix`, `TestDescribeNetworkInterfacesAttachmentFilters`

### DIFF
--- a/providers/aws/ec2/ec2.go
+++ b/providers/aws/ec2/ec2.go
@@ -1379,7 +1379,7 @@ const (
 // SourceDestCheck) that real EC2 permits on a running instance and that are not
 // part of the portable Compute driver. Unlike ModifyInstance it does not
 // require the instance to be stopped.
-func (m *Mock) SetInstanceAttribute(_ context.Context, instanceID, name, value string) error {
+func (m *Mock) SetInstanceAttribute(ctx context.Context, instanceID, name, value string) error {
 	inst, ok := m.instances.Get(instanceID)
 	if !ok {
 		return cerrors.Newf(cerrors.NotFound, "instance %q not found", instanceID)
@@ -1392,7 +1392,15 @@ func (m *Mock) SetInstanceAttribute(_ context.Context, instanceID, name, value s
 	case attrDisableAPITermination:
 		inst.disableAPITermination = parseBool(value)
 	case attrSourceDestCheck:
-		inst.sourceDestCheck = parseBool(value)
+		sourceDestCheck := parseBool(value)
+		inst.sourceDestCheck = sourceDestCheck
+
+		// The networking mock's primary-ENI copy of this flag is what
+		// DescribeInstances/DescribeNetworkInterfaces actually report; keep
+		// it in step so the two describe paths never disagree.
+		if m.networking != nil {
+			_ = m.networking.SetPrimaryNetworkInterfaceSourceDestCheck(ctx, instanceID, sourceDestCheck)
+		}
 	case attrEbsOptimized:
 		inst.ebsOptimized = parseBool(value)
 	case attrDisableAPIStop:
@@ -1654,7 +1662,7 @@ func (m *Mock) AttachVolume(_ context.Context, volumeID, instanceID, device stri
 
 	if state := inst.readState(); state != compute.StateRunning && state != compute.StateStopped {
 		return cerrors.Newf(cerrors.FailedPrecondition,
-			"IncorrectInstanceState: instance %q is in state %q; a volume can only attach to a running or stopped instance",
+			"instance %q is in state %q; a volume can only attach to a running or stopped instance",
 			instanceID, state)
 	}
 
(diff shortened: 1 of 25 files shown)


NitinKumar004:
Follow-up commit addressing review findings: the initial fix left two perpetual Terraform diffs.

**Root causes fixed (commit c09ae941):**
1. `SourceDestCheck` was only added to the nested `networkInterfaceSet` item, not the top-level `Instance.sourceDestCheck` element aws-sdk-go-v2/terraform-provider-aws actually reads — added it at the top level too (`server/aws/ec2/xml.go`, `operations.go`).
2. `ModifyInstanceAttribute(SourceDestCheck)` only updated the compute driver's own copy, not the primary ENI's copy that `DescribeInstances`/`DescribeNetworkInterfaces` report — added `Networking.SetPrimaryNetworkInterfaceSourceDestCheck` so both stores stay in sync (`providers/aws/ec2/ec2.go`, `providers/aws/ec2/networking.go`, `providers/aws/vpc/network_interface.go`).
3. `DescribeAddresses` didn't return an EIP's tags even though `AllocateAddress` stored them — added `Tags` to `addressXML` (`server/aws/ec2/address.go`).
4. Cleaned the internal sub-code text (`VolumeInUse:`, `ZoneMismatch:`, `IncorrectInstanceState:`, `InvalidAttachment.NotFound:`, `CannotDelete:`, `InvalidNetworkInterface.InUse:`) that was baked into driver error messages; the wire layer now matches on the clean message text to pick the AWS error CODE (unchanged), and `error_shape_test.go`'s `leakedPrefixes` list now also guards these sub-codes. Added/extended regression tests in `volume_test.go`, `network_interface_attribute_test.go`, `security_group_test.go`.

**Re-verification (full Terraform lifecycle: vpc/subnet/igw/rt/sg/keypair/instance(source_dest_check=false)/ebs volume+attachment/eip+tags):**
```
$ terraform apply -auto-approve
Apply complete! Resources: 11 added, 0 changed, 0 destroyed.

$ terraform plan -detailed-exitcode
No changes. Your infrastructure matches the configuration.
$ echo $?
0
```

Also confirmed live via curl that `VolumeInUse`, `InvalidVolume.ZoneMismatch`, `IncorrectInstanceState` (via test), `InvalidAttachment.NotFound` (via test), `Client.CannotDelete`, and `InvalidNetworkInterface.InUse` all now return clean human-sentence messages with no internal code prefix.

**Note:** during re-verification I also found `user_data` produces a diff on every plan **only when `user_data` is omitted from the Terraform config** — reproduced that this is independent of the wire response (which is empty and consistent in both the omitted and explicit-`user_data` cases; only the omitted-config case triggers the provider-side mismatch). This is a client-side `terraform-provider-aws` v5.100.0 quirk, not a cloudemu bug — confirmed by setting `user_data` explicitly in config, which makes the diff disappear entirely. Filing this as a known workaround (set `user_data` explicitly, or `lifecycle { ignore_changes = [user_data] }`) rather than a cloudemu fix.

Gates: `go build ./...`, `go test ./providers/aws/ec2/... ./providers/aws/vpc/... ./server/aws/ec2/... ./persist/... -race -count=1` all green; `gofmt -l` clean; `golangci-lint` 23 issues before and after (0 new).

PR #1079, opened 2026-09-04
fix(security): resolve CodeQL high-severity alerts (allocation bounds + RSA key floor)

NitinKumar004:
Clears the 6 high-severity CodeQL alerts so the release code-scanning gate passes. All are pre-existing hardening items; each bound is now in a form CodeQL's dataflow follows.

## Fixes
- **`go/uncontrolled-allocation-size` ×5** — a request-controlled count feeding an allocation is now clamped by a `min(value, CONST)` on the same path, so the allocation is provably bounded:
  - `providers/aws/ec2/spot.go`, `providers/azure/virtualmachines/spot.go`, `providers/gcp/compute/spot.go` — spot request slice capacity → `min(cfg.Count, maxSpotRequests)`.
  - `providers/aws/elasticache/replication_group.go` — member-cluster synthesis → `min(nodes, maxReplicationGroupNodes)`.
  - `server/aws/elasticache/types.go` — node count → `min(numNodes, maxCacheNodesPerCluster)`.
- **`go/weak-crypto-key` ×1** — `providers/aws/acm/certificate_gen.go` now passes one of two compile-time constants (`rsaBits2048`/`rsaBits4096`, never the caller's raw value) to `rsa.GenerateKey`, so no path can reach a sub-2048 key. A caller requesting a larger key still gets 4096; everything else floors to 2048.

Valid inputs are unaffected (normal counts and key sizes behave exactly as before; only pathological values are bounded). Emulator-only, no behavior change for real usage.

## Verification
`go build ./...` + `go test` on all six touched packages green; gofmt clean.

### DIFF
--- a/providers/aws/acm/certificate_gen.go
+++ b/providers/aws/acm/certificate_gen.go
@@ -120,10 +120,10 @@ func rsaKeyMaterial(bits int) (signer crypto.Signer, keyPEM, sigAlg string, err 
 	// Never generate a key weaker than 2048 bits, even if a caller requests the
 	// legacy RSA_1024 algorithm: a sub-2048 RSA key is rejected by modern TLS
 	// stacks and flagged as weak crypto. The emulator issues a fake cert, so the
-	// exact bit count carries no semantic weight — clamp to the safe minimum.
-	if bits < rsaBits2048 {
-		bits = rsaBits2048
-	}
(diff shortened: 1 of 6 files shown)


NitinKumar004:
Superseded by #1078, which uses explicit if-guards — CodeQL does not treat builtin min()/max() as a taint sanitizer, so the min()-based approach here would not clear the alerts.

PR #1091, opened 2026-09-04
fix(azure): eliminate internal error-string prefix leak across all ARM handlers

NitinKumar004:
## The bug

The shared Azure ARM error codec (`server/wire/azurearm`) wrote `err.Error()` into the ARM error envelope's `message` field in `WriteCErr` and `WriteParentNotFound`. For a `*cerrors.Error`, `err.Error()` bakes in the internal code-taxonomy prefix (e.g. `"NotFound: vm 'foo' not found"`, `"AlreadyExists: ..."`). Real Azure never prefixes error messages this way — only the `code` field carries that information. `cerrors.Message` (which strips the prefix) had zero usages under `server/azure`.

Because `WriteCErr`/`WriteParentNotFound` are the shared error-dispatch helpers used by every Azure ARM handler, the leak was repo-wide across the 36 service packages that call them (acr, ai, aks, cache, containerapps, containerinstances, cosmosaccount, cosmosdb, cosmospostgresql, costmanagement, databricks, disks, dns, eventgrid, functions, images, keyvault, loadbalancer, loganalytics, managedcassandra, managedidentity, monitor, mysqlflex, notificationhubs, postgresflex, resourcegraph, resourcegroups, search, servicebus, snapshots, sql, sqlvirtualmachine, sshpublickeys, storageaccount, virtualmachines, vnet).

**A follow-up review found the shared-codec fix alone was incomplete**: several handlers have their own local `err.Error()`-based error mappers (a `writeCErr`/`writeErr`-style switch on `cerrors.IsNotFound`/`IsAlreadyExists`/etc.) that bypass the shared codec entirely, plus a few data-plane handlers with the same bug in a different envelope shape. Those still leaked the prefix. This PR closes that gap.

## The fix

`WriteCErr`/`WriteParentNotFound` in `server/wire/azurearm/azurearm.go` build the message with `cerrors.Message(err)` instead of `err.Error()`.

Additionally, every local ARM/data-plane error mapper that duplicated this switch-on-cerrors-code pattern now also uses `cerrors.Message(err)` instead of `err.Error()`:

- ARM local mappers: `iam/operations.go` (writeCErr helper, used by roleDefinitions/roleAssignments CRUD, plus the RoleAssignmentExists conflict path), `virtualmachines/instances.go` (NetworkInterfaceNotFound), `sql/operations.go` (SourceDatabaseNotFound / TargetElasticPoolDoesNotExist), `vnet/handler.go` (InUseSubnetCannotBeDeleted, PublicIPAddressCannotBeDeleted), `vnet/network_interface.go` (PublicIPAddressCannotBeAssignedToMultipleIpConfigs, InUseNetworkInterfaceCannotBeDeleted), `dns/operations.go` (etag PreconditionFailed on record-set PUT/DELETE), `notificationhubs/registrations.go`, `acr/handler.go`, `kusto/dataplane_mgmt.go`, `servicebus/dataplane.go` (MessageLockLost).
- Data-plane mappers (different envelope, same bug class): `keyvault/handler.go`, `keyvault/certs_operations.go`, `keyvault/keys_handler.go`, `blobstorage/handler.go`, `cosmosdb/handler.go`, `queue/handler.go`, `tablestorage/handler.go`/`batch.go`/`helpers.go`, `databricks/dataplane.go`, `search/dataplane.go`, `ai/dataplane.go`.

A full grep of every `err.Error()` call site under `server/azure/` was reviewed one-by-one to confirm this list is exhaustive. Everything left as `err.Error()` is either a plain JSON/body-decode error (no cerrors taxonomy involved), a substring-match branch condition (not a wire message), or a local sentinel error type (`vnet/subnet_validate.go`'s `azurearmSubnetErr`) whose `.Error()` is already a bare, non-cerrors message.

`code` field and HTTP status are untouched everywhere — only the human-readable `message` text changes.

## Tests

- `server/wire/azurearm/azurearm_test.go`: `TestWriteCErrMessageHasNoInternalPrefix`, `TestWriteParentNotFoundMessageHasNoInternalPrefix` (shared codec).
- New focused regression tests for the local-mapper fixes, each verified to fail before the fix and pass after: `server/azure/iam/error_prefix_test.go` (DELETE missing roleDefinition), `server/azure/virtualmachines/error_prefix_test.go` (VM create referencing a missing NIC), `server/azure/vnet/error_prefix_test.go` (delete a NIC still attached to a VM), `server/azure/keyvault/error_prefix_test.go` (GET missing secret, data-plane).

## Black-box verification

Built the binary and ran `serve -providers=azure` against real ARM + data-plane paths.

- IAM: `DELETE roleDefinitions/deadbeef` → `{"code":"ResourceNotFound","message":"role \"deadbeef\" not found"}` (was `"NotFound: role \"deadbeef\" not found"`)
- VM create with a `networkProfile` referencing a missing NIC → `{"code":"NetworkInterfaceNotFound","message":"networkProfile references network interface \"missing-nic\" which does not exist"}` (was prefixed `"InvalidArgument: "`)
- Key Vault data-plane `GET` missing secret → `{"code":"SecretNotFound","message":"secret \"missing-secret\" not found"}` (was prefixed `"NotFound: "`)
- Blob Storage `GET` missing blob → `<Code>BlobNotFound</Code><Message>container "mystorageacct" not found</Message>` (was prefixed `"NotFound: "`)
- vnet: delete a NIC still attached to a VM → `{"code":"InUseNetworkInterfaceCannotBeDeleted","message":"network interface \"nic-1\" cannot be deleted because it is attached to \"...\""}` (was prefixed `"FailedPrecondition: "`)

`code`/status unchanged in every case — only the message text lost its prefix.

## Gates

- `go build ./...`
- `go test ./server/azure/... ./server/wire/azurearm/... -race -count=1` — all green
- `go test ./providers/azure/... -count=1` — all green
- `gofmt -l` — clean on changed files
- `golangci-lint run --new-from-rev=origin/development ./server/wire/azurearm/... ./server/azure/...` — 0 new issues
- `go generate ./...` — no `docs/coverage` diff
- `go mod tidy` — no-op

### DIFF
--- a/server/azure/acr/handler.go
+++ b/server/azure/acr/handler.go
@@ -213,12 +213,12 @@ func writeErr(w http.ResponseWriter, status int, code, msg string) {
 func writeCErr(w http.ResponseWriter, err error) {
 	switch {
 	case cerrors.IsNotFound(err):
-		writeErr(w, http.StatusNotFound, "NAME_UNKNOWN", err.Error())
+		writeErr(w, http.StatusNotFound, "NAME_UNKNOWN", cerrors.Message(err))
 	case cerrors.IsInvalidArgument(err):
-		writeErr(w, http.StatusBadRequest, "INVALID_REQUEST", err.Error())
+		writeErr(w, http.StatusBadRequest, "INVALID_REQUEST", cerrors.Message(err))
 	case cerrors.IsFailedPrecondition(err):
-		writeErr(w, http.StatusConflict, "DENIED", err.Error())
+		writeErr(w, http.StatusConflict, "DENIED", cerrors.Message(err))
 	default:
-		writeErr(w, http.StatusInternalServerError, "INTERNAL", err.Error())
+		writeErr(w, http.StatusInternalServerError, "INTERNAL", cerrors.Message(err))
 	}
 }
--- a/server/azure/ai/dataplane.go
+++ b/server/azure/ai/dataplane.go
@@ -340,12 +340,12 @@ func codeForStatus(status int) string {
 func dpCErr(w http.ResponseWriter, err error) {
 	switch {
 	case errors.IsNotFound(err):
-		dpErr(w, http.StatusNotFound, err.Error())
+		dpErr(w, http.StatusNotFound, errors.Message(err))
 	case errors.IsInvalidArgument(err):
-		dpErr(w, http.StatusBadRequest, err.Error())
+		dpErr(w, http.StatusBadRequest, errors.Message(err))
 	case errors.IsAlreadyExists(err), errors.IsFailedPrecondition(err):
-		dpErr(w, http.StatusConflict, err.Error())
+		dpErr(w, http.StatusConflict, errors.Message(err))
 	default:
-		dpErr(w, http.StatusInternalServerError, err.Error())
+		dpErr(w, http.StatusInternalServerError, errors.Message(err))
 	}
 }
--- a/server/azure/blobstorage/handler.go
+++ b/server/azure/blobstorage/handler.go
@@ -1450,14 +1450,14 @@ func writeErr(w http.ResponseWriter, err error) {
 
 	switch {
 	case cerrors.IsNotFound(err):
-		writeError(w, http.StatusNotFound, "BlobNotFound", err.Error())
+		writeError(w, http.StatusNotFound, "BlobNotFound", cerrors.Message(err))
 	case cerrors.IsAlreadyExists(err):
-		writeError(w, http.StatusConflict, "ContainerAlreadyExists", err.Error())
+		writeError(w, http.StatusConflict, "ContainerAlreadyExists", cerrors.Message(err))
 	case cerrors.IsInvalidArgument(err):
-		writeError(w, http.StatusBadRequest, "InvalidInput", err.Error())
+		writeError(w, http.StatusBadRequest, "InvalidInput", cerrors.Message(err))
 	case cerrors.IsFailedPrecondition(err):
-		writeError(w, http.StatusConflict, "Conflict", err.Error())
+		writeError(w, http.StatusConflict, "Conflict", cerrors.Message(err))
 	default:
-		writeError(w, http.StatusInternalServerError, "InternalError", err.Error())
+		writeError(w, http.StatusInternalServerError, "InternalError", cerrors.Message(err))
 	}
 }
--- a/server/azure/cosmosdb/handler.go
+++ b/server/azure/cosmosdb/handler.go
@@ -1425,14 +1425,14 @@ func writeError(w http.ResponseWriter, status int, code, msg string) {
 func writeErr(w http.ResponseWriter, err error) {
 	switch {
 	case cerrors.IsNotFound(err):
-		writeError(w, http.StatusNotFound, "NotFound", err.Error())
+		writeError(w, http.StatusNotFound, "NotFound", cerrors.Message(err))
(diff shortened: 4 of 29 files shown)


NitinKumar004 (commented):
The extension (a9a99240) fixed the shared-codec fallout and most local mappers correctly, but the sweep is not complete. One genuine leak site remains, in a file the commit itself claims to have fixed.

**Remaining leak: `server/azure/tablestorage/batch.go:44`**

```go
ops, table, err := parseBatch(w, r)
if err != nil {
    writeError(w, http.StatusBadRequest, "InvalidInput", err.Error())
    return
}
```

`parseBatch` can return a genuine `*cerrors.Error` on several paths — `batchErr()` (`batch.go:29-31`, used for "missing multipart boundary", "malformed entity path in change set", "malformed key predicate in change set") and `entityOpFromRequest`'s `cerrors.Newf(cerrors.InvalidArgument, "unsupported change-set method %q", ...)` (`batch.go:257`). `err.Error()` on a `*cerrors.Error` is `"<Code>: <Message>"`, so this line leaks the internal taxonomy prefix.

Live repro against the extension commit:

```
$ curl -sk -X POST "https://127.0.0.1:14896/\$batch" \
    -H "Host: mystorageaccount.table.core.windows.net" \
    -H "Content-Type: multipart/mixed" -d "irrelevant body"
{"odata.error":{"code":"InvalidInput","message":{"lang":"en-US","value":"InvalidArgument: missing multipart boundary"}}}
```

Note the same file's `writeBatchFailure` (batch.go:~312) *was* fixed in this commit (`err.Error()` → `cerrors.Message(err)`), but that only covers per-op failures from an already-executed batch — it left the parse-level failure path at line 44 with the old leaking call. Fix: `cerrors.Message(err)` there too.

**Everything else checked out:**
- Build clean (`go build ./...`), full azure suite green with `-race -count=1` (`./server/azure/...`, `./server/wire/azurearm/...`, `./providers/azure/...` — all `ok`).
- Independent grep of all 51 non-test `err.Error()` sites under `server/azure/` and `providers/azure/` classified every other site as safe (JSON/body-decode errors, `strings.Contains` branch conditions with `cerrors.Message` in the actual write, or bare-message custom error types) or already correctly using `cerrors.Message`/`errors.Message` (including the two files using the unaliased `errors` import — `search/dataplane.go`, `ai/dataplane.go` — both fixed).
- `vnet/subnet_validate.go:110` exclusion is correct: `azurearmSubnetErr` is a local sentinel whose `.Error()` is always a hand-authored bare message; every path that could carry a real `*cerrors.Error` (the CIDR-parse `cerrors.Newf` and the `DescribeSubnets` passthrough) fails the `.(*azurearmSubnetErr)` type assertion and falls through to the already-fixed `azurearm.WriteCErr`.
- Black-box repros clean, with unchanged codes/status: role-definition DELETE 404 `ResourceNotFound`; VM create with missing NIC 400 `NetworkInterfaceNotFound`; Key Vault GET missing secret 404 `SecretNotFound`; subnet delete while in use by a NIC 400 `InUseSubnetCannotBeDeleted`.
- `cerrors.Message` is a struct-field read (`e.Message`), not string-prefix stripping, so there's no risk of truncating a legitimate colon in message content.
- The 4 new `error_prefix_test.go` tests are real guards (checked `iam`'s in full — asserts status code + absence of the `"NotFound:"` substring in the actual response body).

Please fix `tablestorage/batch.go:44` and re-run the completeness sweep for that file family before merge.

PR #1121, opened 2026-09-05
fix(gcs): real-user e2e divergences from GCP

NitinKumar004:
Deep real-user e2e audit of Google Cloud Storage driving the **real Terraform `hashicorp/google` provider (v6.50.0)** and the `cloud.google.com/go/storage` SDK against `cloudemu serve`. Terraform was never exercised in the earlier GCS pilot; it surfaced two genuine, high-impact divergences.

## Findings fixed

| # | Sev | Divergence |
|---|-----|-----------|
| 5 | HIGH | `uploadType=multipart` rejected the object name when supplied only via the `?name=` query parameter (400 `metadata.name required`). The Terraform provider, gcloud and gsutil all send the name in the query and omit it from the metadata part, so **every `google_storage_bucket_object` upload failed**. |
| 6 | HIGH | `GET /b/{bucket}/anywhereCaches` returned 404. The provider lists anywhere caches during `force_destroy` and aborts its object-clearing loop if that call errors, then fails the bucket delete with 409 "not empty" — so **`terraform destroy` failed on any bucket containing objects**. |

## Fixes
- `uploadMultipart`: object name = query `name` (overrides metadata name) then metadata name, per the GCS *Objects: insert* contract ("Not required if the request body contains object metadata that includes a name value"; "Overrides the object metadata's name value, if any"). `initResumable` aligned to the same precedence.
- New `listAnywhereCaches` handler returns `200 {"kind":"storage#anywhereCaches"}` (empty list — the correct response for a bucket with no caches; cloudemu does not model Anywhere Cache). Routed for both the bare and SDK trailing-slash path forms.

## Verification
- `terraform apply` -> `terraform plan` (no drift) -> `terraform destroy` now succeeds end-to-end, including a versioning-enabled bucket holding an object and an advanced bucket (region location, NEARLINE, uniform_bucket_level_access, two lifecycle rules, `google_storage_bucket_iam_member`).
- Object content/md5Hash/crc32c/size round-trip; error envelopes (404 notFound, 409 notEmpty/conflict) correct.
- Regression tests added (`server/gcp/gcs/terraform_compat_test.go`): multipart name-from-query, anywhereCaches empty list (both path forms), full force-destroy of a versioned bucket via the SDK.
- Gates: `go build ./...`, `go vet`, `gofmt -l`, `golangci-lint run --new-from-rev=origin/development` (0 issues), `go test -race` on server/gcp/gcs + providers/gcp/gcs + services/storage — all green. `go generate ./...` — no docs/coverage change (no driver interface change).

## Filed, not fixed (fidelity gaps, no Terraform drift today)
`rpo` and `softDeletePolicy` bucket defaults absent (soft delete is a whole subsystem — not half-built); `iamConfiguration.uniformBucketLevelAccess` omitted when disabled; the inert top-level `error.status` envelope field (unverifiable without live GCS).

### DIFF
--- a/server/gcp/gcs/handler.go
+++ b/server/gcp/gcs/handler.go
@@ -77,6 +77,13 @@ const (
 	subresIAM = "iam"
 	// subresNotificationConfigs is the /b/{bucket}/notificationConfigs segment.
 	subresNotificationConfigs = "notificationConfigs"
+	// subresAnywhereCaches is the /b/{bucket}/anywhereCaches sub-collection
+	// segment. cloudemu does not model Anywhere Cache instances, but the list
+	// endpoint must answer 200 with an empty set: the Terraform google provider
+	// lists anywhere caches during force_destroy and aborts its object cleanup
+	// (leaving the bucket non-empty) if that call errors instead of returning a
+	// list.
+	subresAnywhereCaches = "anywhereCaches"
 	// defaultLocation / defaultStorageClass are the GCS defaults a bucket reads
 	// back when none was set at create.
 	defaultLocation     = "US"
@@ -284,6 +291,8 @@ func (h *Handler) serveBucketSubcollection(w http.ResponseWriter, r *http.Reques
 		h.bucketIAM(w, r, parts[1])
 	case subresNotificationConfigs:
 		h.notificationCollection(w, r, parts[1])
+	case subresAnywhereCaches:
+		h.listAnywhereCaches(w, r, parts[1])
 	case "lockRetentionPolicy":
 		h.lockRetentionPolicy(w, r, parts[1])
 	default:
@@ -299,6 +308,11 @@ func (h *Handler) serveBucketSubresource(w http.ResponseWriter, r *http.Request,
 		h.objectOp(w, r, parts[1], strings.Join(parts[3:], "/"))
 	case parts[2] == subresIAM && parts[3] == "testPermissions":
 		h.testIAMPermissions(w, r, parts[1])
+	case parts[2] == subresAnywhereCaches && parts[3] == "":
+		// The Go storage client appends a trailing slash to the list URL
+		// (b/{bucket}/anywhereCaches/), which splits into a trailing empty
+		// segment; route it to the same empty-list handler.
+		h.listAnywhereCaches(w, r, parts[1])
 	case parts[2] == subresNotificationConfigs:
 		h.notificationResourceOp(w, r, parts[1], parts[3])
 	default:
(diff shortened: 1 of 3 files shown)


NitinKumar004 (commented):
Independent adversarial review, real-user e2e with `cloudemu serve` + real `google_storage_bucket`/`google_storage_bucket_object` (Terraform hashicorp/google v6.50.0) + raw wire probes.

**Verdict: APPROVE**

## 1. Build + tests
- `go build ./...` — clean.
- `go test -race -count=1 ./server/gcp/gcs/... ./providers/gcp/gcs/... ./services/storage/...` — all pass.
- `go test ./server/gcp/... -count=1` — all 27 packages pass.
- `golangci-lint run ./server/gcp/gcs/...` — 2 pre-existing issues (govet shadow at handler.go:1945, nolintlint at handler.go:1536), confirmed present on `origin/development` before this diff too (checked out the pre-PR file and relinted) — not introduced by this PR.

## 2. Fix #1 — multipart name precedence (`uploadMultipart`, handler.go)
Black-box 4-combo matrix against a live server (raw multipart POSTs, CRLF-correct bodies):
| query | metadata | result |
|---|---|---|
| `combo1.txt` | (absent) | 200, stored as `combo1.txt` |
| `query-wins.txt` | `metadata-name.txt` (different) | 200, stored as **`query-wins.txt`** — query overrides metadata, per GCS Objects:insert doc |
| (absent) | `metadata-only.txt` | 200, stored as `metadata-only.txt` |
| (absent) | (absent) | 400 `Required parameter: name` |

Matches the GCS doc's "name query param overrides the object metadata's name value, if any" exactly. Media (simple) upload unaffected (`uploadType=media` always took `name` from query, untouched). Resumable (`initResumable`) verified end-to-end: init with `?name=...` and no metadata → 200 + `Location` with `upload_id`, then a `PUT` with `Content-Range: bytes 0-22/23` committed the object under the query name with correct content/size — the "alignment" (`firstNonEmpty(query, meta.Name)`) did not break it.

## 3. Fix #2 — anywhereCaches (handler.go, types.go)
- `GET /b/{bucket}/anywhereCaches` → `200 {"kind":"storage#anywhereCaches"}` (no `items`).
- `GET /b/{bucket}/anywhereCaches/` (trailing slash, the Go SDK's shape) → same.
- `kind` string (`storage#anywhereCaches`) matches real GCS.
- No shadowing: on the same bucket, `/o` (objects list) and `/iam` still route and answer correctly (checked directly after adding the anywhereCaches route).

## 4. Terraform full lifecycle (the core value)
- `apply` — basic bucket+object, versioning-enabled bucket+object, and an advanced bucket (US-CENTRAL1, NEARLINE, `uniform_bucket_level_access=true`, lifecycle rule) all created; the multipart uploads (previously 400 "metadata.name required") now succeed. **6 to add → 6 added, 0 errors.**
- `plan` immediately after — "No changes. Your infrastructure matches the configuration." (no drift).
- `destroy` with `force_destroy=true` on a bucket holding an object **not tracked in TF state** (uploaded raw, name-only-in-query, i.e. exactly the wire shape of the Terraform provider's own object writes) — this is the real force_destroy path (list objects + list anywhereCaches + delete objects + delete bucket), previously failing 409 not-empty because the anywhereCaches 404 aborted cleanup. **Destroy succeeded**; confirmed bucket gone (`GET` → 404) afterward. Both HIGH repros reproduced and fixed end-to-end via the real provider.

## 5. Object round-trip
Content byte-exact on every object (`hello world from terraform`, `advanced bucket object content`, `v1`, raw multipart/resumable/media test objects). `md5Hash` independently recomputed and matches (`Af+Svq5dwIhRtFPp1mYPlQ==` for the 26-byte terraform object). `crc32c` present on every object. `size` returned as a string, correct byte count.

## 6. Test honesty (revert-fails-test)
- Reverted `uploadMultipart`'s name resolution to `meta.Name`-only → `TestMultipartUploadNameFromQueryParam` fails with the old `400 metadata.name required`.
- Reverted the `anywhereCaches` route (removed the case from `serveBucketSubcollection`) → `TestAnywhereCachesListReturnsEmpty` fails with `404 unknown sub-collection`.
- Both reverts restored; full `server/gcp/gcs` suite green again afterward.

## 7. No regression (spot-checked)
- Duplicate bucket create → `409 conflict`.
- Nonexistent bucket GET → `404 notFound`.
- Delete non-empty bucket without `force_destroy` → `409 notEmpty` ("The bucket you tried to delete was not empty.").
- Object list ordering is lexicographic/deterministic.

## 8. Filed/deferred items
Not part of this diff's scope; the full Terraform `plan` after `apply` showed zero drift on the buckets exercised (including the advanced bucket's `uniform_bucket_level_access`/lifecycle), so nothing here regresses this PR's value — reasonable to leave for a separate pass.

No defects found. Kill of the test `serve` instance confirmed — no orphan left on the scratch port.

PR #1265, opened 2026-09-07
fix(aws): DeleteSubnet drops the subnet's route-table associations

### LINKED ISSUE #1261: [Bug] EC2 DeleteSubnet leaves route-table associations behind, so DeleteRouteTable fails with DependencyViolation
## Describe the Bug

EC2 `DeleteSubnet` does not remove the subnet's route-table associations. The association row stays in `rtAssocs` after the subnet is gone, so a later `DeleteRouteTable` on that table fails with:

```
DependencyViolation: route table "rtb-0000009a" has a subnet association "rtbassoc-0000009d" and cannot be deleted
```

Real AWS deletes a subnet's route-table associations implicitly when the subnet is deleted (the association has no existence apart from the subnet). The standard teardown order — delete subnets, then delete the route tables that served them — therefore succeeds on AWS and fails on cloudemu.

Network ACL associations are already cascaded correctly in the same function, so this looks like an oversight rather than a design choice.

## To Reproduce

Against `cloudemu serve` (or the Docker image), with any AWS SDK:

1. `CreateVpc` → `CreateSubnet` in it → `CreateRouteTable` in it
2. `AssociateRouteTable(RouteTableId, SubnetId)`
3. `DeleteSubnet(SubnetId)` → succeeds
4. `DeleteRouteTable(RouteTableId)` → **DependencyViolation** naming the association from step 2
5. `DescribeRouteTables` still lists the association, pointing at a subnet that no longer exists

Found while running a full EKS provision-and-teardown lifecycle from the zopdev/zopnight provisioner against cloudemu (`AWS_ENDPOINT_URL` override, no code changes). Every other resource tore down cleanly; the route tables and therefore the VPC were left behind, and the deletion job reported a failure that does not happen on real AWS.

## Expected Behavior

After `DeleteSubnet`, `DescribeRouteTables` shows no association for that subnet, and `DeleteRouteTable` on a table whose only associations were to deleted subnets succeeds.

## Actual Behavior

The association survives the subnet. `DeleteRouteTable` refuses with the FailedPrecondition above; the route table and its VPC cannot be deleted without an explicit `DisassociateRouteTable` that real-AWS callers never need.

## Code Sample

`providers/aws/vpc/vpc.go`, `(*Mock).DeleteSubnet`: the loop after `m.subnets.Delete(id)` cascades `m.aclAssocs` only. The equivalent sweep over `m.rtAssocs` (delete every non-main association whose `SubnetID == id`) is missing.

`providers/aws/vpc/route_table.go`, `DeleteRouteTable`: the guard at the `has a subnet association` check is correct for live subnets; it only misfires because the stale rows above are never removed.

```go
// DeleteSubnet, after m.subnets.Delete(id) — mirrors the existing aclAssocs sweep
for assocID, a := range m.rtAssocs.All() {
    if a.SubnetID == id && !a.Main {
        m.rtAssocs.Delete(assocID)
    }
}
```

## Environment

- Go version: 1.26.0 (client side, aws-sdk-go-v2 v1.43.5 / config v1.32.16)
- CloudEmu version: Docker `ghcr.io/stackshy/cloudemu:latest` (2.10.0-engines); source at `d19d141718c77b9652d12d7c9827f3c2dde25200`
- OS: macOS (Docker, linux/amd64 image)

arunesh-j:
Fixes #1261.

## What was wrong

EC2 removes a subnet's route-table associations when the subnet is deleted — the association has no existence apart from the subnet. The AWS mock kept the `rtAssocs` row, so the standard teardown order (delete subnets, then the route tables that served them) hit `DeleteRouteTable`'s dependency guard on a table that was actually free:

```
DependencyViolation: route table "rtb-0000009a" has a subnet association "rtbassoc-0000009d" and cannot be deleted
```

The same function already cascaded network ACL associations, so this was an oversight in the cascade, not a design choice. Found while running a full EKS provision-and-teardown lifecycle from a real provisioner against cloudemu: every other resource tore down cleanly, the route tables and therefore the VPC were left behind, and the run reported a failure that does not happen on AWS.

## Change

- **AWS** `(*Mock).DeleteSubnet`: after removing the subnet, drop every non-main `rtAssocs` entry whose `SubnetID` matches, mirroring the existing `aclAssocs` sweep. The main association carries no subnet and is left alone, so `DeleteRouteTable`'s main-table protection is unchanged.
- **Azure** and **GCP** `DeleteSubnet`: the same cascade, for portable-API parity per CONTRIBUTING. Neither guards `DeleteRouteTable` on associations so nothing was failing there, but a stale association to a deleted subnet lingered in their stores too. In Azure the association is literally a subnet property, so this also matches the real service.

No wire-layer change: the server handlers already call the mock.

## Tests

Per provider, `TestDeleteSubnet_CascadesRouteTableAssociations`, each watched failing before the fix:

- After `DeleteSubnet`, `DisassociateRouteTable(assoc)` returns not-found.
- **AWS additionally:** `DeleteRouteTable` on that table now succeeds; and a table still associated with a *different* live subnet keeps refusing deletion, so the guard is intact.

Verified at the wire level too: built `cmd/cloudemu`, ran `serve` on spare ports, and replayed the issue's five ec2query steps — `CreateVpc → CreateSubnet → CreateRouteTable → AssociateRouteTable → DeleteSubnet → DeleteRouteTable`. `DescribeRouteTables` shows no association after the subnet delete and `DeleteRouteTable` returns `<return>true</return>`.

`golangci-lint run --new-from-rev=development` on the three packages: 0 issues. `go test ./...` on the branch: no failures.

### DIFF
--- a/providers/aws/vpc/vpc.go
+++ b/providers/aws/vpc/vpc.go
@@ -762,6 +762,14 @@ func (m *Mock) DeleteSubnet(_ context.Context, id string) error {
 		}
 	}
 
+	// EC2 drops a subnet's route-table associations with the subnet; a stale
+	// one would make DeleteRouteTable refuse a table that is really free.
+	for assocID, a := range m.rtAssocs.All() {
+		if a.SubnetID == id && !a.Main {
+			m.rtAssocs.Delete(assocID)
+		}
+	}
+
 	return nil
 }
 
--- a/providers/aws/vpc/vpc_test.go
+++ b/providers/aws/vpc/vpc_test.go
@@ -1579,3 +1579,38 @@ func assertNotEmpty(t *testing.T, s string) {
 		t.Error("expected non-empty string")
 	}
 }
+
+// Real EC2 drops a subnet's route-table associations with the subnet, so the
+// standard teardown (delete subnets, then their route tables) must succeed.
+func TestDeleteSubnet_CascadesRouteTableAssociations(t *testing.T) {
+	m := newTestMock()
+	ctx := context.Background()
+	v := createTestVPC(m)
+	s, _ := m.CreateSubnet(ctx, driver.SubnetConfig{VPCID: v.ID, CIDRBlock: "10.0.1.0/24"})
+	rt, _ := m.CreateRouteTable(ctx, driver.RouteTableConfig{VPCID: v.ID})
+	assoc, err := m.AssociateRouteTable(ctx, rt.ID, s.ID)
+	requireNoError(t, err)
+
+	requireNoError(t, m.DeleteSubnet(ctx, s.ID))
+
+	t.Run("association is gone", func(t *testing.T) {
+		assertError(t, m.DisassociateRouteTable(ctx, assoc.ID), true)
+	})
+
+	t.Run("route table can now be deleted", func(t *testing.T) {
+		requireNoError(t, m.DeleteRouteTable(ctx, rt.ID))
+	})
+
+	t.Run("other subnets' associations survive", func(t *testing.T) {
+		s2, _ := m.CreateSubnet(ctx, driver.SubnetConfig{VPCID: v.ID, CIDRBlock: "10.0.2.0/24"})
+		s3, _ := m.CreateSubnet(ctx, driver.SubnetConfig{VPCID: v.ID, CIDRBlock: "10.0.3.0/24"})
+		rt2, _ := m.CreateRouteTable(ctx, driver.RouteTableConfig{VPCID: v.ID})
+		_, _ = m.AssociateRouteTable(ctx, rt2.ID, s2.ID)
+		keep, _ := m.AssociateRouteTable(ctx, rt2.ID, s3.ID)
+
+		requireNoError(t, m.DeleteSubnet(ctx, s2.ID))
+		// s3 is still associated, so the table must still refuse to go.
+		assertError(t, m.DeleteRouteTable(ctx, rt2.ID), true)
+		requireNoError(t, m.DisassociateRouteTable(ctx, keep.ID))
+	})
+}
--- a/providers/azure/vnet/vnet.go
+++ b/providers/azure/vnet/vnet.go
@@ -276,6 +276,13 @@ func (m *Mock) DeleteSubnet(_ context.Context, id string) error {
 		return cerrors.Newf(cerrors.NotFound, "subnet %q not found", id)
 	}
 
+	// A route-table association belongs to the subnet and must not outlive it.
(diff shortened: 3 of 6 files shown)


NitinKumar004 (commented):
## Review — DeleteSubnet drops the subnet's route-table associations

**No blocking findings — clean, correct, and well-tested.**

The fix removes a subnet's route-table associations on `DeleteSubnet`, mirroring the pre-existing ACL-association cleanup two lines above it (`providers/aws/vpc/vpc.go`) — `rtAssocs` was the one association type left dangling, which is exactly the gap that made `DeleteRouteTable` refuse a table that was really free. Applied consistently across all three providers (AWS/Azure/GCP).

### Verified
- **Correctness / concurrency:** `memstore.Store.All()` returns a snapshot, so iterating it while calling `Delete` on the store is safe; the change follows the exact existing pattern and per-store lock model, so it introduces no new race.
- **AWS `!a.Main` guard** is right — a main route-table association is VPC-level and must survive a subnet delete; Azure/GCP have no `Main` concept, so deleting all matching associations there is the correct mirror-appropriate difference.
- **Tests are thorough** — `TestDeleteSubnet_CascadesRouteTableAssociations` asserts the association is gone, the route table can then be deleted (the cascade this targets), and — importantly — that *other* subnets' associations survive (the over-delete negative case, incl. a table that must still refuse to delete while another subnet is associated). Azure/GCP carry parallel tests.
- `go build` and `go test -race` pass on all three touched packages.

### Blast radius
Safe: the change only *adds* cleanup. No caller depended on the old (stale-association) behavior; `DeleteRouteTable`'s association check now correctly sees the subnet's association removed.

### CI note
The current red checks (Build & Vet / Test / Test 4-4 / Tidy) are all the same transient `proxy.golang.org` module-download outage from that run (`stream error: INTERNAL_ERROR` fetching costexplorer/kusto/elasticache/vpclattice/appflow), not this change — a re-run should clear them. The branch is also a bit behind `development`; a rebase before merge is worthwhile.

NitinKumar004 (approved):
Approving. Verified the fix builds and vets clean against current `development` (the stale branch merges zero-conflict), `-race` clean across all three providers under concurrent DeleteSubnet/Associate/Describe, and confirmed end-to-end on the wire: after `DeleteSubnet`, the subnet's route table can be deleted, while another subnet's association is correctly retained (`DependencyViolation`). The `!a.Main` guard is correct — a main association carries no subnet id.

The current red checks are a transient `proxy.golang.org` module-download flake (SDK zip fetch `INTERNAL_ERROR`), not this change — re-running them.

